PR #6, opened 2022-12-01
Circuit propagator

mmcilree:
See notes attached to email.


ciaranm:
I've broken this a bit. Problem::create_proof_only_variable will need to go into State instead, and Circuit will need a clone() function.

ciaranm:
I couldn't sleep because my tooth hurts, so I implemented proof-only variables and took the opportunity to rename some of the unhelpful create_variable-related functions. The way I've set it up is a bit different, and doesn't involve fake state for these variables any more. You can't use these variables inside Linear constraints, but you can use them inside pseudo-Boolean inequalities (which now also support half reification and equalities). So, something like this appears to work:

```
--- gcs/constraints/circuit.cc	2022-12-02 23:25:32.470107487 +0000
+++ gcs/constraints/circuit.cc	2022-12-02 23:19:45.397221315 +0000
@@ -14,13 +16,18 @@
 using std::nullopt;
 using std::to_string;
 using std::pair;
-
+using std::unique_ptr;
 
 Circuit::Circuit(const vector<IntegerVariableID> & v) :
     _succ(v)
 {
 }
 
+auto Circuit::clone() const -> unique_ptr<Constraint>
+{
+    return make_unique<Circuit>(_succ);
+}
+
 namespace {
     // Very rubbish basic propagator to get started with
     auto propagate_circuit_using_check(vector<IntegerVariableID> succ, State& state) -> Inference {
@@ -69,33 +76,33 @@
 
     if (propagators.want_definitions()) {
         // Define encoding to eliminate sub-cycles
-        vector<IntegerVariableID> position;
+        vector<ProofOnlySimpleIntegerVariableID> position;
         auto constN = ConstantIntegerVariableID{Integer{static_cast<long long>(_succ.size()-1)}};
 
         // Need extra proof variable: pos[i] = j means "the ith node visited in the circuit is the jth var"
         // WLOG start at node 0, so pos[0] = [0]
-        position.emplace_back(propagators.create_proof_only_variable(0_i, 0_i, "pos0"));
-        auto cv = Linear{{1_i, position[0]}, {-1_i, constN}};
-        propagators.define_linear_eq(initial_state, cv, 0_i, _succ[0] == 0_i);
+        position.emplace_back(propagators.create_proof_only_integer_variable(0_i, 0_i, "pos0"));
+        auto cv = WeightedPseudoBooleanTerms{{1_i, position[0]}, {-1_i, constN}};
+        propagators.define_pseudoboolean_eq(initial_state, move(cv), 0_i, _succ[0] == 0_i);
 
         for(unsigned int idx = 1; idx < _succ.size(); ++idx) {
-            position.emplace_back(propagators.create_proof_only_variable(0_i, Integer(_succ.size() - 1), "pos" + to_string(idx)));
+            position.emplace_back(propagators.create_proof_only_integer_variable(0_i, Integer(_succ.size() - 1), "pos" + to_string(idx)));
         }
 
         for(unsigned int idx = 1; idx < _succ.size(); ++idx) {
             // (succ[0] = i) -> pos[i] = 1
-            auto cv1 = Linear{{1_i, position[idx]}, {-1_i, 1_c}};
-            propagators.define_linear_eq(initial_state, cv1, 0_i, _succ[0] == Integer{idx});
+            auto cv1 = WeightedPseudoBooleanTerms{{1_i, position[idx]}, {-1_i, 1_c}};
+            propagators.define_pseudoboolean_eq(initial_state, move(cv1), 0_i, _succ[0] == Integer{idx});
 
             // (succ[i] = 0) -> pos[i] = n-1
-            auto cv2 = Linear{{1_i, position[idx]}, {-1_i, constN}};
-            propagators.define_linear_eq(initial_state, cv2, 0_i, _succ[idx] == 0_i);
+            auto cv2 = WeightedPseudoBooleanTerms{{1_i, position[idx]}, {-1_i, constN}};
+            propagators.define_pseudoboolean_eq(initial_state, move(cv2), 0_i, _succ[idx] == 0_i);
 
             // (succ[i] = j) -> pos[j] = pos[i] + 1
             for(unsigned int jdx = 1; jdx < _succ.size(); ++jdx) {
                 //if (idx == jdx) continue;
-                auto cv3 = Linear{{1_i, position[jdx]}, {-1_i, position[idx]}, {-1_i, 1_c}};
-                propagators.define_linear_eq(initial_state, cv3, 0_i, _succ[idx] == Integer{jdx});
+                auto cv3 = WeightedPseudoBooleanTerms{{1_i, position[jdx]}, {-1_i, position[idx]}, {-1_i, 1_c}};
+                propagators.define_pseudoboolean_eq(initial_state, move(cv3), 0_i, _succ[idx] == Integer{jdx});
             }
         }
 
```

I am very confused by the inequality you're actually writing for `WLOG start at node 0, so pos[0] = [0]` which looks far more complicated than it needs to be to me, but this may be lack of sleep...

mmcilree:
Ah whoops, the comment `WLOG start at node 0, so pos[0] = 0` was just meant to refer only to the line directly after, i.e. fact that we force pos[0] to be 0 in its definition: domain [0 ... 0]

Then there's the inequality: 

$x_0  = 0 \implies x_0 = n - 1$

which should have had a comment explaining that if $x[0] = 0$, then the only time that is allowed is if we have the trivial graph with just one node (i.e.$ n = 1, n - 1 = 0$)

mmcilree:
I've realised that even the simple check propagator for circuit may not be as straightforward to proof log as I originally thought. This is because the $pos_j = pos_i + 1$ equations don't necessarily immediately collapse down to contradiction. Minimal example:

```
2 x1 1 x0 -2 y1 -1 y0 = 1 ;
2 y1 1 y0 -2 x1 -1 x0 = 1 ;
```

These two equations are clearly contradictory, since they say $x = y + 1$ and $y = x + 1$ and we can sub one into the other to get $0 = 2$. However, they don't *unit propagate* to contradiction

```
pseudo-Boolean proof version 1.2
f 4
u >= 1;
c -1
```
gives:
```
Failed in proof file line 3.
Hint: Failed to show '>= 1' by reverse unit propagation.
```
I think we need the additional step:
```
p 1 3 +
```
Which allows:

```
=== begin trace ===
line 002: f 4
  ConstraintId 001: 1 x0 2 x1 1 ~y0 2 ~y1 >= 4
  ConstraintId 002: 1 ~x0 2 ~x1 1 y0 2 y1 >= 2
  ConstraintId 003: 1 ~x0 2 ~x1 1 y0 2 y1 >= 4
  ConstraintId 004: 1 x0 2 x1 1 ~y0 2 ~y1 >= 2
line 003: p 1 3 +
  ConstraintId 005: >= 2
line 004: u >= 1;
  ConstraintId 006: >= 1
line 005: c -1
=== end trace ===

Verification succeeded.
```

For the actual circuit propagator then, I've added a map to keep track of the PB constraints that say 
$pos[j] = pos[i] + 1$, for each $i, j$, and then for each small cycle, output a pol step that manually tells the verifier to add up the relevant constraints to get the collapse down to contradiction.

mmcilree:
I've also implemented prevent, and it seems to take about 2/3 the time of check (on some very haphazard unscientific benchmarking), so that's encouraging. I suspect the need for pol steps outlined above probably also applies here, will check that tomorrow.

ciaranm:
How vital is it that alldifferent is definitely propagated before this is called?

ciaranm:
Thanks. A few little things:

Can you rebase against main? I don't usually mind so much about clean history, but I think in this case it's worth tidying up because of the merges and the files you added then removed, and some of the stray left-overs from adding proof-only variables. Also, it looks like clang-format has only picked up your most recent changes, so doing a clang-format on every commit might be helpful.

gcs/CMakeLists.txt still needs a fix.

When installing the circuit constraint, can you add in a call to propagators.trim_lower_bound, similar to how we do element?

Can you add a comment with the source of the tour example?

I missed this before, but `const std::vector<IntegerVariableID> & _succ;` as a reference for Circuit won't work any more due to the changes to clone() and how constraints are handled. It needs to be a copy rather than a reference now, or it might crash if the caller's vector gets destroyed before the search is done.

PR #9, opened 2023-01-17
Smart Table Constraint

mmcilree:
Uses a version of SmartSTR as described in ["The Smart Table Constraint"](https://link.springer.com/chapter/10.1007/978-3-319-18008-3_19) to implemented disjunction of conjuctions of "SmartEntry" constraints:

Smart entries can have the following forms:
1. $<\text{var}> <\text{op}> a$
2. $<\text{var}> \in S$ or $<\text{var}> \notin S$
3. $<\text{var}><\text{op}><\text{var}>$ 
4. $<\text{var}><\text{op}><\text{var}> + \;b$

where $<\text{var}>$ is a variable, $S$ is a set of constants, $<\text{op}>$ is one of $\{<, \leq, =, \neq, \geq, >\}$.


gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+                continue;
+            }
+
+            for (const auto & tree : forests[tuple_idx]) {
+                // Initialise supported by tree to current variable domains
+                VariableDomainMap supported_by_tree{};
+
+                for (const auto & var : vars) {
+                    state.for_each_value(var, [&](Integer value) {
+                        if (! supported_by_tree.contains(var)) {
```
ciaranm, line 316: Also not needed?

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+            if (state.optional_single_value(selectors[tuple_idx]) == 0_i) {
+                continue;
+            }
+
+            for (const auto & tree : forests[tuple_idx]) {
+                // Initialise supported by tree to current variable domains
+                VariableDomainMap supported_by_tree{};
+
+                for (const auto & var : vars) {
+                    state.for_each_value(var, [&](Integer value) {
```
ciaranm, line 315: Coming back to sorting: as it happens, `state.for_each_value` will always return them in sorted order. Perhaps we should make this an explicit guarantee? Maybe also we should have a `state.values_as_vector` which might be much faster?

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+
+            if (state.optional_single_value(selectors[tuple_idx]) != 0_i) {
+                const auto unrestricted = get_unrestricted(vars, tuples[tuple_idx]);
+                for (const auto & var : unrestricted) {
+                    unsupported[var] = vector<Integer>{};
+                }
+            }
+        }
+
+        bool some_tuple_still_feasible = false;
```
ciaranm, line 349: This feels potentially slow if there are lots of tuples and nearly all of them are infeasible. With Table and its use of an integer variable, we get fast iteration over sparse sets for free (allegedly). Perhaps this is worth considering?

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+        for (unsigned int tuple_idx = 0; tuple_idx < tuples.size(); ++tuple_idx) {
+            if (state.optional_single_value(selectors[tuple_idx]) != 0_i) {
+                some_tuple_still_feasible = true;
```
ciaranm, line 352:
```suggestion
                some_tuple_still_feasible = true;
                break;
```

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+                    entry_tree[current_level].emplace_back(edge);
+                },
+                [&](const UnaryValueEntry &) {
+                    entry_tree[current_level].emplace_back(edge);
+                }}
+                .visit(edge);
+        }
+    }
+
+    auto build_forests(SmartTuples & tuples) -> vector<SmartTable::Forest>
```
ciaranm, line 409:
```suggestion
    [[nodiscard]] auto build_forests(SmartTuples & tuples) -> vector<SmartTable::Forest>
```

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+            unordered_map<IntegerVariableID, bool> node_visited;
+            unordered_map<IntegerVariableID, vector<SmartEntry>> adjacent_edges;
+
+            // Get all the vars in the tuple and record adjacencies
+            for (const auto & entry : current_tuple) {
+                overloaded{
+                    [&](const BinaryEntry & binary_entry) {
+                        node_visited[binary_entry.var_1] = false;
+                        node_visited[binary_entry.var_2] = false;
+                        adjacent_edges.try_emplace(binary_entry.var_1, vector<SmartEntry>{});
```
ciaranm, line 422: Not sure you need this due to default initialisation?

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+            }
+
+            forests.emplace_back(forest);
+        }
+        return forests;
+    }
+}
+
+// For PB model
+auto make_binary_entry_flag(State & state, Propagators & propagators, const IntegerVariableID & var_1, const IntegerVariableID & var_2, const ConstraintType & c) -> ProofFlag
```
ciaranm, line 459: Should this be in the anonymous namespace?

ciaranm:
```suggestion
[[nodiscard]] auto make_binary_entry_flag(State & state, Propagators & propagators, const IntegerVariableID & var_1, const IntegerVariableID & var_2, const ConstraintType & c) -> ProofFlag
```

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+    switch (c) {
+    case EQUAL:
+        // f => var1 == var2
+        flag = propagators.create_proof_flag("bin_eq");
+        propagators.define_pseudoboolean_eq(state, move(var_1_minus_var_2), 0_i, flag);
+
+        // !f => var1 != var2
+        flag_lt = propagators.create_proof_flag("lt");
+        flag_gt = propagators.create_proof_flag("gt");
+        propagators.define_pseudoboolean_ge(state, move(var_1_minus_var_2), 1_i, flag_gt);
```
ciaranm, line 473: I think this is only working by fluke: `move` can be a destructive operation. You can't rely upon a variable having a sensible value after you move it, but you're using `var_1_minus_var_2` twice.

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+            forests.emplace_back(forest);
+        }
+        return forests;
+    }
+}
+
+// For PB model
+auto make_binary_entry_flag(State & state, Propagators & propagators, const IntegerVariableID & var_1, const IntegerVariableID & var_2, const ConstraintType & c) -> ProofFlag
+{
+    ProofFlag flag, flag_lt, flag_gt; // Is this initialisation okay?
```
ciaranm, line 461: Yes. You don't need to `{}` initialise things when you're declaring variables.

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+        return flag;
+    case NOT_EQUAL:
+        // !f => var1 == var2
+        flag = propagators.create_proof_flag("bin_eq");
+        propagators.define_pseudoboolean_eq(state, move(var_1_minus_var_2), 0_i, ! flag);
+
+        // f => var1 != var2
+        flag_lt = propagators.create_proof_flag("lt");
+        flag_gt = propagators.create_proof_flag("gt");
+        propagators.define_pseudoboolean_ge(state, move(var_1_minus_var_2), 1_i, flag_gt);
```
ciaranm, line 501: Again, I think you're reusing moved variables.

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+        flag = propagators.create_proof_flag("bin_gt");
+        propagators.define_pseudoboolean_ge(state, move(var_1_minus_var_2), 0_i, flag);
+        propagators.define_pseudoboolean_ge(state, move(var_2_minus_var_1), 1_i, ! flag);
+        return flag;
+    default:
+        throw UnexpectedException{"Unexpected SmartEntry type encountered while creating PB model."};
+    }
+}
+
+auto literal_from_unary_entry(UnaryValueEntry & unary_entry) -> Literal
```
ciaranm, line 517: Anonymous namespace?

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+        propagators.define_pseudoboolean_ge(state, move(var_2_minus_var_1), 1_i, flag_lt);
+        propagators.define_pseudoboolean_ge(state, move(var_1_minus_var_2), 0_i, ! flag_lt);
+        propagators.define_pseudoboolean_ge(state, {{1_i, flag_lt}, {1_i, flag_gt}}, 1_i, flag);
+        return flag;
+    case GREATER_THAN_EQUAL:
+        flag = propagators.create_proof_flag("bin_gt");
+        propagators.define_pseudoboolean_ge(state, move(var_1_minus_var_2), 0_i, flag);
+        propagators.define_pseudoboolean_ge(state, move(var_2_minus_var_1), 1_i, ! flag);
+        return flag;
+    default:
```
ciaranm, line 512: I try to avoid using default in a switch, because it turns off compiler warnings for missing case entries. I would move this outside the switch instead.

ciaranm: Also I think there's a specific NonExhaustiveSwitch exception I made for this kind of thing?

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+        return var < value + 1_i;
+    case EQUAL:
+        return var == value;
+    case NOT_EQUAL:
+        return var != value;
+    case GREATER_THAN:
+        return var >= value + 1_i;
+    case GREATER_THAN_EQUAL:
+        return var >= value;
+    default:
```
ciaranm, line 534: Would move outside.

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+        std::map<BinaryEntryData, ProofFlag> smart_entry_flags;
+
+        for (unsigned int tuple_idx = 0; tuple_idx < _tuples.size(); ++tuple_idx) {
+            WeightedPseudoBooleanTerms entry_flags_sum{};
+            WeightedPseudoBooleanTerms entry_flags_neg_sum{};
+            for (const auto & entry : _tuples[tuple_idx]) {
+                overloaded{
+                    [&](BinaryEntry binary_entry) {
+                        auto binary_entry_data = std::make_tuple(binary_entry.var_1, binary_entry.var_2, binary_entry.constraint_type);
+                        if (! smart_entry_flags.contains(binary_entry_data))
```
ciaranm, line 577: I think modern C++ has a function that does these two lines a bit more efficiently, `try_emplace` or something like that?

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+                            smart_entry_flags[binary_entry_data] = make_binary_entry_flag(initial_state, propagators, binary_entry.var_1, binary_entry.var_2, binary_entry.constraint_type);
+
+                        WeightedPseudoBooleanTerm t{1_i, smart_entry_flags[binary_entry_data]};
+                        WeightedPseudoBooleanTerm minus_t{-1_i, smart_entry_flags[binary_entry_data]};
+                        entry_flags_sum.emplace_back(t);
```
ciaranm, line 582: Should be able to inline these rather than using variables.

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+                    }}
+                    .visit(entry);
+            }
+            auto tuple_len = Integer{static_cast<long long>(entry_flags_sum.size())};
+            propagators.define_pseudoboolean_ge(initial_state, move(entry_flags_sum), tuple_len, pb_selectors[tuple_idx]);
+            propagators.define_pseudoboolean_ge(initial_state, move(entry_flags_neg_sum), -tuple_len + 1_i, ! pb_selectors[tuple_idx]);
+        }
+    }
+
+    // Trigger when any var changes? Is this overkill?
```
ciaranm, line 623: I guess if variables only appear in, e.g. less-than tuples, they can get weaker triggers. Perhaps this is another argument in favour of calculating all variables ourselves rather than relying upon the user passing them in.

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+    Triggers triggers;
+    triggers.on_change = {_vars.begin(), _vars.end()};
+
+    vector<Forest> forests = build_forests(_tuples);
+
+    propagators.install(
+        [selectors, vars = _vars, tuples = _tuples, forests = forests](State & state) -> pair<Inference, PropagatorState> {
```
ciaranm, line 630: You should be able to `move(_tuples)` and forests here.

gcs/constraints/smart_table.hh
```diff
@@ -0,0 +1,26 @@
...
+#include <gcs/constraint.hh>
+#include <gcs/extensional.hh>
+#include <gcs/smart_entry.hh>
+#include <gcs/variable_id.hh>
+namespace gcs
+{
+    class SmartTable : public Constraint
+    {
+    private:
+        const std::vector<IntegerVariableID> & _vars;
```
ciaranm, line 13: With the new clone constraints interface, these should be stored as values, not references. In fact, I think probably they should be a `shared_ptr` to the const data, and the user API should accept both a value and a  `shared_ptr`.

gcs/constraints/smart_table.hh
```diff
@@ -0,0 +1,26 @@
...
+        SmartTuples & _tuples;
+
+    public:
+        explicit SmartTable(const std::vector<IntegerVariableID> & vars, SmartTuples & tuples);
```
ciaranm, line 17: Pass by copy, use `move` more, and accept a `shared_ptr` as another option?

gcs/constraints/smart_table.hh
```diff
@@ -0,0 +1,26 @@
...
+        virtual auto clone() const -> std::unique_ptr<Constraint> override;
+        virtual auto describe_for_proof() -> std::string override;
+        virtual auto install(innards::Propagators &, innards::State &) && -> void override;
+        using TreeEdges = std::vector<std::vector<SmartEntry>>;
```
ciaranm, line 21: Can these be private?

ciaranm: Or even just in the cc file?

gcs/smart_entry.hh
```diff
@@ -0,0 +1,130 @@
+#ifndef GLASGOW_CONSTRAINT_SOLVER_GUARD_GCS_SMART_ENTRY_HH
+#define GLASGOW_CONSTRAINT_SOLVER_GUARD_GCS_SMART_ENTRY_HH
+
+#include <variant>
+#include <vector>
+#include <gcs/variable_id.hh>
+#include <gcs/integer.hh>
+#include <algorithm> //std::sort
```
ciaranm, line 8: Blank line before

gcs/smart_entry.hh
```diff
@@ -0,0 +1,130 @@
...
+using namespace gcs;
```
ciaranm, line 9: Blank line before

gcs/smart_entry.hh
```diff
@@ -0,0 +1,130 @@
...
+
```
ciaranm, line 10: All of this should be in `namespace gcs`. I think it might even be worth having a sub-namespace for some of it, so that shorter class names can be used?

gcs/smart_entry.hh
```diff
@@ -0,0 +1,130 @@
...
+enum ConstraintType {
```
ciaranm, line 11: Prefer `enum class` to `enum`.

gcs/smart_entry.hh
```diff
@@ -0,0 +1,130 @@
...
+    EQUAL,
+    NOT_EQUAL,
+    GREATER_THAN,
+    GREATER_THAN_EQUAL,
+    IN,
+    NOT_IN
+};
+
+/**
+ * \brief Smart entries for smart tables
```
ciaranm, line 23: These should all get a Doxygen group.

gcs/smart_entry.hh
```diff
@@ -0,0 +1,130 @@
...
+ */
+struct BinaryEntry {
+    IntegerVariableID var_1;
+    IntegerVariableID var_2;
+    ConstraintType constraint_type;
+
+    // Constructor
```
ciaranm, line 30: Your ctor is what would be created by default, you don't need to specify it.

gcs/smart_entry.hh
```diff
@@ -0,0 +1,130 @@
...
+    : var(var), value(value),constraint_type(constraint_type) {};
+};
+
+struct UnarySetEntry {
+    IntegerVariableID var;
+    std::vector<Integer> values;
+    ConstraintType constraint_type;
+
+    // Constructor
+    UnarySetEntry(const IntegerVariableID &var, const std::vector<Integer> &values, ConstraintType constraint_type)
```
ciaranm, line 51: Again, no ctor needed, but if you did need one, I think it's best to pass the vector by value and then use a move.

ciaranm (commented):
Comments from a first pass, we can discuss on Thursday or whenever you're free.

General C++ thing: when using auto to declare functions, specify `-> void` if it doesn't return something, and `[[nodiscard]]` if it does.

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+                    },
+                    [&](const UnaryValueEntry & unary_val_entry) {
+                        remove_supported(unsupported, unary_val_entry.var, supported_by_tree[unary_val_entry.var]);
+                    }}
+                    .visit(edge);
+            }
+        }
+    }
+
+    auto get_unrestricted(const vector<IntegerVariableID> & vars, const vector<SmartEntry> & tuple) -> vector<IntegerVariableID>
```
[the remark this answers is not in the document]
mmcilree: Don't think this one is right. Have fixed the others

gcs/smart_entry.hh
mmcilree: For some reason I get an error with the child structures if I don't have a constructor

`No matching constructor for initialization of 'BinaryEntry' candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 3 were provided candidate constructor (the implicit move constructor) not viable: requires 1 argument, but 3 were provided candidate constructor (the implicit default constructor) not viable: requires 0 arguments, but 3 were provided`

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+SmartTable::SmartTable(const vector<IntegerVariableID> & v, SmartTuples & t) :
+    _vars(v),
+    _tuples(t)
+{
+}
+
+namespace
+{
+    // Shorthands
+    using VariableDomainMap = unordered_map<IntegerVariableID, vector<Integer>>;
```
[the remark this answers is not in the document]
mmcilree: Not sure... I just did it to save line space - could replace with the longhand everywhere if necessary, but could be painful to read.
Particularly the   
```
using TreeEdges = std::vector<std::vector<SmartEntry>>;
using Forest = std::vector<TreeEdges>;
```
which I did so I could write `vector<Forest>` rather than `vector<vector<vector<vector<SmartEntry>>>>` which seems ludicrous

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+    //                cout << &us.var;
+    //            },
+    //            [&](UnaryValueEntry us) {
+    //                cout << &us.var;
+    //            }}
+    //            .visit(edge);
+    //    }
+    // // ---
+
+    auto filter_edge(const SmartEntry & edge, VariableDomainMap & supported_by_tree)
```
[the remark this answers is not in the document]
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+    {
+        // Currently filter both domains - might be overkill
+        // If the tree was in a better form, think this can be optimised to do less redundant filtering.
+        overloaded{
+            [&](const BinaryEntry & binary_entry) {
+                vector<Integer> new_dom_1{};
+                vector<Integer> new_dom_2{};
+                auto dom_1 = supported_by_tree[binary_entry.var_1];
+                auto dom_2 = supported_by_tree[binary_entry.var_2];
+                std::sort(dom_1.begin(), dom_1.end());
```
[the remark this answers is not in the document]
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+                        [&](Integer dom_val) { return dom_val >= value; });
+                    break;
+                default:
+                    throw UnexpectedException{"Unexpected SmartEntry type encountered."};
+                }
+            }}
+            .visit(edge);
+    }
+
+    auto filter_and_check_valid(const SmartTable::TreeEdges & tree, VariableDomainMap & supported_by_tree) -> bool
```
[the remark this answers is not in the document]
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+        auto unsupported_set = std::set(unsupported[var].begin(), unsupported[var].end());
+        auto to_remove_set = std::set(to_remove.begin(), to_remove.end());
+        std::set_difference(unsupported_set.begin(), unsupported_set.end(),
+            to_remove_set.begin(), to_remove_set.end(),
+            back_inserter(new_unsupported));
+
+        unsupported[var] = new_unsupported;
+    }
+
+    auto filter_again_and_remove_supported(const SmartTable::TreeEdges & tree, VariableDomainMap & supported_by_tree, VariableDomainMap & unsupported)
```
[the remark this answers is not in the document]
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+        auto vars_set = std::set(vars.begin(), vars.end());
+        auto vars_in_tuple_set = std::set(vars_in_tuple.begin(), vars_in_tuple.end());
+
+        std::set_difference(vars_set.begin(), vars_set.end(),
+            vars_in_tuple_set.begin(), vars_in_tuple_set.end(),
+            back_inserter(unrestricted));
+        return unrestricted;
+    }
+
+    auto propagate_using_smart_str(const vector<IntegerVariableID> & selectors, const vector<IntegerVariableID> & vars,
```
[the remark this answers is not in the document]
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/constraints/smart_table.cc
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/constraints/smart_table.cc
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/constraints/smart_table.cc
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/constraints/smart_table.cc
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/constraints/smart_table.cc
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/constraints/smart_table.cc
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/constraints/smart_table.cc
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/constraints/smart_table.cc
```diff
@@ -0,0 +1,642 @@
...
+        vector<ProofFlag> pb_selectors;
+
+        for (unsigned int i = 0; i < _tuples.size(); ++i) {
+            pb_selectors.emplace_back(propagators.create_proof_flag("t" + std::to_string(i)));
+        }
+        WeightedPseudoBooleanTerms sum_pb_selectors{};
+
+        for (const auto & s : pb_selectors) {
+            WeightedPseudoBooleanTerm t{1_i, s};
+            sum_pb_selectors.emplace_back(t); // No idea why, but this errors if I put the initialiser list directly in emplace_back()
```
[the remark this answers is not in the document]
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/constraints/smart_table.cc
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/constraints/smart_table.hh
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/smart_entry.hh
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/smart_entry.hh
mmcilree: Fixed in c62e8914eb9e01df2a147f01713ff505d7e905e0

gcs/constraints/smart_table.cc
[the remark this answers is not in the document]
ciaranm: Ah, I didn't mean the using declaration, I meant avoiding the need to look up variables in a hash, rather than knowing where they are directly.

CMakeLists.txt
```diff
@@ -7,11 +7,13 @@ add_compile_options(-W)
 add_compile_options(-Wall)
 add_compile_options(-Wno-restrict) # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104336
 add_compile_options(-Wno-stringop-overread) # ditto
-add_compile_options(-O3)
+#add_compile_options(-O3)
```
ciaranm, line 10: Is `-O3` causing problems, or was this for debugging? In general it might be handy if we can keep this sort of thing separated out from feature commits.

mmcilree: Ah whoops yes sorry this was just for debugging

PR #11, opened 2023-02-13
Smart table improvements

mmcilree:
As discussed in last week's meeting and in emails, the current proof logging for Smart Table is not enough, for some specific examples.

I've added random_table.cc that randomly generates small smart tables and tests them, and for all of them, the proof works without the extra logging. However, it's possible to construct plenty of problems where more inference is needed or else the proof fails. So I've added conditional subproofs logging all the inferences for binary entries when the trees are filtered. Very much hoping this is definitely enough now, although proving it formally, and having a condition to differentiate the "boring" cases where it's not necessary remains an open question.


examples/lex_table/lex_table.cc
```diff
@@ -71,7 +71,7 @@ auto main(int, char *[]) -> int
                      << endl;
                 return true;
             }},
-        ProofOptions{"lex_table.opb", "lex_table.veripb"});
```
ciaranm, line 74: I think you changed this in the wrong file?

gcs/constraints/smart_table.cc
```diff
@@ -90,31 +81,62 @@ namespace
 
                 switch (binary_entry.constraint_type) {
                 case ConstraintType::LESS_THAN:
```
ciaranm, line 83: clang-format? The spacing breaks my brainparser.

gcs/constraints/smart_table_test.cc
```diff
@@ -123,9 +123,9 @@ auto run_lex_test(int length, vector<pair<int, int>> ranges, bool reverse = fals
                 lex_violated = lex_violated || (reverse ? (! check_lex(y_sols, x_sols, or_equal)) : (! check_lex(x_sols, y_sols, or_equal)));
                 return true;
             }},
```
ciaranm, line 125: Should probably use a unique filename here too.

gcs/constraints/smart_table.cc
```diff
@@ -50,32 +54,20 @@ namespace
...
-    //                cout << &us.var;
-    //            }}
-    //            .visit(edge);
-    //    }
-    // // ---
-
-    auto filter_edge(const SmartEntry & edge, VariableDomainMap & supported_by_tree) -> void
+    auto log_filtering_inference(const ProofFlag & tuple_selector, const Literal & lit, State & state)
+    {
+        state.infer(TrueLiteral{}, JustifyExplicitly{[&](Proof & proof, vector<ProofLine> &) -> void {
```
ciaranm, line 59: Compiler warning, should check state.infer's return value and throw an exception if it's not Unchanged.

examples/random_smart_tables/random_smart_tables.cc
```diff
@@ -0,0 +1,269 @@
...
+            if (count_in_prufer[v2] == 0) {
+                count_in_prufer[v2]--;
+                count_in_prufer[v1]--;
+                edges.emplace_back(v2 + offset, v1 + offset);
+                break;
+            }
+        }
+    }
+    auto v1 = -1;
+    int v2;
```
ciaranm, line 63: Initialise because the compiler isn't smart enough to realise it will always be set at some point?

gcs/constraints/smart_table.cc
```diff
@@ -50,32 +54,20 @@ namespace
...
+            proof.need_proof_variable(lit);
+            stringstream proof_step;
+            proof_step << "u" << proof.trail_variables(state, 1_i);
+            proof_step << " 1 " << proof.proof_variable(! tuple_selector);
+            proof_step << " 1 " << proof.proof_variable(lit);
+            proof_step << " >= 1 ;\n";
```
ciaranm, line 65: No need for a newline here, emit_proof_line does one.

gcs/constraints/smart_table.cc
```diff
@@ -90,31 +82,62 @@ namespace
 
                 switch (binary_entry.constraint_type) {
                 case ConstraintType::LESS_THAN:
+                    if (state.maybe_proof()) {
```
ciaranm, line 85: OK, so for this and all the other calls to log_filtering_inference: I think you are outputting proof steps even if no filtering ends up being done? I tried the following dumb instance:

```
    auto x = p.create_integer_variable(-1_i, 3_i, "x");
    auto z = p.create_integer_variable(-1_i, 3_i, "z");
    auto y = p.create_integer_variable(-1_i, 3_i, "y");
    auto tuples = SmartTuples{ {NotEqualsVar{x, y}} };
    p.post(SmartTable{{x, y}, tuples});
    p.post(Equals{x, 3_c});
```

which I would expect to log one inference, once, and then not do anything again. But it looks like it's logging `u guesses 1 ~flag0_t0 1 ~iv2_eq_3 >= 1 ;` every time the propagator is called, even if no inference is done?

ciaranm: I guess this might be why you have massive proofs?

mmcilree: That's a good point. I think I've fixed it now

mmcilree: Proof sizes are still quite large but that makes a significant improvement. For largest smart table on 6 variables reduced to 96,020 lines from 136,314

PR #12, opened 2023-02-21
Regular improvements

mmcilree:
Have made the regular propagator incremental, as described in the paper, and tidied the proof logging up a bit. For this I needed a way for propagators to access some sort of restorable persistent data structures - in this case the graph and associated edge degrees etc - so I've made innards/constraint_state.hh to handle this. Probably there's a better way of doing this, I'm not sure. There's a lot of copying going on every time a new guess is made.


ciaranm:
How about `std::any` rather than `std::variant` for holding the state data?

ciaranm:
The way I decide is something like:
- If you're dealing with choices where the class itself deals with most of the dependent behaviour, use subclassing.
- Otherwise, if you're dealing with a fixed number of choices, where every user of the API has to know how to deal with every choice, use `std::variant` (or possibly an `enum class`). The difference here is that all the dependent behaviour is moved out of the class, and into the code that uses it.
- If you want to be able to store, copy, and pass something around without knowing what it is, use `std::any`, but be careful.
- Don't use unions, except to implement `std::variant`.
Usually `std::any` is a bit of a pain because in principle, code that's given a `std::any` is supposed to be able to deal with getting any type back. But in this case, you're relying upon `State` behaving itself and always giving back the thing it was given.

gcs/innards/constraint_state.hh
```diff
@@ -0,0 +1,37 @@
...
+#include <gcs/integer.hh>
+#include <set>
+#include <tuple>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+#include <any>
+
+using namespace gcs;
+struct RegularGraph
```
ciaranm, line 12: The nice thing about `std::any` is that this can now go in `regular.cc`, and be completely invisible to anyone who isn't specifically looking at your constraint.

mmcilree: I guess in that case I wouldn't really need `constraint_state.hh` at all then?

ciaranm: Yeah, it's probably fine just as a `using` in `state.hh`.

gcs/constraints/regular.hh
```diff
@@ -22,16 +22,21 @@ namespace gcs
...
 
     public:
         explicit Regular(std::vector<IntegerVariableID> vars,
             std::vector<Integer> symbols,
             long num_states,
-            std::vector<std::vector<long>> transitions,
+            std::vector<std::unordered_map<Integer, long>> transitions,
             std::vector<long> final_states);
 
+        Regular(std::vector<IntegerVariableID> v,
```
ciaranm, line 35: `explicit`? Doesn't really matter for multi-argument constructors but I try to get into the habit of using `explicit` by default unless I have a good reason not to.

gcs/innards/state.cc
```diff
@@ -1232,6 +1236,18 @@ auto State::infer_on_objective_variable_before_propagation() -> Inference
         return Inference::NoChange;
 }
 
+auto innards::State::add_constraint_state(const ConstraintState c) -> unsigned long
+{
+    // Copying because I'm not sure if making it a reference is a bad idea... (want it to persist)
```
ciaranm, line 1241: Yes, copying is correct.

gcs/innards/state.hh
```diff
@@ -516,6 +519,9 @@ namespace gcs::innards
         [[nodiscard]] auto current() -> CurrentState;
 
         ///@}
+
```
ciaranm, line 522: Is it worth having a very cheap wrapper struct rather than using `unsigned long` directly? I kind of dislike having naked integers floating around in an API because it's a pain to find and change them later on.

gcs/innards/state.cc
ciaranm: Maybe one day in the future we can switch this to use COWs, but I don't think that's a priority.

mmcilree: What are COWs? Aside from the things that eat grass and make milk (just googled it and felt silly)

ciaranm: Copy On Write. We do it in `State` to avoid copying a `std::set` on every branch if it's not actually going to be changed.

gcs/innards/state.hh
mmcilree: As in, something like:
```c++
struct DFAStateID { 
     unsigned long raw_value;
}
```

and then just accessing `.raw_value` everywhere?

Although, sometimes the long is representing a state ID and sometimes it's representing a position in the sequence of variables... not sure what's best to do in the latter case?

Or I could replace everything with `gcs::Integer`?

ciaranm: I think this comment might have moved slightly after some commits, but it refers to the return value of `add_constraint_state` and the value passed to `get_constraint_state`. I feel like these should be a `struct ConstraintStateHandle { unsigned long index };` just so the user can't easily pass in something else as an argument.

mmcilree: Ahh I see. Okay yes that's much less painful to change

mmcilree:
Can I go ahead and merge? I'd like to mess about with some things for the graph data structures but probably better to branch off main than branch off this branch.

ciaranm:
Yep, go for it.

PR #46, opened 2024-01-09
This is a clone of the Circuit SCC Branch

mmcilree:
Feel free to review and/or merge.


examples/circuit_random/circuit_random.cc
```diff
@@ -0,0 +1,198 @@
+#include <gcs/problem.hh>
```
ciaranm, line 1: These includes are in a bit of a strange order.

examples/circuit_random/circuit_random.cc
```diff
@@ -0,0 +1,198 @@
...
+    //            }
+
+    for (int loc1 = 0; loc1 < n; loc1++) {
+        for (int loc2 = 0; loc2 < n; loc2++) {
+            if (distances[loc1][loc2] < 0) {
+                p.post(NotEquals{x[loc1], ConstantIntegerVariableID{Integer{loc2}}});
+            }
+        }
+    }
+    SCCOptions options = {false, true, false, false, false, true};
```
ciaranm, line 79: I wonder whether it would be helpful to have these as command line options, so we can run a bunch of configurations automatically as part of the tests?

examples/circuit_random/circuit_random.cc
```diff
@@ -0,0 +1,198 @@
...
+
+        for (int n = smallest_n; n <= largest_n; ++n) {
+            for (int r = 0; r < repetitions; ++r) {
+                auto [distances, seed] = generate_random_graph(n, EDGE_PROBABILITY);
+
+                Stats stats = run_circuit_problem(n, distances);
+
+                cout << "Num solutions: " << stats.solutions << endl;
+
+                if (0 != system("veripb circuit_random.opb circuit_random.veripb")) {
```
ciaranm, line 187: I'd leave this out, in favour of using the CMake test runner to check the proofs.

gcs/constraints/circuit/circuit.cc
```diff
@@ -0,0 +1,288 @@
+#include <gcs/constraints/all_different.hh>
+#include <gcs/constraints/circuit/circuit.hh>
+#include <gcs/innards/propagators.hh>
+
+#include <iostream>
```
ciaranm, line 5: sus

gcs/constraints/circuit/circuit.cc
```diff
@@ -0,0 +1,288 @@
...
+#include <util/enumerate.hh>
+#include <utility>
+
+using namespace gcs;
+using namespace gcs::innards;
+
+using std::cmp_equal;
+using std::cmp_less;
+using std::cmp_not_equal;
+using std::cout;
```
ciaranm, line 19: sus again

gcs/constraints/circuit/circuit.cc
```diff
@@ -0,0 +1,288 @@
...
+    while (! known_ends.empty()) {
+        auto i = known_ends.back();
+        known_ends.pop_back();
+        auto length = chain_lengths.back();
+        chain_lengths.pop_back();
+        if (cmp_less(length, succ.size() - 1)) {
+            increase_inference_to(result,
+                state.infer(succ[end[i]] != Integer{i}, JustifyExplicitly{[&](Proof & proof) {
+                    output_cycle_to_proof(succ, i, length, pos_var_data, state, proof, Integer{end[i]}, Integer{i});
+                }}));
```
ciaranm, line 140: Should probably check for contradiction and return immediately after each inference. The API's a bit icky here, but after you do something that infers contradiction, State might end up in an invalid state. Perhaps this should be an exception, which is how some other solvers do it, except that's slow.

gcs/constraints/circuit/circuit.cc
```diff
@@ -0,0 +1,288 @@
...
+        }
+        else {
+            increase_inference_to(result, state.infer(succ[end[i]] == Integer{i}, JustifyUsingRUP{}));
+        }
+    }
+    return result;
+}
+
+auto gcs::propagate_non_gac_alldifferent(const ConstraintStateHandle & unassigned_handle, State & state) -> innards::Inference
```
ciaranm, line 149: Should we split the AllDifferent code up, and have both versions available as constraints?

gcs/constraints/circuit/circuit.cc
```diff
@@ -0,0 +1,288 @@
...
+{
+    auto & unassigned = any_cast<list<IntegerVariableID> &>(state.get_constraint_state(unassigned_handle));
+
+    auto result = Inference::NoChange;
+    vector<pair<IntegerVariableID, Integer>> to_propagate;
+    {
+        // Collect any newly assigned values
+        auto i = unassigned.begin();
```
ciaranm, line 157: I'd write this as a for loop with no increment condition inside the `for(;;)`. Possibly more readable, or at least idiomatic?

gcs/constraints/circuit/circuit.hh
```diff
@@ -0,0 +1,130 @@
+
```
ciaranm, line 1: Spaaaaaaaaaaaaace

gcs/constraints/circuit/circuit.hh
```diff
@@ -0,0 +1,130 @@
...
+#include <gcs/innards/state.hh>
+#include <gcs/variable_id.hh>
+#include <list>
+#include <map>
+#include <optional>
+#include <vector>
+
+namespace gcs
+{
+    struct ProofFlagData
```
ciaranm, line 19: These should be in a sub namespace, not in `gcs::`. Possibly `gcs::circuit::innards::` to avoid possible clashes with other constraints, and to make it really clear to users they shouldn't be touching it?

gcs/constraints/circuit/circuit.hh
```diff
@@ -0,0 +1,130 @@
...
+    protected:
+        const bool _gac_all_different;
+        const std::vector<IntegerVariableID> _succ;
+        virtual auto set_up(innards::Propagators &, innards::State &) -> PosVarDataMap;
+
+    public:
+        explicit CircuitBase(std::vector<IntegerVariableID> var, bool gac_all_different = false);
+        [[nodiscard]] auto clone() const -> std::unique_ptr<Constraint> override = 0;
+        auto describe_for_proof() -> std::string override;
+        //        virtual auto install(innards::Propagators &, innards::State &) && -> void = 0;
```
ciaranm, line 85: No need to keep this commented line in I think

gcs/constraints/circuit/circuit.hh
```diff
@@ -0,0 +1,130 @@
...
+        [[nodiscard]] auto clone() const -> std::unique_ptr<Constraint> override;
+        auto install(innards::Propagators &, innards::State &) && -> void override;
+    };
+    
+    using Circuit = CircuitSCC;
+
+    auto propagate_non_gac_alldifferent(
+        const innards::ConstraintStateHandle & unassigned_handle, innards::State & state) -> innards::Inference;
+
+    auto output_cycle_to_proof(const std::vector<IntegerVariableID> & succ,
```
ciaranm, line 116: These should also be namespaced off.

gcs/constraints/circuit/circuit.hh
```diff
@@ -0,0 +1,130 @@
...
+    };
+
+    /**
+     * \brief Circuit constraint: requires the variables, representing graph nodes, take values
+     * such that each variable's value represents the index of the next node in a tour that visits
+     * all variables.
+     *
+     * \ingroup Constraints
+     */
+    class CircuitBase : public Constraint
```
ciaranm, line 74: Maybe more header files would be useful? So `CircuitBase`, `CircuitPrevent`, etc get their own headers, and `circuit.hh` just includes them and has the final `using` declaration? I find it less confusing if there's a direct correspondence between the `cc` and `hh` files. Also some of our source files are probably getting a bit long and this might help break them up.

gcs/constraints/circuit/circuit_scc.cc
```diff
@@ -0,0 +1,1250 @@
+#include <gcs/constraints/circuit/circuit.hh>
+#include <gcs/innards/propagators.hh>
+#include <iostream>
```
ciaranm, line 3: Sus

gcs/constraints/circuit/circuit_scc.cc
```diff
@@ -0,0 +1,1250 @@
...
+#include <string>
+#include <utility>
+#include <variant>
+#include <vector>
+
+using std::cmp_equal;
+using std::cmp_less;
+using std::cmp_less_equal;
+using std::cmp_not_equal;
+using std::cout;
```
ciaranm, line 17: More sus

gcs/constraints/circuit/circuit_scc.cc
```diff
@@ -0,0 +1,1250 @@
...
+
+    struct OrderingAssumption
+    {
+        ProofFlag assumption_flag;
+        long first;
+        long middle;
+        long last;
+    };
+
+    auto select_root(long n) -> long
```
ciaranm, line 55: This gives an unused variable compiler warning. Either tag it as `[[unused]]` or don't name it.

gcs/constraints/circuit/circuit_scc.cc
```diff
@@ -0,0 +1,1250 @@
...
+    {
+        PosVarDataMap & pos_var_data;
+        ConstraintStateHandle proof_flag_data_handle;
+        ConstraintStateHandle pos_alldiff_data_handle;
+    };
+
+    struct PLine
+    {
+        // Represents a pol line in the proof that we can add terms to.
+        // Maybe this could be generalised (e.g. to other operations) and live in proof.cc?
```
ciaranm, line 96: Yes please (although `proof.cc` is also too big, and should probably be its own directory rather than living directly in innards)

gcs/constraints/circuit/circuit_scc.cc
```diff
@@ -0,0 +1,1250 @@
...
+        PosVarDataMap pos_var_data, bool using_shifted_pos,
+        const SCCOptions & options, State & state) -> ProofLine
+    {
+        // Prove that at most one (shift)pos[node] == v is true for v in values
+        ProofLine am1line;
+
+        state.infer_true(JustifyExplicitly{[&](Proof & proof) {
+            stringstream proofline;
+
+            // We should document properly why this works somewhere now that it's in more than one place
```
ciaranm, line 231: Yes, this feels like another candidate for our improved proofs API

gcs/constraints/circuit/circuit_scc.cc
```diff
@@ -0,0 +1,1250 @@
...
+            create_flag_for_greater_than(next_node, middle, flag_data[next_node], pos_var_data, pos_alldiff_data,
+                succ, options, state);
+
+            PLine temp_p_line;
+
+            auto & shifted_pos_geq = flag_data[root].shifted_pos_geq;
+            auto & shifted_pos_eq = flag_data[root].shifted_pos_eq;
+
+            state.infer_true(JustifyExplicitly{[&](Proof & proof) {
+                // TODO: All these should be ProofLevel::Temporary, but unfortunately that fails in specific instances
```
ciaranm, line 639: Did you ever figure that out?

gcs/innards/proof.cc
```diff
@@ -316,11 +316,11 @@ auto Proof::set_up_integer_variable(SimpleIntegerVariableID id, Integer lower, I
     _imp->solution_variables.push_back(id);
 }
 
-auto Proof::create_proof_flag(const string &) -> ProofFlag
+auto Proof::create_proof_flag(const string & n) -> ProofFlag
 {
     ProofFlag result{_imp->flags.size() / 2, true};
 
-    string name = xify("f" + to_string(result.index)); // + "_" + n);
+    string name = xify("f" + to_string(result.index) + "_" + n);
```
ciaranm, line 323: Hrm, we should decide what to do about this.

gcs/innards/proof.cc
```diff
@@ -1178,24 +1178,56 @@ auto Proof::emit_red_proof_line(const SumLessEqual<Weighted<PseudoBooleanTerm>>
...
 
-auto Proof::emit_red_proof_lines_reifying(const SumLessEqual<Weighted<PseudoBooleanTerm>> & ineq, ProofFlag reif,
-    ProofLevel level) -> pair<ProofLine, ProofLine>
+auto Proof::emit_subproofs(const map<string, JustifyExplicitly> & subproofs)
 {
-    need_all_proof_names_in(ineq.lhs);
+    _imp->proof << " begin\n";
+    ++_imp->proof_line;
+    for (const auto & [proofgoal, proof] : subproofs) {
+        ++_imp->proof_line;
```
ciaranm, line 1186: This looks fragile and prone to going horribly wrong. Perhaps it's better to pass in a lambda function that gets called, and deal with constraint IDs that way? It's probably useful for the subproof-writing code to have access to constraint IDs.

gcs/innards/proof.hh
```diff
@@ -341,7 +339,23 @@ namespace gcs::innards
             const std::vector<std::pair<ProofLiteralOrFlag, ProofLiteralOrFlag>> & witness, ProofLevel level) -> ProofLine;
 
         /**
-         * Emit a pair of RED proofs step for the specified expression, reified on the specified flag.
+         * Emit subproofs, using a map of proofgoals to explicit justifications
+         */
+        auto emit_subproofs(const std::map<std::string, JustifyExplicitly> & subproofs) -> auto;
```
ciaranm, line 344: Should this be public?

gcs/innards/proof.hh
```diff
@@ -365,7 +379,8 @@ namespace gcs::innards
          * \name Helpful things for making proof text.
          */
         ///@{
-
+        [[nodiscard]] auto proof_name(const VariableConditionFrom<SimpleOrProofOnlyIntegerVariableID> &) const -> const std::string &;
```
ciaranm, line 382: Hrm, I'm not sure whether making these public again is a good idea. I guess we could do this for now and then I'll tidy up the API again later?

gcs/innards/state.hh
```diff
@@ -535,8 +535,36 @@ namespace gcs::innards
...
+         * Store a given std::any value as a constraint state that is accessible via
+         * the returned handle and restores on backtrack.
+         */
         [[nodiscard]] auto add_constraint_state(const ConstraintState c) -> ConstraintStateHandle;
+
+        /**
+         * Store a given std::any value as a constraint state that is accessible via
+         * the returned handle and does not restore on backtrack.
+         */
+        [[nodiscard]] auto add_persistent_constraint_state(const ConstraintState c) -> ConstraintStateHandle;
```
ciaranm, line 553: What exactly is the need for this? In other constraints I've been moving the appropriate data structure into the lambda, so it's stored there instead. Does this get too ugly?

gcs/innards/state.hh
```diff
@@ -535,8 +535,36 @@ namespace gcs::innards
...
         [[nodiscard]] auto get_constraint_state(const ConstraintStateHandle h) -> ConstraintState &;
+
+        /**
+         * Return the persistent constraint state for the given handle.
+         *
+         */
+        [[nodiscard]] auto get_persistent_constraint_state(const ConstraintStateHandle h) -> ConstraintState &;
+        ///@}
+
+        auto current_domains_debug_strings(std::vector<IntegerVariableID> vars) -> std::vector<std::string>;
```
ciaranm, line 567: Probably should remove

gcs/constraints/circuit.hh
ciaranm, line 1: I think circuit.hh should still exist for consistency, even if it just includes the sub-headers.

gcs/constraints/circuit/circuit.hh
ciaranm: Clarifying, I mean, `constraints/circuit.hh` rather than `constraints/circuit/circuit.hh`

test/break_temporary/break_temporary.cc
ciaranm, line 1: What's this one? Is this the saturation bug example? I don't think that's useful keeping here, if so.

test/cpaior_circuit_experiments/cpaior_circuit_experiments.cc
ciaranm, line 1: I don't think I'd include this in the published branch

test/disconnected/disconnected.cc
ciaranm, line 1: Can we call this circuit_disconnected or similar? Also if this is specific to the circuit branch, probably it should live in the circuit directory, rather than tests?

test/scc_multiple_sccs/scc_multiple_sccs.cc
ciaranm, line 1: Ditto

test/verify_circuit_proofs/verify_circuit_proofs.cc
ciaranm, line 1: I think this belongs elsewhere

ciaranm (commented):
Thoughts so far. Some of this is minor cleanup things, some of it might be about you telling me to go and design a better API for you.

gcs/constraints/circuit/circuit.hh
ciaranm: You could bring `ProofFlag` into `gcs::circuit::innards` with a `using`, I guess, although that's visible for everyone using the header, which isn't great. Maybe the namespace should be `gcs::innards::circuit` rather than `gcs::circuit::innards`?

mmcilree: Ah yes gcs::innards::circuit would make sense

gcs/constraints/circuit/circuit_scc.cc
mmcilree: Yes, the way I was emitting comments was messing with the proof levels. Should be fixed now.

gcs/innards/proof.cc
mmcilree: Going to leave these for now since it will need rewritten anyway with the new proof API?

gcs/innards/state.hh
mmcilree: Was trying to simplify the arguments being passed around and keep the persistent and dynamic data together as handles until they are needed. Maybe not the best but would require quite a bit of refactoring to change

gcs/constraints/circuit/circuit.cc
mmcilree: I can split these up, however, both prevent and non_gac_alldiff need to quickly access a list of unassigned variables which I currently have as a constraint state. If I make non_gac_alldiff it's own class "VCAllDifferent", both the VCAllDifferent constraint and the CircuitPrevent constraint are keeping track of unassigned variables separately which feels silly. Would it be reasonable to keep track of this in State itself and have a get_unassigned_variables method?

mmcilree: Additionally I'm a little bit worried we won't be able to guarantee VCAllDifferent gets propagated before CircuitPrevent if it is a separate constraint, and we rely on this to stop infinite loops...

ciaranm: For finding a constraint's unassigned variables quickly... This feels like it would be useful in a bunch of places. Potentially a constraint could also remove its triggers on certain variables during search. I think this might be a "solve it properly later" problem.

mmcilree: > If you have a look on the `reasons` branch in `gcs/constraints/linear/propagate.hh`, there are functions that can be used to propagate a linear inequality without using a `Constraint` directly. Doing that for alldiff would let you call the propagation function at the start of your propagator, without needing to worry about it being a constraint that might not be called.

That would work, although since we also need the PB encoding would I duplicate the encoding logic both in Circuit and in VCAllDifferent? Or should the encoding function be available somewhere too?

ciaranm: I don't think it would hurt to make the encoding function semi-public.

mmcilree:
Okay @ciaranm that should be all the comments addressed and I've refactored AllDifferent into VCAllDifferent and GACAllDifferent. Might want to sit down at the same time to merge and fix the conflicts as I think we have both made conflicting changes to proof.cc and it needs a bit of untangling.

PR #51, opened 2024-04-10
MiniZinc support for regular.

mmcilree:


ciaranm:
Can you get the nurses test to run automatically as well?

mmcilree:
> Can you get the nurses test to run automatically as well?

Can do but it's relatively slow (like 30s) not sure why...

ciaranm:
> Can do but it's relatively slow (like 30s) not sure why...

Hah, then adding it in might motivate me to benchmark it at some point.

mmcilree:
ok :)

PR #63, opened 2024-11-26
A refactor of reification logic (patched in from multiply branch)

mmcilree:
If I've done this correctly, these should just be the changes from the multiply branch related to reification in the logging API.

If possible could you review/amend this so it can be merged in? It's a more urgent feature than the rest of the multiply branch changes, as I'd like to use the `logger.reify` method in the TU justification code.

Basically the changes are:
- Move reification logic to its own function so it is only done in one place
- Similarly factor out the code that gets a list of Literals from a reason
- Refactor `emit_<rule>_line_under_reason` (while still maintaining the original function signatures). RUP, IA, and A lines should all be emittable with or without reasons, and we should avoid duplicated logic for these since the only differences are the proof rule string, and possibly trailing id lists for IA or annotated RUP. 
- Provide a `logger.reify()` method that allows arbitrary reification of `WeightedPBLessEqual{}` (it returns a new `WeightedPBLessEqual` with correct reification.

### DIFF
--- a/gcs/innards/proofs/proof_logger.cc
+++ b/gcs/innards/proofs/proof_logger.cc
@@ -17,6 +17,7 @@ using std::deque;
 using std::flush;
 using std::fstream;
 using std::ios;
+using std::make_optional;
 using std::map;
 using std::max;
 using std::nullopt;
@@ -35,6 +36,12 @@ using namespace gcs::innards;
 
 namespace
 {
+    // Is this a bad idea?
+    const char * proof_rule_str[] = {
+        "u",
+        "a",
+        "ia"};
+
     [[nodiscard]] auto deview(const VariableConditionFrom<ViewOfIntegerVariableID> & cond) -> VariableConditionFrom<SimpleIntegerVariableID>
     {
         switch (cond.op) {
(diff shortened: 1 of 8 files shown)


ciaranm (commented):
These, plus I think the compile is broken

gcs/innards/proofs/proof_logger.cc
```diff
@@ -387,24 +432,74 @@ auto ProofLogger::emit_rup_proof_line_under_reason(const Reason & reason, const
     variable_constraints_tracker().need_all_proof_names_in(ineq.lhs);
 
 #ifdef GCS_TRACK_ALL_PROPAGATIONS
-    _imp->proof << "* emit rup proof line from " << where.file_name() << ":" << where.line() << " in " << where.function_name() << '\n';
+    _imp->proof << "* emit " << proof_rule_str[rule] << " proof line from " << where.file_name() << ":" << where.line() << " in " << where.function_name() << '\n';
```
ciaranm, line 435: I think this should be `()` not `[]` now?

python/CMakeLists.txt
```diff
@@ -20,4 +20,4 @@ target_link_libraries(gcspy PRIVATE fmt::fmt)
 install(TARGETS gcspy DESTINATION .)
 
 configure_file(${CMAKE_SOURCE_DIR}/python/python_test.py ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/python/python_test.py COPYONLY)
-add_test(NAME python_bindings COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/python/python_test.py)
```
ciaranm, line 23: Should this be in the same commit?

ciaranm (commented):
Nearly there I think

mmcilree:
Sorted now? :-)

PR #64, opened 2024-12-16
More constraint names

ciaranm:
Not sure what I think of this and whether it might benefit from more indirection. But:

```
{
"1": {"type":"variable_lower_bounds","variables":["i0_s_b0","i0_s_b1","i0_s_b2","i0_s_b3"]},
"2": {"type":"variable_upper_bounds","variables":["i0_s_b0","i0_s_b1","i0_s_b2","i0_s_b3"]},
"3": {"type":"variable_lower_bounds","variables":["i1_e_b0","i1_e_b1","i1_e_b2","i1_e_b3"]},
// ...
"17": {"name":"AllDifferent","role":"not equals because lower","type":"constraint_definition"},
"18": {"name":"AllDifferent","role":"not equals because higher","type":"constraint_definition"},
"19": {"name":"AllDifferent","role":"not equals because lower","type":"constraint_definition"},
"20": {"name":"AllDifferent","role":"not equals because higher","type":"constraint_definition"},
// ...
}
```

For example, is it worth trying to group together constraints from the same "instantiation" of, e.g. AllDifferent? We'd have to have a way of giving constraints unique IDs then, which is doable, and possibly have yet another mapping from those IDs to constraint names.

### DIFF
--- a/gcs/constraints/element.cc
+++ b/gcs/constraints/element.cc
@@ -142,7 +142,7 @@ auto ElementConstantArray::install(Propagators & propagators, State & initial_st
     if (optional_model) {
         for (const auto & [idx, v] : enumerate(*_vals))
             if (initial_state.in_domain(_idx, Integer(idx)))
-                optional_model->add_constraint({_idx != Integer(idx), _var == v});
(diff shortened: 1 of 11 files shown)


ciaranm:
Let's do this a bit differently

Issue #66, opened 2025-01-13
`Filename too long` on Windows makes `pip install` fail


ThomSerg:
The `range-v3` dependency has a submodule called `gh-pages` under the `/docs` directory. It gets recursively cloned upon build during `pip install gcspy`. This causes the pip install to fail on Windows, since `gh-pages` submodule contains image files with filenames which are too large for Windows (without making a registry edit). E.g.:

```
CUSTOMBUILD : error : unable to create file 
structstd_1_1tuple__element_3_01I_00_01_1_1ranges_1_1compressed__tuple__detail_1_1compressed__tu8fb1bf77fd01ee20b7f8b0fad28a83c7.html: Filename 
too long
```

The following change in `CMakeLists.txt` prevents the troublesome submodule from being cloned:

```
FetchContent_Declare(
        range-v3
        GIT_REPOSITORY https://github.com/ericniebler/range-v3.git
        GIT_TAG a81477931a8aa2ad025c6bda0609f38e09e4d7ec # v0.12.0
        GIT_SUBMODULES "" # Disable submodule installation <- ADDED FIX
)
```

Not sure if this fix causes other issues, since I haven't been able to install from source yet.

ciaranm:
That looks plausible. I'm not a CMake expert, but there might also be a way of getting range-v3 to not bother building docs, tests, etc.

Please can you submit a pull request for this? It'll automatically get checked by the integration tests, and it will give an error if it actually breaks something.

I suspect this won't be enough to get things working on Windows. I had a go a while ago but couldn't figure out how to install Boost on Windows and gave up.

PR #67, opened 2025-01-13
Disable downloading range-v3  gh-pages submodule

ThomSerg:
Causes 'Filename too long' on Windows, see #66.

Test pull request to see if integration tests might fail.


ciaranm:
Thanks, looks like this doesn't obviously break anything

PR #74, opened 2025-07-01
Switch to Matthew's encoding which I hate because now I have to do this

ciaranm:


ciaranm:
Aha, looks like there's at least one situation where this RUP isn't enough. Let me run the tests a few thousand times and see what it is.

ciaranm:
Is this enough? It's passing more tests at least.

Issue #101, opened 2026-04-07
Split up `Constraint::install`


ciaranm:
Currently `Constraint::install` does two things: it defines the proof encoding (usually inside an `if (optional_model)` block), and it sets up the actual propagators. This would be better split in two: a `Constraint::define_proof_model` that takes a `ProofModel` by reference (not pointer) and that is only called if we actually care, and a `Constraint::install_propagators` that (destructively) turns the constraint into actual initialiser and propagation code.

Known complications:

- Some of the constraints pass data such as `ProofLine` from defining the model into the propagators. This will need to be turned into class member data for the constraints, and then copied or moved into the propagation functions as needed.
- Probably in a few cases there will need to be some refactoring just to make this separation clean.
- I haven't thought about how this connects to the s-expr code. I guess that remains an independent, third function.

philrod1:
I had a wee look at this yesterday, and I have had another look today.  Here's where I am.  My first naive approach was to the function split like so:  (I am using Abs for all this, btw)
 
`virtual auto install(innards::Propagators &, innards::State &, innards::ProofModel * const) && -> void override;`

becomes

```
virtual auto define_proof_model(ProofModel & model) -> void override;
virtual auto install_propagators(Propagators & propagators, State & initial_state) && -> void override;
```

Implementing define_proof_model something like this:

```
std::vector<ProofLine> _proof_lines;

auto define_proof_model(ProofModel & model) -> void override
{
    _proof_lines.push_back(model.add_constraint("Abs", "non-negative", ...));
    _proof_lines.push_back(model.add_constraint("Abs", "negative", ...));
}
```

That bit is easy enough, assuming that you are happy with each constraint having its own std::vector<ProofLine> _proof_lines

However, inside the old install function we have code like this:

`propagators.trim_lower_bound(initial_state, optional_model, _v2, 0_i, "Abs");`

leading to code like this where ProofLines are discarded, or at least never make their way back to the Constraint:

```
            if (optional_model)
                optional_model->add_constraint({var >= val});
```

Is the goal to have ::define_proof_model() generate ALL ProofLines before calling ::install_propogators() and not having the propagator generating ProofLines?

From what I can tell, the current flow looks like this:

Abs::install gets ProofModel* 
- passes it to trim_lower_bound(optional_model)
  - trim_lower_bound calls optional_model->add_constraint() 
  - generates a ProofLine (but doesn't return it)
- passes it to optional_model->add_constraint() directly
[...]

PR #102, opened 2026-04-08
Better reification

ciaranm:


ciaranm:
I think this is ready to go, unless you can spot anything obviously silly?

Later, it would be good to split up some of the duplicated logic even further, so that to implement a fully reified constraint we could just define the "definitely true", "definitely false", and "undecided" cases and not have the same logic repeated everywhere. However, that should probably wait until after issue:#101.

philrod1 (approved):
I mean, yeah.  The code looks good and the test run.  If the code does what you intended for it to do ... no idea.  As far as I can tell, from looking at the code and running the tests, everything is fine.  Matthew would be much better qualified to know if the intent is correct or not.

Edit: This will definitely clash with the formally_verified_encodings branch, but I can pull your changes into that and fix the issues in-branch.  The branch I have for issue:#101 has nothing of consequence on it yet, and I'm happy to hold off until this is merged.

PR #106, opened 2026-04-28
Split up install

philrod1:
I _think_ I have all the constraints' install functions split into three parts: prepare(), define_proof_model(), and install_propagators().  The new install() functions all do exactly the same thing:

```
auto Abs::install(Propagators & propagators, State & initial_state, ProofModel * const optional_model) && -> void
{
    if (! prepare(propagators, initial_state, optional_model))
        return;

    if (optional_model)
        define_proof_model(*optional_model);

    install_propagators(propagators);
}
```

So much so that the function could be implemented once in constraint.cc.  That would be nice.  They also retain the previous install() signature.

Most constraints were willing, but some needed to be persuaded to adapt to the decoupling.  MultBC and SmartTable in particular bested me.  They fit the pattern from an API point of view, but they are not cleanly decoupled at all.


ciaranm:
Hi Phil — we've rebased this onto current main as #114 with a handful of fixes and a narrower scope.

Highlights:
- VCAllDifferent had a `define_proof_model` bug (used moved-from `_vars`); fixed.
- Carried over the `Reason → HalfReifyOnConjunctionOf` change from 218b76d that the merge here had silently dropped.
- Dropped the splits for the delegators / tightly-coupled cases (GACArithmetic, AtMostOneSmartTable, LexSmartTable, Table, In, MultBC, SmartTable, both Circuits) — agreed it's better to leave them un-split until their helpers can be split too, rather than have `prepare()` do everything and return false.
- Style fixes: `_` prefix on new private members, header default-init for `_evaluated_cond`, alphabetical includes, etc.
- Dropped the `test/capture_encodings.py` per-arg-set change since separate ctest entries on `proper-lex-propagator` make it unnecessary.

All your authorship is preserved on the per-constraint commits in #114. Full breakdown of what's lost vs this PR is in the #114 description.

Happy to close this in favour of #114 once you're satisfied — or feel free to push more on top of #114 directly.

ciaranm:
And following up as me rather than Claude: for the constraints that don't fit nicely, some careful thinking and restructuring is in order. The important thing here is we have a bunch of nice easy constraints that will let us start working on the "replace line numbers with labels" task, and maybe during doing that we'll figure out how the odd cases should be designed.

PR #112, opened 2026-04-30
Fix file I/O error handling in proof output (issue #111)

### LINKED ISSUE #111: Check proof exists before using in ProofLogger
I came across this when doing unrelated code review:

```
auto ProofLogger::start_proof(const ProofModel & model) -> void
{
    _imp->proof.open(_imp->proof_file, ios::out);

    _imp->proof << "pseudo-Boolean proof version 3.0\n";

    _imp->proof << "f " << model.number_of_constraints() << " ;\n";
    _imp->proof_line.number += model.number_of_constraints().number;

    if (! _imp->proof)
        throw ProofError{"Error writing proof file to "
                         " + _imp->proof_file + "
                         ""};
}
```

The guard and exception should be at the start of the function.  Also, the formatting of the error message is off, which is what caught my eye in the first place.

ciaranm:
## Summary

- `ProofLogger::start_proof` had a broken error message — the intended string concatenation `" + _imp->proof_file + "` was accidentally quoted, so the filename never appeared in the exception text
- File I/O failures were detected with manual `if (!stream)` checks that missed write errors between open and check, and in some cases missed the open failure itself
- `NamesAndIDsTracker` wrote its closing `"\n}\n"` in the destructor, making it impossible to propagate a write failure as a `ProofError`

All file I/O in the proof output path (`proof_logger.cc`, `proof_model.cc`, `names_and_ids_tracker.cc`, `solve.cc`) now uses `stream.exceptions(ios::failbit | ios::badbit)` before `open()`, with each I/O section wrapped in `try/catch(const ios_base::failure &)` that rethrows as `ProofError` with the filename included.

The closing write in `NamesAndIDsTracker` is moved out of the destructor into a new `finalise()` method (matching `ProofModel::finalise()`), called explicitly at the end of `solve_with()`. To catch any future code that forgets to call `finalise()`, both `ProofModel` and `NamesAndIDsTracker` now set a `bool finalised` flag in their `finalise()` methods and abort with a diagnostic message in their destructors if the flag is unset — guarded by `std::uncaught_exceptions() == 0` so it does not fire during stack unwinding from an unrelated exception.

## Test plan

- [ ] CI passes (sanitize build exercises proof output paths with ASan + UBSan)
- [ ] `reification_test` passes (directly constructs `NamesAndIDsTracker` and `ProofModel`, now calls `tracker.finalise()` explicitly)
- [ ] `tutorial_proof` passes (end-to-end proof generation and VeriPB verification)

Closes #111

### DIFF
--- a/gcs/innards/proofs/proof_model.cc
+++ b/gcs/innards/proofs/proof_model.cc
@@ -10,7 +10,9 @@
 #include <gcs/innards/proofs/simplify_literal.hh>
 
 #include <algorithm>
+#include <cstdlib>
 #include <deque>
+#include <exception>
 #include <fstream>
 #include <iterator>
 #include <map>
@@ -26,6 +28,8 @@ using namespace gcs::innards;
 
 using std::deque;
 using std::fstream;
+using std::ios;
+using std::ios_base;
 using std::istreambuf_iterator;
 using std::map;
 using std::nullopt;
@@ -59,6 +63,7 @@ struct ProofModel::Imp
     stringstream opb;
 
     bool always_use_full_encoding = false;
+    bool finalised = false;
 
     explicit Imp(NamesAndIDsTracker & t) :
         tracker(t)
@@ -73,7 +78,13 @@ ProofModel::ProofModel(const ProofOptions & proof_options, NamesAndIDsTracker & 
     _imp->always_use_full_encoding = proof_options.always_use_full_encoding;
 }
 
-ProofModel::~ProofModel() = default;
+ProofModel::~ProofModel()
+{
+    if (! _imp->finalised && std::uncaught_exceptions() == 0) {
+        fmt::print(stderr, "ProofModel destroyed without calling finalise()\n");
+        std::abort();
+    }
+}
 
 auto ProofModel::advance_constraint_counter() -> ProofLineNumber
 {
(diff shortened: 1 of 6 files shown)


gcs/innards/proofs/proof_model.cc
```diff
@@ -73,7 +78,13 @@ ProofModel::ProofModel(const ProofOptions & proof_options, NamesAndIDsTracker &
     _imp->always_use_full_encoding = proof_options.always_use_full_encoding;
 }
 
-ProofModel::~ProofModel() = default;
+ProofModel::~ProofModel()
+{
+    if (! _imp->finalised && std::uncaught_exceptions() == 0) {
```
philrod1, line 83: Oh, that's clever!

philrod1: Well that issue took an unexpected turn.  Did Claude do all this, or did you need to intervene?

ciaranm: It took a few iterations. It fixed the original issue but without the finalise logic, and using some ugly code in the destructor instead. Then I told it to identify other places where files are used and to suggest a similar fix, which it did. Then I told it to use a finalise() instead of the ugly destructor code. Finally, I told it to add in code to check when finalise() wasn't called in case we forgot about it when making future changes. So, I'd say it was fairly heavily assisted in the "what good design looks like" side, but on the "using C++" side it didn't need any help, including remembering how C++ is weird with exceptions inside exceptions and exceptions inside destructors.

philrod1: "exceptions inside destructors"
That sounds nasty.  I/O is hard.

philrod1: What happens if Claude changes a bunch of stuff and you don't like it?  Can you undo?

ciaranm: The wording I'm using is "suggest changes", which prompts on each change (similar to git add -p) and I can reply "No, instead do X" on changes. I've also told it not to commit anything without explicit instructions, so I can review there. All it has is access to a checkout of the git repository and the ability to read from GitHub, and it needs permission for anything else.

PR #121, opened 2026-05-06
Move table install logic into Table constraint

philrod1:
This code removes table-specific code from propagators and puts it into table.cc.  It uses that to do a proper split of prepare -> define proof model -> install propagators.  With  that done, we can use a table to model the `In` and `GACArithmetic` constraints.


ciaranm (changes requested):
Just the commented out bit in the header

ciaranm:
I think that failing test is fixed on the branch I'm about to merge. Don't think that's you.

PR #129, opened 2026-05-07
Rewrite NegativeTable with watched literals

### LINKED ISSUE #117: Better NegativeTable constraint implementation
Replace the current quick-and-dirty `NegativeTable` implementation with a clean, from-scratch, fully-featured, efficient implementation.

ciaranm:
## Summary
- **Commit 1:** moves Table and NegativeTable into a `gcs/constraints/table/` family directory mirroring `all_different/`. Pure restructure, no behavioural changes.
- **Commit 2:** rewrites `NegativeTable` with 2-watched-literals. Each tuple has two watch positions; each propagator fire just rechecks those. Replacement search only when a watch breaks (a variable is instantiated to its tuple value). Same propagation strength as before (unit propagation per clause), but O(forbidden_tuples) per fire in the common case instead of O(forbidden_tuples × n_vars).
- Wildcards fall out for free — `var == Wildcard` returns `TrueLiteral` via the existing operator overloads, which tests as `DefinitelyTrue`, so wildcard positions naturally fail the watch-validity check.
- Watches are non-backtrackable (`shared_ptr`); leaving them moved across backtracks is sound and skips restoration overhead.
- Root-level contradictions (e.g. all-wildcard tuple) and unit-at-init propagations are handled in `install_initialiser`.

Tests expanded: two- and three-variable cases, larger forbidden tables, dedicated wildcard tests, root-level contradictions (all-forbidden + all-wildcard), and a unit-at-init case.

Closes #117.

## Test plan
- [x] `ctest -j 32` — 161/161 pass.
- [x] Proof verification on every NegativeTable test case (including new ones) under VeriPB.
- [x] Confirmed sanitize-build failures in `smart_table`/`count`/`at_most_one` are pre-existing on `main` (libstdc++ debug-mode `vector::operator[]` assertions, unrelated to this work).

### DIFF
--- a/gcs/constraints/table/negative_table.cc
+++ b/gcs/constraints/table/negative_table.cc
@@ -9,6 +9,9 @@
 
 #include <util/enumerate.hh>
 
+#include <cstddef>
+#include <limits>
+#include <memory>
 #include <optional>
 #include <sstream>
 #include <utility>
@@ -25,7 +28,12 @@
 using namespace gcs;
 using namespace gcs::innards;
 
+using std::make_shared;
+using std::nullopt;
 using std::optional;
+using std::pair;
+using std::shared_ptr;
+using std::size_t;
 using std::string;
 using std::stringstream;
 using std::unique_ptr;
@@ -69,21 +77,7 @@ namespace
     {
         return visit([](auto v) { return tuple_entry_as_string(v); }, v);
     }
-}
 
-NegativeTable::NegativeTable(vector<IntegerVariableID> v, ExtensionalTuples t) :
-    _vars(move(v)),
-    _tuples(move(t))
-{
-}
-
-auto NegativeTable::clone() const -> unique_ptr<Constraint>
-{
-    return make_unique<NegativeTable>(_vars, ExtensionalTuples{_tuples});
-}
-
-namespace
-{
     auto add_literal(Literals & lits, const IntegerVariableID & var, const Integer & val)
     {
         lits.emplace_back(var != val);
@@ -97,10 +91,7 @@ namespace
     {
         visit([&](const auto & val) { add_literal(lits, var, val); }, val);
     }
-}
 
-namespace
-{
     auto operator==(const IntegerVariableID &, const Wildcard &) -> Literal
     {
         return TrueLiteral{};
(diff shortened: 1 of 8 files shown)


gcs/constraints/table/negative_table.hh
```diff
@@ -0,0 +1,37 @@
...
+#include <gcs/extensional.hh>
+#include <gcs/variable_id.hh>
+
+#include <vector>
+
+namespace gcs
+{
+    /**
+     * \brief Constrain that the specified variables are not equal to one of the specified
+     * tuples.
```
mmcilree: Doesn't negative table usually mean the variables are not equal to *all* of the specified tuples?

gcs/constraints/table/negative_table.cc
```diff
@@ -0,0 +1,334 @@
...
+                        // any rescue search will discover the inference is now redundant.
+                        watches->emplace_back(*w1, *w1);
+                    }
+                    else {
+                        watches->emplace_back(*w1, *w2);
+                    }
+                }
+            });
+
+        propagators.install(
```
mmcilree, line 233: Since it looks like CNF TWL is what is implemented then yes I guess the description should say all. Side note, it doesn't look like there's much here that's specific to not-equals literals. Would it be worth asking claude to make a general CNF propagator that can take any literals (and implementing NegativeTable as an instantiation of this)?

mmcilree (approved):
Minor comments.

gcs/constraints/table/negative_table.hh
ciaranm: Good catch — fixed in f264296. Reworded to spell out the semantics directly: "the assignment to the specified variables does not match any of the specified tuples — equivalently, for every tuple t, at least one vars[i] differs from t[i]".

gcs/constraints/table/negative_table.cc
ciaranm: Good observation — agreed there's nothing structurally tied to !=. Filed #131 to track it as a follow-up rather than scope-creeping this PR; the doc is fixed here. The two design points worth thinking about for the generalisation are (a) whether to add a per-variable reverse map (this version iterates all clauses on every fire, fine for short tables but wasteful at scale) and (b) likely consumers in `logical.cc` (Or, Implies) and possibly the per-clause encodings in `at_most_one` and Lex decompositions.

Issue #134, opened 2026-05-07
Simplify internal data structures for variable domains


ciaranm:
Currently we have multiple ways of storing variable domains: a constant value, a lower/upper bounds pair, a bitset, or an interval set. Ideally we would only use the interval set representation, because this would a) simplify a lot of code, and b) make it much easier to start refactoring some of the places that do things like intersecting domains.

The big reason not to do this is that IntervalSet is moderately slow: it does memory allocations thanks to using std::vector. However, this is potentially an avoidable cost, by switching from vector to a vector implementation that does the "small value optimisation".

Rough plan:

- Change State to only use IntervalSet. Remove most of the branching and overloading. Be careful with any of the code that looks to see whether a variable has only a single value by assuming that this is the case iff it is a constant.
- Measure slowdown on representative benchmarks.
- Compare different "small value optimisation" library implementations. Select one and convert IntervalSet to use it.
- Re-measure slowdown on representative benchmarks.
- If the performance is similar, also identify other places where we could use the optimised vector (probably Reasons, HalfReifyOnConjunctionOf, others?)
- Identify places where we can make use of more efficient algorithms to do unions, intersections, etc, now that we know for sure we'll always be working with interval set domains.

ciaranm:
## Implementation plan

### Design decisions

- **Collapse `IntegerVariableState` entirely.** No more `std::variant<Constant, Range, SmallSet, IntervalSet>`. Storage is plain `IntervalSet<Integer>`. `ConstantIntegerVariableID` queries short-circuit at the `deview()` level for every read-only entry point — they never reach `state_of()`.
- **Inline storage, no `shared_ptr`.** `_imp->integer_variable_states` becomes `list<vector<IntervalSet<Integer>>>`. The COW-via-refcount pattern (the `use_count() != 1` checks in `state.cc`) goes away — mutate in place. `new_epoch()` does a real per-variable copy of the interval vector; for the typical "single contiguous range" variable that's two ints. Phase 2 numbers will tell us whether this hurts in practice.
- **`Bits` and `gcs/innards/bits.hh` are deleted.** Confirmed (via grep) that nothing outside the soon-to-be-deleted SmallSet path uses it. `IntegerVariableProofRepresentation::Bits` is unrelated.
[...]

PR #136, opened 2026-05-07
State: store all variable domains as IntervalSet (issue #134, phase 1)

ciaranm:
## Summary

Phase 1 of issue #134: collapse the four-alternative `IntegerVariableState` variant (Constant / Range / SmallSet / IntervalSet) into plain `IntervalSet<Integer>` storage. `ConstantIntegerVariableID` short-circuits at the `deview()` level rather than materialising a placeholder state; the `shared_ptr` COW indirection on stored domains is gone in favour of inline storage. `bits.hh` and `integer_variable_state.{hh,cc}` are deleted; `Bits` is no longer needed since the SmallSet bitset path is gone.

Net diff: +180/-921 in `state.cc`, `state.hh`, `CMakeLists.txt`, `problem.cc` (plus three deleted files).

## Correctness — `Inference::NoChange`

The most error-prone part of this refactor: spuriously returning `BoundsChanged` / `Instantiated` / `InteriorValuesChanged` when nothing actually moved would cause infinite propagation. Each `change_state_for_*` is structured to test the before-state and decide `NoChange` / `Contradiction` before any mutation:

- `infer_equal(value)` → `NoChange` if domain is already singleton equal to `value`; `Contradiction` if `value` not in domain; otherwise `Instantiated`.
- `infer_not_equal(value)` → `NoChange` if `value` already not in domain; `Contradiction` if domain is the singleton `{value}`; else mutate, then `Instantiated` / `BoundsChanged` / `InteriorValuesChanged`.
- `infer_less_than(value)` → `NoChange` if `upper() < value`; else mutate.
- `infer_greater_than_or_equal(value)` → `NoChange` if `lower() >= value`; else mutate.

`domain_has_holes()` is now exact (`intervals.size() > 1`); the SmallSet "always true" over-approximation is gone, but the API still permits over-approximation, so this is a tightening that no consumer cares about.

## What this PR is *not*

No performance work. `new_epoch()` now deep-copies the per-variable `IntervalSet` vectors instead of bumping a `shared_ptr` refcount; this is the change the planned small-buffer-optimised vector (Phase 3) is meant to mitigate. Phase 2 (benchmarking) and Phase 3+ (SBO survey, library evaluation, application to `Reason` / `HalfReifyOnConjunctionOf`, algorithmic wins) come in follow-up PRs.

Plan and decisions are in the comment thread on issue #134.

## Test plan

- [x] `state_test` and `interval_set_test` pass under both gcc 15.2 and clang 21
- [x] Full `ctest --preset release` (168 tests, with VeriPB proof verification): all pass
- [x] Full `ctest` under sanitize (asan + ubsan, 115 tests, ~7 minutes): all pass
- [ ] Phase 2 benchmarking against pre-refactor baseline — separate task


gcs/innards/state.cc
```diff
@@ -873,101 +421,51 @@ auto State::each_value_immutable(const VarType_ & var) const -> generator<Intege
...
-            }(r, apply);
-        },
-        [&](const IntegerVariableIntervalSetState & s) {
-            return [](IntegerVariableIntervalSetState s, auto apply) -> generator<Integer> {
-                for (auto i : s.values->each())
-                    co_yield apply(i);
-            }(s, apply);
-        }}
-        .visit(var_ref);
+    return visit_actual(actual_var, [&](const SimpleIntegerVariableID & v) -> generator<Integer> { return [](IntervalSet<Integer> set, auto apply) -> generator<Integer> {
```
ciaranm: Indenting here is pretty rough, push the braces onto a newline and see if clang-format does a decent job of it.

gcs/innards/state.cc
```diff
@@ -873,101 +421,51 @@ auto State::each_value_immutable(const VarType_ & var) const -> generator<Intege
...
-            }(r, apply);
-        },
-        [&](const IntegerVariableIntervalSetState s) {
-            return [](IntegerVariableIntervalSetState s, auto apply) -> generator<Integer> {
-                for (auto i : s.values->each())
-                    co_yield apply(i);
-            }(s, apply);
-        }}
-        .visit(var_ref);
+    return visit_actual(actual_var, [&](const SimpleIntegerVariableID & v) -> generator<Integer> { return [](IntervalSet<Integer> set, auto apply) -> generator<Integer> {
```
ciaranm: Likewise on indenting

gcs/innards/state.cc
```diff
@@ -873,101 +421,51 @@ auto State::each_value_immutable(const VarType_ & var) const -> generator<Intege
...
 
 auto State::operator()(const IntegerVariableID & i) const -> Integer
 {
     if (auto result = optional_single_value(i))
         return *result;
-    auto [actual_var, negate_first, then_add] = deview(i);
-    IntegerVariableState space = IntegerVariableConstantState{0_i};
-    throw VariableDoesNotHaveUniqueValue{"Integer variable " + visit([&](const auto & i) { return debug_string(i) + " " + debug_string(state_of(i, space)); }, actual_var)};
+    auto [actual_var, _1, _2] = deview(i);
+    throw VariableDoesNotHaveUniqueValue{"Integer variable " + overloaded{[&](const SimpleIntegerVariableID & v) {
```
ciaranm: More bad indenting

gcs/innards/state.cc
```diff
@@ -1097,6 +593,19 @@ auto innards::State::get_persistent_constraint_state(const ConstraintStateHandle
     return _imp->persistent_constraint_states[h.index];
 }
 
+namespace
+{
+    auto debug_string_for(const IntervalSet<Integer> & set) -> string
```
ciaranm: Is this basically dead code now? Showing the representation isn't as important now that we only have one way of storing domains.

ciaranm:
## Phase 2 results — Phase 1 perf impact vs main

3 trials per build per benchmark, median solve time:

| Benchmark              | Baseline | After | Δ | % |
|------------------------|---------:|------:|--:|--:|
| `magic_series_300`     | 7.074s   | 6.008s   | −1.066s  | **−15.1%** |
| `magic_square_5`       | 41.189s  | 32.767s  | −8.422s  | **−20.4%** |
| `langford_11`          | 12.812s  | 11.385s  | −1.426s  | **−11.1%** |
| `n_queens_14_all`      | 22.307s  | 17.817s  | −4.490s  | **−20.1%** |
| `ortho_latin_6_all`    | 20.706s  | 20.000s  | −0.705s  | **−3.4%**  |
| `qap_12`               | 6.359s   | 6.939s   | +0.580s  | **+9.1%**  |
| `tsp_default`          | 73.471s  | 60.712s  | −12.758s | **−17.4%** |
| `n_queens_88`          | 375.715s | 312.917s | −62.798s | **−16.7%** |

**7 of 8 benchmarks faster** (3.4 % to 20.4 %). Only `qap_12` regresses (+9.1 %, very stable: 6.91 / 7.16 / 6.94 against baseline 6.36 / 6.67 / 6.35). Recursion and propagation counts match **exactly** between builds for every benchmark — semantics correctly preserved.

This is the opposite of the cost we anticipated when planning the issue. The expected story was "unification will be a slowdown, Phase 3 SBO recovers it." The actual story is closer to "unification is a net win; SBO matters specifically for small-domain variables in QAP-like workloads."

### Likely explanations

For the speedups (most benchmarks):
- `std::variant<4>` `std::visit` dispatch on every domain query → direct IntervalSet method calls.
- `shared_ptr` atomic-refcount manipulation on the IntervalSet alternative → direct mutation, no atomics.
- Smaller code (~700 lines deleted from `state.cc`) — better icache behaviour.
- IntervalSet operations inline more cleanly without crossing variant boundaries.

For the QAP regression:
- QAP-12 has 12 small-domain (1..12) assignment variables. Under the variant the `SmallSetState` bitset path was inline; now they are heap-allocated 1-interval `IntervalSet`s. With 123k recursions and `new_epoch` deep-copying the per-variable IntervalSets, that's significant `malloc`/`free` churn that wasn't there before.
- This is exactly the case Phase 3 SBO is designed to recover.

### Phase 3 implications

The decision shape has shifted:

- **Phase 1 alone is already a net improvement** on the curated set. There's no longer a "must mitigate slowdown" story driving Phase 3.
- Phase 3 SBO is still worth doing **specifically to bring QAP-style small-domain workloads back into line** — that is a real and stable 9 % regression — but the urgency is lower than originally planned. We could merge Phase 1 on its own merits and tackle Phase 3 / 5 / 6 as separate, optional improvements rather than "completing" the issue plan.

How would you like to proceed?

1. Merge Phase 1 as-is, tackle Phase 3 (SBO) as a follow-up if QAP-style workloads matter.
2. Hold Phase 1 until Phase 3 is also done, ship together.
3. Investigate the QAP regression in more detail first (profile, confirm root cause is the malloc traffic and not something else) before deciding.

(My weak preference is option 1: ship the visible win now, unblock Phase 5 / 6 separately.)

PR #137, opened 2026-05-07
dev_docs: document the curated benchmark set for perf changes

ciaranm:
## Summary

Adds `dev_docs/benchmarking.md` recording the eight-benchmark set settled on while doing Phase 2 of #134, the rationale for each pick, the harness pattern for comparing two builds, and what to capture (solve time + node counts as a correctness cross-check).

The intent is that future performance work re-uses this set, so results stay comparable across PRs over time.

Independent of the Phase 1 refactor PR (#136). Doc-only change.

## Test plan

- [x] doc-only — no tests required
- [x] referenced from `dev_docs/README.md`

### DIFF
--- a/dev_docs/benchmarking.md
+++ b/dev_docs/benchmarking.md
@@ -1,0 +1,154 @@
+# Benchmarking performance changes
+
+This document describes the curated set of benchmarks we use when evaluating
+a performance-sensitive change (data structure refactor, propagator rewrite,
+search heuristic change, etc.) and how to run them. It is **not** a continuous
+performance test suite — it is a checklist for when a maintainer needs to
+quantify the wall-time impact of a non-trivial change.
+
+The set was settled on while doing Phase 2 of issue #134. The intent is that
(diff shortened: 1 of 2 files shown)


dev_docs/benchmarking.md
```diff
@@ -0,0 +1,154 @@
...
+git worktree add ../baseline-worktree main
+cd ../baseline-worktree && cmake --preset release && cmake --build --preset release --parallel 32
+```
+
+Then run a script that alternates trials between the two builds. Per-build
+3-trial sweep is usually enough; the long benchmarks (`tsp`, `n_queens_88`)
+have low variance anyway, the short ones can be noisy. Take the median.
+
+A minimal harness (the one used for issue #134 Phase 2 lives in
+`/tmp/bench-phase2/run.sh`):
```
ciaranm: Pointing to a file in `/tmp` isn't helpful. Just the code below is fine.

PR #141, opened 2026-05-08
IntervalSet-aware idioms in element.cc and search_heuristics (issue #134, phase 6)

ciaranm:
## Summary

Phase 6 of issue #134, stacked on #140. Three localised cleanups now that variable domain storage is uniformly `IntervalSet`:

### A. `search_heuristics.cc` — drop `copy_of_values` where unnecessary

Four of the five `copy_of_values` + iterate sites become `each_value()` (smallest_first) or `domain_size()` + `each_value()` (the split_* trio). The fifth (largest_first) keeps `copy_of_values` for now because `IntervalSet::each_reversed()` has no analog on `CurrentState`.

Behaviourally a wash for the bench set (the heuristics most of our benchmarks use are `smallest_in` / `lower_bound`-based, not `smallest_first`); included for code clarity.

### B. `element.cc` — hoist invariants out of the per-`test_val` loop

`state.copy_of_values(result_var)` and `state.bounds(result_var)` were being recomputed on every iteration of a loop whose body only mutates `index_vars[fixed_dim]` (never `result_var`). Hoist both out of the loop.

### C. New `State::domain_intersects_with(var, IntervalSet)` primitive

`element.cc`'s inner loop did `looking_for.contains_any_of(state.copy_of_values(array_var))` — a full copy of `array_var`'s domain just to test intersection. The new primitive walks the stored `IntervalSet` against the given set via the existing `contains_any_of` merge-walk, no copy. Constant short-circuits via `set.contains`; view-with-offset and constant-with-offset cases also handled.

## Bench (Phase 6 vs Phase 5)

3 trials per build, 2 for `n_queens_88`. Median solve time:

| Benchmark           | phase5 | phase6 | Δ | % |
|---------------------|-------:|-------:|---:|---:|
| `magic_series_300`  | 5.128 s | 5.146 s | +0.018 s | +0.4 % |
| `magic_square_5`    | 30.711 s | 30.502 s | −0.209 s | −0.7 % |
| `langford_11`       | 10.325 s | 10.114 s | −0.211 s | **−2.0 %** |
| `n_queens_14_all`   | 17.072 s | 17.163 s | +0.091 s | +0.5 % |
| `ortho_latin_6_all` | 18.295 s | 18.152 s | −0.143 s | −0.8 % |
| `qap_12`            | 6.723 s | 6.550 s | −0.174 s | **−2.6 %** |
| `tsp_default`       | 54.931 s | 54.074 s | −0.857 s | **−1.6 %** |
| `n_queens_88`       | 248.043 s | 247.839 s | −0.204 s | −0.1 % |

**The wins land where Element is in the propagation loop.** `langford`, `qap`, and `tsp` all post Element-family constraints (`Element`, `Element2DConstantArray`, `ElementConstantArray` respectively). The other benchmarks see run-to-run noise because items A (search-heuristic) and C (Element) don't apply to them.

Recursion and propagation counts identical between builds — semantics preserved.

A future Element-heavy benchmark (with non-constant arrays in the inner loop) would show the C-side win more dramatically. Out of scope for this PR.

## Cumulative since pre-Phase-1 baseline

Combining Phase 4's baseline numbers with the deltas through Phases 3, 5, 6:

| Benchmark           | baseline (pre-#134) | phase6 (this) | total Δ |
|---------------------|--------------------:|--------------:|--------:|
| `magic_series_300`  | 7.226 s             | 5.146 s       | **−28.8 %** |
| `magic_square_5`    | 41.069 s            | 30.502 s      | **−25.7 %** |
| `langford_11`       | 12.642 s            | 10.114 s      | **−20.0 %** |
| `n_queens_14_all`   | 21.759 s            | 17.163 s      | **−21.1 %** |
| `ortho_latin_6_all` | 20.440 s            | 18.152 s      | **−11.2 %** |
| `qap_12`            | 6.333 s             | 6.550 s       | +3.4 % |
| `tsp_default`       | 72.945 s            | 54.074 s      | **−25.9 %** |
| `n_queens_88`       | 377.174 s           | 247.839 s     | **−34.3 %** |

7/8 between 11 % and 34 % faster end-to-end. QAP regression now down to +3.4 % (from +9.1 % at Phase 1, +7.2 % at Phase 3).

## Deferred

Phase 6D — interval-level inference primitives (`infer_not_in_interval` etc.) — punted to a separate follow-up issue. Design space needs a dedicated pass: which primitives, how proof-logging reasons should look, which propagators benefit. Issue to be filed alongside this PR.

## Test plan

- [x] `state_test`, `interval_set_test` pass under release.
- [x] Full `ctest --preset release`: 168/168 with VeriPB proof verification.
- [ ] Sanitize ctest deferred (disk constraints during the cumulative bench setup).

## Stacking

Targets `sbo-interval-set` (#138). Once #136, #138, #140 land in order, this rebases cleanly. Diff here is Phase-6-only.

### DIFF
--- a/gcs/constraints/count.cc
+++ b/gcs/constraints/count.cc
@@ -197,9 +197,17 @@ auto Count::install_propagators(Propagators & propagators) -> void
             if (highest_how_many_might) {
                 auto just = JustifyExplicitlyThenRUP{
                     [&](const ReasonFunction & reason) -> void {
-                        for (const auto & voi : state.each_value_immutable(value_of_interest)) {
-                            for (const auto & [idx, var] : enumerate(vars)) {
-                                if (! state.in_domain(var, voi)) {
+                        // Per-(voi, var) conditional pairs: emit when voi is
+                        // in value_of_interest's domain but not var's. Outer
+                        // loop is over var so we can call each_interval_minus
+                        // once per var; the (A, B) pair for one (voi, var) is
+                        // order-sensitive but the (voi, var) iterations
+                        // themselves are not.
+                        auto voi_set = state.copy_of_values(value_of_interest);
+                        for (const auto & [idx, var] : enumerate(vars)) {
+                            auto var_set = state.copy_of_values(var);
+                            for (auto [lo, hi] : voi_set.each_interval_minus(var_set))
+                                for (Integer voi = lo; voi <= hi; ++voi) {
                                     logger->emit_rup_proof_line_under_reason(reason,
                                         WPBSum{} + 1_i * (value_of_interest != voi) + 1_i * (! get<0>(flags[idx])) >= 1_i,
                                         ProofLevel::Temporary);
@@ -207,12 +215,15 @@ auto Count::install_propagators(Propagators & propagators) -> void
                                         WPBSum{} + 1_i * (value_of_interest != voi) + 1_i * (var != voi) >= 1_i,
                                         ProofLevel::Temporary);
                                 }
-                            }
+                        }
 
+                        // Per-voi unconditional lines: emit after all the
+                        // conditionals so they have the full set of pairwise
+                        // facts to RUP-derive against.
+                        for (const auto & voi : voi_set.each())
                             logger->emit_rup_proof_line_under_reason(reason,
                                 WPBSum{} + 1_i * (value_of_interest != voi) + 1_i * (how_many < *highest_how_many_might + 1_i) >= 1_i,
                                 ProofLevel::Temporary);
-                        }
                     }};
                 inference.infer(logger, how_many < *highest_how_many_might + 1_i, just, generic_reason(state, all_vars));
             }
--- a/gcs/constraints/count.cc
+++ b/gcs/constraints/count.cc
@@ -106,33 +106,17 @@ auto Count::install_propagators(Propagators & propagators) -> void
             int how_many_definitely_do_not = 0;
             auto viable_places = 0_i;
             for (auto & var : vars) {
-                bool seen_any = false;
-                for (const auto & voi : state.each_value_immutable(value_of_interest)) {
-                    if (state.in_domain(var, voi)) {
-                        seen_any = true;
-                        break;
-                    }
-                }
-
-                if (! seen_any)
-                    ++how_many_definitely_do_not;
-                else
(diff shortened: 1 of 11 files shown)


ciaranm:
Pushed an additional commit while looking at ReifiedEquals: `State::domains_intersect(var, var)` primitive that walks both stored interval sets via merge for the common case of two SimpleIntegerVariableIDs with no view offsets — no copy of either side. Used at three sites with the same per-value-membership-scan pattern as element.cc:276:

- `equals.cc` (ReifiedEquals's "force not-equals if domains don't overlap" check)
- `count.cc` (two sites — propagator + RUP-justification callback)

Bench-neutral on the 8-benchmark set (neither Equals nor Count is a bottleneck on those workloads); the win is asymptotic on wide-domain workloads where the per-value membership scan was the cost.

ciaranm:
Pushed another commit while looking at `enforce_equality`: new `IntervalSet::each_interval_minus(other)` primitive that yields each maximal closed interval of values present in `*this` and absent from `other`, walking both interval lists via merge in `O(intervals(this) + intervals(other) + |output|)` instead of the per-value membership scan.

Applied at two propagator sites that had the same shape:

- `equals.cc` (enforce_equality's symmetric-difference arm — the original site)
- `min_max.cc` (ArrayMinMax's tighten-supporting-variable inference)

Doxygen on the new method spells out the closed-interval convention (matches `each_interval()`, distinct from the half-open `each_gap_interval()`).

Unit tests: 13 new TEST_CASEs covering edge cases plus a brute-force cross-check that yielded intervals enumerate exactly the values a per-value membership scan reports.

`count.cc:200` has the same shape but inside a justification callback; flipping its outer loop would change the order of emitted RUP proof lines (may or may not be load-bearing under VeriPB). Left for a separate, more careful pass.

ciaranm:
Pushed the count.cc:200 follow-up: same `each_interval_minus` substitution as equals/min_max, with the outer voi loop flipped to be the inner one. The (A, B) emit pair for one (voi, var) stays in its original relative order (A is load-bearing for B's RUP derivation); the (voi, var) iterations themselves are not order-dependent. Per-voi unconditional lines move to a final pass after all conditionals.

Verified: `count_test` passes under VeriPB with the new ordering; full release ctest 168/168.

Issue #181, opened 2026-05-09
Prepare, bounds, and model_contradiction changes and cleanup


ciaranm:
Currently, `Propagators::model_contradiction` is used in two ways:

- To enforce a contradiction of the kind that "by definition, what you asked for does not match the constraint semantics, to the extent that I cannot even write a sensible OPB encoding of this".
- To say "although the constraint you asked for makes sense, I can immediately detect it is unsatisfiable".

Similarly, `Propagators::trim_lower_bound` and `trim_upper_bounds` can mean:

- That by definition, a variable must have certain bounds (for example, the index of an Element constraint) and so this must be forced in the OPB file.
- That as an obvious consequence, a variable must have certain bounds (for example, Abs).

In both cases, the latter leads to the OPB file containing some constraints that aren't really "definitions". This is morally wrong.

This is also somehow linked to Constraint::prepare and it having a somewhat awkward API currently.

I think a fixed API should separate these two cases cleanly. We need to be able to say "by definition, these variable's bounds cannot be wider than ..." and "by definition, this constraint is contradictory". This is definitely a "prepare" thing. Separately, anything that immediately infers bounds or a contradiction should probably be an initialiser, that includes a proof if necessary (but this should be run with higher priority than "slow" initialisers like autotabulation, because trimming the variable bounds is hugely important for efficiency there).

There is the slightly annoying case where one constraint lets us infer a good bound on a variable, and then a second constraint is posted on that variable where it should really talk about the defined bounds, even though this isn't efficient. I'm not sure how we should handle that.

First task: audit all the constraints, work out which falls into which cases, determine if there are unknown unknowns, and design an API.

ciaranm:
## Audit + plan

After auditing every `model_contradiction`, `trim_*_bound`, and `prepare()` call site, here's the proposed shape and per-constraint migration.

### Audit findings

`prepare()` today is doing six orthogonal jobs across the constraints: input sanity-check, definitional bound tightening, consequence bound tightening, pre-computation read by `define_proof_model`/propagators, state-handle allocation, and full reformulation. The conflation that makes the OPB file morally wrong is in jobs 2 and 3 — `trim_*_bound` writes an OPB axiom regardless of whether the bound is part of the constraint definition or just a consequence of it (sometimes one that depends on bounds set by *other* posted constraints).

Classification of `trim_*_bound` call sites:
[...]

PR #186, opened 2026-05-10
constraints: stop special-casing degenerate inputs in prepare() (issue #181)

ciaranm:
## Summary

Second of four PRs implementing the plan in #181, stacked on top of #185. Replaces every `Propagators::model_contradiction` call with `install_initial_contradiction` (or its clique-duplicate variant), and lets each constraint's `define_proof_model` handle the degenerate input itself. The OPB file becomes fully self-describing — no more unlabelled empty-clause axioms emitted out-of-band.

- **AllDifferent (VC, GAC, Symmetric, and AllDifferentExcept's dup+empty-excluded leg).** On duplicate variables, `prepare()` records a flag and returns `true`; `define_proof_model` runs the existing clique-of-not-equals encoding unchanged (it already emits a self-contradicting half-reified pair for the duplicated pair); `install_propagators` installs a `SimpleDefinition`-priority initialiser instead of the matching propagator. The proof RUPs `selector` and `!selector` citing one duplicated pair's flag (drawn from the encoding's return value), then RUPs false. The two-step shape is necessary because RUP for false directly cannot pick a selector polarity from two non-unit half-reifications.
- **Table with no allowed tuples.** `prepare()` records the flag and skips the selector allocation (an empty range `[0, -1]` would be invalid). `define_proof_model` emits `WPBSum{} >= 1_i` directly — morally what an empty-domain selector would produce. `install_propagators` uses `install_initial_contradiction` with plain RUP, since a literal `0 >= 1` is unit-derivable.

`define_clique_not_equals_encoding` now returns the duplicate-pair selectors, mirroring the existing `_except` variant. The shared contradiction-initialiser body lives in `encoding.cc` as `install_clique_duplicate_contradiction_initialiser`; VC, GAC, Symmetric, and AllDifferentExcept (dup+empty-excluded leg) all use it.

A small follow-up commit adds explicit `run_alldiff_dup_test` cases for VC and GAC AllDifferent — the contradiction path was previously only exercised indirectly via Symmetric and AllDifferentExcept. Six new VeriPB-verified scenarios.

## Test plan

- [x] `cmake --build --preset release --parallel 8` clean
- [x] `ctest --preset release -j 8` — 194/194 pass
- [x] New `run_alldiff_dup_test` cases for both `GACAllDifferent` and `VCAllDifferent` verify UNSATISFIABLE under VeriPB
- [x] Existing duplicate-vars / empty-table tests in `symmetric_all_different_test`, `all_different_except_test`, `table_test` continue to pass with proof logging

### DIFF
--- a/gcs/constraints/all_different/encoding.hh
+++ b/gcs/constraints/all_different/encoding.hh
@@ -5,19 +5,29 @@
 #include <gcs/innards/inference_tracker-fwd.hh>
 #include <gcs/innards/proofs/proof_logger-fwd.hh>
 #include <gcs/innards/proofs/proof_only_variables.hh>
+#include <gcs/innards/propagators-fwd.hh>
 #include <gcs/innards/state.hh>
 #include <gcs/integer.hh>
 #include <gcs/variable_id.hh>
 
 #include <map>
+#include <string>
 #include <vector>
 
 namespace gcs
 {
     namespace innards
     {
+        // Emits the clique-of-not-equals encoding. Where the same variable
+        // appears more than once in `vars`, the resulting pair-of-half-reified
+        // constraints is jointly inconsistent — both polarities of the
+        // selector are forced false. The returned map gives one of the
+        // per-pair selector flags for each such duplicated variable, so
+        // callers can cite it from a justification that derives
+        // contradiction explicitly (RUP alone cannot pick a polarity, since
+        // both half-reifications are non-unit).
         auto define_clique_not_equals_encoding(ProofModel & model,
-            const std::vector<IntegerVariableID> & vars) -> void;
+            const std::vector<IntegerVariableID> & vars) -> std::map<IntegerVariableID, ProofFlag>;
 
         // Emits the AllDifferentExcept clique encoding. Where the same variable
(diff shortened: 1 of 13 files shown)


gcs/constraints/all_different/encoding.hh
```diff
@@ -5,19 +5,29 @@
...
+        // appears more than once in `vars`, the resulting pair-of-half-reified
+        // constraints is jointly inconsistent — both polarities of the
+        // selector are forced false. The returned map gives one of the
+        // per-pair selector flags for each such duplicated variable, so
+        // callers can cite it from a justification that derives
+        // contradiction explicitly (RUP alone cannot pick a polarity, since
+        // both half-reifications are non-unit).
         auto define_clique_not_equals_encoding(ProofModel & model,
-            const std::vector<IntegerVariableID> & vars) -> void;
+            const std::vector<IntegerVariableID> & vars) -> std::map<IntegerVariableID, ProofFlag>;
```
ciaranm: Rather than returning lots of duplicate selectors, would this be cleaner if it just returned one `optional<pair<IntegerVariableID, ProofFlag>>`? None of the callers seem to need more than that, and it's not like the map returns the full set anyway.

PR #190, opened 2026-05-11
Constraint names

philrod1:
Looks like a rebase is in order.

This PR is implementing the named constraints functionality.  It looks like this:

`using ConstraintName = std::variant<CurrentlyUnnamedConstraint, NumberedConstraint, NamedConstraint>;`

CurrentlyUnamedConstraint is the default, and is what constraints get when the constructor is called.  Cloning a constraint loses any set names back to CurrentlyUnamedConstraint.

I added plumbing to Problem::post() to support users adding named constraints.  I also added supporting code to enable "automatic" formatting of names in format strings.

### DIFF
--- a/examples/money/money.cc
+++ b/examples/money/money.cc
@@ -77,13 +77,14 @@ auto main(int argc, char * argv[]) -> int
     auto y = p.create_integer_variable(0_i, 9_i, "y");
 
     vector<IntegerVariableID> vars{s, e, n, d, m, o, r, y};
-    p.post(AllDifferent{vars});
+    p.post(AllDifferent{vars}, "_money_a_d");
 
     // clang-format off
     p.post(WeightedSum{}
             +                 1000_i * s +  100_i * e +  10_i * n +  1_i * d
             +                 1000_i * m +  100_i * o +  10_i * r +  1_i * e
-            + -10000_i * m + -1000_i * o + -100_i * n + -10_i * e + -1_i * y == 0_i);
+            + -10000_i * m + -1000_i * o + -100_i * n + -10_i * e + -1_i * y
+            == 0_i , "_money_sum");
     // clang-format on
 
     auto stats = solve(
--- a/examples/money/money.cc
+++ b/examples/money/money.cc
@@ -77,14 +77,14 @@ auto main(int argc, char * argv[]) -> int
     auto y = p.create_integer_variable(0_i, 9_i, "y");
 
     vector<IntegerVariableID> vars{s, e, n, d, m, o, r, y};
-    p.post(AllDifferent{vars}, "_money_a_d");
+    p.post_named(AllDifferent{vars}, "money_a_d");
 
     // clang-format off
-    p.post(WeightedSum{}
+    p.post_named(WeightedSum{}
             +                 1000_i * s +  100_i * e +  10_i * n +  1_i * d
             +                 1000_i * m +  100_i * o +  10_i * r +  1_i * e
             + -10000_i * m + -1000_i * o + -100_i * n + -10_i * e + -1_i * y
-            == 0_i , "_money_sum");
+            == 0_i , "money_sum");
     // clang-format on
 
     auto stats = solve(
--- a/gcs/constraint.hh
+++ b/gcs/constraint.hh
@@ -17,6 +17,7 @@ namespace gcs
 
     struct CurrentlyUnnamedConstraint final
     {
+        auto as_string() const -> std::string { return "unnamed"; }
     };
 
     struct NumberedConstraint final
@@ -26,15 +27,19 @@ namespace gcs
         // Claude (web) says this is a good idea for comparisons.
         // Do we need comparisons? (C++20 spaceship operator)
         auto operator<=>(const NumberedConstraint &) const = default;
+        auto as_string() const -> std::string { return "C" + std::to_string(number); }  // TODO: change "C" to "_" at some point before merge
     };
 
     struct NamedConstraint final
     {
         std::string name;
         auto operator<=>(const NamedConstraint &) const = default;
+        auto as_string() const -> std::string { return name; }
     };
 
     using ConstraintName = std::variant<CurrentlyUnnamedConstraint, NumberedConstraint, NamedConstraint>;
+
+    auto as_string(const ConstraintName &) -> std::string;
 
     /**
      * \brief Subclasses of Constraint give a high level way of defining
@@ -83,7 +88,7 @@ namespace gcs
         /**
          * Return an s-expr representation of the constraint. To be used internally.
          */
-        [[nodiscard]] virtual auto s_exprify(const std::string & name, const innards::ProofModel * const) const -> std::string = 0;
+        [[nodiscard]] virtual auto s_exprify(const innards::ProofModel * const) const -> std::string = 0;
     };
 }
 
--- a/gcs/constraint.hh
+++ b/gcs/constraint.hh
@@ -9,6 +9,12 @@
 #include <string>
 #include <variant>
 
+#if defined(__cpp_lib_format)
+#include <format>
+#else
+#include <fmt/core.h>
+#endif
+
 namespace gcs
 {
     /**
@@ -91,7 +97,6 @@ namespace gcs
 
 // The following is needed to allow ConstraintName to be used in format strings.
 #if defined(__cpp_lib_format)
-#include <format>
 template <>
 struct std::formatter<gcs::ConstraintName> : std::formatter<std::string>
 {
@@ -101,7 +106,6 @@ struct std::formatter<gcs::ConstraintName> : std::formatter<std::string>
     }
 };
 #else
-#include <fmt/format.h>
 template <>
 struct fmt::formatter<gcs::ConstraintName> : fmt::formatter<std::string>
 {
--- a/gcs/constraint.hh
+++ b/gcs/constraint.hh
@@ -5,6 +5,7 @@
 #include <gcs/innards/propagators-fwd.hh>
 #include <gcs/innards/state-fwd.hh>
 
+#include <format>
 #include <memory>
 #include <string>
 #include <variant>
@@ -91,5 +92,29 @@ namespace gcs
         [[nodiscard]] virtual auto s_exprify(const innards::ProofModel * const) const -> std::string = 0;
     };
 }
+
+// The following is needed to allow ConstraintName to be used in format strings.
+// We do this quite a lot in s_exprify().  If we didn't do this, we'd have to write
+// `format("{}", as_string(_name))` everywhere instead of just `format("{}", _name)`.
+// It's comfort over clarity.  What's the best option?
+#if defined(__cpp_lib_format)
+template <>
+struct std::formatter<gcs::ConstraintName> : std::formatter<std::string>
+{
+    auto format(const gcs::ConstraintName & name, std::format_context & ctx) const
+    {
+        return std::formatter<std::string>::format(gcs::as_string(name), ctx);
+    }
+};
+#else
+template <>
+struct fmt::formatter<gcs::ConstraintName> : fmt::formatter<std::string>
+{
+    auto format(const gcs::ConstraintName & name, fmt::format_context & ctx) const
+    {
+        return fmt::formatter<std::string>::format(gcs::as_string(name), ctx);
+    }
+};
+#endif
 
 #endif
--- a/gcs/constraint.hh
+++ b/gcs/constraint.hh
@@ -31,9 +31,6 @@ namespace gcs
     struct NamedConstraint final
     {
         std::string name;
-
-        // Claude (web) says this is a good idea for comparisons.
-        // Do we need comparisons? (C++20 spaceship operator)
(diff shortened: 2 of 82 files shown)


gcs/constraints/abs.cc
```diff
@@ -23,16 +23,20 @@
 using namespace gcs;
 using namespace gcs::innards;
 
+using gcs::as_string;
```
philrod1: I don't need this any more

gcs/constraint.hh
```diff
@@ -5,15 +5,43 @@
...
+    };
+
+    struct NumberedConstraint final
+    {
+        unsigned long long number;
+
+        // Claude (web) says this is a good idea for comparisons.
+        // Do we need comparisons? (C++20 spaceship operator)
+        auto operator<=>(const NumberedConstraint &) const = default;
+        auto as_string() const -> std::string { return "C" + std::to_string(number); }  // TODO: change "C" to "_" at some point before merge
```
ciaranm: Change!

ciaranm: Also this should be `[[nodiscard]]`.

gcs/constraint.hh
```diff
@@ -5,15 +5,43 @@
...
+        auto as_string() const -> std::string { return "unnamed"; }
+    };
+
+    struct NumberedConstraint final
+    {
+        unsigned long long number;
+
+        // Claude (web) says this is a good idea for comparisons.
+        // Do we need comparisons? (C++20 spaceship operator)
+        auto operator<=>(const NumberedConstraint &) const = default;
```
ciaranm: Comparisons are useful to have if you're potentially going to want to sort by constraint, or put them in a container. Also, this can be `[[nodiscard]]`.

gcs/constraint.hh
```diff
@@ -5,15 +5,43 @@
...
+    struct NamedConstraint final
+    {
+        std::string name;
+        auto operator<=>(const NamedConstraint &) const = default;
+        auto as_string() const -> std::string { return name; }
+    };
+
+    using ConstraintName = std::variant<CurrentlyUnnamedConstraint, NumberedConstraint, NamedConstraint>;
+
+    auto as_string(const ConstraintName &) -> std::string;
```
ciaranm: `[[nodiscard]]`

gcs/constraint.hh
```diff
@@ -38,7 +69,8 @@ namespace gcs
 
     public:
         virtual ~Constraint() = 0;
-
+        auto name() const -> const ConstraintName & { return _name; }
```
ciaranm: `[[nodiscard]]`

gcs/constraint.hh
```diff
@@ -57,8 +89,32 @@ namespace gcs
...
          */
-        [[nodiscard]] virtual auto s_exprify(const std::string & name, const innards::ProofModel * const) const -> std::string = 0;
+        [[nodiscard]] virtual auto s_exprify(const innards::ProofModel * const) const -> std::string = 0;
     };
 }
 
+// The following is needed to allow ConstraintName to be used in format strings.
+// We do this quite a lot in s_exprify().  If we didn't do this, we'd have to write
+// `format("{}", as_string(_name))` everywhere instead of just `format("{}", _name)`.
+// It's comfort over clarity.  What's the best option?
```
ciaranm: Yeah, I'm generally in favour of making things formattable as we go forward.

gcs/problem.cc
```diff
@@ -1,3 +1,4 @@
+#include "gcs/constraint.hh"
```
ciaranm: Should be angle brackets

gcs/problem.hh
```diff
@@ -87,16 +87,31 @@ namespace gcs
          */
         auto post(const Constraint &) -> void;
 
+        /**
+         * \brief Add a named clone of this Constraint to the model.
+         */
+        auto post(const Constraint &, const std::string &) -> void;
```
ciaranm: I think actually these should be called `post_named(...)`. Although overloads work here, I can see this being a bit awkward in the future, and this is part of the public API so we should aim for stability.

examples/money/money.cc
```diff
@@ -77,13 +77,14 @@ auto main(int argc, char * argv[]) -> int
     auto y = p.create_integer_variable(0_i, 9_i, "y");
 
     vector<IntegerVariableID> vars{s, e, n, d, m, o, r, y};
-    p.post(AllDifferent{vars});
+    p.post(AllDifferent{vars}, "_money_a_d");
```
ciaranm: Any particular reason for the leading underscore?

ciaranm (changes requested):
Some little bits. Hoping this rebases fairly easily on top of all the other mucking around with constraints I'm doing...

examples/money/money.cc
philrod1: That just happened to be the last thing I tested.  I'll change it to a non-underscore value.

gcs/problem.hh
philrod1: We will probably need to update the documentation to include post_named()

ciaranm:
Want me to run the format fix through Claude? It seems to know this one.

philrod1:
By all means!  I was getting frustrated by trying to fix a bug that doesn't happen on my machine, so I have set up an Ubuntu 24.04 VM for testing and debugging in the correct environment.  If Claude knows how to fix it, that's great.

gcs/constraint.hh
```diff
@@ -7,13 +7,43 @@
 
 #include <memory>
 #include <string>
+#include <variant>
```
ciaranm, line 10: The Ubuntu 24.04 build failure (errors at `arithmetic.cc:120` from `std::format` not finding a formatter for `std::variant<...>`) traces back to this `#if defined(__cpp_lib_format)`. There's no `#include <version>` before it, and `<memory>`/`<string>`/`<variant>` don't pull `<version>` in transitively on libstdc++ 13. So when this header is included near the top of a translation unit, `__cpp_lib_format` is undefined, the `#else` branch is taken, and only `fmt::formatter<gcs::ConstraintName>` gets defined.

Then something later in the TU (e.g. `gcs/integer.hh`, which does include `<version>` before its own check) pulls in `<format>`. ADL on the `std::variant` argument in `format("{}", _name, ...)` resolves the unqualified call to `std::format`, which then looks for `std::formatter<std::variant<...>>` — never defined in this TU — and falls back to the deleted primary template. macOS escapes this because libc++ in clang 21 defines `__cpp_lib_format` early enough.

`gcs/integer.hh` and `gcs/constraints/innards/constraints_test_utils.hh` already use the right pattern: `#include <version>` before the `#if`. One-line suggestion below.

```suggestion
#include <variant>
#include <version>
```

ciaranm: Here's what Claude suggests. Let's try it and see.

ciaranm:
Well that's a new error now, at least. Not sure what's happened, will see if our AI overlord knows.

gcs/constraints/abs.cc
```diff
@@ -28,12 +28,11 @@ using std::max;
 using std::string;
 using std::stringstream;
 using std::unique_ptr;
```
ciaranm, line 36: This PR dropped `using std::vector;` because it was unused on the branch at the time, but `main` has since gained `pieces_to_set(vector<pair<Integer, Integer>> &)` and two call-sites in `abs.cc` (commit 181034b, the interval-wise abs reasoning). The PR-into-`main` merge therefore reintroduces `vector` usage without the using-declaration, hence the `'vector' was not declared` errors on every platform.

```suggestion
using std::unique_ptr;
using std::vector;
```

Heads-up: once this lands there's a second, Ubuntu-only error already visible in the same log (`fmt::v11::detail::type_is_unformattable_for<std::variant<...>, char> has incomplete type` in `abs.cc:299` and `all_different_except.cc:209`). It's a conditional mismatch I missed when suggesting the `<version>` fix: the formatter specialization in `constraint.hh` gates on `__cpp_lib_format` alone, but every `.cc` using it gates on `__cpp_lib_print && __cpp_lib_format`. On GCC 13 `__cpp_lib_format` is defined and `__cpp_lib_print` is not, so the call sites take the `fmt` branch while `constraint.hh` defines only `std::formatter<gcs::ConstraintName>` — `fmt::format` then can't find a formatter. Easiest fix is to tighten the `#if` in `constraint.hh` to `__cpp_lib_print && __cpp_lib_format` to match. Happy to push that as a separate suggestion if useful.

gcs/constraint.hh
```diff
@@ -7,13 +7,44 @@
...
+#include <version>
+
+#if defined(__cpp_lib_format)
```
ciaranm: Tighten the gate to match the call-site condition used in `arithmetic.cc`, `abs.cc`, etc. (`__cpp_lib_print && __cpp_lib_format`). Otherwise on GCC 13 — where `__cpp_lib_format` is defined but `__cpp_lib_print` is not — this header includes `<format>` (and the specialization below defines `std::formatter<gcs::ConstraintName>`), while every `.cc` falls into the fmt branch and tries to use `fmt::format` / `fmt::print` with no `fmt::formatter<gcs::ConstraintName>` in sight. That's the source of the second wave of errors in this build (`type_is_unformattable_for<std::variant<...>>` in `abs.cc:299` and `all_different_except.cc:209`).

Same change is needed at line 100 for the specialization branch — see follow-up suggestion below.

```suggestion
#if defined(__cpp_lib_print) && defined(__cpp_lib_format)
```

gcs/constraint.hh
```diff
@@ -57,8 +92,29 @@ namespace gcs
         /**
          * Return an s-expr representation of the constraint. To be used internally.
          */
-        [[nodiscard]] virtual auto s_exprify(const std::string & name, const innards::ProofModel * const) const -> std::string = 0;
+        [[nodiscard]] virtual auto s_exprify(const innards::ProofModel * const) const -> std::string = 0;
     };
 }
 
+// The following is needed to allow ConstraintName to be used in format strings.
+#if defined(__cpp_lib_format)
```
ciaranm: Same change as line 13 — the gate for picking `std::formatter` vs `fmt::formatter` must match the gate the call sites use (`__cpp_lib_print && __cpp_lib_format`), otherwise GCC 13 ends up with `std::formatter` defined but `fmt::format` called.

```suggestion
#if defined(__cpp_lib_print) && defined(__cpp_lib_format)
```

gcs/constraint.hh
philrod1: I already tried that :(
I think it is correct to add that check, but the problem persists.

ciaranm:
Let's let the checks run on this and see how far they get, and if that doesn't work I'll take a proper look.

philrod1:
I feel like I'm making progress, only to be slapped again by the compiler.  gcs/integer.hh needed fixing, too, but I'm still chasing errors.  I'll look for other instances of not using the double condition.

ciaranm:
I'll let Claude iterate on it. So far it looks like a purely mechanical "are the headers, guards, and usings set up correctly" problem that probably isn't helped by a few recent changes on main, which is probably less frustrating for a machine to solve than us trying to track it down. Assuming Claude doesn't get stuck, I'll ask it to produce and push a clean rebase on top of main after this, and then try to merge this one before I cause any more problems by merging other stuff :)

gcs/constraints/arithmetic.cc
```diff
@@ -117,7 +117,7 @@ auto GACArithmetic<op_>::s_exprify(const std::string & name, const innards::Proo
     case ArithmeticOperator::Power: op_str = "power"; break;
     }
 
-    return format("({} {} {} {} {})", name, op_str,
+    return format("({} {} {} {} {})", _name, op_str,
```
ciaranm: The remaining Ubuntu error (`std::formatter must be specialized` at this line) is fallout from my earlier `constraint.hh` tightening, plus a quirk specific to this file: every other constraint formats `_name` via `print(s, "{} ...", _name)` where `print` resolves to `fmt::print` or `std::print` through an explicit `using`. This file has no `using std::format;` / `using fmt::format;`, so the unqualified `format(...)` here is found by ADL on the `std::variant` argument — which always picks `std::format`, even on GCC 13 where the gate now only defines `fmt::formatter<gcs::ConstraintName>`.

I tried suggesting a conditional `using ::format;` block alongside the existing `std::vector` line, but that line isn't in the PR's diff hunk so GitHub won't accept the suggestion there. The simpler local fix: pre-convert via `as_string(_name)` (already declared in `constraint.hh`), so what reaches `format` is just a `std::string` and no `ConstraintName` formatter is needed on either side.

```suggestion
    return format("({} {} {} {} {})", as_string(_name), op_str,
```

philrod1:
I have found the culprit.  arithmetic.cc was using format() in s_exprify() not print().  the format() function wasn't being pulled in, so I have modified the s_exprify() to use print().  Once I have the branch commits merged, I'll push

ciaranm:
Aha, this is looking promising. What's the status now? Should I go over the final diff, or would you like to rebase first, or should I?

philrod1:
I have successfully rebased on main.  If you are happy with the diff, it should be ready to go.

Issue #198, opened 2026-05-14
Extract documentation information on what exactly our OPB encoding procedures are


ciaranm:
Currently, the OPB encoding procedure is a source of weakness for verification: we produce proofs that verify that the solver output is correct, *if* we trust that the OPB encoding "means the same thing" as the input problem. There is ongoing work to close this gap, by developing an external, formally-verified encoder. This encoder will take an s-expr file as input, and generate its own OPB encoding. So that the proofs actually work, this encoding needs to line up closely with the encodings our solver uses (we are not going to be reading this OPB encoding in, just using it, and also having independent code so that proof verification can still be done with weaker guarantees without the external encoder).

Current challenge: what exactly is the OPB encoding procedure for each of the constraints we support? We need a document detailing this so we can discuss what to change (both in our solver and the external encoder).

Rough plan for an initial pass:

- Identify every constraint we support, noting both its C++ class and the name it generates via `s_exprify`. In some cases (e.g. `less_than_iff`, `greater_than_if`, etc) we have a single class that does a family of related constraints. In this case, it's not a one-to-one correspondence. Rather than listing the s-expr names in full, using notation like `{less_than,greater_than}{,_if,_iff}` is appropriate.
- For each of these constraints, identify the "encoding procedure". This is usually the "define the proof model" code. Take care to also include any bounds that are "by definition" (e.g. `Element`'s index var), but not "enforced as a consequence" (e.g. `Abs`). I use the term "encoding procedure" here because there are often loops. Examples of encoding procedures can be found in Matthew McIlree's PhD thesis: as a sub-step, it may be useful to read this and summarise a few of these as examples inside this issue. Anything ugly or weird should be flagged for intervention.
- Document this either within the code as a detailed comment inside each constraint that can easily be extracted, or possibly we should have an associated developer-oriented markdown file for each constraint with this and other information if this seems cleaner.
- Collate these into a summary.
- We'll need to iterate on design a bunch of times, but in the long term we want to end up with a sustainable procedure that can be partially automated, either by scripts or by an AI agent (as is done for the MiniZinc, XCSP3, and CPMpy bindings currently -- although this isn't a language binding, it is a similar "integration with an external language" issue).

ciaranm:
After discussion, here's a refined version of the plan.

## Scope

"The encoding" = exactly the content of the OPB file that the (future, formally-verified) external encoder will produce from an s-expr input.

**In scope:**
[...]

PR #199, opened 2026-05-14
opb-encoding: pilot in-code doc blocks for Equals, Abs, AllDifferent, Element

ciaranm:
## Summary

First pilot for the OPB-encoding documentation effort (#198): a structured
in-code `OPB-ENCODING-BEGIN` / `OPB-ENCODING-END` block above each
`define_proof_model` (or shared encoding helper), one block per s-expr
name. The block captures clauses (with `(name, sub_name)` labels and
half-reification guards), `define_bound` calls, CP literals referenced,
auxiliary PB flags introduced, and notes for anything non-obvious.

This is enough material to look at the format end-to-end and share with
the rest of the project for feedback before fanning out to every
constraint.

Constraints covered:

- **Equals** family: six blocks inline in the visitor for `equals`,
  `not_equals`, `equals_if`, `not_equals_if`, `equals_iff`,
  `not_equals_iff`. The `Iff` arm covers two s-expr names; each is given
  its own self-contained block rather than a "see X with substitution"
  pointer.
- **Abs**: one block; the textbook two-half-reified definition.
- **AllDifferent**: one block on the shared
  `define_clique_not_equals_encoding` helper (used by GAC, VC, and
  `SymmetricAllDifferent`). Records honestly that all per-pair
  `notequals` flags share a string name; per-pair distinguishing
  protocol is flagged TBD.
- **Element**: one parametric block describing `element_1d` /
  `element_2d` / `element_3d` as a D-dimensional family, plus a matching
  `s_exprify` rename so the s-expr name carries the dimension suffix.

Pilot-level design choices made along the way (open to revision once the
team reviews):

- In-code structured comment blocks, one per s-expr name; extractor
  script can come later once the format has settled.
- Inline placement above the relevant visitor arm / function. When two
  s-expr names share one C++ arm (Iff covers `equals_iff` and
  `not_equals_iff`), they get two consecutive blocks.
- Parametric pseudocode for constraints that loop over arity or
  dimension (AllDifferent, Element). Concrete per-name spelling-out
  would also work; can switch if reviewers prefer it.
- Clause syntax mirrors `WPBSum` source form rather than raw OPB
  syntax; the encoder team converts.
- Constraint labels in OPB (`"NDimensionalElement"` etc.) are kept as
  the source already has them; they aren't required to match s-expr
  names.

## Test plan

- [x] `abs_test`, `equals_test`, `all_different_test`, `element_test`
      all build and pass (Release).
- [x] No tests check s-expr file output, so the `element_1d` /
      `element_2d` / `element_3d` rename has no in-tree fixtures to
      regenerate.

### DIFF
--- a/gcs/constraints/abs.cc
+++ b/gcs/constraints/abs.cc
@@ -88,6 +88,22 @@ auto Abs::install(Propagators & propagators, State &, ProofModel * const optiona
     install_propagators(propagators);
 }
 
+// OPB-ENCODING-BEGIN: abs
+//   s-expr:  abs v1 v2
+//   Clauses:
+//     ("Abs", "non-negative")
+//         1*v2 + -1*v1 == 0            half-reified on { v1 >= 0 }
+//     ("Abs", "negative")
+//         1*v2 + 1*v1 == 0             half-reified on { v1 < 0 }
+//   Bounds:                 (none)
+//   CP literals referenced: v1 >= 0  (v1 < 0 is its negation, handled
+//                                     syntactically by the encoder)
+//   Auxiliary PB flags:     (none)
+//   Notes:
+//     Textbook abs definition: v2 = |v1| means (v1 >= 0 => v2 = v1) and
+//     (v1 < 0 => v2 = -v1). v2's non-negativity is a consequence, not
+//     part of the encoding -- it's derived inside the proof.
+// OPB-ENCODING-END
 auto Abs::define_proof_model(ProofModel & model) -> void
 {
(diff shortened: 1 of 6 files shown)


philrod1:
Is the plan to merge this as a pilot?  If so I see no reason not to merge, unless there's something missing I am not aware of. 
EDIT: What I mean is is the plan to merge stage 1, or wait for the other stages?

ciaranm:
The plan is to discuss this at / before Friday's encodings meeting, to see whether there's anything else we should be adding in here (in particular labels). I'm hoping we'll be able to get the Crystal Maze example as our first full verification test: we should chat about that (this afternoon / tomorrow?).

ciaranm:
We're doing this a different way now.

Issue #200, opened 2026-05-15
Grand unified MDD propagator framework


ciaranm:
Conceptually, there's a lot of overlap between `MDD` (issue #149), `Knapsack`, `BinPacking` (issue #148), `Regular`, and potentially other propagators: they're all looking for paths through a DAG. In the case of `MDD`, the DAG is given explicitly. For `Regular`, it's also close to explicit, but with different syntax. For `Knapsack`, it's constructed on the fly (the current propagator does this once per propagation rather than upfront, which is likely far less efficient than something cleverer), and the same will be true for `BinPacking`.

What we want: a generic `MDD` (and `CostMDD`?) propagator framework that can implement all of the above, sharing the code for checking paths and proof logging paths disappearing.

ciaranm:
Scoped the cost-arc case as a sub-plan: #653 (`CostRegular` / `CostMDD`). Survey findings that bear on this issue:

- Adding costs needs **per-edge flags** in the OPB. `MDD::define_proof_model` and `Regular::define_proof_model` both encode node flags only, with transitions as implications, so there is nothing to attach `cost = sum_e w_e * e` to. That is a shared change to the family, not a CostMDD-local one.
- `MDDGraph` maintains in/out edges and degrees but no path costs; shortest-path-to/from each node has to join it.
- The equality form (`cost = ...`) is **NP-hard** to propagate to domain consistency (Gange, Stuckey, Van Hentenryck, CP 2013), so both MiniZinc globals' shape has to be installed as `<=` /\ `>=`.

#654 covers `CostGCC`, which shares the LP-duality-over-a-graph certificate shape but is otherwise a separate problem.

PR #213, opened 2026-05-17
Regular: upfront-DAG reimpl with Top-level proof scaffolding (#200)

ciaranm:
## Summary

Reimplements `Regular` along the upfront-scaffolding pattern from MDD (#211) and Knapsack (#210): build the DFA support graph once in `prepare()`, emit per-val backward chains and statically-dead-state lines once at `ProofLevel::Top` via `install_initialiser`, and slim the per-call propagator to a single cache-gated `~state[i][q]` line at `ProofLevel::Current` per state-death. The OPB encoding is unchanged from `RegularLegacy`.

The legacy implementation is preserved verbatim as `RegularLegacy` in `gcs/constraints/regular/regular_legacy.{hh,cc}` and exercised by `regular_legacy_test` for as long as we want side-by-side benchmarking.

`Regular` is structurally a layer-uniform MDD, so it's essentially the MDD-reimpl pattern specialised to a shared transition function across layers. Same Top-scaffolding shape, same `DeadCache` plumbing, same per-call sweep.

## Default: the upfront `Regular` is now the default

Everything posts the upfront `Regular` — both frontends (`xcsp_glasgow_constraint_solver`, `fzn_glasgow`), the constraint tests, and the `regular_random` example. `RegularLegacy` (the older per-call propagator) is kept only as the `--legacy` opt-in twin, for side-by-side benchmarking and as a correctness reference.

Upfront is the right default because it **wins on both proof-logging axes**, measured this session *within-branch* (same search tree to the digit — only the propagator is toggled):

- **Proof size: 13–55× smaller** — 9 MB vs `RegularLegacy`'s 496 MB at the largest instance (7660 solutions).
- **VeriPB verify time: 2.3–7× faster** — 1.66 s vs 11.69 s on that instance.

Both effects grow with search size (the fixed root scaffold amortises; the displaced per-call volume compounds). Mechanistically a `Regular` automaton is *narrow*, so the permanent Top-level scaffold is tiny and adds almost no per-line "DB-tax", while the per-call `RegularLegacy` baseline is very verbose — large displacement, small tax, so upfront wins size *and* time. This is the narrow-diagram case of the cross-cutting analysis in `dev_docs/decision-diagram-proof-strategies.md`, which lands MDD (#211), Knapsack (#210) and BinPacking (#212) on their respective defaults.

## What's where

- **`gcs/constraints/regular/regular.{hh,cc}`** — new implementation.
- **`gcs/constraints/regular/regular_legacy.{hh,cc}`** — old implementation, class renamed to `RegularLegacy`; `s_exprify` label is now `regular_legacy`.
- **`gcs/constraints/regular.hh`** — thin shim including both, so existing callers (`gcs/gcs.hh`, xcsp, minizinc, examples) compile unchanged.
- **`gcs/constraints/regular_test.cc`** — tests the new `Regular` (same seven cases as before).
- **`gcs/constraints/regular_legacy_test.cc`** — same seven cases retargeted at `RegularLegacy`, separate proof basename so both run side by side under VeriPB.
- **`examples/regular_random/regular_random.cc`** — adds `--legacy`; combined with the existing `--short-reasons` (which only affects the legacy), one binary now covers all three variants.
- **`dev_docs/regular.md`** — design note: layered-DAG view, OPB encoding, Top scaffolding derivation, per-call propagator shape, bench numbers, #200 pointer.

The `short_reasons` constructor arg is retained on the new `Regular` as a dummy for API parity with `RegularLegacy` and to let `regular_random` cover all three variants from a single command line. The new propagator emits at most one proof line per state-death, so the flag's effect on the new path is small or zero.

## Bench (regular_random, seed `-1115540197`)

One binary, three variants:

| n  | variant            | solve | proof    | verify |
|---:|--------------------|------:|---------:|-------:|
| 10 | new                | 0.00s | 86 KB    | 0.00s  |
| 10 | legacy             | 0.01s | 1.76 MB  | 0.03s  |
| 10 | legacy + short-rsn | 0.00s | 454 KB   | 0.04s  |
| 14 | new                | 0.00s | 230 KB   | 0.02s  |
| 14 | legacy             | 0.06s | 8.76 MB  | 0.13s  |
| 14 | legacy + short-rsn | 0.01s | 1.75 MB  | 0.24s  |
| 18 | new                | 0.01s | 490 KB   | 0.06s  |
| 18 | legacy             | 0.25s | 33.3 MB  | 0.51s  |
| 18 | legacy + short-rsn | 0.04s | 4.95 MB  | 0.98s  |

The new `Regular` produces proofs 4–68× smaller than the legacy default and 1.5–10× smaller than the legacy with `short_reasons`, and VeriPB verifies them faster on every sampled size. This is the MDD-reimpl shape (proof shrink + slim per-call); it doesn't follow the BinPacking-#212 regression because `RegularLegacy` was emitting per-(parent, val) intermediates on every propagation call, so it's far from the already-lean category.

## Risks / trade-offs

- Same Top-vs-Current bet as MDD/Knapsack: if a search uses very few of the dead-state flags the scaffolding bought, we still emit the chains at root. With Regular's relatively small DFAs (vs Knapsack's wide DP) and the layer-uniform sharing, root cost is small in practice — visible in the verify-time table above being lower than the legacy at every sampled size.
- `RegularLegacy::s_exprify` now emits `regular_legacy` instead of `regular`. No callers consume `s_exprify` output as a parser of constraint identity, but worth flagging.

## Test plan

- [x] `cmake --build --preset release --target regular_test regular_legacy_test` clean.
- [x] `./build/regular_test` and `./build/regular_legacy_test` — all 7 subtests verify under VeriPB for each.
- [x] `cmake --build --preset sanitize` clean.
- [x] `./build-sanitize/regular_test` and `./build-sanitize/regular_legacy_test` — clean under ASan + UBSan.
- [x] `ctest --preset release -j 32` — 205/205 pass.
- [x] `ctest --preset sanitize -j 32 -R regular` — 4/4 pass (incl. `regular_random` and `regular_random-seeded`).
- [x] `regular_random --prove --seed -1115540197` — verifies for all three variants (`new`, `--legacy`, `--legacy --short-reasons`).
- [x] `xcsp_regular` passes through the new `Regular` (no constructor change at the frontend).

### DIFF
--- a/dev_docs/README.md
+++ b/dev_docs/README.md
@@ -49,6 +49,12 @@ library. For an introduction to *using* the solver, start with the top-level
   in `constraints.md`; this doc spells them out for one concrete propagator
   and flags the bits that should carry across to `Disjunctive` and
   `BinPacking`.
+- [`Regular`: design and proof scaffolding](regular.md) — working-design
+  note for the upfront-DAG `Regular` propagator: layered-DAG view, OPB
+  encoding, Top-level scaffolding (per-val backward chains + statically-
+  dead-state lines), slim per-call propagator, and bench numbers vs
+  `RegularLegacy`. Cross-references the broader unification path of
+  issue #200.
 - [Proof logging for `Disjunctive`](disjunctive-proof-logging.md) — companion
   to the cumulative writeup, focused on the new bridge pattern:
   keeping the OPB encoding declarative (pairwise non-overlap clauses
--- a/dev_docs/README.md
+++ b/dev_docs/README.md
@@ -57,6 +57,12 @@ library. For an introduction to *using* the solver, start with the top-level
   in `constraints.md`; this doc spells them out for one concrete propagator
   and flags the bits that should carry across to `Disjunctive` and
   `BinPacking`.
+- [`Regular`: design and proof scaffolding](regular.md) — working-design
+  note for the upfront-DAG `Regular` propagator: layered-DAG view, OPB
+  encoding, Top-level scaffolding (per-val backward chains + statically-
+  dead-state lines), slim per-call propagator, and bench numbers vs
+  `RegularLegacy`. Cross-references the broader unification path of
+  issue #200.
 - [Proof logging for `Disjunctive`](disjunctive-proof-logging.md) — companion
   to the cumulative writeup, focused on the new bridge pattern:
   keeping the OPB encoding declarative (pairwise non-overlap clauses
--- a/dev_docs/regular.md
+++ b/dev_docs/regular.md
@@ -1,0 +1,171 @@
+# `Regular`: design and proof scaffolding
+
+Working-design note for the `Regular` propagator. The legacy
+`RegularLegacy` is preserved alongside for side-by-side benchmarking
+and as a correctness reference; new code should post `Regular`. The
+high-level approach mirrors [`mdd.md`](mdd.md) (when written) and
+[`knapsack.md`](knapsack.md) (when written) — build the layered graph
+once in `prepare()`, emit the proof scaffolding (per-val backward
+chains, statically-dead-state lines) once at `ProofLevel::Top` via
+`install_initialiser`, and run a slim per-call sweep that emits at
+most one cache-gated `~state[i][q]` line at `ProofLevel::Current`
+per state-death.
+
+For the broader unification goal across `MDD`, `Regular`, `Knapsack`
+and `BinPacking`, see issue #200.
+
+## Constraint definition
+
+A `Regular` constraint over a sequence `vars[0..n-1]` is parameterised
+by a deterministic finite automaton:
+
+- `num_states`, the number of DFA states (state `0` is the start)
+- `transitions[q][val]`, a partial map from `(state, symbol) -> next
+  state` shared across all positions
+- `final_states`, the set of accepting states
+
+The sequence `vars[0..n-1]` is accepted iff feeding the symbols from
+left to right starting at state `0` ends in a state in `final_states`.
+
+Two constructors are supported, mirroring `RegularLegacy`:
+
+- `Regular(vars, num_states, vector<unordered_map<Integer,long>>
+  transitions, final_states)` — sparse map form.
+- `Regular(vars, num_states, vector<vector<long>> transitions,
+  final_states)` — dense table form (`-1` for no transition).
+
+The `short_reasons` constructor parameter is kept on the new `Regular`
+for API parity with `RegularLegacy` and for benchmarking
+(`regular_random --legacy --short-reasons` covers the
+short-reasons-on-legacy variant). It is forwarded to the propagator
+but the new design emits at most one proof line per state-death, so
+its effect is small or zero on this path.
+
+## Layered DAG view
+
+`Regular` is structurally a layer-uniform MDD: each input position
+yields a copy of the same state set `0..num_states-1`, and the same
+transition function applies at every position. Layer `i` holds the
+DFA state after consuming the first `i` symbols, so there are
+`n + 1` layers in total.
+
+This makes `Regular` a degenerate case of `MDD` where every layer
+shares its node set and transition function. The new implementation
+exploits that fact only at the level of OPB encoding (one shared
+alphabet rather than per-layer) and the scaffolding emission loop
+(no per-layer state-count or transition lookup); the proof
+derivations are the same as in `MDD`.
+
+## OPB encoding (unchanged from `RegularLegacy`)
+
+`define_proof_model` emits:
+
+```
+state_i_is_q          (proof flag for each (i, q), i ∈ 0..n, q ∈ 0..num_states-1)
+∑_q state_i_is_q == 1  for every layer i              (exactly-one per layer)
+state_0_is_0 >= 1                                     (start state pinned)
+∑_{f ∈ final_states} state_n_is_f >= 1                (accept at the end)
+```
+
+and, for every layer `i ∈ 0..n-1`, every state `q`, and every value
+`val` in the OPB alphabet (the union of all transition keys and every
+variable's initial domain):
+
+- If `transitions[q]` has no entry for `val`:
+  `(vars[i] != val) + ~state_i_is_q >= 1`
+- Otherwise, where `q' = transitions[q][val]`:
+  `~state_i_is_q + (vars[i] != val) + state_{i+1}_is_q' >= 1`
+
+A human reading the OPB sees DFA semantics, not propagator internals.
+
+## Top-level scaffolding (new in `Regular`)
+
+`install_initialiser` emits, **once** at search root and only if proof
+logging is enabled:
+
+1. **Per-val backward chains.** For each `(i, q', val)` with
+   `val ∈ initial dom(vars[i])`:
+   ```
+   ~state_{i+1}_is_q' + (vars[i] != val) + ∑_{q : T(q,val)=q'} state_i_is_q >= 1
+   ```
+   These are RUP-derivable from the OPB forward chains plus the
+   per-layer exactly-one: assume `state_{i+1}_is_q' = 1` and
+   `vars[i] = val`; AMO at layer `i+1` forces every other
+   `state_{i+1}_*` to zero; layer-`i` ALO forces some non-parent
+   `state_i_is_q* = 1`; the OPB forward chain for `(q*, val)`
+   UP-contradicts.
+
+2. **Statically-dead-state lines.** A state is statically dead iff
+   it has no path from the root forward, or no path forward to any
+   accepting state, under initial domains. For each dead `(i, q)`
+   `Regular` emits `~state_i_is_q >= 1` at Top:
+   - Forward-unreachable nodes in ascending layer order — their RUP
+     uses the new per-val backward chains plus already-emitted
+     earlier-layer `~state_{i-1}_*` lines.
+   - The rest (forward-reachable but no path forward to a final
+     state) in descending layer order — their RUP uses the OPB
+     forward chains plus already-emitted later-layer `~state_{i+1}_*`
+     lines, plus the `final_states` ALO at layer `n`.
+
+The per-subtree `DeadCache` is pre-populated with the static-dead set
+so the per-call propagator's first emission for those states is a
+no-op.
+
+## Per-call propagator
+
+For each propagation call:
+
+1. On the first call, build the support graph (forward-reachability,
+   then accepting-state filter, then backward-reachability) under the
+   current domain. Each newly-dead `(i, q)` triggers one cache-gated
+   `~state_i_is_q >= 1` line at Current.
+2. On every call, for each value `val` that the propagator's record
+   of `states_supporting[i][val]` knows still has support but is now
+   absent from `vars[i]`'s current domain, drop the corresponding
+   edges; the resulting `out_deg`/`in_deg` reductions cascade through
+   `decrement_outdeg` / `decrement_indeg`, each step optionally
+   emitting one cache-gated `~state_i_is_q >= 1` line at Current when
+   a state's last edge disappears.
+3. Final inferences: for every `(i, val)` where
+   `states_supporting[i][val]` is empty, infer `vars[i] != val` with
+   `JustifyUsingRUP{}` against the generic reason. The OPB forward
+   chains plus the Top-level backward chains plus the cached
+   `~state_*_*` lines let UP close it without any per-call
+   intermediate emissions.
+
+The contrast with `RegularLegacy` is exactly the same as the
+MDD/Knapsack contrast: the legacy implementation emitted per-(parent
+state, val) intermediate aggregations on every propagation call;
+`Regular` does that work once at root.
+
+## Benchmark (regular_random)
+
+`regular_random --legacy [--short-reasons]` exposes all three
+variants from one binary. Sampled at the seed already hard-coded into
+`add_test(regular_random-seeded ...)`:
+
+| n  | variant            | solve | proof    | verify |
+|---:|--------------------|------:|---------:|-------:|
+| 10 | new                | 0.00s | 86 KB    | 0.00s  |
+| 10 | legacy             | 0.01s | 1.76 MB  | 0.03s  |
+| 10 | legacy + short-rsn | 0.00s | 454 KB   | 0.04s  |
+| 14 | new                | 0.00s | 230 KB   | 0.02s  |
+| 14 | legacy             | 0.06s | 8.76 MB  | 0.13s  |
+| 14 | legacy + short-rsn | 0.01s | 1.75 MB  | 0.24s  |
+| 18 | new                | 0.01s | 490 KB   | 0.06s  |
+| 18 | legacy             | 0.25s | 33.3 MB  | 0.51s  |
+| 18 | legacy + short-rsn | 0.04s | 4.95 MB  | 0.98s  |
+
+The new `Regular` produces proofs that are 4–68× smaller than the
+legacy default and 1.5–10× smaller than the legacy with
+`short_reasons` enabled, and VeriPB verifies them faster on every
+sampled size. This matches the MDD-reimpl shape (proof shrink + slim
+per-call); it doesn't follow the `BinPacking` regression because
+`RegularLegacy`'s per-call work is far from already-lean.
+
+## #200 follow-up
+
+`Regular` and `MDD` now share the same Top-scaffolding shape modulo
+layer-uniform vs per-layer transitions. Folding both into a single
+layered-DAG scaffolder is the natural next step; that's the scope of
+issue #200.
--- a/dev_docs/regular.md
+++ b/dev_docs/regular.md
@@ -1,15 +1,50 @@
 # `Regular`: design and proof scaffolding
 
-Working-design note for the `Regular` propagator. The legacy
-`RegularLegacy` is preserved alongside for side-by-side benchmarking
-and as a correctness reference; new code should post `Regular`. The
-high-level approach mirrors [`mdd.md`](mdd.md) (when written) and
+Working-design note for the `Regular` propagator. The high-level
+approach mirrors [`mdd.md`](mdd.md) (when written) and
 [`knapsack.md`](knapsack.md) (when written) — build the layered graph
 once in `prepare()`, emit the proof scaffolding (per-val backward
 chains, statically-dead-state lines) once at `ProofLevel::Top` via
 `install_initialiser`, and run a slim per-call sweep that emits at
 most one cache-gated `~state[i][q]` line at `ProofLevel::Current`
 per state-death.
+
+## Default: the upfront `Regular`
+
+**`Regular` (this upfront-scaffolding propagator) is the default.**
+Everything posts it — both frontends
+(`xcsp_glasgow_constraint_solver`, `fzn_glasgow`), the constraint
+tests, and the `regular_random` example. `RegularLegacy` — the older
+per-call propagator that re-emits per-`(parent state, val)`
+intermediate aggregations on every propagation call — is preserved
+only as the `--legacy` opt-in twin, for side-by-side benchmarking and
+as a correctness reference.
+
+The reason for defaulting to upfront is that it **wins on both axes
+that matter for proof logging**, measured within-branch (same search
+tree to the digit — only the propagator is toggled):
+
+- **Proof size: 13–55× smaller.** At the largest sampled instance the
+  upfront `Regular` produced a 9 MB proof against `RegularLegacy`'s
+  496 MB (7660 solutions).
+- **VeriPB verification time: 2.3–7× faster** (1.66 s vs 11.69 s on
+  that instance).
+
+Both effects grow with search size: the fixed root scaffold amortises,
+and the displaced per-call volume compounds. (At trivially small search
+— tens of solutions — the fixed root scaffold is not yet amortised, so
+upfront can be marginally bigger/slower; the crossover is well below
+any real instance.)
+
+This is the *narrow-diagram* case of the general decision-diagram
+proof-strategy analysis: because a `Regular` automaton is narrow, the
+permanent Top-level scaffold is tiny, so it adds almost no per-line
+"DB-tax" to the rest of the proof, while the per-call `RegularLegacy`
+baseline is very verbose (large displacement). Displacement far
+exceeds the tax, so upfront wins size *and* time. See
+[`decision-diagram-proof-strategies.md`](decision-diagram-proof-strategies.md)
+for the cost model, the predictive rule, and how the same analysis
+lands MDD, Knapsack and BinPacking on different defaults.
 
 For the broader unification goal across `MDD`, `Regular`, `Knapsack`
 and `BinPacking`, see issue #200.
--- a/examples/regular_random/regular_random.cc
+++ b/examples/regular_random/regular_random.cc
@@ -35,7 +35,7 @@ auto index_of(const IntegerVariableID & val, const vector<IntegerVariableID> & v
     return (int)pos;
 }
 
-auto post_random_regular(Problem & p, const int & n, mt19937 & rng, bool short_reasons)
+auto post_random_regular(Problem & p, const int & n, mt19937 & rng, bool short_reasons, bool legacy)
 {
     stringstream string_rep;
 
@@ -78,7 +78,10 @@ auto post_random_regular(Problem & p, const int & n, mt19937 & rng, bool short_r
         }
     }
 
-    p.post(Regular{x, num_states, transitions, final_states, short_reasons});
+    if (legacy)
+        p.post(RegularLegacy{x, num_states, transitions, final_states, short_reasons});
+    else
+        p.post(Regular{x, num_states, transitions, final_states, short_reasons});
 }
 
 auto main(int argc, char * argv[]) -> int
@@ -94,7 +97,8 @@ auto main(int argc, char * argv[]) -> int
             ("n", "Number of variables in the sequence", cxxopts::value<int>()->default_value("6"))                                //
             ("stats", "Print stats, including solve time")                                                                         //
             ("proof-files-basename", "Alternative path and name for the proof", cxxopts::value<string>()->default_value("regular_random")) //
-            ("short-reasons", "Use short reasons method");
+            ("short-reasons", "Use short reasons method (RegularLegacy only)")                                                      //
+            ("legacy", "Post RegularLegacy instead of the new Regular");
 
         options_vars = options.parse(argc, argv);
     }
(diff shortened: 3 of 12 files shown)


mmcilree:
```
﻿./run_test_and_verify.bash build/regular_random -n 10 --seed 88
```

mmcilree (changes requested):
Not convinced at all by this. Want me to review?

ciaranm:
Yes please.

ciaranm:
> ```
> ﻿./run_test_and_verify.bash build/regular_random -n 10 --seed 88
> ```

Annoyingly I can't reproduce this one, but seed 56 is consistently failing for me.

ciaranm:
And 230, 289, 303. Hopefully this is enough to narrow things down.

ciaranm:
Claude believes it was a "RUP in the wrong order" problem. I've asked it to do further checks across Regular, MDD, Knapsack, and BinPacking.

mmcilree:
Okay, I've done some more thinking about this, and I think (despite my best efforts to break it)... I agree with Claude that this level of proof scaffolding is in fact, sufficient, although I think it could be explained better why the RUP checks will always succeed. I wonder if it can write a meta-proof in the style of my thesis...? 

Here's my thinking aloud for what it's worth.

The basic difference is that instead of maintaining the invariant:
- Every explicitly removed edge is justified with 
$\mathcal R \Rightarrow \bar s_{i = q}  + \bar x_{i = j} \geq 1$ 

Claude maintains the invariant
- Every explicitly removed state is justified with 
$\mathcal R \Rightarrow \bar s_{i = q} \geq 1$ (2)

I had to think a bit about why these are equivalent, but basically, the regular propagator only removes edges when 
1. all incoming edges are removed => all outgoing edges are removed => state is removed 
2. all outgoing edges are removed => all incoming edges are removed => state is removed 
3. the value labelling the edge is removed => state may not be removed, but this is OK

I thought that 3. might cause problems, but when it comes to actually justifying inferences, any value still present with no supporting edges must by definition only be lacking edges due to 1 and 2, (or disallowed transitions). So, providing all the necessary (2) are logged, we have for each edge represented by 

$s_{i = q} \land x_{i = j}$ 

either

1. $\bar s_{i = q} \geq 1$
is a removed state and has been logged => $\bar s_{i = q}$ propagates

2. $s_{i = q} \land x_{i = q} \Rightarrow 0 \geq 1$ is in the OPB (disallowed transition) => $\bar s_{i = q}$ propagates

3. Otherwise,
$s_{i = q} \land x_{i = q} \Rightarrow s_{i = r} \geq 1$ exists, and since $s_{i = q}$ is not a removed state, $s_{i = r}$ must be a removed state (otherwise this would be a supporting edge), and so $\bar s_{i = r} \geq 1$ has already been logged => $\bar s_{i = q}$ propagates

The edge being removed due to $x_{i = j}$ having been pruned is not possible since by definition the value has not been removed yet. 

So in all cases $\bar s_{i = q}$ propagates and hence we contradict the AL1 for these. 

Then in terms of seeing why each $\bar s_{i = q} \geq 1$ is RUP: there are two cases
1. $s_{i = q}$ has lost all of its incoming edges (forwards pass). This means for any value of 
$X_{i - 1}$, either $X_{i-1}= v$ has been pruned (and hence $\bar x_{i -1 = v}$ propagates), or else
$\bar s_{i -1 = r}$ has been logged for every $r$ where $\delta(r, v) = q$. 
Since at the top 
of the proof, we logged
$x_{i - 1 = v} \land s_{i = q} \Rightarrow \sum_{r : \delta(r, v) = q} s_{i = r} \geq 1$
the latter case will propagate $\bar x_{i = v}$ and so we reach a contradiction by domain-wipeout.
2. $s_{i = q}$ has lost all of its outgoing edges (backwards pass). This mean for any value of $X_i$ either $X_{i} = v$ has been pruned (and hence $\bar x_{i = v}$ propagates) or else $\delta(q, v)$ is a disallowed transition and hence $x_{i = v}$ propagates, or else 
$\bar s_{i + 1 = r}$ has been logged where $\delta(q, v) = r$. 
In the latter case, $\bar x_{i = v}$ propagates due to the formula constraint 
$x_{i = v} \land s_{i = q} \Rightarrow s_{i + 1 = r}$.

This is basically the same recursive justification as we had previously for edge removals, except instead of having the per-previous state intermediate steps ("scaffolding") for the forwards pass during propagation, it is done as an initialiser. 

So all that's left is to think carefully about why the top-level 
$x_{i = v} \land s_{i + 1 = q} \Rightarrow \sum_{r : \delta(r, v) = q} s_{i = r} \geq 1$
will always be RUP, but I think I'm convinced: any $s_{i = r}$ where $\delta(r, v) = q$ are already excluded by negation, and if $\delta(r, v) = p \neq q$ then we have 
$s_{i = r} \land x_{i = v} \Rightarrow s_{i + 1 = p}$ which propagates $\bar s_{i = r}$ due to $s_{i + 1 = q}$
or else if $\delta(r, v)$ is disallowed we have 
$s_{i = r} \land x_{i = v} \Rightarrow 0 \geq 1$
which also propagates $s_{i + 1 = q}$.

mmcilree (approved):

ciaranm:
# Justifying the Upfront-Scaffolding `Regular` Propagator

A meta-proof in the style of §5.1 of Matthew McIlree's PhD thesis, adapted to the
new `Regular` design in PR #213. We keep the OPB encoding from
Encoding Procedure 5.1 of the thesis, and we keep the high-level idea — that
domain-consistent prunings made by `Regular` are justified by maintaining a
proof-side invariant about the support graph — but we replace the *edge-level*
invariant of Justification Procedure 5.1 with a strictly weaker *state-level*
invariant, supplemented by a one-shot top-level initialiser that supplies a
family of per-symbol backward chains.

We use the notation of the thesis throughout:

- $X_i$ ($i = 1, \dots, n$) the constraint's CP variables, $\Sigma$ the
  alphabet, $Q$ the DFA states, $\delta$ the partial transition function,
  $q_0$ the start state, $F$ the accepting states.
- $s_{i,q}$ the proof flag "after consuming the first $i$ symbols, the DFA is
  in state $q$" (so layers $0, \dots, n$).
- $x_{i=v}$, $\overline{x}_{i=v}$ the equality literal and its negation.
- $\mathrm{dom}_0(X_i)$, $\mathrm{dom}_t(X_i)$ the initial and time-$t$ domain.
- $\mathrm{GenericR}_t(\mathcal X)$ as in §3.4.1 (the bound and equality literals
  describing the current domain).
- $\mathrm{dom}_R(X)$, $\mathrm{ImpLits}$, Theorems 3.2 and 3.3 as in §3.3.1.

We assume throughout that $\mathrm{Regular}(\mathcal X, M)$ has been encoded as
in Encoding Procedure 5.1, so the OPB contains, for every layer $i$, the
per-layer ALO $\sum_{q \in Q} s_{i,q} \geq 1$ and AMO
$-\sum_{q \in Q} s_{i,q} \geq -1$; the root pin $s_{0,q_0} \geq 1$; the accept
constraint $\sum_{f \in F} s_{n,f} \geq 1$; the forward chain
$\overline{s}_{i,q} + \overline{x}_{i=v} + s_{i+1, \delta(q,v)} \geq 1$ for
every $(q, v)$ with $\delta(q,v)$ defined; and the disallowed-transition
constraint $\overline{s}_{i,q} + \overline{x}_{i=v} \geq 1$ for every $(q, v)$
with $\delta(q,v)$ undefined.

## 5.1' Strategy: a state-level proof invariant

Justification Procedure 5.1 of the thesis maintains an *edge-level* invariant:
for every explicitly removed edge $(q^i_k, q^{i+1}_l)$ labelled with $v$, the
proof contains an edge-deletion line
$$ \mathcal R_e \;\Rightarrow\; \overline{s}_{i,k} + \overline{x}_{i=v} \geq 1. \qquad \text{(Edge-Inv)} $$

We instead maintain the strictly weaker *state-level* invariant: for every
"explicitly removed state" $(i, q)$, the proof contains
$$ \mathcal R_t \;\Rightarrow\; \overline{s}_{i,q} \geq 1. \qquad \text{(State-Inv)} $$

Call any $(i, q)$ for which (State-Inv) has been derived at or before time $t$ a
*logged dead state at time* $t$. Logged dead states form a monotonically growing
set along any branch of the search tree (Top-level lines are unconditionally
present; Current-level lines persist until the relevant decision is undone).

The point is that (Edge-Inv) implies (State-Inv): an edge can be removed for
three reasons (cf. §5.1.2 of the thesis), and in cases (1) "all incoming edges
gone" and (2) "all outgoing edges gone", the propagator does explicitly remove
the state; case (3) "labelling value pruned from $\mathrm{dom}_t(X_i)$" does not remove
the state, but in that case the equality literal $\overline{x}_{i=v}$ is
already in $\mathrm{GenericR}_t(X_i)$, so any consumer that needed
(Edge-Inv) for that edge can substitute $\overline{x}_{i=v}$ directly. We make
this precise in Justification Procedures 5.1'.4 below.

The price of the weaker invariant is that some derivations that were RUP under
(Edge-Inv) are no longer RUP under (State-Inv) alone. We pay this price once,
at the root of the search tree, by emitting an *Initialiser* that supplies a
family of *per-symbol backward chains*.

## Initialiser Procedure 5.1' (Top-Level Scaffolding for Regular)

**Preconditions.** $\mathrm{Regular}(\mathcal X, M)$ encoded as in
Encoding Procedure 5.1; $\mathrm{dom}_0(X_i)$ the initial domain of each $X_i$.

**Procedure.**

*Part A (per-symbol backward chains).* For every $i \in \{0, \dots, n-1\}$,
every $q' \in Q$, and every $v \in \mathrm{dom}_0(X_i)$:

  1: $\mathbf{rup} \quad \overline{s}_{i+1, q'} + \overline{x}_{i=v} + \sum_{q : \delta(q,v) = q'} s_{i,q} \;\geq\; 1 \quad (\star)$

(When no $q$ satisfies $\delta(q, v) = q'$ the empty sum is zero and $(\star)$
collapses to $\overline{s}_{i+1, q'} + \overline{x}_{i=v} \geq 1$.)

*Part B (statically dead states).* Compute under $\mathrm{dom}_0$ the set
$\mathrm{Dead}_0 \subseteq \{0, \dots, n\} \times Q$ of *statically dead states*:
$(i, q) \in \mathrm{Dead}_0$ iff either there is no path from $(0, q_0)$ to
$(i, q)$ in the layered graph under $\mathrm{dom}_0$ (forward-unreachable), or there is
no path from $(i, q)$ to any $(n, f)$ with $f \in F$
(no-path-forward-to-accepting). Partition
$\mathrm{Dead}_0 = \mathrm{Dead}_0^{\mathrm{fwd}} \cup \mathrm{Dead}_0^{\mathrm{bwd}}$
where $\mathrm{Dead}_0^{\mathrm{fwd}}$ is the forward-unreachable part.

  Part B1: for $i = 1, 2, \dots, n$, in ascending order, and for each
  $q$ with $(i, q) \in \mathrm{Dead}_0^{\mathrm{fwd}}$:
  
  2: $\mathbf{rup}\quad \overline{s}_{i,q} \;\geq\; 1$

  Part B2: for $i = n, n-1, \dots, 0$, in descending order, and for each
  $q$ with $(i, q) \in \mathrm{Dead}_0^{\mathrm{bwd}}$:
  
  3: $\mathbf{rup}\quad \overline{s}_{i,q} \;\geq\; 1$

All lines are emitted at proof level $\mathrm{Top}$.

**Correctness Proof.**

*Part A.* Fix $(i, q', v)$ and consider the negation of $(\star)$, which
propagates $s_{i+1, q'}$, $x_{i=v}$, and $\overline{s}_{i, q}$ for every $q$
with $\delta(q, v) = q'$. We show this leads to contradiction by unit
propagation against the OPB.

By the per-layer AMO $-\sum_{p \in Q} s_{i+1, p} \geq -1$, the assumption
$s_{i+1, q'}$ propagates $\overline{s}_{i+1, p}$ for every $p \neq q'$.
Now fix any $q^* \in Q$ with $\delta(q^*, v) \neq q'$ (in the sense of either
"defined and equal to some $p \neq q'$" or "undefined"). Two subcases:

- If $\delta(q^*, v)$ is undefined, the disallowed-transition constraint
  $\overline{s}_{i, q^*} + \overline{x}_{i=v} \geq 1$ reduces under $x_{i=v}$ to
  $\overline{s}_{i, q^*} \geq 1$, so $\overline{s}_{i, q^*}$ propagates.
- If $\delta(q^*, v) = p$ with $p \neq q'$, the forward chain
  $\overline{s}_{i, q^*} + \overline{x}_{i=v} + s_{i+1, p} \geq 1$ reduces under
  $x_{i=v}$ and $\overline{s}_{i+1, p}$ (already propagated by AMO) to
  $\overline{s}_{i, q^*} \geq 1$, so $\overline{s}_{i, q^*}$ propagates.

Thus $\overline{s}_{i, q^*}$ propagates for every $q^* \in Q$ with
$\delta(q^*, v) \neq q'$. Combined with the assumed $\overline{s}_{i, q}$ for
every $q$ with $\delta(q, v) = q'$, the per-layer ALO
$\sum_{q \in Q} s_{i, q} \geq 1$ has all its literals falsified and propagates
to contradiction.

*Part B1, base case $i = 1$.* We use AMO together with the root pin
$s_{0, q_0} \geq 1$: by the same argument as in Part A, AMO at layer $0$
propagates $\overline{s}_{0, q}$ for every $q \neq q_0$. So whenever $(1, q)$
is forward-unreachable, every parent $(0, q')$ with $\delta(q', v) = q$ for
some $v \in \mathrm{dom}_0(X_0)$ must have $q' \neq q_0$ (otherwise $(1, q)$ would be
reachable), so $\overline{s}_{0, q'}$ is already propagated. Now consider the
negation of $\overline{s}_{1, q} \geq 1$, which gives $s_{1, q}$. For every
$v \in \mathrm{dom}_0(X_0)$, the backward chain $(\star)$ at $(0, q, v)$ reduces under
$s_{1, q}$ to
$\overline{x}_{0=v} + \sum_{q' : \delta(q', v) = q} s_{0, q'} \geq 1$.
Every $q'$ in the sum has $\overline{s}_{0, q'}$ already propagated, so
$\overline{x}_{0=v}$ propagates. This holds for every $v \in \mathrm{dom}_0(X_0)$,
hence $\mathrm{dom}_R(X_0) = \emptyset$ and Theorem 3.2 gives the conflict.

*Part B1, inductive step at layer $i \geq 2$.* By induction, every
$(i - 1, q')$ in $\mathrm{Dead}_0^{\mathrm{fwd}}$ has $\overline{s}_{i-1, q'}$
already in the proof. We claim that if $(i, q)$ is forward-unreachable, then
every $q'$ with $\delta(q', v) = q$ for some $v \in \mathrm{dom}_0(X_{i-1})$ is itself
forward-unreachable at layer $i - 1$: otherwise, extending a witness path to
$(i - 1, q')$ by $v$ would witness reachability of $(i, q)$. So the same
argument as the base case applies: $(\star)$ at $(i - 1, q, v)$, together with
$s_{i, q}$ and the already-derived $\overline{s}_{i-1, q'}$ for every $q'$ in
the sum, propagates $\overline{x}_{i-1 = v}$. Doing this for every
$v \in \mathrm{dom}_0(X_{i-1})$ wipes out the domain of $X_{i-1}$ and Theorem 3.2
gives the conflict.

*Part B2, base case $i = n$.* $(n, q) \in \mathrm{Dead}_0^{\mathrm{bwd}}$ means
$(n, q)$ is forward-reachable but $q \notin F$. Assume $s_{n, q}$; AMO at
layer $n$ propagates $\overline{s}_{n, p}$ for every $p \neq q$, in particular
for every $f \in F$ (since $q \notin F$). The accept constraint
$\sum_{f \in F} s_{n, f} \geq 1$ is then violated by unit propagation.

*Part B2, inductive step at layer $i < n$.* By induction (reverse), every
$(i', q')$ in $\mathrm{Dead}_0$ with $i' > i$ has $\overline{s}_{i', q'}$ in
the proof, irrespective of which sub-part of $\mathrm{Dead}_0$ it belongs to.
Assume $s_{i, q}$ for $(i, q) \in \mathrm{Dead}_0^{\mathrm{bwd}}$. For every
$v \in \mathrm{dom}_0(X_i)$:
- if $\delta(q, v)$ is undefined, the disallowed-transition constraint
  propagates $\overline{x}_{i=v}$ under $s_{i, q}$;
- if $\delta(q, v) = q'$, then $(i+1, q')$ has no path forward to accepting
  (otherwise so would $(i, q)$), so $(i+1, q') \in \mathrm{Dead}_0$ and by the
  induction hypothesis $\overline{s}_{i+1, q'}$ is in the proof; the forward
  chain $\overline{s}_{i, q} + \overline{x}_{i=v} + s_{i+1, q'} \geq 1$ then
  propagates $\overline{x}_{i=v}$.

Hence $\overline{x}_{i=v}$ propagates for every $v \in \mathrm{dom}_0(X_i)$, domain
wipeout, Theorem 3.2.   $\Box$

A noteworthy property of the Initialiser Procedure 5.1' is that it does not
depend on the propagator's internal state at all, only on the OPB encoding and
$\mathrm{dom}_0$. Consequently every line it emits is unconditional and may be reused
along every branch of the search tree.

## Justification Procedures 5.1' (per-call state deaths and value pruning)

The new design splits the work of Justification Procedure 5.1 into three small
procedures, each of which is a single RUP step justified by Initialiser
Procedure 5.1' together with the cumulative (State-Inv) lines previously
derived on the current search path.

We will say that a logged dead state $(i, q)$ from time $t' \leq t$ is
*available at time $t$* iff its reason $\mathcal R_{t'}$ is dominated by
$\mathcal R_t$. Whenever both $\mathcal R_{t'}$ and $\mathcal R_t$ are
$\mathrm{GenericR}$ functions evaluated on the same search path, with $t' \leq t$,
monotonicity of search guarantees
$\mathrm{dom}_{\mathcal R_t} \sqsubseteq \mathrm{dom}_{\mathcal R_{t'}}$, and by Theorem 3.3
every literal of $\mathcal R_{t'}$ propagates under $\mathcal R_t$; we use this
without further comment below.

### Justification Procedure 5.1'.1 (Backward-Pass State Death)

**Preconditions.** $i \in \{0, \dots, n-1\}$; in the propagator's support graph,
the state $(i, k)$ has just lost its last outgoing edge; for every
$v \in \mathrm{dom}_t(X_i)$ with $\delta(k, v) = l$ defined, the state $(i+1, l)$ is
available as a logged dead state at time $t$.

**Procedure.**
  1: $\mathcal R \leftarrow \mathrm{GenericR}_t(\mathcal X)$
  
  2: $\mathbf{rup} \quad \mathcal R \;\Rightarrow\; \overline{s}_{i, k} \geq 1$

**Correctness Proof.** The negation of line 2 propagates $\mathcal R$ and
$s_{i, k}$. For every value $v$ in the OPB alphabet, either
$v \notin \mathrm{dom}_t(X_i)$ and $\overline{x}_{i=v}$ propagates directly from
$\mathcal R$, or $v \in \mathrm{dom}_t(X_i)$ and we split:

- If $\delta(k, v)$ is undefined, the disallowed-transition constraint with
  $s_{i, k}$ propagates $\overline{x}_{i=v}$.
- If $\delta(k, v) = l$, by precondition $(i+1, l)$ is available, so
  $\overline{s}_{i+1, l}$ propagates by Theorem 3.3, and then the forward chain
  with $s_{i, k}$ propagates $\overline{x}_{i=v}$.

Hence $\mathrm{dom}_R(X_i) = \emptyset$ and Theorem 3.2 gives the conflict.   $\Box$

### Justification Procedure 5.1'.2 (Forward-Pass State Death)

**Preconditions.** $i \in \{1, \dots, n\}$; in the propagator's support graph,
the state $(i, k)$ has just lost its last incoming edge; for every
$v \in \mathrm{dom}_t(X_{i-1})$ and every $q \in Q$ with $\delta(q, v) = k$, the state
$(i - 1, q)$ is available as a logged dead state at time $t$.

**Procedure.**
  1: $\mathcal R \leftarrow \mathrm{GenericR}_t(\mathcal X)$
  
  2: $\mathbf{rup} \quad \mathcal R \;\Rightarrow\; \overline{s}_{i, k} \geq 1$

**Correctness Proof.** As before, negate to get $\mathcal R \wedge s_{i, k}$.
For every value $v$ in the OPB alphabet, either $v \notin \mathrm{dom}_t(X_{i-1})$ and
$\overline{x}_{i-1=v}$ propagates from $\mathcal R$, or $v \in \mathrm{dom}_t(X_{i-1})$
and the per-symbol backward chain $(\star)$ from Part A of the Initialiser
Procedure, namely
$$\overline{s}_{i, k} + \overline{x}_{i-1=v} + \sum_{q : \delta(q, v) = k} s_{i-1, q} \geq 1,$$
reduces under $s_{i, k}$ to
$\overline{x}_{i-1=v} + \sum_{q : \delta(q, v) = k} s_{i-1, q} \geq 1$.
By precondition every $(i-1, q)$ in the sum is available, so
$\overline{s}_{i-1, q}$ propagates for every such $q$, hence
$\overline{x}_{i-1=v}$ propagates.

Domain wipeout on $X_{i-1}$, Theorem 3.2.   $\Box$

Observe that this is exactly where Initialiser Procedure 5.1'.A "pays for"
the weaker (State-Inv): the backward chain $(\star)$ is the proof-side
substitute for the missing edge-deletion lines that (Edge-Inv) would have
provided.

### Justification Procedure 5.1'.3 (Value Pruning)

**Preconditions.** A variable-value pair $(X_i, v)$ has no support in the
propagator's support graph at time $t$ (no edge $(q^i, q^{i+1}_l)$ labelled
with $v$ for any $q, l$); for every $q \in Q$ such that $\delta(q, v)$ is
defined, $(i, q)$ is available as a logged dead state at time $t$.

**Procedure.**
  1: $\mathcal R \leftarrow \mathrm{GenericR}_t(\mathcal X)$
  
  2: $\mathbf{rup} \quad \mathcal R \;\Rightarrow\; \overline{x}_{i=v} \geq 1$

**Correctness Proof.** Negate to get $\mathcal R \wedge x_{i=v}$. We show that
$\overline{s}_{i, q}$ propagates for every $q \in Q$.

- If $\delta(q, v)$ is undefined, the disallowed-transition constraint
  $\overline{s}_{i, q} + \overline{x}_{i=v} \geq 1$ with $x_{i=v}$ propagates
  $\overline{s}_{i, q}$.
- If $\delta(q, v)$ is defined, by precondition $(i, q)$ is available, so
  $\overline{s}_{i, q}$ propagates by Theorem 3.3.

The per-layer ALO $\sum_{q \in Q} s_{i, q} \geq 1$ is then violated and unit
propagation closes the proof.   $\Box$

### Justification Procedure 5.1'.4 (Value Pruning, Edge-Form Bridge)

This procedure does not appear in the propagator code but is implicit in the
discussion above. It witnesses formally that (State-Inv) suffices in place of
(Edge-Inv) for *any* value-pruning context that (Edge-Inv) would have served.

**Preconditions.** $\delta(k, v)$ defined and equal to $l$; either
(a) $\overline{x}_{i=v} \in \mathrm{GenericR}_t(\mathcal X)$, or
(b) $(i, k)$ is available as a logged dead state at time $t$, or
(c) $(i+1, l)$ is available as a logged dead state at time $t$.

**Procedure.**
  1: $\mathcal R \leftarrow \mathrm{GenericR}_t(\mathcal X)$
  
  2: $\mathbf{rup} \quad \mathcal R \;\Rightarrow\; \overline{s}_{i, k} + \overline{x}_{i=v} \geq 1$

**Correctness Proof.** Negate to get $\mathcal R \wedge s_{i, k} \wedge x_{i=v}$.
Case (a): $\overline{x}_{i=v} \in \mathcal R$ then propagates and contradicts
$x_{i=v}$. Case (b): $\overline{s}_{i, k}$ propagates by Theorem 3.3 and
contradicts $s_{i, k}$. Case (c): $\overline{s}_{i+1, l}$ propagates by
Theorem 3.3, the forward chain
$\overline{s}_{i, k} + \overline{x}_{i=v} + s_{i+1, l} \geq 1$ has its first
two literals falsified by $s_{i, k}$ and $x_{i=v}$, and then $s_{i+1, l}$ is
forced true, contradicting the propagated $\overline{s}_{i+1, l}$.   $\Box$

So whenever an edge $(q^i_k, q^{i+1}_l)$ has been "explicitly removed" in the
sense of §5.1.2 of the thesis, one of (a)–(c) holds, and Procedure 5.1'.4
recovers the edge-deletion line (Edge-Inv) by a single RUP step against the
proof state we have already built. We never actually invoke 5.1'.4 in the
implementation because Procedures 5.1'.1 through 5.1'.3 are sufficient to
justify all of the propagator's prunings directly.

## Maintenance discipline

To ensure that the preconditions of 5.1'.1, 5.1'.2 and 5.1'.3 are met, the
propagator imposes a topological emission order:

- Initialiser Procedure 5.1' is invoked once at the root, before the propagator
  runs. Parts B1 and B2 are themselves topologically ordered with respect to
  each other (B1 ascends, B2 descends), and within each part, layer-by-layer in
  the chosen direction. The per-subtree dead-state cache is then pre-populated
  with $\mathrm{Dead}_0$, so the per-call propagator never re-emits a static
  death.

- On a backward-pass state death (Procedure 5.1'.1), the propagator emits the
  $\overline{s}_{i, k}$ line *before* recursing into any parent
  $(i - 1, l)$ whose out-degree may drop to zero as a consequence of $(i, k)$'s
  death: this is the "emit before recurse" invariant of `decrement_outdeg`.
  When the recursion reaches a parent and invokes 5.1'.1 on it, $(i, k)$ is
  already available, so the precondition is satisfied for the child slot.

- Symmetrically, on a forward-pass state death (Procedure 5.1'.2), the
  propagator emits $\overline{s}_{i, k}$ before recursing into any child whose
  in-degree may drop to zero, so when the recursion reaches a child the
  parent slot is already available. This is the corresponding invariant of
  `decrement_indeg`.

- Finally, value pruning (Procedure 5.1'.3) is invoked only after the support
  graph has reached its current fixpoint, by which time the entire wave of
  state-deaths cascading from the current change has been emitted.

Together with the dead-state cache, this discipline guarantees by simple
induction on the order of emission that whenever a line is emitted, every
logged dead state its RUP relies on is already in the proof and available
under the current reason. The proof obligations of Procedures 5.1'.1, 5.1'.2
and 5.1'.3 are therefore met, and by Procedure 5.1'.4 the resulting state-level
invariant suffices to certify every domain-consistent pruning the propagator
might make.

## Cost

Let $D = \max_i |\mathrm{dom}_0(X_i)|$. Initialiser Procedure 5.1' emits
$O(n |Q|^2 D)$ lines in Part A (although the empty-sum and disallowed-transition
collapses make the constant small), and $O(n |Q|)$ lines in Part B. Each
per-call state death emits exactly one line at proof level $\mathrm{Current}$,
and each value pruning one line at $\mathrm{Current}$. By contrast,
Justification Procedure 5.1 of the thesis emits an edge-deletion line per
edge-removal and a $|Q|$-step inner loop for the forward-pass case, paid on
every propagation call. The trade-off is therefore: more lines at the root
(amortised across the whole search tree), and at most one line per propagation
event, with no per-(parent-state, value) intermediates.

PR #216, opened 2026-05-20
regular: add RegularBacchus (issue #215)

### LINKED ISSUE #215: Regular Proof Logging Based on Bacchus Encoding
Based on Bacchus' CP 2007 paper "[GAC via unit propagation](https://www.cs.toronto.edu/~fbacchus/Papers/B-CP2007.pdf)", if we used the following encoding, then domain consistency should follow by plain unit propagation on the OPB file and inferences for regular would require no justification at all. 

```
allowedTransitions(delta) = { (q, v) in Q x Sigma : delta(q, v) is defined }

for i in {0, ..., n-1}:
  for q in Q:
    for v in Sigma such that (q, v) in allowedTransitions(delta):
      DEFINE t[i][q][v] => x[i]=v + state[i]=q + state[i+1]=delta(q,v) >= 3

    if i > 0:
      inTransitions = { t[i-1][p][v] : (p, v) in allowedTransitions and delta(p,v) = q }
      DEFINE state[i]=q => sum_{t[i-1][p][v] in inTransitions} t[i-1][p][v] >= 1

    outTransitions = { t[i][q][v] : (q, v) in allowedTransitions }
    DEFINE state[i]=q => sum_{t[i][q][v] in outTransitions} t[i][q][v] >= 1

  for v in Sigma:
    DEFINE x[i]=v => sum_{q : (q,v) in allowedTransitions} t[i][q][v] >= 1

DEFINE sum_{i=1}^{|Q|} ~state[i]=0 >= |Q| - 1
DEFINE sum_{i in Q\F} ~state[n]=i >= |Q\F|
```

Try the following:
1. Using this encoding directly via mode->add_constraint() and replace all regular justifications with NoJustificationNeeded and test/benchmark it compared with current Regular (both the original and Claude-improved #213 versions)
2. Derive this encoding from the current encoding as an initialiser step, and then use NoJustificationNeeded and benchmark this as well against the other version. Hints: 
- the transition variable definitions (t[i][q][v] => ) can be introduced as auxiliary/extension variables via redundance.
- The AL1s for inTransitions and outTransitions and variable supports are probably derivable via resolution using AL1s over the state[i]=q vars and the x_[i]=j vars along with the current (state[i]=q /\ x[i]=v =>) lines
- The last two constraints are probably just RUP

ciaranm:
## Summary

Resolves #215. Adds `RegularBacchus` as a sibling to `Regular` and
`RegularLegacy`. The OPB is unchanged from `Regular` (natural DFA
encoding); the initialiser derives a Bacchus transition-variable
encoding so that the per-call propagator emits zero proof lines (every
inference is `NoJustificationNeeded`).

Step 1 of the issue (Bacchus encoding directly in the OPB +
NoJustificationNeeded everywhere) was implemented locally as a sanity
check on the encoding then reverted — the OPB stays natural in the
landed code.

For the step-2 derivation:
- Each `t[i][q][v]` flag created via `create_proof_flag_reifying`
  (redundance) at `ProofLevel::Top`.
- One `pol(forward_chain + reverse_t_def)` per `(i, q, v ∈ dom_q)` to
  derive the per-`(q, v)` intermediate that the AL1 RUPs need —
  VeriPB's UP collapses the complementary `state[i+1][δ] +
  ~state[i+1][δ]` pair during the RUP check.
- Outgoing AL1, incoming AL1, variable-support AL1 via RUP at
  `ProofLevel::Top`.

Also added a `--all` flag to `regular_random` that enumerates every
accepting string instead of stopping at the first solution. This was
needed to actually exercise per-call work — without `--all`, random
DFAs resolve in `n+1` recursions and almost no propagation.

### Decision / shallow search (default, no `--all`)

`rec` = `recursions`, `props₀` = first integer of `propagations:`
from `--stats` (identical across variants since they make the same
propagator decisions). All veripb-clean.

| n  | seed         | rec | props₀ | variant | solve | verify | opb     | pbp     |
|---:|-------------:|----:|-------:|---------|------:|-------:|--------:|--------:|
| 10 |  1           |  11 |     17 | Regular | 0.00s | 0.01s  |  ~      |  167 KB |
| 10 |  1           |  11 |     17 | Bacchus | 0.00s | 0.06s  |  184 KB |  508 KB |
| 14 |  1           |  15 |     23 | Regular | 0.01s | 0.04s  |  ~      |  443 KB |
| 14 |  1           |  15 |     23 | Bacchus | 0.02s | 0.31s  |  462 KB | 1366 KB |
| 18 |  1           |  19 |     30 | Regular | 0.02s | 0.14s  |  ~      |  953 KB |
| 18 |  1           |  19 |     30 | Bacchus | 0.06s | 1.18s  |  970 KB | 2974 KB |
| 22 |  1           |  23 |     34 | Regular | 0.06s | 0.32s  |  ~      | 1761 KB |
| 22 |  1           |  23 |     34 | Bacchus | 0.12s | 4.61s  | 1744 KB | 5635 KB |
| 22 | -1115540197  |  23 |     45 | Regular | 0.02s | 0.12s  |  ~      |  843 KB |
| 22 | -1115540197  |  23 |     45 | Bacchus | 0.05s | 1.01s  |  909 KB | 2526 KB |

Bacchus's PBP is 2–3× larger than `Regular`'s and verifies 5–20×
slower. `recursions ≈ n+1`, `propagations` in the tens — almost no
search, so the bench reflects scaffolding cost, not per-call savings.

### Search-intensive (`--all`)

`--all` enumerates every accepting string, so per-call work dominates
even at small `n`.

| n | seed         | rec    | props₀ | sols   | variant | solve | verify | pbp      |
|--:|-------------:|-------:|-------:|-------:|---------|------:|-------:|---------:|
| 4 |  1           |    238 |    293 |    165 | Regular | 0.00s | 0.10s  |   74 KB  |
| 4 |  1           |    238 |    293 |    165 | Bacchus | 0.00s | 0.10s  |   69 KB  |
| 5 |  1           |   2260 |   2484 |   1746 | Regular | 0.02s | 0.20s  |  730 KB  |
| 5 |  1           |   2260 |   2484 |   1746 | Bacchus | 0.02s | 0.20s  |  461 KB  |
| 5 | 42           |   1877 |   2259 |   1364 | Regular | 0.02s | 0.10s  |  660 KB  |
| 5 | 42           |   1877 |   2259 |   1364 | Bacchus | 0.01s | 0.20s  |  376 KB  |
| 6 |  1           |  40654 |  43763 |  32985 | Regular | 0.50s | 18.71s | 14394 KB |
| 6 |  1           |  40654 |  43763 |  32985 | Bacchus | 0.40s | 22.62s |  8578 KB |
| 6 | 42           |  17597 |  21336 |  13785 | Regular | 0.21s |  3.60s |  6572 KB |
| 6 | 42           |  17597 |  21336 |  13785 | Bacchus | 0.17s |  4.80s |  3690 KB |
| 6 | -1115540197  |   3620 |   4473 |   2600 | Regular | 0.04s | 0.30s  |  1838 KB |
| 6 | -1115540197  |   3620 |   4473 |   2600 | Bacchus | 0.02s | 0.30s  |  753 KB  |

Bacchus's PBP is 40–60% **smaller** than `Regular`'s — the
all-NoJustNeeded shape is paying its way. Verify is still ~20–30%
slower per byte (wider proof DB means each backtrack RUP chases more
constraints).

### Step-1 cheat (Bacchus encoding in OPB, never landed)

Verified in 0.00–0.06s with 1.7–9.7 KB PBPs across n=10..22 in
decision mode — confirms the encoding + UP path is fast when the
scaffolding is OPB-side rather than PBP-side. Step-2's cost is the
size of the derived PBP scaffolding, not the encoding itself.

## Takeaway

The result is shape-dependent:
- **Decision / shallow search**: Bacchus loses on every axis.
- **Search-intensive**: Bacchus's proof shrinks by ~half, verify
  still slightly slower per byte.

Same dynamic as the
[BinPacking/Knapsack pattern regression](https://github.com/ciaranm/glasgow-constraint-solver/issues/212):
the upfront-only design only wins when per-call work dominates.

`RegularBacchus` is kept as a sibling — not made the default — for
the issue #200 unified layered-DAG follow-up. The t-flag definitions,
per-`(q, v)` pol intermediates, and AL1 derivations are useful
primitives whose value is now demonstrated on the search-intensive
side, even if a decision-only bench would have missed it.

## Test plan
- [x] `ctest -j 32` clean on `release` build (207/207 pass).
- [x] `regular_bacchus_test`: 7 functional cases × {no proof, with
      proof + veripb}; all `s VERIFIED *`.
- [x] `regular_random --bacchus --prove` veripb-clean across 150
      ad-hoc seeds (`n ∈ {6, 8, 10, 12, 14, 16, 18, 20, 22, 24}` ×
      15 seeds each) — zero failures.
- [x] `regular_random --all --bacchus --prove` veripb-clean at
      n=4..6 across 3 seeds each.
- [x] Built with both presets (default release).

### DIFF
--- a/examples/regular_random/regular_random.cc
+++ b/examples/regular_random/regular_random.cc
@@ -104,7 +104,8 @@ auto main(int argc, char * argv[]) -> int
             ("proof-files-basename", "Alternative path and name for the proof", cxxopts::value<string>()->default_value("regular_random")) //
             ("short-reasons", "Use short reasons method (RegularLegacy only)")                                                             //
             ("legacy", "Post RegularLegacy instead of the new Regular")                                                                    //
-            ("bacchus", "Post RegularBacchus instead of the new Regular");
+            ("bacchus", "Post RegularBacchus instead of the new Regular")                                                                  //
+            ("all", "Enumerate all solutions instead of stopping at the first");
 
         options_vars = options.parse(argc, argv);
     }
@@ -128,6 +129,7 @@ auto main(int argc, char * argv[]) -> int
     auto short_reasons = options_vars.contains("short-reasons");
     auto legacy = options_vars.contains("legacy");
     auto bacchus = options_vars.contains("bacchus");
+    auto all_solutions = options_vars.contains("all");
     auto proof_prefix = options_vars["proof-files-basename"].as<string>();
 
     if (legacy && bacchus) {
@@ -144,8 +146,8 @@ auto main(int argc, char * argv[]) -> int
     auto p = Problem{};
     post_random_regular(p, n, rng, short_reasons, legacy, bacchus);
 
-    auto stats = solve_with(p,                                                           //
-        SolveCallbacks{.solution = [&](const CurrentState &) -> bool { return false; }}, //
+    auto stats = solve_with(p,                                                                   //
+        SolveCallbacks{.solution = [&](const CurrentState &) -> bool { return all_solutions; }}, //
         prove ? make_optional(ProofOptions{proof_prefix}) : nullopt);
 
     if (print_stats) {
--- a/examples/regular_random/regular_random.cc
+++ b/examples/regular_random/regular_random.cc
@@ -101,7 +101,8 @@ auto main(int argc, char * argv[]) -> int
             ("proof-files-basename", "Alternative path and name for the proof", cxxopts::value<string>()->default_value("regular_random")) //
             ("short-reasons", "Use short reasons method (RegularLegacy only)")                                                      //
             ("legacy", "Post RegularLegacy instead of the new Regular")                                                              //
-            ("bacchus", "Post RegularBacchus instead of the new Regular");
+            ("bacchus", "Post RegularBacchus instead of the new Regular")                                                            //
+            ("all", "Enumerate all solutions instead of stopping at the first");
 
         options_vars = options.parse(argc, argv);
     }
@@ -125,6 +126,7 @@ auto main(int argc, char * argv[]) -> int
     auto short_reasons = options_vars.contains("short-reasons");
     auto legacy = options_vars.contains("legacy");
     auto bacchus = options_vars.contains("bacchus");
+    auto all_solutions = options_vars.contains("all");
     auto proof_prefix = options_vars["proof-files-basename"].as<string>();
 
     if (legacy && bacchus) {
@@ -146,7 +148,7 @@ auto main(int argc, char * argv[]) -> int
     auto stats = solve_with(p,
         SolveCallbacks{
             .solution = [&](const CurrentState &) -> bool {
-                return false;
+                return all_solutions;
             }},
         prove ? make_optional(ProofOptions{proof_prefix}) : nullopt);
 
--- a/examples/regular_random/regular_random.cc
+++ b/examples/regular_random/regular_random.cc
@@ -35,7 +35,7 @@ auto index_of(const IntegerVariableID & val, const vector<IntegerVariableID> & v
     return (int)pos;
 }
 
-auto post_random_regular(Problem & p, const int & n, mt19937 & rng, bool short_reasons, bool legacy)
+auto post_random_regular(Problem & p, const int & n, mt19937 & rng, bool short_reasons, bool legacy, bool bacchus)
 {
     stringstream string_rep;
 
@@ -80,6 +80,8 @@ auto post_random_regular(Problem & p, const int & n, mt19937 & rng, bool short_r
 
     if (legacy)
         p.post(RegularLegacy{x, num_states, transitions, final_states, short_reasons});
+    else if (bacchus)
+        p.post(RegularBacchus{x, num_states, transitions, final_states, short_reasons});
     else
         p.post(Regular{x, num_states, transitions, final_states, short_reasons});
 }
@@ -98,7 +100,8 @@ auto main(int argc, char * argv[]) -> int
             ("stats", "Print stats, including solve time")                                                                         //
             ("proof-files-basename", "Alternative path and name for the proof", cxxopts::value<string>()->default_value("regular_random")) //
             ("short-reasons", "Use short reasons method (RegularLegacy only)")                                                      //
-            ("legacy", "Post RegularLegacy instead of the new Regular");
+            ("legacy", "Post RegularLegacy instead of the new Regular")                                                              //
+            ("bacchus", "Post RegularBacchus instead of the new Regular");
 
         options_vars = options.parse(argc, argv);
     }
@@ -121,7 +124,13 @@ auto main(int argc, char * argv[]) -> int
     auto print_stats = options_vars.contains("stats");
     auto short_reasons = options_vars.contains("short-reasons");
     auto legacy = options_vars.contains("legacy");
+    auto bacchus = options_vars.contains("bacchus");
     auto proof_prefix = options_vars["proof-files-basename"].as<string>();
+
+    if (legacy && bacchus) {
+        cerr << "Error: --legacy and --bacchus are mutually exclusive" << endl;
+        return EXIT_FAILURE;
+    }
 
     if (seed == -1) {
         random_device rand_dev;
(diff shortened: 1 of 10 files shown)


examples/regular_random/regular_random.cc
```diff
@@ -98,7 +100,9 @@ auto main(int argc, char * argv[]) -> int
             ("stats", "Print stats, including solve time")                                                                         //
             ("proof-files-basename", "Alternative path and name for the proof", cxxopts::value<string>()->default_value("regular_random")) //
             ("short-reasons", "Use short reasons method (RegularLegacy only)")                                                      //
-            ("legacy", "Post RegularLegacy instead of the new Regular");
+            ("legacy", "Post RegularLegacy instead of the new Regular")                                                              //
+            ("bacchus", "Post RegularBacchus instead of the new Regular")                                                            //
+            ("all", "Enumerate all solutions instead of stopping at the first");
```
mmcilree, line 108: We should definitely have this option for smart_table_random and in the other up-front Regular implementation, since it makes such a difference for benchmarking this one. Also we should probably have a more interesting example that actually uses regular -- maybe we (Claude) could model [nonogram from mzn-challenge 2013](https://github.com/MiniZinc/mzn-challenge/blob/develop/2013/nonogram/non.mzn), or just include the fzn as a test?

gcs/constraints/regular/regular_bacchus.cc
```diff
@@ -0,0 +1,497 @@
...
+                for (long q = 0; q < ns; ++q) {
+                    for (const auto & [val, _tflag] : ts[i][q]) {
+                        auto fwd_it = fwd_lines[i][q].find(val);
+                        auto rev_it = rev_tdef_lines[i][q].find(val);
+                        if (fwd_it == fwd_lines[i][q].end() || rev_it == rev_tdef_lines[i][q].end())
+                            continue;
+                        PolBuilder{}
+                            .add(fwd_it->second)
+                            .add(rev_it->second)
+                            .emit(*logger, ProofLevel::Top);
```
mmcilree: Nitpicky but may as well delete this one the AL1s have been derived.

gcs/constraints/regular/regular_bacchus.cc
```diff
@@ -0,0 +1,497 @@
...
+                }
+            }
+
+            // Per-(i, q, v) pol step: sum the OPB forward chain
+            //   ~state[i][q] + ~vars=v + state[i+1][delta(q,v)] >= 1
+            // with the t-def reverse line
+            //   t[i][q][v] + ~vars=v + ~state[i][q] + ~state[i+1][delta(q,v)] >= 1
+            // to derive
+            //   2*~state[i][q] + 2*~vars=v + state[i+1][delta(q,v)] + ~state[i+1][delta(q,v)] + t[i][q][v] >= 2.
+            // VeriPB's UP cancels the complementary state[i+1][delta]/~state[i+1][delta]
```
mmcilree: It's interesting Claude doesn't quite seem to realise this is resolution, even when I mentioned resolution in the issue hints.

gcs/constraints/regular/regular_bacchus.cc
```diff
@@ -0,0 +1,497 @@
...
+                    bool any = false;
+                    for (long q = 0; q < ns; ++q) {
+                        auto it = ts[i][q].find(val);
+                        if (it != ts[i][q].end()) {
+                            sum += 1_i * it->second;
+                            any = true;
+                        }
+                    }
+                    if (any)
+                        logger->emit_rup_proof_line(move(sum) >= 1_i, ProofLevel::Top);
```
mmcilree: Could be doing pols rather than RUPs for these AL1 derivations, but probably doesn't matter too much for an up-front derivation.

mmcilree (commented):
I guess this is roughly what we expected in terms of performance — but it's very cool that Claude can just do this, would have definitely taken me a day or two to implement. I would say this is just as worth having as a reference alongside the up-front DAG version of regular, particularly since it wins on logging overhead for deeper-search instances.

PR #236, opened 2026-05-26
Adding constraint labels

philrod1:
As per the "semantics_of_cp" document and issue #214, add support for labelled constraints, like this example for Abs:

```
@c[name][posge] X>=0 ==> X >= Y
@c[name][posle] X>=0 ==> X <= Y
@c[name][negle] X>=0 ==> -X <= Y
@c[name][negge] X>=0 ==> -X >= Y
```

I have implemented enough of this for the Abs example.  I have also renamed ConstraintName to ConstraintID to better reflect the intent and to avoid confusion with constraint (type) names.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,5 @@ out/
 Testing/
 verification_pipeline.sh
 crystal_maze_pipeline.sh
+verification/
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,4 @@ compile_commands.json
 out/
 .claude
 Testing/
-verification_pipeline.sh
-crystal_maze_pipeline.sh
-verification/
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,8 @@ Doxyfile
 *.pbp
 *.varmap
 *.scp
+*.corepb
+*.verifiedopb
 *~
 XCSP3-CPP-Parser
 generator
@@ -24,3 +26,5 @@ compile_commands.json
 out/
 .claude
 Testing/
+verification_pipeline.sh
+crystal_maze_pipeline.sh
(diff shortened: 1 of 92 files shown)


ciaranm:
Yep, this is looking sensible to me.

philrod1:
I'm a wee bit concerned about the wording "optional annotation" in the doc.  
`@c[name][optional annotation] C`
In the `Abs` example, the annotations `posge`, `posle`, `negle` and `negge` are needed to make the labels unique.  Uniqueness is not something I am enforcing at this point, but I think it is needed.  When I was using `fwd` and `back`, the test failed because multiple constraints ended up with the same label.  If the annotations are optiona, how do we ensure uniqueness?

ciaranm:
I think the unit testing should catch this, once we have that in place: if labels aren't set up correctly, the proofs won't pass.

Longer term, we can add in a check for duplicate constraint IDs and labels if we need to, just to get a less weird error.

ciaranm:
Could you focus on the constraints needed to get the Crystal Maze example to work? That's probably a nice easy within-reach actual runnable example we can use.

philrod1:
We have various label formats coming out:
`before_{1_4}`, `before{1_2}`, `lt_0_fwd`, and various `_{i}`
Including a mix of 0-indexed and 1-indexed.  It would be good to settle on a consistent format.

ciaranm:
Agreed. Before we get too invested, do we now have enough to start trying Crystal Maze out, and to see what breaks?

philrod1:
> Agreed. Before we get too invested, do we now have enough to start trying Crystal Maze out, and to see what breaks?

This works from my end:

```
$ build/crystal_maze --prove
  3 5
7 1 8 2
  4 6

  4 6
7 1 8 2
  3 5

  5 3
2 8 1 7
  6 4

  6 4
2 8 1 7
  5 3
```
What else do we need to do to check it?

ciaranm:
The ideal sequence would be to see how far along this gets before it fails:

```
crystal_maze --prove
veripb crystal_maze.opb crystal_maze.pbp
cake_pb_cp crystal_maze.scp > crystal_maze.verifiedopb
veripb crystal_maze.verifiedopb crystal_maze.pbp --elaborate crystal_maze.corepb
cake_pb_cp crystal_maze.scp crystal_maze.corepb
```

PR #237, opened 2026-05-28
view-as-own-pb-var: stages 0-4 (registry, lazy alloc, body+literal in V's bits, atom-level links, projection)

ciaranm:
## Summary

Implements the view-as-own-pb-var design for proof-logging views in `gcs/innards/proofs/`, across six commits. Working state through stage 4 reviewed and ready; pol-using justifications under views are the only known regression (stage-6 scope).

The design: give each `ViewOfIntegerVariableID` its own binary-encoded proof-only variable `BinEnc(V)`, emit constraint bodies and atomic literals in V's bits, link `BinEnc(V) = s*BinEnc(X) + c` in OPB. Pol-derived atom-level links between V's and X's atoms (both GE and EQ) bridge UP across the link so propagator inferences whose reasons mix V- and X-atoms close as RUP. Empirically grounded in the worked OPB/PBP at `~/claude/tmp/view_demo.{opb,pbp}` (Theorem 2.9 / B=0 shape closes where B!=0 / Example 2.14 fails).

## Commits

- `22e83ab` **stage 0**: `NamesAndIDsTracker::need_view` skeleton (throws).
- `0e12245` **stage 1**: lazy allocation of view bit vectors + bit-vector link `BinEnc(V) = s*BinEnc(X) + c` emitted by `need_view` during model writing. OPB augmented but unused in proof emission.
- `b275aee` **stage 2**: `emit_inequality_to` and the objective branch emit V's own bits when the view is registered (find_view), bringing constraint body mixing coefficient into the Theorem-2.9-friendly `B in {0,1}` zone. Preserve walk left as deview (correct for the projection design).
- `4e279d2` **stage 3**: `simplify_literal` returns `ProofVariableCondition` over V's atoms for view literals (when registered); deview as fallback. All call sites threaded with `const NamesAndIDsTracker &`. Pol-derived GE-atom-level links (`v>=k <=> x_cond`) emitted eagerly in `need_gevar`, built via `PolBuilder` (Def fwd + link half + Def [s-flipped] = saturate), queued via `emit_proof_line_now_or_at_start` to land at the top of the proof. `need_pol_item_defining_literal`'s Eq/NotEq throws for views are replaced with normal lookups.
- `6096bef` **stage 3+**: eager EQ-atom-level links `V=v <=> X=k_x` emitted from `need_direct_encoding_for`, two RUP lines per atom. Fixes backtrack-Bt failures that needed UP to chase `X=k -> V=v` through Def(X=k) reverse (which has a disjunctive consequent UP can't handle alone).
- `04c3ead` **stage 4**: projection set documentation + dedup. Views never appear in the preserved walk in normal use (because `Problem::all_normal_variables()` returns only `SimpleIntegerVariableID`s), and the existing view branch already emitted underlying bits. Comment freshened to call out the projection-extension design; an `already_emitted` set added so two views of the same X (or X + a view of X) don't duplicate.

## State

- **Default ctest: 205/205 pass.**
- **equals_test verifies under every view-wrap 0..18** with seed=1.
- **All constraint tests except `all_different_test` verify under views.** Confirmed empirically by a sweep with `JustifyExplicitlyOnly` and `JustifyExplicitlyThenRUP` hacked to `AssertRatherThanJustifying` (suppressing pol-using justifications): only all_different had remaining failures, and those are themselves pol-shape (the `justify_all_different_hall_set_or_violator` pol chain), not pure RUP.

## Known regressions (stage-6 scope)

Pol-using constraints under view wraps emit pol steps that were written against the deviewed body shape and so produce incorrect coefficient combinations under V-encoded bodies:

- `abs_test` under views: pol step combining "Abs non-negative" body with Def(i[_1][ge2]) + Def(y_ge0) is shape-mismatched.
- `mult_bc_test` under views: similar.
- `all_different_test` under views: per-value am1 pol-derivations and final pol Hall-set lemma in `justify_all_different_hall_set_or_violator`.

These are per-constraint exercises (update each pol callback to use V-encoded constraints), not framework-level work. Default (non-view) ctest is unaffected.

## Design and implementation notes

- `~/claude/tmp/view-proof-design.md` --- design rationale (extended Ch 3 framework, Theorem 2.9 / Example 2.14 shape analysis, worked example).
- `~/claude/tmp/view-proof-implementation-plan.md` --- staged plan.
- `~/claude/tmp/view_demo.{opb,pbp}` --- hand-crafted OPB/PBP demonstrating the new encoding shape closes by RUP where the deview shape fails.

## Test plan

- [x] Default ctest 205/205 under release preset.
- [x] equals_test under every view-wrap 0..18, seed=1.
- [x] Constraint sweep under views with `AssertRatherThanJustifying` hack to isolate RUP-only failures (only all_different remained, and those are pol-shape).


gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -73,6 +73,33 @@ struct NamesAndIDsTracker::Imp
...
+    // (and thereby the V<->X link) when an X atom is introduced after a
+    // view has been registered. When views are registered AFTER an X atom
+    // already exists, need_view itself backfills via this map's setup.
+    std::map<SimpleIntegerVariableID, std::vector<ProofOnlySimpleIntegerVariableID>> views_of_variable;
+
+    // Constraint-content registry: WPBSum that produced a given proof line.
+    // Populated by ProofModel::add_constraint and the proof-phase emitters.
+    // Used by the deview-form derivation in ProofModel and by PolBuilder
+    // in deview mode.
+    map<ProofLine, SumOf<Weighted<PseudoBooleanTerm>>> constraint_content_by_line;
```
ciaranm: Is this still used?

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -276,10 +347,18 @@ auto NamesAndIDsTracker::need_all_proof_names_in(const SumOf<Weighted<PseudoBool
...
+                    .visit(simplify_literal(*this, lit));
             },
             [&](const ProofFlag &) {},
-            [&](const IntegerVariableID &) {},
+            [&](const IntegerVariableID & var) {
+                // Stage 1: opportunistically register view bit vectors during
+                // model writing only. The proof-logging phase doesn't yet know
+                // how to introduce a view via ext/red lines; later stages may
+                // extend this.
+                if (_imp->model)
```
ciaranm, line 351: Is this still accurate?

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -460,6 +539,53 @@ auto NamesAndIDsTracker::need_direct_encoding_for(SimpleOrProofOnlyIntegerVariab
     }
 
     _imp->eqvars_that_exist[id].emplace(v, pair{forwards_line, reverse_line});
+
+    // Stage 3: if `id` is a view's proof-only var, eagerly emit the
```
ciaranm: Not sure that the Stage comments are still helpful

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -546,6 +672,257 @@ auto NamesAndIDsTracker::need_gevar(SimpleOrProofOnlyIntegerVariableID id, Integ
...
+}
+
+auto NamesAndIDsTracker::find_view(const ViewOfIntegerVariableID & view) const -> optional<ProofOnlySimpleIntegerVariableID>
+{
+    if (auto it = _imp->view_proof_only_vars.find(view); it != _imp->view_proof_only_vars.end())
+        return it->second;
+    return std::nullopt;
+}
+
+auto NamesAndIDsTracker::register_constraint_content(const ProofLine & line, SumOf<Weighted<PseudoBooleanTerm>> content) -> void
```
ciaranm: Is this still used?

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -546,6 +672,257 @@ auto NamesAndIDsTracker::need_gevar(SimpleOrProofOnlyIntegerVariableID id, Integ
...
+}
+
+auto NamesAndIDsTracker::deviewed_line_for(const ProofLine & line) const -> ProofLine
+{
+    if (auto it = _imp->deviewed_line_by_v_form.find(line); it != _imp->deviewed_line_by_v_form.end())
+        return it->second;
+    return line;
+}
+
+auto NamesAndIDsTracker::constraint_content_for(const ProofLine & line) const -> const SumOf<Weighted<PseudoBooleanTerm>> &
```
ciaranm: Is this still used?

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -546,6 +672,257 @@ auto NamesAndIDsTracker::need_gevar(SimpleOrProofOnlyIntegerVariableID id, Integ
...
+    // negative V.
+    //
+    // We deliberately do NOT saturate. Downstream consumers (PolBuilder in
+    // deview mode) use this line as the starting constraint in their own
+    // pol + divide chains. Saturating here would clip bit-level coefficients
+    // on wide-range variables (those encoded with a sign bit + magnitude bits),
+    // which then leaks an uncancelled residual into the consumer's pol when
+    // it adds a reif on the same variable. The unsaturated form has the full
+    // bit-level coefficient mass needed for clean cancellation.
+    stringstream pol;
```
ciaranm: Is there any reason not to use PolBuilder here? I guess it could infinite recurse if it's in rewrite mode?

gcs/innards/proofs/pol_builder.hh
```diff
@@ -67,11 +67,33 @@ namespace gcs::innards
...
+         * `deviewed_line_for(line)` returns the input unchanged for them,
+         * so deview-mode builders behave identically to the default builder
+         * for constraints with no view operands.
+         *
+         * The tracker reference must outlive the builder. `add(ProofLine, ...)`
+         * calls perform the lookup; `add(XLiteral, ...)` and
+         * `add_for_literal` are unaffected (they push raw literals, not
+         * constraint references).
+         */
+        explicit PolBuilder(const NamesAndIDsTracker & tracker);
```
ciaranm: From an API perspective: I think having the two constructors and the unobvious link to how they control behaviour is really unintuitive. I'd prefer something very explicit, like a `enable_deview_mode(...)`.

gcs/innards/proofs/proof_model.cc
```diff
@@ -363,30 +376,46 @@ auto ProofModel::finalise() -> void
...
+                            full_opb << (v.negate_first ? -bit_value : bit_value) << " " << names_and_ids_tracker().pb_file_string_for(bit_name) << " ";
+                    }
                 }}
                 .visit(*_imp->optional_minimise_variable);
 
             full_opb << ";\n";
         }
 
         if (_imp->preserved_variables) {
+            // Stage 4: the projection set for solx/soli only includes the
```
ciaranm: Again, are Stage comments useful? (Several places)

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: No — it was write-only. `constraint_content_for` (the only reader) had no callers, so this map and `register_constraint_content` were dead. Removed the map, the decl/def, and all five call sites in 256c358.

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: Not quite — corrected in 256c358. The "later stages may extend this" was stale. It now reads: registration happens during model writing only, because `need_view` throws during the proof-logging phase, so the `_imp->model` gate is the real reason.

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: Agreed — dropped the "Stage N:" prefixes across all sites (this one and the others) in 256c358, keeping the substantive content.

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: `find_view`, `need_view`, `view_link_lines_for` etc. are all still used. But `register_constraint_content` (just below here) was not — its only reader `constraint_content_for` had no callers, so the whole content registry is removed in 256c358.

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: No — `constraint_content_for` had zero callers, making the `constraint_content_by_line` registry write-only. Removed it and `register_constraint_content` in 256c358. (`register_deviewed_line` / `deviewed_line_for` just above are still used and stay.)

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: Good call — switched to `PolBuilder` in 256c358 (byte-identical output). Your recursion instinct is right: I used a *plain* builder, not deview mode. A deview-mode builder would call `deviewed_line_for(v_form_line)` for the very line being derived here, so plain mode is both correct and sidesteps that self-reference; noted in the comment.

gcs/innards/proofs/pol_builder.hh
ciaranm: Done in 256c358 — dropped the second constructor and added an explicit, chainable `enable_deview_mode(tracker)` method. The two linear call sites now read `PolBuilder pol; pol.enable_deview_mode(...);`.

gcs/innards/proofs/proof_model.cc
ciaranm: Removed throughout in 256c358 — all eight "Stage N:" comments across names_and_ids_tracker, simplify_literal, emit_inequality_to and proof_model are rewritten without the staging references.

Issue #144, opened 2026-05-08
Add infer_not_in_range / infer_in_range / IntervalSet::erase_range


ciaranm:
## Background

Phase 6 of issue #134 added `IntervalSet::each_interval_minus(other)` plus `State::domains_intersect` / `domain_intersects_with`, which let propagators express set-difference reasoning at the interval level. What's still done value-by-value is the *inference* itself: every site that wants to remove a contiguous range of values from a domain still loops calling `infer_not_equal` per value.

A natural next step would be:

```cpp
template <IntegerVariableIDLike VarType_>
auto State::infer_not_in_range(const VarType_ & var, Integer lo, Integer hi,
    std::function<Justification(Integer value)> justification_for,
    std::function<Reason(Integer value)> reason_for) -> Inference;
```

…plus a complementary `infer_in_range` and a supporting `IntervalSet::erase_range` primitive.

## Concrete callers for `infer_not_in_range`

All four are post-Phase-6 sites that today loop `for (Integer val = lo; val <= hi; ++val) inference.infer_not_equal(...)` over an interval yielded by `each_interval_minus` or already in IntervalSet form:

| Site | Range source | Per-value reason | Per-value justification |
|------|--------------|------------------|--------------------------|
| `equals.cc:65-73` | `v1.each_interval_minus(v2)` | `reason + (v2 != val)` | `JustifyUsingRUP{}` (uniform) |
| `equals.cc:77` (mirror) | `v2.each_interval_minus(v1)` | `reason + (v1 != val)` | `JustifyUsingRUP{}` (uniform) |
| `min_max.cc:179-194` | `support_1.each_interval_minus(result)` | `var != value` collection | `JustifyExplicitlyThenRUP{...}` per-value scaffolding |
| `element.cc:477-505` | `still_to_find_support_for.each()` | `(var != value)` per-`var` collection | `JustifyExplicitlyThenRUP{...}` per-value scaffolding |

Notably **every Pattern-A site already has a value-dependent reason** — the literals in the reason reference the value being removed. So the API has to take per-value callbacks; we can't promote the reason to a single static `Reason` for the whole range.

## Sites that wouldn't fit

- `regular.cc:319` — per-value condition (`states_supporting[i][val].empty()`), not a range.
- `smart_table.cc:525, 535` — iterates `unsupported[var]` (a `vector` of values, not necessarily contiguous).
[...]

PR #240, opened 2026-05-30
dev_docs: range-literal proof-logging design note

ciaranm:
Adds `dev_docs/range-literal-proof-logging.md`, recording the proof-logging investigation behind the `infer_not_in_range` half of #144.

## What this is

A **design note, not implemented** — clearly marked as such at the top and in the index. It documents whether a whole-interval domain removal `X ∉ [a,b]` can be logged as one compact proof object instead of one RUP line per removed value, and the VeriPB experiments that establish the approach is sound at the checker level.

## Contents

- **Why the naive single-Boolean fails** — asserting `¬r` for a fresh `r ⇔ ([X≥a] ∧ ¬[X≥b+1])` is inert (the backward clause stays binary, propagates nothing) and breaks the chapter-3 backtracking invariant Inv3. The disjunctive-reification trap.
- **The encoding that works** — fully reify the range (the backward clause is a guarded long-range link on the order chain), plus a lazy per-value link `r ∨ [X≠v]` emitted when `[X=v]` is defined inside a defined range, plus a hole-aware `domL`.
- **VeriPB 3.0.2 confirmation** — per-value link RUP-derivable (pinned and sign-crossing), mixed-vocabulary empty-domain sweep → UNSAT, bound-jumps both signs / unpinned, bogus claims rejected.
- **Exhaustive single-constraint ablation** — 9/16 constraints individually necessary; the **backward reification clause is load-bearing while both forward clauses are redundant**, with the `--elaborate`/confluence caveat that motivates the ablation.
- **The one obligation VeriPB can't witness** (forward propagation / Inv3) and the three metatheory items still owed before implementation.

Updates the relevant section's pessimistic conclusion in #144 (which had it that the per-value cost was unavoidable). Probe scripts and the ablation harness referenced from #144.

Docs-only; no code touched.

### DIFF
--- a/dev_docs/README.md
+++ b/dev_docs/README.md
@@ -57,6 +57,14 @@ library. For an introduction to *using* the solver, start with the top-level
   `recover_am1`-based pairwise at-most-one derivation that replaces
   cumulative's encoded `C_t`, and the third reusable proof-logging
   idea that comes out of the design.
+- [Range-literal proof logging](range-literal-proof-logging.md) — **design
+  note, not implemented.** Whether a whole-interval domain removal
+  (`X ∉ [a,b]`) can be logged as one compact proof object instead of one RUP
+  line per value: why the naive single-Boolean breaks the chapter-3
+  backtracking invariant, the fully-reified-range + lazy per-value-link
+  encoding that fixes it, the VeriPB experiments (and ablation) that show it is
+  sound at the checker level, and the metatheory still owed. The proof half of
+  issue #144 (`infer_not_in_range`).
 
 More documents will be added here as we build up coverage of other parts of
 the codebase.
--- a/dev_docs/range-literal-proof-logging.md
+++ b/dev_docs/range-literal-proof-logging.md
@@ -148,30 +148,116 @@ straight through the `BinEnc` bit layer, and two contradictory bounds on the
 same two's-complement bit-vector already conflict (chapter 3 Lemma 3.2 /
 Theorem 2.7). So for *derivability* the chains are an efficiency aid, not a
 correctness requirement — the bit layer does the linking the chains would.
-(They still matter for keeping the *up-front* order encoding compact and for
-forward propagation; see the next section.)
+**This is a statement about Theorem 3.2 (empty domain → conflict) only.** It
+does *not* transfer to forward propagation: for Theorem 3.3 (next section) the
+per-value links are strictly load-bearing, as the failure example there shows.
+The two theorems have different load-bearing constraint sets; do not read
+"chains are cheap for the conflict" as "linking is cheap in general."
 
-## The one obligation VeriPB cannot witness
+## The central obligation VeriPB cannot witness: forward propagation (Theorem 3.3)
 
-VeriPB only ever checks RUP **derivability**. The remaining hole is about
-forward **unit-propagation**: that `[X≠v]` actually *propagates* from `¬r`
-(so it lands in the propagated set `L`, keeping `domL ⊆ dom_t` and making the
-literal reason-valid per Lemma 3.4). That is exactly what the standing
-per-value link `r ∨ [X≠v]` is *for* — but "this clause causes that forward
-propagation" is a framework property, not a checkable proof step, so it cannot
-be probed here. It has to be argued in the metatheory.
+VeriPB only ever checks RUP **derivability** — that a claimed clause *follows*.
+It says nothing about which literals a solver's unit propagation will *reach*.
+The empty-domain sweep and the ablation above are both about Theorem 3.2 (a set
+of literals defining an empty domain unit-propagates to conflict). The harder
+theorem — and the one that is actually *central* to making `infer_not_in_range`
+usable — is **Theorem 3.3**:
+
+> if a set of literals `L` defines a domain `D`, then unit propagation of `L`
+> propagates **every defined literal implied by `D`**.
+
+This is where a naive `¬r` silently fails *as a reason source for other
+propagators*, even though the `¬r` inference itself is fine. The example is
+Matthew's. A propagator logs `reason_1 ⇒ ¬r` for `r = [X ∈ [10,20]]`, so the
+solver now knows `X ≠ 15`. A *different* propagator — `X = Y`, or `Abs(X,Y)` —
+carries the reason clause
+
+```
+[X ≠ 15] ⇒ [Y ≠ 15]        i.e.   [X=15] ∨ [Y≠15]
+```
+
+and `15` is the last value left for `Y`. We expect to backtrack now with the
+clause `[guesses] ⇒ 0 ≥ 1`. For that to be RUP, unit propagation under
+`[guesses]` must reach the conflict:
+
+- `[guesses]` propagates `reason_1`, which propagates `¬r`. So far so good.
+- But `¬r` is the **inert disjunction** again (`¬[X≥10] ∨ [X≥21]`, two
+  unassigned literals): on its own it propagates **nothing** about `[X=15]`. The
+  reif clauses are all satisfied or dormant under `¬r` (§"Why the naive
+  single-Boolean fails"). So `[X=15] ∨ [Y≠15]` never fires, `[Y≠15]` never
+  propagates, `Y=15` is never refuted — **no conflict, and the backtracking
+  clause is not RUP.**
+
+The fix is exactly ingredient 2, the per-value link `r ∨ [X≠v]`. With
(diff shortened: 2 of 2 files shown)


mmcilree:
It's still not enough. We need the implied literals to go the other way as well to get something equivalent to my theorem 3.3 (assuming claude knows the right theorem number, I didn't check).

There's nothing in Claude's design currently that will ensure a set of not equals literals will propagate a negated range literal. 

To try to make an example where this might actually matter:
Suppose we separately infer

`R => ~[x=10], R => ~[x =11] ... R=> ~[x=20]`

and then some constraint tells us that without the range x in {10..20} it is infeasible. We might want to justify this by deriving
`~[x in {10..20}] => 0 >= 1` somehow.

But then, backtracking with `G => 0 >= 1` will fail: `G` propagates `R` which propagates  `~[x=10] ... ~[x=20]`, but none of these propagate anything further and in particular `~[x in {10..20}]` is not propagated and we don't get the contradiction.

The linking constraint `~[x=10] /\ ... /\ ~[x=20] => ~[x in {10..20}]` is RUP so maybe it's enough to just say we try to introduce reverse links like these too. But it's bit tricky to say exactly what the invariant needs to be, would need to think about it (or Claude can think some more about it).

mmcilree (changes requested):
See above comment

mmcilree:
A completely different idea: @ciaranm if we're convinced views work, maybe do range literals by just reusing the soundness of views?

So if we want `[x in {10, 20}]` we just use the literal `[v <= 30]` where 
v is the view `v = x + (min dom X + 10)`. If we preserve the view invariants and prove these work, we get range literals for free.

... of course this would require Wietze's introducing BinEnc vars on the fly

ciaranm:
This is increasingly convincing me that the "formally verify exactly the statements we need" approach is a good idea...

ciaranm:
You're right, and it's a clean catch. The forward links `r ∨ [X≠v]` only give `¬r ⇒ [X≠v]` — I conflated *introduction-time* symmetry (we add that clause on both the new-range and new-equality paths) with *propagation-direction* symmetry, which it isn't. Nothing reconstitutes `¬r` from the not-equals.

The reverse needs the covering clause `r ∨ [X=a] ∨ … ∨ [X=b]` ( `{[X≠v]} ⇒ ¬r` ). It's RUP — under the negation `r` itself gives `[X≥a]`, then the equality–bound chain marches up to `[X≥b+1]` against `r ⇒ ¬[X≥b+1]` — but, same as the forward case, RUP-derivable ≠ forward-propagating, so it has to actually be posted. Two refinements: (1) bound-driven emptying is already covered by the forward reif clauses (`¬[X≥a] ⇒ ¬r`, `[X≥b+1] ⇒ ¬r`), so only your value-by-value case needs it; (2) a constant-width form exists, dual to the forward containment chain — nest `r_k = [X∈[a,k]]` and post `¬r_k ∨ r_{k-1} ∨ [X=k]` so UP composes `¬[X=a] → ¬r_a → … → ¬r_b`, trading the wide clause for `O(width)` sub-range literals.

The thing this convinces me of: the win is *directional*. `¬r ⇒` {many not-equals} is cheap and is the use case; {many not-equals} `⇒ ¬r` is inherently `Ω(width)` whichever encoding, because you have to read all the exclusions back. So the reverse links cost the same order as the per-value lines we were avoiding, and they only bite when something *else* emptied the range one value at a time — exactly when the range literal wasn't earning its keep. Stating the invariant for *when* a range coexists with all its interior equality literals (and so needs the reverse clause) is the unresolved bit.

On views: I'm not convinced it gains us much. A bound literal `[V ≤ c]` is a real atom and would inherit both directions for free, but a two-sided interval isn't a single affine-view bound — it needs an abs-like view, which is just the same difficulty moved into the `Abs` bit-composition gap. It relocates the work rather than dissolving it, so I don't think it buys us a way out here.

Recorded all of this in the design note on #240. And the fact that we keep turning up these gaps by hand is itself the argument for pinning down the exact Theorem 3.3 instances formally.

mmcilree:
Yeah I agree the views thing doesn't quite work. Just a half-baked thought. If we did manage to solve abs views though it's worth noting it'd be two with one stone. 

In terms of getting a clean invariant though, we're still not there yet with above.

Even with both forwards and reverse channelling between 
negated range and negated equality literals, we're still going to get problems channelling between range literals. I think claude is sort of acknowledging this above, but it's making it sound like it's just performance problem rather than an invariant breaker, which I think it is.


For example we'd need the two negated range literals `~[x ∈ {10..20}]` and `~[x ∈ {15..25}]` to propagate `~[x ∈ {12..24}]` if these things are all defined and no amount of channelling between neq-literals is going to make that happen.

We'd be stuck with 

`[x ≤ 10] ∨ [x ≥ 20]` (after propagating `~[x ∈ {10..20}]`)

`[x ≤ 15] ∨ [x ≥ 25]` (after propagating `~[x ∈ {15..25}]`)

and the definition of `[x ∈ {12, 24}]` is untouched (all resulting constraints are "inert" in Claudespeak).

I agree formally verifying would be very nice, but there's not much point in trying to do that until we're at least semi-confident we have something that works.

mmcilree:
# A Proposal for a Range Literal Logging Invariant

Here's a proposal for an invariant that seems like it might work and might be worth trying to prove.

First, we treat eqvars as just unit range literals (RLs) for simplicity. 
So `[x = 3]`  is considered to be `[x in {3..3}]`.

Then we ensure the following. 

**Invariant:**
For every range literal `[x in R]`:
1. if `R` disjoint with the ranges of all other RLs no linking constraints are needed
2. if `R` is *fully* contained in the range of another RL, then 
let `C` be the narrowest range for a RL currently defined and ensure that `~[x in C] => ~[x in R]` is derived (RUP)
3. if `R` is not disjoint with the ranges of all other RLs but is not fully contained in any single one ensure 
that `~[x in R] => ~[x in A] ∧ ~[x in B] ∧ ~[x in C]` is derived (pol?) where `A` `B` `C` partition `R` and either disjoin from or fully contained in ranges for previously defined range literals. `B` is allowed to be empty here in which case the literal is omitted.
4. if the ranges of any set of defined RLs fully partition `R`, ensure that for at least one partitioning set `R1 ... Rn` we define
`~[x in R1] ∧ ... ∧ ~[x in Rn]=> ~[x in R]`.


**What the this gives us**:
Not a proper proof (may well be wrong) but my intuition:
- Negated ranges propagate negated subranges: easy induction on the case 2 requirement.
e.g. `~[x in [10..20]]` will propagate `[x != 15]` as we by assumption a chain of containing ranges
`~[x in [10..20]] => .. => .. [x != 15]`

- Negated subranges partitioning a range propagate their negated union, if it's defined. This is an induction on case 4:.

- Overlapping ranges propagate consistently: this is tricky but I think will work due to the combination of case 3 and case 4. 

Any overlapping ranges must be defined in terms of common subranges at some level
so e.g if we start with
`~[x in [10..20]]` 
Then introduce
`~[x in [15.. 25]]` 
then we are obliged to define
`~[x in [15..25]] => ~[x in [15..20]] ∧ ~[x in [21..25]]` (by Case 3)

then if we define 
`~[x in [10..25]]`
we we are obliged to derive
`~[x in [10..25]] => ~[x in [10..20]] ∧ ~[x in [21..25]] ` (by Case 3)
And now if we propagate `~[x in [10..20]] ∧ ~[x in [15..25]]` 
`~[x in [10..25]]` will propagate, as required.

- Is this enough for full propagation consistency and a Thm 3.3. style claim? I'm not sure

**How expensive is it to maintain?**:
Also not sure.
I've given myself a bit of confidence by drawing boxes on a piece of paper: basically we store all the ranges in a tree of strict containment and reuse previously defined ranges for the partitioning where possible. 

Case 2 costs 1 constraint for the parent of the new range 
and potentially k constraints for however many children it became the parent of (worst case all RLs).
Case 4 is 1 constraint whenever this happens, but at most per new RL.
Case 3 I think can always be done with at most two new literals each time, but that's just based on drawing boxes.

This is all very half baked but putting out here in case it helps Claude find a better solution, or gives a candidate for something we might want to try to formalise.

ciaranm:
Superseded by #303 (merged): the authoritative range-literal design now lives in `dev_docs/range_literals_spec.md`. This early proof-logging note predates the from-scratch reimplementation, and several of its central conclusions (e.g. "both forward reification clauses are redundant", "a range removal is one RUP line with per-value facts free") are catalogued as refuted in the spec's Appendix B (B1/B2/B3), each with the witness test that catches it. Branch kept unmerged as archaeology.

PR #255, opened 2026-06-01
Sort and ArgSort constraints with a fully VeriPB-certified Mehlhorn–Thiel propagator

ciaranm:
## Summary

Adds the `Sort` (value sort — MiniZinc `sort`, Gecode `sorted`) and `ArgSort`
(stable sorting permutation — MiniZinc `arg_sort`) global constraints, with
MiniZinc bindings, unit + bindings tests, and `dev_docs/sortedness.md`.

`Sort` ships a **Mehlhorn–Thiel bounds-consistency propagator** with **fully
VeriPB-certified proof logging** — every inference is checked, no
`AssertRatherThanJustifying` / `AssertProofRule` remains. Per the survey in
`dev_docs/sortedness.md`, no certifying/explained sortedness propagator existed
previously, so this appears to be the first.

## Encoding

Compact and domain-width-independent: a proof-only **stable rank**
`pos[i] = Σ before[·][i]` (fully-reified pairwise comparison flags), channelled
`(pos[i]=k) → y[k]=x[i]`, plus `y` sorted. `pos` is a function of `x`, so a full
`x`-assignment pins everything by UP (solutions verify) and a violated leaf is
plain RUP. `ArgSort` keeps a lean checker (its permutation is a real variable).

## Propagator + proof

- MT: Y-normalization, Glover down/up-sweep matchings, Tarjan SCC for the
  x-bounds; bounds(Z) on both `x` and `y`.
- Root permutation lemmas (totality, antisymmetry, transitivity, rank gaps,
  `recover_am1` injectivity → surjectivity) derived once at `ProofLevel::Top`
  and reused (the Cumulative/Disjunctive bridge pattern).
- Bound tightenings: normalization (sortedness chain) and order-statistic
  (count is RUP under the reason) cases, plus a shared **Hall-violator
  pigeonhole over the rank line** (`find_band`) — used unconditionally for the
  no-matching contradiction and under the negated-goal assumption (route a /
  Cor 2.1) for the Hall sub-cases.
- Also fixes a **pre-existing x-side bounds-consistency bug** found via the seed
  sweep: the SCC x-bound used each component's whole y-span instead of each
  x's reduced-graph neighbours, under-propagating `lb(x)`/`ub(x)`.

## Testing

`sort_test` / `arg_sort_test` (BC consistency-checked, nested-band + seeded
random batch), the MiniZinc binding tests, and `examples/sort_count_probe` (a
proof stress test). The whole suite, the n=20 probe, and a 200-seed random
sweep all verify `s VERIFIED` with **zero unchecked assertions**.

The first commit is labelled "WIP / staged proof logging": that reflects the
development history (each inference was made honest in stages, each commit
verifiable under the then-current assertions). The final state is fully honest.


ciaranm:
Follow-up commits (a762b9e, a188e0f, cae0adc): **ArgSort upgraded from a leaf-checker to a certified bounds-consistency propagator**.

- **Reuse, not duplication.** Sort's `before`/`pos` witness build + Mehlhorn–Thiel propagator are factored into shared helpers (`sortedness.hh`, `SortednessWitness`); Sort's methods become thin wrappers (sort_test still fully certifies). ArgSort allocates internal real `y[j]`, runs the helpers on `{x, y}` for `y = sort(x)` for free, and channels them to the permutation `p`.
- **p reaches certified bounds(Z) consistency.** Channel pruning + GAC `all_different` (reused) + an **achievable-rank-set** propagator. The reachable rank set can have *holes* inside `[a_k, b_k]` (ties make the "number below k" jump), so per-element intervals are not enough. This is exactly BC on `p`: a reachable rank always has a full supporting solution. Matches Gecode `sorted`+`Perm` (Thiel) strength — but VeriPB-certified.
- **Proofs, no asserts.** Outside-interval prunings use a `rank_ge`/`rank_le` pol; hole prunings use a threshold-pivot case split (`x[k] ≤ U` vs `≥ U+1`) mirroring this PR's pivot-bridge, closed via a definitional inverse channel `p[j]=offset+k ↔ pos[k]=j`.
- Full GAC on `p` is NP-hard, so rank-feasible-but-integer-infeasible values can survive — that residual is deliberately left.

Verification: tests assert `p` at BC with explicit tie-hole cases; 260-seed sweep (n≤5) + n≤6 stress all `s VERIFIED`; sort + `minizinc-argsort` regression green. Details in `dev_docs/sortedness.md` (ArgSort section).

examples/sort_count_probe/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+add_executable(sort_count_probe sort_count_probe.cc)
```
ciaranm: Is this still useful? The comments suggest it's a throwaway rather than a useful example to keep around.

mmcilree:
Was having a look at this and my immediate thought was that the OPB encoding
looks dangerously complicated.

Reading Claude's design docs, it agrees that the more obvious encoding would be
to have $n^2$ $0-1$ aux vars representing a permutation, so

$\sum_{i} p[i][j] = 1$

$\sum_{j} p[i][j] = 1$

Then the $2n^2$ enforcing $X$ is some permutation of $Y$

$p[i][j] \Rightarrow X_i = Y_j$

Plus $Y$ sorted is enough.

However, as it spotted, 
this encoding does not meet the requirement for partial solution witnessing 
(setting all $X_i$ and $Y_i$ does not necessarily set all the $p$ vars) 
due to symmetric permutations.

It therefore defines a much more complicated "canonical-stable-rank" encoding with lots of 
extra auxiliary variables to get an encoding that's AC in the SAT sense. 

I suppose this is fair enough given what is done everywhere else and what's in my thesis. 

---

However; here's a research idea / a chance to see if we can get Claude to undersand dominance.

Why not we try to introduce a canonicalisation of the $p[i][j]$ via a dominance step at the 
start of the proof? This shouldn't break enumeration proofs since $p$ vars are not preserved, 
but it might need to use the new "orders with auxiliary variables" feature of VeriPB.

gcs/constraints/sort/arg_sort.cc
```diff
@@ -0,0 +1,508 @@
...
+                        auto [li, ui] = state.bounds(x[i]);
+                        bool f = (i < k) ? (ui.raw_value <= vk) : (ui.raw_value < vk);
+                        bool c = (i < k) ? (li.raw_value <= vk) : (li.raw_value < vk);
+                        if (f)
+                            ++forced;
+                        if (c)
+                            ++possible;
+                    }
+                    for (long long r = forced; r <= possible; ++r)
+                        if (r >= 0 && r < static_cast<long long>(n))
```
ciaranm: Rather than the casts here (and through the rest of the commits), can we use `std::cmp_less` etc? (Check compiler support)

gcs/constraints/sort/arg_sort.cc
```diff
@@ -0,0 +1,508 @@
...
+        propagate_gac_all_different(p, vals, vector<Integer>{}, *am1_lines, state, inference, logger);
+        return PropagatorState::Enable;
+    },
+        ad_triggers);
+
+    Triggers triggers;
+    triggers.on_instantiated.insert(triggers.on_instantiated.end(), _x.begin(), _x.end());
+    triggers.on_instantiated.insert(triggers.on_instantiated.end(), _p.begin(), _p.end());
+
+    // Checking-only propagator: act only once every element of x and p is
```
ciaranm: Does the checking-only propagator still do anything now that the other propagators exist?

gcs/constraints/sort/sort.cc
```diff
@@ -0,0 +1,1118 @@
...
+    // Derive the permutation facts once at the proof root, at ProofLevel::Top so
+    // they persist across the whole search, and reuse them in every bound
+    // justification (the Cumulative/Disjunctive bridge pattern). The chain:
+    //   totality      before[a][b] + before[b][a] >= 1      (order is total)
+    //   antisymmetry  !before[a][b] + !before[b][a] >= 1     (at most one way)
+    //   transitivity  before[i][i'] & before[k][i] -> before[k][i']
+    //   rank gap      before[i][i']  ->  pos[i'] >= pos[i] + 1
+    // The gap lines make the pairwise rank-distinctness clauses RUP, which
+    // recover_am1 folds into per-value injectivity (built in the next step).
+    propagators.install_initialiser(
```
ciaranm, line 978: I think we have some way of marking initialisers as cheap / expensive? This one's expensive.

ciaranm:
Addressed the review in d60e6e2:

- **`sort_count_probe`** — removed. The large-degree count-line regime it guarded is described in `dev_docs/sortedness.md`; the throwaway example wasn't worth keeping.
- **Casts → `std::cmp_*`** — done across `sort.cc` / `arg_sort.cc` for the signed/unsigned *comparisons* (`<utility>`, well supported on GCC 13+ / clang 21). Construction and indexing casts (`Integer(...)`, `inj_lines[...]`) are left, since `std::cmp` can't express those.
- **Checking-only propagator** — you were right to question it: it's **dead**, removed. Once `x` is fixed each element's reachable-rank set collapses to its single stable rank, so the achievable-rank-set propagator (with GAC `all_different`) already prunes `p` to the unique permutation; a wrong `p` is a domain wipeout before any leaf. Verified empirically — full enumeration + BC-consistency checks + VeriPB all unchanged with it gone, across the suite and a 60-seed sweep.
- **Expensive initialiser** — marked `InitialiserPriority::Expensive` (the O(n³) root permutation lines), so the definitional initialisers run first.

@mmcilree — thanks, the canonical-stable-rank rationale you reconstructed is exactly it. I've spun your dominance-canonicalisation idea out to #262 (tagged `scary`) so it doesn't block this PR.

PR #256, opened 2026-06-02
Add a regular-expression string constructor to Regular

ciaranm:
## What

`Regular` can now be posted from a **regex string** in addition to an explicit automaton:

```cpp
p.post(Regular{vars, "0+ 1+ 0+ | 2+"});
```

The syntax matches MiniZinc/Gecode exactly (the authoritative reference is libminizinc's flex/bison grammar, not the prose docs): integer literals, concatenation by juxtaposition, `|`, groups `()`, wildcard `.`, classes `[3-6 7]`, negated classes `[^3 5]`, and the `*` `+` `?` `{n}` `{n,}` `{n,m}` quantifiers. Two non-obvious semantics are replicated: `.`/`[^…]` expand over the domain's contiguous `min..max`, and literals are non-negative only.

## How

- **New, dependency-free compilation unit** `gcs/constraints/regular_regex.{hh,cc}`: a hand-written lexer + recursive-descent parser → Thompson construction → ε-elimination, producing an ε-free **NFA**. No subset construction, so the automaton stays linear in the regex size and avoids determinisation blow-up. (No Gecode dependency — MiniZinc's string regex is actually a Gecode-backed compiler builtin, which we deliberately don't pull in.)
- **`Regular` generalised to a multi-target transition table** (`set<long>` per `(state, value)`). The existing DFA constructors insert singletons, so a DFA is just the `k==1` special case. The propagator's layered-graph reachability and the proof model's transition clause (now a disjunction over the target set) handle `k>=1` uniformly. The regex is compiled to its NFA in `prepare()`, where the variable domains — and hence the alphabet for `.`/`[^…]` — are known.
- One genuinely-multi-target bug fixed along the way: `decrement_outdeg` now uses a `still_supports()` guard so a source state keeps its support (and emits no spurious justification) while any live edge on that value remains.

## Proof logging

No new VeriPB rule types — same layered-graph encoding; the regex never appears in the proof. The transition clause generalises to a disjunction over the NFA target set, and the multi-target RUP justifications are validated empirically (below). As with MiniZinc, regex→NFA compilation is trusted/unverified: the proof attests the automaton, not that the automaton matches the regex.

## Testing

- `regular_regex_test` (Catch2): exhaustive cross-check of the compiled NFA against an independent reference matcher over all short sequences, plus explicit precedence/semantics and parse-error cases — 383 assertions.
- `regular_test`: end-to-end regex cases including a **genuinely non-deterministic NFA** (`"1 2|1 3"`), all GAC-checked and veripb-verified (`COMPLETE ENUMERATION`).
- Existing DFA and `xcsp_regular` coverage unchanged and green.

### DIFF
--- a/examples/regex/regex.cc
+++ b/examples/regex/regex.cc
@@ -28,12 +28,13 @@ auto main(int argc, char * argv[]) -> int
     cxxopts::ParseResult options_vars;
 
     try {
-        options.add_options("Program Options")                                                       //
-            ("help", "Display help information")                                                     //
-            ("prove", "Create a proof")                                                              //
-            ("proof-files-basename", "Basename for the .opb and .pbp files",                         //
-                cxxopts::value<string>()->default_value("regex"))                                    //
-            ("stats", "Print solve statistics")                                                      //
+        options.add_options("Program Options")                                                                 //
+            ("help", "Display help information")                                                               //
+            ("prove", "Create a proof")                                                                        //
+            ("proof-files-basename", "Basename for the .opb and .pbp files",                                   //
+                cxxopts::value<string>()->default_value("regex"))                                              //
+            ("regex", "Post the constraint from a regular-expression string instead of an explicit automaton") //
+            ("stats", "Print solve statistics")                                                                //
             ;
 
         options_vars = options.parse(argc, argv);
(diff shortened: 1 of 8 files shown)


examples/regex/regex.cc
```diff
@@ -28,12 +28,13 @@ auto main(int argc, char * argv[]) -> int
...
-            ("prove", "Create a proof")                                                              //
-            ("proof-files-basename", "Basename for the .opb and .pbp files",                         //
-                cxxopts::value<string>()->default_value("regex"))                                    //
-            ("stats", "Print solve statistics")                                                      //
+        options.add_options("Program Options")                                                                 //
+            ("help", "Display help information")                                                               //
+            ("prove", "Create a proof")                                                                        //
+            ("proof-files-basename", "Basename for the .opb and .pbp files",                                   //
+                cxxopts::value<string>()->default_value("regex"))                                              //
+            ("regex", "Post the constraint from a regular-expression string instead of an explicit automaton") //
```
ciaranm, line 36: Add a test run that uses this option

gcs/constraints/regular_regex.cc
```diff
@@ -0,0 +1,564 @@
+#include <gcs/constraints/regular_regex.hh>
```
ciaranm, line 1: Worth instead making a `gcs/constraints/regular/` directory as we've done for other constraints that have started to span more files?

ciaranm:
Ooops didn't mean to click the merge button on this one yet.

ciaranm:
Addressed both in 34479c1:

- **`gcs/constraints/regular/` directory** — done, mirroring the `sort/` layout. `regular.cc`/`regular.hh`/`regular_test.cc` plus the regex→NFA helper as `regex.cc`/`regex.hh`/`regex_test.cc` (dropping the redundant `regular_` prefix, as `sort/` does with `arg_sort`/`sortedness`). The public `<gcs/constraints/regular.hh>` is now a thin forwarding header, so `gcs.hh` and the other includers are untouched. Moves done with `git mv` to keep history. Renamed the `regular_regex_test` target to `regex_test`.
- **Test run for the `--regex` option** — added a `regex-regex` ctest that runs the example with `--regex` and verifies the proof (`s VERIFIED COMPLETE ENUMERATION OF 7 SOLUTIONS`); previously the regex-string path was only built, never run. Both runs write `regex.{opb,pbp}`, so they share a `RESOURCE_LOCK` to avoid racing under parallel ctest (the `crystal_maze` pattern).

Full build clean (lib + `fzn-glasgow` + xcsp); `regular_constraint`, `regular_constraint_view_mixed`, `regex_test`, `regex`, `regex-regex` all green.

Issue #146, opened 2026-05-08
`Disjunctive` propagator


ciaranm:
## Background

XCSP3-core defines `noOverlap` (1D and 2D, with constant or variable lengths, with a
`zeroIgnored` flag). MiniZinc's stdlib ships `disjunctive` / `disjunctive_strict` and the
rectangle-packing family `diffn` / `diffn_nonstrict` / `diffn_k` / `diffn_nonstrict_k`.
CPMpy ships `NoOverlap` and `NoOverlapOptional`. All map to the same underlying
non-overlap-in-time capability. We currently have no propagator.

This issue covers the **1D Disjunctive** propagator. 2D (`Disjunctive2D` /
`noOverlap`-k-D / `diffn`) and optional-task variants (`NoOverlapOptional`,
MiniZinc `*_opt`) are parked as follow-ups under this same issue and listed at the end.

### MiniZinc

Two flavours, distinguished by whether zero-duration tasks must still avoid co-occupying
a time point:

```minizinc
% disjunctive_strict — d[i] = 0 still must not overlap with anything
predicate fzn_disjunctive_strict(array[int] of var int: s, array[int] of var int: d) =
    forall (i in index_set(d)) (d[i] >= 0) /\
    forall (i,j in index_set(d) where i<j) (
        s[i]+d[i] <= s[j] \/ s[j]+d[j] <= s[i]);

% disjunctive — d[i] = 0 floats freely
predicate fzn_disjunctive(array[int] of var int: s, array[int] of var int: d) =
    forall (i in index_set(d)) (d[i] >= 0) /\
    forall (i,j in index_set(d) where i<j) (
        d[i]=0 \/ d[j]=0 \/ s[i]+d[i] <= s[j] \/ s[j]+d[j] <= s[i]);
```

The user-facing `disjunctive` short-circuits to `disjunctive_strict` whenever the lower
bound of all durations is `>= 1`.

### XCSP3

`noOverlap` has its own constraint family (independent of `cumulative`), with four parser
callbacks already stubbed at `xcsp/xcsp_glasgow_constraint_solver.cc:475-497`:

1. 1D, constant lengths
2. 1D, variable lengths
3. k-D, constant lengths (per-task origin/length tuples; outer index = task, inner = dim)
4. k-D, variable lengths

A `zeroIgnored` XML attribute selects the strict / non-strict semantics.

### CPMpy

```python
class NoOverlap:
    def __init__(self, start, duration, end=None): ...
[...]

PR #260, opened 2026-06-04
Disjunctive2D: variable rectangle sizes (#146)

ciaranm:
Extends `Disjunctive2D` (#259) to **variable widths/heights** — rotation, strip-packing — with constant *or* variable sizes, every inference VeriPB-certified. Reachable from MiniZinc (`diffn` with `var dx`/`dy`) and XCSP3 (variable-size 2D `noOverlap`). Closes the variable-sizes follow-up under #146.

**Stacked on #259** (base `disjunctive-2d`); GitHub will retarget to `main` once that merges. The 4 commits here are VS-a … VS-d.

## Design — the Cumulative end-proxy, one dimension up ×2

This is the 2D analogue of the Cumulative variable-duration work (#258). With a variable size the `before` flag stays linear (`pos_i + size_i ≤ pos_j`), but the bridge "after" flag `pos_i + size_i ≥ p+1` is two-variable and pinning it hits the cross-variable RUP wall. So when both position and size vary, "after" is reified on a proof-only **`end = pos + size`** (single variable):

- pinning an "after" first **materialises** `end ≥ s_lo + lb(size)` with a `pol` over the captured `end ≥ pos+size` line and the two operand order-literal defs (`s_lo` = `lb(pos)` for a mandatory span, the chain running bound / `¬ext_lit` for the pushed rectangle — the analogue of Cumulative's `pin_pushed`);
- the `before`-flag pol adds the `end ≤ pos+size` line so `end` cancels back to `pos+size` and cancels the `before` flag's term exactly as in the constant case.

A constant size folds into the OPB and keeps the single-variable "after" (no `end`) — constant-size proofs are byte-identical. The propagator is bounds-based (mandatory parts use `lb(size)`, windows `ub(size)`).

Commits: VS-a (size API → `IntegerVariableID`), VS-b (variable-size OPB + bridge + contradiction proof), VS-c (variable-size pushes), VS-d (frontends + docs).

## Scope note

Non-strict mode with a size that **can be 0** needs a reified zero-size escape disjunct in the separation clause (per `fzn_diffn_nonstrict`); that is rejected with a clear message and left as a follow-up. Strict mode, and non-strict with always-positive sizes (the rotation/carpet-cutting case), are fully supported.

## Testing

- `disjunctive_2d_test`: variable + mixed const/var sizes, rotation, wide ranges (>8-bit `end`), a forced variable-size clash (UNSAT), a possibly-zero strict case — all verify full enumeration under VeriPB. Stress-tested 40× strict + non-strict; sanitizer-clean.
- MiniZinc `diffnrotatetest` (rotatable rectangles, var `dx`/`dy`): matches the default solver (232 solutions), proof-verified.
- XCSP `no_overlap_2d_var` (variable sizes): matches an ACE reference (896 solutions), proof-verified.
- Full suite (275 tests) green.


ciaranm:
Update: the non-strict zero-size escape (the scope note's deferred follow-up) is now included in this PR (commit 93679cf). A rectangle whose width/height can be 0 gets a reified `size<=0` escape flag in its separation clauses, matching `fzn_diffn_nonstrict`; the contradiction/reduced-clause pols pin those flags false (sizes are >=1 whenever an inference fires). So **all variable-size modes are now supported** — strict and non-strict, including zero-area. Added a `diffn_nonstrict` end-to-end test (1236 sols, matches the default solver, VeriPB-verified); full suite (276 tests) green.

gcs/constraints/disjunctive_2d.cc
```diff
@@ -90,16 +116,52 @@ auto Disjunctive2D::prepare(Propagators &, State & initial_state, ProofModel * c
...
+    // Non-strict mode: a rectangle whose width or height can be 0 may be
+    // zero-area for some assignments, in which case it does not constrain. Note
+    // which sizes can be zero so define_proof_model can add a zero-size escape
+    // to the separation clause; the propagator already ignores zero-mandatory
+    // rectangles via lb(size).
+    _can_be_zero_w.assign(n, false);
+    _can_be_zero_h.assign(n, false);
+    if (! _strict)
+        for (auto i : _active_rects) {
+            _can_be_zero_w[i] = ! is_constant_variable(_widths[i]) && initial_state.lower_bound(_widths[i]) == 0_i;
```
ciaranm, line 170: Check: where is "not negative" dealt with?

gcs/constraints/disjunctive_2d.cc
```diff
@@ -127,28 +189,80 @@ auto Disjunctive2D::define_proof_model(ProofModel & model) -> void
...
-            flag);
+            "Disjunctive2D", "first rectangle precedes second on this axis", ineq, flag);
         if (! fwd || ! rev)
             throw UnexpectedException{"Disjunctive2D: pairwise reification half missing"};
         return BeforeFlagData{flag, *fwd, *rev};
     };
 
+    // For each rectangle with a variable size on an axis, a proof-only
+    // end = pos + size with both definition directions captured.
+    _end_x.assign(_xs.size(), std::nullopt);
```
ciaranm: These should probably have a `using` rather than `std::` everywhere

gcs/constraints/disjunctive_2d.hh
```diff
@@ -46,11 +48,21 @@ namespace gcs
...
 
+        // Size snapshots resolved in prepare(). _*_vals holds the constant
+        // value for a constant size (0 for a variable one, where the variable
+        // is used instead); _*_ub holds the initial upper bound (used for the
+        // possible-active window and the active-rect filter). Until the
+        // variable-size milestone lands, sizes must be constant.
+        std::vector<Integer> _width_vals, _width_ub;
+        std::vector<Integer> _height_vals, _height_ub;
+        // Non-strict mode: whether each rectangle's width/height can be 0.
+        std::vector<bool> _can_be_zero_w, _can_be_zero_h;
```
ciaranm: `vector<bool>` is usually a bad idea, even if we're not really doing much with it

PR #267, opened 2026-06-05
Verified-encoding workflow: sign-bit alignment + cake_pb_cp end-to-end test

ciaranm:
Stacked on #266 (base is `scp-term-model`; review/merge that first). Brings up the first end-to-end run of the verified-encoding workflow (workflow 2) on a real constraint.

## Commits
1. **Name the integer sign bit `[sign]`, not `[neg]`** — the two's-complement sign bit was named `i[v][neg]` (and `p[v][neg]` for proof-only vars), but `cake_pb_cp` and the encoding spec call it `i[v][sign]`. Since VeriPB matches variables by name, the mismatch meant a solver proof couldn't be checked against cake's re-derived OPB at all. Display-name only; OPB/proof structure unchanged. (Cherry-pickable on its own.)
2. **`examples/cake_abs`** — a tiny example + `run_cake_pb_cp.bash` driver + skippable ctest that exercises the whole chain:
   ```
   cake_abs --prove                                       # scp + opb + pbp
   cake_pb_cp cake_abs.scp > cake_abs.verifiedopb         # re-derive OPB
   veripb cake_abs.verifiedopb cake_abs.pbp --elaborate cake_abs.corepb
   cake_pb_cp cake_abs.scp cake_abs.corepb                # check the core
   ```
   plus an `opbdiff` oracle (solver OPB == cake OPB modulo the `preserved:` line). Both cake checks report **VERIFIED UNSATISFIABLE**; opbdiff reports **10 matches, 0 differing**.

## Notes
- The model is a deliberately **UNSAT** abs instance. An enumeration/optimisation proof logs solutions with `solx`, which veripb can only check against an OPB carrying a `preserved:` line — and `cake_pb_cp` doesn't emit one yet. That's a documented cake limitation to lift later, not an encoding disagreement; until then, refutation proofs are what verify through the chain.
- The test exits 77 (`SKIP_RETURN_CODE`) when `cake_pb_cp`/`veripb` are absent, so it's a no-op on machines without them (`CAKE_PB_CP` overrides the path).
- A companion LaTeX encoding spec (with this abs example auto-regenerated and re-verified) lives in a separate repo, not in this PR.


verified_encodings/run_cake_pb_cp.bash
```diff
@@ -0,0 +1,67 @@
...
+#   2. cake_pb_cp <name>.scp     -> <name>.verifiedopb   (re-derive OPB from scp)
+#   3. veripb verifiedopb pbp --elaborate corepb         (check solver proof
+#                                                         against cake's OPB)
+#   4. cake_pb_cp <name>.scp <name>.corepb               (cake checks the core)
+#   5. opbdiff <name>.opb <name>.verifiedopb (oracle)    (solver OPB == cake OPB,
+#                                                         modulo the preserved set)
+#
+# Exits 77 (the ctest SKIP_RETURN_CODE for this test) when a *required* external
+# tool is missing, so the test is skipped rather than failed on machines without
+# cake_pb_cp or veripb. The opbdiff oracle (steps needing opbdiff + jq) is an
```
ciaranm: Is jq still used?

verified_encodings/abs/abs.cc
```diff
@@ -0,0 +1,104 @@
+#include <gcs/constraints/abs.hh>
```
ciaranm, line 1: I realise I'm changing my mind yet again on this, but if there's going to be a very close mapping between constraints and these verified encodings tests, maybe this just belongs in `gcs/constraints/abs/` as an `abs_verified_encoding_test.cc` (and we take the opportunity to move over any constraints that don't fully make use of a directory yet). Maybe to decide, we need to look at a few more constraints and see whether there's a lot of overlap (as there are already for propagator tests)?

Issue #262, opened 2026-06-04
Research: canonicalise permutation-matrix aux vars via a dominance prelude (instead of bespoke determined-by-construction encodings)


ciaranm:
Spun out of review discussion on #255 (Sort/ArgSort, Mehlhorn–Thiel). Filing as a research idea, **not** a required change to that PR.

## Background

`ArgSort(x, p, offset)` needs an OPB encoding of "`p` is the stable sorting permutation of `x`". The obvious encoding uses `n²` 0/1 aux vars `p[i][j]` for a permutation matrix:

- `Σ_i p[i][j] = 1`, `Σ_j p[i][j] = 1`  (permutation)
- `p[i][j] ⇒ X_i = Y_j`  (`2n²` channel implications)
- plus `Y` sorted

The problem is **partial-solution witnessing**: fixing all `X_i` and `Y_i` does *not* necessarily fix all the `p[i][j]` when there are equal keys (symmetric permutations), so a full assignment to the core variables is not propagated to a unique witness — which the proof framework requires (cf. Matthew's thesis, and `[[proof-only-aux-must-be-determined]]`).

The PR therefore uses a more complex **canonical-stable-rank** encoding (`pos[i]` = number of elements strictly before `i` under value-then-index order; channel `Y[pos[i]] = X_i`). This is AC in the SAT sense and `pos` is a function of `x` alone, so it *is* determined by the core assignment. See `dev_docs/sortedness.md`. Matthew's verdict: "fair enough given what's done everywhere else and what's in my thesis."

## The idea

Rather than design a bespoke determined-by-construction encoding, **keep the simple `n²` permutation-matrix encoding and canonicalise the `p[i][j]` via a dominance step at the start of the proof**. A dominance rule picks a canonical witness among the symmetric permutations, restoring uniqueness without the stable-rank scaffolding.

- This shouldn't break enumeration proofs, since the `p` vars are not solution-preserved (they're aux, not output).
- It likely needs VeriPB's newer **"orders with auxiliary variables"** feature for the dominance order.

If it works, it's a general recipe: simple symmetric encodings + a canonicalising dominance prelude, instead of hand-crafted determined-by-construction encodings, anywhere this partial-witness problem shows up.

## Why "scary"

- Dominance (`G1`) is the strongest VeriPB rule and easy to get subtly wrong; a bad dominance step can make an *unsound* proof verify.
[...]

PR #269, opened 2026-06-05
Sort: permutation-based OPB encoding with direct pigeonhole proofs

mmcilree:
# Matthew's Comment

Based on #262 , I tried to get Claude to implement the dominance idea. 
It churned for hours and basically decided dominance is too hard.
So I guess it probably needs more hand-holding to manage that. 

It has claimed to me though that the simpler encoding I proposed 
(p[i][j] as n^2 binary variables for the permutation rather than pos[i] vars)
can be repurposed into the existing implementation and makes the 
proofs shorter. 

I haven't particularly reviewed this - @ciaranm maybe you could get 
your Claude (with it's better memories) to review what my Claude did 🤪. 

I'm also not sure if I've set up Claude code to work correctly with 
github, but here's the MR it generated and hopefully the commits 
are correctly attributed.

# Claude's MR Body:
## Summary

This PR replaces Sort's `pos[i]` integer-variable witness with a matrix of
`n²` binary `p[i][j]` flags and rewrites the propagation proofs to use a
direct doubly-stochastic pigeonhole argument. ArgSort is unchanged — it
continues to use `SortednessWitness` / `install_sortedness_propagator`.

---

## Background: the encoding problem

VeriPB requires every proof-only auxiliary variable to be uniquely determined
from the real variables (x, y) by unit propagation (UP), so that `solx`
solution-checking lines can be verified. The naive permutation encoding — a
doubly-stochastic binary matrix `p[i][j]` plus the channel `p[i][j] → y[j] =
x[i]` — fails this test: with duplicate x values, multiple valid permutation
matrices exist and UP cannot choose the canonical one.

The previous fix was to use a single proof-only integer `pos[i]` equal to the
*stable rank* of `x[i]` (the count of elements that come before it under the
value-then-index order). `pos[i]` is a function of x alone, so UP pins it.
The downside was that the propagation proofs needed to certify `pos` as a
permutation — an O(n³) root derivation (totality → antisymmetry → transitivity
→ rank-gap → `recover_am1`) — before any bound inference could reuse
surjectivity. Every individual bound proof then required a *pivot bridge* (an
O(n²) step) to propagate the rank structure through the before flags.

---

### New OPB model for Sort (`SortPermWitness`)

The stable-rank canonicalisation is now encoded directly in the OPB model via
three layers of proof-only flags:

```
before[ip][i]   iff  x[ip] comes before x[i] in stable order
                     (fully reified against the x comparison, both halves in OPB)

rank[i][j]      iff  exactly j of the n-1 before-flags for row i are set
                     (forward-ge + forward-le OPB constraints + at-least-one)

p[i][j]         iff  x[i] is placed at sorted position j  (y[j] = x[i])
                     (p-rank link  ~p[i][j] + rank[i][j] >= 1  in OPB)
```

Plus doubly-stochastic (`p row = 1`, `p col = 1`) and the channel
`p[i][j] → y[j] = x[i]`.

**UP-determinism chain:** given x values, UP derives before → rank →
(p-rank link + row-al1) → p → (channel) → y. VeriPB can verify every
solution without any proof-time canonicalisation step.

### Simpler propagation proofs (`propagate_sortedness_perm`)

With doubly-stochastic model constraints in hand:

- **Surjectivity** (`Σ_i p[i][j] ≥ 1`) is the counting pol
  `Σ_i row_al1[i] + Σ_{k≠j} col_am1[k]` — no root initialiser needed.
- **Injectivity** (`Σ_i p[i][j] ≤ 1`) is `col_am1[j]` directly.

Every bound inference — contradiction, y-bounds (normalization / order-statistic
/ Hall), x-bounds (intersection / Hall) — reduces to a single pattern:

1. RUP exclusions: `~p[i][k]` from channel + normalized y-bounds.
2. Restricted row-al1: `Σ_{k∈band} p[i][k] ≥ 1` (RUP from model `row_al1[i]`
   + exclusions), with the goal literal ORed in for bound sub-cases.
3. Counting pol: `Σ_{i∈S} restricted + Σ_{k∈band} col_am1[k]` → contradiction.

There is no pivot bridge, no transitivity chain, and no geqvar/eqvar machinery.

### Why not VeriPB `dom`?

The alternative of keeping a bare doubly-stochastic p model and using VeriPB's
`def_order` / `dom` mechanism to canonicalise p at proof time was fully
explored. The key insight (from the reviewer) that "defining the flags before
introducing the order" avoids the `$`-prefixed auxiliary-variable witness
restriction is correct: before/rank flags introduced via `red` steps before
`def_order` are regular proof variables the `dom` witness can reference.

However, the `def_order` transitivity proof for the canonical-condition
constraints (`~lrank_i_j + lp_i_j + ~rp_i_j ≥ 1`) requires establishing
`v_rank = u_rank` across the two O⪯ instantiations. This chain — x-equality
→ bef-consistency → rank-consistency — is not autoproved by VeriPB's augmented
format, and writing an explicit generated pol chain for each of the n² goals is
complex enough that encoding the flags directly in the OPB model is clearly
simpler. The `dom` / `def_order` / `load_order` infrastructure added to
`ProofLogger` remains for future use.

### New `ProofLogger` methods

- `write_raw_to_proof(string)` — emit a raw block without incrementing the line counter.
- `emit_dom_step(C, witness, level)` — emit `dom C : v1 -> l1 ... ;`.
- `emit_load_order(name, vars)` — emit `load_order name v1 v2 ... ;`.

---

## Verification

All proof-checked sort tests pass:

```
s VERIFIED COMPLETE ENUMERATION OF N SOLUTIONS   (for every satisfiable case)
s VERIFIED UNSATISFIABLE                         (for every infeasible case)
```

ArgSort is unaffected.


ciaranm:
Immediate question: does this work if we have two or more different Sort constraints, potentially on overlapping variables? That's the main reason I've been scared of dominance stuff for this kind of thing.

mmcilree:
Well this specific MR is an alternative encoding that doesn't use dominance at all (because Claude gave up on it).

ciaranm:
Ah, I see, the dominance stuff is just "left over in case we want to review it".

mmcilree:
But I suspect (without thinking that hard about it) that is should be possible: all the separate Sort constraints would use separate p vars (even if they have overlapping variables) so you'd just have to define an order that is the conjunction of all the orders on all the p vars defined.

ciaranm:
Yeah, so we'd need to coordinate orders across all consumers somehow.

mmcilree:
If we could resurrect the idea of using dominance for aux p vars in circuit, there's potentially an interesting research project on a general framework for canonicalising extension variables using dominance across multiple constraints.

mmcilree:
The backwards reifs for the t vars in table would be another candidate.

ciaranm:
I'm burning all my attention and token on trying to get Phil's encodings stuff merged and automated currently. But yes, we should come back to this.

PR #279, opened 2026-06-07
Fix Element .scp encoding; add element/count to scp_reader

ciaranm:
Part of #1 (extending the verified-encoding round-trip). Starts with `element` and `count`, as requested.

## The Element .scp bug (the flagged concern)
`Element::s_exprify` wrote a form `cake_pb_cp` can't parse:
```
(_1 element (9,0 5,1 -1,2 -10,3 9,4 4,5) (_2,0) _1)   <- array as value,index pairs; comma index
```
cake rejected it with `invalid array index; expected (Y offset)`. cake wants a **bare varc list** and a **space-separated** index:
```
(_1 element (9 5 -1 -10 9 4) (_2 0) _1)               <- now emitted; cake parses + encodes it
```
`s_exprify` now emits cake's form (and `element_2d` for the 2-D case). The offset already matched our index-start semantically (both subtract it: chosen entry is `Xs[val(Y) − off]`); only the format was wrong. (`count`'s writer was already correct.)

## Reader support
Adds `element` / `count` to `scp_reader`, so a `.scp` using them can be solved (workflow 3) and round-tripped. `Element` refers to its array **by pointer** and the reader returns before the solve, so new `Problem::keep_array` takes ownership of the backing array and returns a stable pointer (a `deque` never relocates). `scp_reader_test` gains enumeration + write→read→write coverage for both.

## Not yet in the workflow-2 chain harness — and why
The format fix makes cake *parse* element, but the solver's **OPB encoding** diverges from cake's, so the solver's proof doesn't apply to cake's re-derived OPB:
- **element**: 26 vs 30 constraints — cake emits a per-position guarded pair (`R ≥ arr[j]` / `R ≤ arr[j]` under selector `i[I][eqj]`); the solver's formulation is different.
- **count**: OPB counts match, but the solver uses different aux selectors than cake's `x[_1][i][eq]`; `count_unsat` verifies through the chain, `count_sat`'s enumeration trips a RUP step at the aux mismatch.

That's encoding-conformance work (akin to the deferred all_different selector item), separate from this format/reader fix. Flagging for a direction decision (conform the solver to cake, or add to the cake-side request list).

All `element` / `count` / `scp` ctests pass (40/40); `scp_reader_test` 233 assertions / 20 cases.

### DIFF
--- a/dev_docs/scp_s_expr_migration.md
+++ b/dev_docs/scp_s_expr_migration.md
@@ -46,9 +46,10 @@ instead of a string), then finish the cutover:
      `(name op ...)` *with* outer parens, so the writer double-wraps it to
      `((name op ...))`. The bridge preserves this faithfully today; fix it when
      migrating arithmetic (return the body, or a single list).
-   - **`element` index form `_4,0`** — the comma-joined `elem,idx` atom is
-     suspect and likely to change; decide its structured form when migrating
-     `element`.
+   - **`element`** — the old comma-joined `elem,idx` atoms (and `(var,offset)`
+     index) were wrong: `cake_pb_cp` wants a bare varc list `(X0 ... Xn-1)` and a
+     space-separated `(var offset)` index. `s_exprify()` now emits that form (and
+     `element_2d` for the 2-D case); carry it over when migrating to `s_expr()`.
    - **Multi-line tabular constraints** (`table`, `smart_table`, `regular`,
      `knapsack`, …) currently emit pretty multi-line strings; via the bridge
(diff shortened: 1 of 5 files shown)


ciaranm:
Rebased onto #280 (ArrayParam) and **dropped `Problem::keep_array`** — the reader now hands Element its array by value and the constraint owns it via `ArrayParam`. This PR is now **stacked on #280**; merge that first (or together). Net change vs #280: Element `s_exprify` format fix + element/count in `scp_reader` + tests; no `problem.*` changes anymore.

PR #281, opened 2026-06-09
Range/interval literals (staged, WIP)

ciaranm:
**Status: draft / WIP.** Staged work toward first-class range ("in") literals `[x in a..b]` in the CP→PB proof encoding, usable in reasons, inferences, and branching.

### Foundation (done, green)
The whole constraint test suite verifies under `reject_random_interval` branching — VeriPB checks a proof for every constraint while the solver branches by accepting/rejecting random *intervals*. A range literal is reified to its two order cuts: `[a,b] ⇔ (x≥a) ∧ ¬(x≥b+1)`. Two linking axes:
1. **Bound / RUP axis** — the order chain (Inv1) gives range↔range and range↔order deductions on a single variable. No covering needed.
2. **Hole / backtrack axis** — an interval *reject* `¬[a,b]` moves no bound; the chain won't forward-propagate it. Fixed by **laminar containment** (`child → parent` edges down to `eq` leaves), each a `rup` line.

### Stage 3 — `infer_not_in_range` (issue #144): the key finding
Goal: collapse an interval of per-value `infer_not_equal` into one `infer_not_in_range`. **Built and committed** (general, view-safe, green, unused by default): `IntervalSet::erase_range`; `State::infer_not_in_range`; `InferenceTracker`/`ProofLogger::infer_not_in_range`.

**The single-line range *inference* is NOT pure RUP for the constraints that actually do interval removals.** A range flag asserts only *order* atoms; with `lo<hi`, **Theorem 2.8 never pins bits**, so `~[X in lo..hi]` can't cross a **bit-sum equality** `X−Y==0` by RUP (the per-value version works only because asserting `X=v` pins bits). Confirmed via VeriPB `--trace-failed` + McIlree's thesis.

**Scope audited** — the break is exactly *interior interval removal across a primary bit-sum equality*:

| Affected (bit-sum equality) | Unaffected (eq-atom/selector channelled — forward-propagate) |
|---|---|
| `Equals`/`ReifiedEquals`, `Element` (index fixed), `AllEqual`, `Abs` | Table, Regular, GCC, Count, Among, In, Inverse, MinMax, AtMostOne, ValuePrecede, Circuit, AllDifferent |

(Bounds-consistent arithmetic constraints — Linear, Comparison, Plus — never do interior removals, so the question doesn't arise.)

**The fix that works — an explicit bridge (case-split, each case RUP), no re-encoding.** To remove `[lo,hi]` from `v1` because `v2` lacks it (`v1=v2`):
```
1)  ~[v1 in lo..hi] ∨ (v2 ≥ lo)      # RUP: ¬ gives the Theorem 2.9 contradictory-binary-sums config
2)  ~[v1 in lo..hi] ∨ ¬(v2 ≥ hi+1)   #      across the equality (the case-split supplies the opposing bound)
⇒  ~[v1 in lo..hi]                    # RUP from (1),(2) + the disjunctive reason ~[v2 in lo..hi]
```
This is Justification Procedure 3.2 (Comparison) materialising each bound across the equality; `justify_abs_hole` is the per-value sign-split form. **Verified** end to end (full `equals_test`, unmodified bit-sum model). Wired behind **`GCS_RANGE_INFERENCES` (default off)**, with a `range_infer_test`. **Default behaviour byte-identical; full suite 324/324.**

**Proof size** — width-independent inference (3 lines). Measured: narrow ranges break-even (1306 vs 1269), wide hole a ~24% win (445 vs 582). The per-flag `need_invar` setup (reification + O(width) containment) amortises heavily — 23 flags / 932 references (~40× reuse) in `equals_test`. Hoisting the bridge lemmas to `Top` and verify-time effects are not yet measured.

### Stage 3b — the bridge helper, and the headline result
The two bound-lemmas are a **reusable free helper** `gcs::innards::justify_not_in_range_across_equality` (`gcs/constraints/innards/justify_not_in_range.{hh,cc}`), `ProofLogger&` first (the `abs/justify`, `linear/justify` convention, so it travels with the scoopable proof-pattern layer). It lives in `enforce_equality`, so `Equals` and `Element` (index-fixed) inherit it and verify gate-on **when posted in isolation** (`Element`: 61 instances).

**Headline: interval inference does NOT compose across the equality family — and this is general, not `AllEqual`-specific.** A star of plain `Equals` constraints over three transitively-linked variables, with no `AllEqual` anywhere, breaks gate-on identically (same rate, same instances). Reducing `AllEqual` to literally n−1 single-edge `Equals` calls breaks too. So the inference justification was never the problem. The cause: once ≥3 variables share an equality class, the solver's **backtracking justification** needs unit propagation to thread a *bound* across a bit-sum equality (e.g. `i₃≥3 ⟹ i₂≥3`), and UP across a bit-sum equality moves only **values** (Thm 2.8 bit-pinning) and **contradictions** (Thm 2.7/2.9), **never a lone bound** (cf. the thesis's own Example 2.14). Per-value removals never raise the obligation (every one is a single value); a wide interval literal asserts only order atoms, never pins bits, so its obligation is a bound and the missing step is forced. **Two variables in isolation never trigger it — which is the only reason `Equals`/`Element` pass their tests.**

In thesis terms: we **preserve** Theorem 3.3 (intra-variable completeness, extended to interval literals — hence reasons/inference/branching work) but **lose** Inv2 / Theorem 3.4 (the backtracking-RUP invariant) for any constraint that channels across a bit-sum equality. Full thesis-grounded write-up: **`dev_docs/range_literals_theory.md`** (preserved-vs-lost table, the why, the cautionary history). Three refuted "fixes" (pairwise OPB, durable `Top` lemmas, single-hop star) are recorded in `dev_docs/range_literals.md`.

### Status / what's left
- **No code changes pending.** The bridge helper + `infer_not_in_range` primitives are committed; `GCS_RANGE_INFERENCES` stays **off** by default; the 324/324 `ctest` (gate-off) is unchanged. Awaiting review of the theory doc before deciding direction.
- **Decision, not a code task:** interval inference across the equality family needs a cross-variable Inv1 (`(x≥k) ⇔ (y≥k)` per equality at boundary values) — a *covering*, applied family-wide, i.e. the per-value work interval literals exist to avoid. Worth it or not is a design call.
- **Unaffected and usable now:** interval **reasons** (`GCS_RANGE_REASONS`) and interval **branching** (`reject_random_interval`) — both intra-variable, both green.
- **Safe target for interval inference (if pursued):** the eq-atom/selector-channelled family (Table, Regular, GCC, Count, …), where every cross-variable obligation stays a single value and the problem never arises.

Full settled write-up: `dev_docs/range_literals.md` (status/design) and `dev_docs/range_literals_theory.md` (the theory). Relates to #144.

### DIFF
--- a/dev_docs/range_literals.md
+++ b/dev_docs/range_literals.md
@@ -257,3 +257,17 @@ reasons, inferences, AND branching (today reasons and branching use it; inferenc
   verification is manual per constraint. An earlier note that Element was "proven by the suite"
   was unsubstantiated until checked directly (it does pass: `element_test <mode> --prove` with
   `GCS_RANGE_INFERENCES=1`).
+- **Two refuted "more elaborate proof" attempts for AllEqual n≥3** (the OPB is a *line* of n-1
+  consecutive `vars[i]=vars[i+1]`, not a star):
+  1. **Full pairwise (n²/2) OPB instead of the line** — *does not help* (10/40 still fail). The
+     failing backtrack step (`i₂≤4 ⟹ i₃≤4`) is across an equality that is *already adjacent* in
+     the line; UP can't cross it, and adding the other pairs doesn't change that. Not a
+     connectivity problem.
+  2. **Bridge lemmas durable at `Top` (unconditional) instead of `Temporary`** — *necessary but
+     insufficient* (13/40 still fail). They fire and resolve flags (the trail gets much further),
+     and are RUP-derivable at `Top` for a direct equality, but the residual failures are raw
+     var-to-var bound propagations *not mediated by any flag* (the solver's backtrack clauses,
+     e.g. `i₂==5 ∨ i₃==3`).
+  The only thing that would close it is durably materialising the order-atom covering of the
+  equalities (`(vᵢ≥k) ⟺ (vⱼ≥k)`) — exactly the O(width) covering range literals exist to avoid.
+  So for a multi-hop chain, "coalesce the inference" and "keep the proof cheap" are in tension.
--- a/dev_docs/range_literals.md
+++ b/dev_docs/range_literals.md
@@ -7,14 +7,20 @@ off) pending productionisation — see "Stage 3" below for the key finding (inte
 across a bit-sum equality need an explicit bridge proof, not RUP). The bridge now lives in a
 reusable free helper `justify_not_in_range_across_equality`
 (`gcs/constraints/innards/justify_not_in_range.{hh,cc}`) and is wired into `enforce_equality`,
-so **`Equals` and `Element` (index-fixed) both verify it gate-on**. Attempting to extend it to
-**`AllEqual` surfaced an axis-1/axis-2 coupling**: the bridge *inference* is sound, but
-coalescing a multi-variable chain's per-value removals into one range flag breaks *backtrack-
-clause* reconstruction (n≥3 fails; n==2 works) — see "Generalising across the family" below.
-Branch `range-literals`, draft PR #281.
+so `Equals` and `Element` (index-fixed) verify it gate-on **in isolation**. **CORRECTED (and
+the headline): this does NOT compose.** A star of plain `Equals` over three transitively-linked
+variables — no `AllEqual` involved — breaks gate-on identically. So interval *inference* across
+a bit-sum equality is a **general** limitation of the equality family, not an `AllEqual` bug:
+the moment ≥3 variables share an equality class, the backtracking invariant needs a *bound*
+threaded across the equality, which unit propagation cannot do. The two-variable tests pass only
+because two variables in isolation never raise that obligation. The full thesis-grounded
+explanation (which thesis properties we preserve vs lose, and why) is in
+**[`range_literals_theory.md`](range_literals_theory.md)** — read that first. Branch
+`range-literals`, draft PR #281.
 
 This document is the authoritative current-state summary. The git history and PR have the
-blow-by-blow; this is the settled picture so we don't re-derive it.
+blow-by-blow; this is the settled picture so we don't re-derive it. For the *theory* of why the
+equality family breaks, see [`range_literals_theory.md`](range_literals_theory.md).
 
 ## What a range literal is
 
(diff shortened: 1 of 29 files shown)


mmcilree:
> A range literal does not need covering/containment/tree machinery. Reify it to its own two
order-encoding cuts — [x in a..b] ⇔ (x≥a) ∧ ¬(x≥b+1), a "wide equality literal" — and the
existing order chain (need_gevar/Inv1) is what makes propagation complete. 

press X to doubt

ciaranm:
Yes, I totally don't believe this, but I think the only way I'm getting a useful answer out of this is to see how far we get and then really carefully analyse the cause.

ciaranm:
And as suspected, it fails on the "hole in the middle of the interval" problem.

mmcilree:
> And as suspected, it fails on the "hole in the middle of the interval" problem.

It''s kind of funny it can read all of the discussion in #240 and still decide it knows better.

ciaranm:
I'm wondering whether I'm giving bad instructions or whether there's a bad combination of memories / documentation somewhere because we've kept converging on it suggesting the same proposals for fixing it, that are all things I *really* don't want to do. I've asked for a replan: Claude now believes it's eq-atom containment specifically that's the problem, whereas I think it's more likely to be intervals within intervals in general that break. Let's see...

mmcilree:
> I'm wondering whether I'm giving bad instructions or whether there's a bad combination of memories / documentation somewhere because we've kept converging on it suggesting the same proposals for fixing it, that are all things I _really_ don't want to do. I've asked for a replan: Claude now believes it's eq-atom containment specifically that's the problem, whereas I think it's more likely to be intervals within intervals in general that break. Let's see...

Yeah I agree. The eq-atoms are just the specific case of unit 1 intervals, don't think there's anything special about them.

ciaranm:
I think this new plan is actual progress. Whether it's towards where we want to end up remains to be seen.

ciaranm:
Superseded by #303 (curated foundation + reimplementation design spec, `dev_docs/range_literals_spec.md` there). This branch stays unmerged as the archaeological record — its final commit (af3e99f) is labelled throwaway experiments, and the design conclusions in its dev_docs are superseded (the theory doc's §5 "composition requires a cross-variable covering" conclusion was refuted by the 2026-06-10/11 failure analysis; see the spec's Appendix B for the corrected understanding and the witness tests that pin it).

ciaranm:
Superseded by #303 (merged): the range-literal foundation was reimplemented from scratch against `dev_docs/range_literals_spec.md` (always-covered partitions, eq-atom unification, native interval reasons/branching). This staged WIP branch is kept unmerged as archaeology.

PR #302, opened 2026-06-11
Annotated Assertions

mmcilree:
Okay, I believe this is now in a state where I'm happy for it to be Claudited (Claudified? Claudicated? Claudificated?).

## Purpose of these changes 

We are working towards having an external justifier tool that will read a proof with annotated assertions and fill in PB justifications, and also trim the proof. This should hopefully both speed up end-to-end verification time and make it easier to implement proof logging in other solvers. 

As a step towards this, we would like the Glasgow Constraint Solver to have a mode where it asserts all its inferences with annotations sufficient to reconstruct the proof later.

### Assertion Levels
I've decided it will be useful to have several degrees to which assertions will replace proof steps. This is controlled via an ordered enum (lower = more explicit/real derivations, higher = more assertions/omissions). Specifically
- `AssertionLevel::Off`: current behaviour, everything is justified, no assertions. 
- `AsserionLevel::Definitions`: inferences are asserted, no intermediate ad-hoc justification steps are output, but atomic literals and their definitions + links are still derived as normal in the proof, backtracks are still RUP, solutions are logged as normal.
-  `AssertionLevel::Links`: same as Definitions, except the redundance steps introducing geq vars reified on the bit encoding are omitted, and soli/solx now use the full bit assignment rather than eqvars, and assert the blocking constraint. Note that eqvar definitions `x[eq5] <=> x[ge5] /\ ~x[ge6]` are considered "links" for the purpose of this definition.
- `AssertionLevel::Inferences`: no "linking" constraints are output: so no chaining constraints for geqvars,  no containment/covering for range literals, no view linking and no eqvar definitions. This obviously breaks the standard invariants for backtracking search so backtracking steps are asserted at this level
- `AssertionLevel::Backtracking`: only backtracking and solution logging steps are output, asserted. This matches the first level of the Flippo et al. multi-stage framework.

This is controlled by an env var for now, for convenience. If we eventually standardise a command line interface, we could use have a standard `--assertion-level` option.

Some "footguns" (to use Claudespeak):
-  For solution logging to work at AssertionLevel::Links and above, changes are needed to the way VeriPB deals with enumeration proofs, since the asserted constraints in the proof might not necessarily be satisfied by UP. I have an outstanding [work item](https://gitlab.com/MIAOresearch/software/veripb-dev/-/work_items/188) and [merge request](https://gitlab.com/MIAOresearch/software/veripb-dev/-/merge_requests/188) in the VeriPB repo and I think in principle it should be fine to change `solx` semantics to only require constraints 
in the core set to be satisfied.
- I've had to manually add guard statements `if (logger && logger->get_assertion_level() == AssertionLevel:Off)` to everywhere in the solver that emits ad-hoc proof steps outside of infer() or contradiction(). Hopefully I've caught them all.
- `propagate_extensional` is a bit tricky as it relies on the not-real selector variable domain wipeout to justify backtracking and uses `NoJustificationNeeded{}`. The naive approach will either cause a crash (if I try to write something to do with the selector variable without defining it), or a proof logging failure (backtrack rup won't succeed as there's no wipeout of any of the actual variables). I've changed this to explicitly assert a conflict justification when the selector is wiped out, which fixes the problem.
- Similarly, a couple of other places where `NoJustificationNeeded` is used (circuit) need to be special-cased.
- The final rup contradiction line needs to be asserted at `AssertionLevel::Inferences` and above, easy to miss since was old code that doesn't use emit and literally just wrote the string `"rup >= 1;"`.
- I was hoping I could make it possible to assert some inferences and not others (so we can compare immediate vs. deferred justification at some point). This should in theory work at `AssertionLevel::Definitions` (only), but we would need some global method of controlling which assertions were output. Just using the presence of annotation alone to decide whether to output assertions doesn't work, because an un-annotated inference might depend on ad-hoc logging statements (e.g. a gac inference for linear equality) that isn't being logged above `AssertionLevel:Off`.
- The scp tests fail at higher assertion levels. This is because they are looking for the string "s VERIFIED" and veripb outputs "s UNDER ASSERTIONS". This is probably fine.  All other tests succeed at all assertion levels.

### Assertion Hints
I've made up some data structures for actually creating hints. I'm happy with s-expressions as the format. 

Overall, I'd like:
- maximum flexibility for the hint shape - I don't know exactly yet what the external justifier will need and so I'll probably add and remove things as I go along.
- Easy to construct: I've tried to generify some hint_list() constructors for making S expressions.
- Hard to put completely arbitrary data into unilaterally:
    - The hints are going to effectively define the API of the justifier, and will ideally be standardised across solvers that use it.  
    - This is why I've gone for enum specifications all in one place, in addition to attempting to avoid allocated strings in the hints. 
    - However, if there's another way to make a clear API reference while decentralising the hint data definitions, great. 
- Low/zero cost when AssertionHints are not used. This I don't think I currently achieve particularly well - obviously, the enum allocations and empty vectors are low cost, but to extend what I've done to all the constraints would need a lot of guarding to ensure AssertionHins

I'm not massively wedded to the exact implementation details, though, if it can be improved, great!

Which brings me nicely on to...

### Re #347
- I'm on board with the more ambitious refactor idea that solves `Reason` and `AssertionHint` allocation problems at once. 
- I'm not entirely sure about string_view for hints though. I guess it solves the allocation problem and avoids big long enums, but it still allows for arbitrary keywords in hints, whereas I'd rather have it constrained to match the justifier API in one place somewhere.
    - Possibly even one day, the Justifier codebase will provide lightweight libraries in different programming languages that define the unified assertion annotation language. It would be good to design with something like this in mind.

### Re the additional annotation fields
The reason for hint_name as a special colon-separated field separate from the rest of the assertion hints is to give something for vanilla VeriPB to parse, e.g. for a statistics feature or a richer UNDER ASSERTIONS output. It might be nice for veripb to say a bit about what "types" of assertions were used and that's what this field designates. 

Similarly, the reason for derived_from is mostly future-proofing. For now, derived_from won't really be used by the justifier, but in other uses of annotated-assertion, it might be used by a trimmer to determine how to trim assertions properly. For GCS, it could be used if we get to a point where we want to assert some part of a justification (rather than asserting an inference as a whole). In that case, we might want to specify what intermediate derived constraints will be used to derive the assertion, and might become more usable if/when constraint group labels are implemented in VeriPB.

### More things for Claude to do
In addition to #347 refactoring it would be good for Claude to do the tedious work of adding assertion hints everywhere in the solver. 

I'd have a couple of preferences/recommendations for this:
- A lot of Assertion hints will probably have a master hint shape for the whole propagator and then individual inference types will modify it slightly. It would be good to abstract over this interface (some nice builder methods for assertion annotations perhaps).
- Every assertion annotation requires a name, which should correspond to the *model-level* "thing" that explains where the assertion is coming from. So even if we're in a propagate_all_different, if it comes from circuit, it should say circuit. 
- Most (if not all) assertions should specify the constraint_id responsible for the inference in the hints. It'd be good if there was a clean/default way to thread this through. What I've done so far for `gac_all_different` is awkward.
- Many (but not all) assertions will have a main (justifier keyword) pair. E.g. for `propagate_gac_all_different` we have `(justifier hall_set_or_violator)`. This will correspond to the actual justifier function being called, but may not be necessary for every constraint. I would have an easy way to add this "justifier dispatch" keyword but I wouldn't bother adding it for every inference prematurely, as lots of things probably aren't going to need it: for constraints where there's only one way to justify, the constraint id/name will likely be enough.
- I've been a bit sloppy with names: to be consistent, an assertion "annotation" refer to the full 3-field structure, whereas assertion "hints" refers to just the last field.

**IMPORTANT INFO FOR CLAUDE**: If testing proofs at higher assertion levels, main branch VeriPB will fail at the moment, because the assertion_hints feature and the solx behaviour fix are currently outstanding MRs in the development repo. For testing, I have made a temporary branch in VeriPB-dev `temp/annotated-assertions-and-sol-fix` - this is what you should build VeriPB from if building on top of the work in this branch.  

I'm happy for this to be merged and then for Claude to build on top of it, or for Claude to pull in / reuse bits of this branch. Your call @ciaranm. Either way I would like commits based directly on work in the branch credited to me ;-)

### DIFF
--- a/gcs/proof.hh
+++ b/gcs/proof.hh
@@ -38,13 +38,32 @@ namespace gcs
     struct ProofOptions final
     {
         explicit ProofOptions(const std::string &);
-        explicit ProofOptions(const ProofFileNames &, bool, bool);
+        explicit ProofOptions(const ProofFileNames &);
         ProofOptions(const ProofOptions &) = default;
 
         ProofFileNames proof_file_names;       ///< Filenames for OPB, proof, and mapping files
         bool verbose_names = true;             ///< Use verbose names in proofs?
         bool always_use_full_encoding = false; ///< Always write the full variable encoding to the OPB file
         bool use_annotated_assertions = false; ///< Only write annotated assertions instead of full justifications.
+
+        /// Write annotated assertions instead of full justifications.
+        ProofOptions & enable_assertions()
+        {
+            use_annotated_assertions = true;
+            return *this;
+        }
+        /// Always write the full variable encoding to the OPB file.
+        ProofOptions & enable_full_encoding()
+        {
+            always_use_full_encoding = true;
+            return *this;
+        }
+        /// Set whether to use verbose names in proofs.
+        ProofOptions & set_verbose_names(bool v)
+        {
+            verbose_names = v;
+            return *this;
+        }
     };
 
     class Proof
--- a/gcs/proof.hh
+++ b/gcs/proof.hh
@@ -58,11 +58,13 @@ namespace gcs
         bool verbose_names = true;             ///< Use verbose names in proofs?
         bool always_use_full_encoding = false; ///< Always write the full variable encoding to the OPB file
         AssertionLevel assertion_level = AssertionLevel::Off;
+        bool assertion_level_set_explicitly = false; ///< Was assertion_level set in code (so it overrides the env var)?
 
         /// Write annotated assertions instead of full justifications.
         ProofOptions & set_assertion_level(AssertionLevel a = AssertionLevel::Inferences)
         {
             assertion_level = a;
+            assertion_level_set_explicitly = true;
             return *this;
         }
         /// Always write the full variable encoding to the OPB file.
--- a/gcs/proof.hh
+++ b/gcs/proof.hh
@@ -44,6 +44,7 @@ namespace gcs
         ProofFileNames proof_file_names;       ///< Filenames for OPB, proof, and mapping files
         bool verbose_names = true;             ///< Use verbose names in proofs?
         bool always_use_full_encoding = false; ///< Always write the full variable encoding to the OPB file
+        bool use_annotated_assertions = false; ///< Only write annotated assertions instead of full justifications.
     };
 
     class Proof
--- a/gcs/proof.hh
+++ b/gcs/proof.hh
@@ -30,6 +30,19 @@ namespace gcs
     };
 
     /**
+     * \brief Mode setting for annotated assertions. Each option involves successively less
+     * justification.
+     */
+    enum class AssertionLevel
+    {
+        Off = 0,      /// Justify everything; no annotated assertions
+        Definitions,  /// Justify backtracking, links, definitions; assert inferences
+        Links,        /// Justify backtracking; assert inferences & links; omit definitions
+        Inferences,   /// Assert backtracking & inferences; omit links & definitions
+        Backtracking, /// Assert backtracking only; omit everything else
+    };
+
+    /**
      * \brief Options for a Problem telling it how to produce a proof.
      *
      * \sa Problem
@@ -44,18 +57,18 @@ namespace gcs
         ProofFileNames proof_file_names;       ///< Filenames for OPB, proof, and mapping files
         bool verbose_names = true;             ///< Use verbose names in proofs?
         bool always_use_full_encoding = false; ///< Always write the full variable encoding to the OPB file
-        bool use_annotated_assertions = false; ///< Only write annotated assertions instead of full justifications.
+        AssertionLevel assertion_level = AssertionLevel::Off;
 
         /// Write annotated assertions instead of full justifications.
-        ProofOptions & enable_assertions()
+        ProofOptions & set_assertion_level(AssertionLevel a = AssertionLevel::Inferences)
         {
-            use_annotated_assertions = true;
+            assertion_level = a;
             return *this;
         }
         /// Always write the full variable encoding to the OPB file.
-        ProofOptions & enable_full_encoding()
+        ProofOptions & enable_full_encoding(bool e = true)
         {
-            always_use_full_encoding = true;
+            always_use_full_encoding = e;
             return *this;
         }
         /// Set whether to use verbose names in proofs.
--- a/examples/cake/cake.cc
+++ b/examples/cake/cake.cc
@@ -92,8 +92,9 @@ auto main(int argc, char * argv[]) -> int
                 variable_order::dom_then_deg(vector<IntegerVariableID>{banana, chocolate}),
                 value_order::largest_first())},
         options_vars.contains("prove")
-            ? make_optional<ProofOptions>(ProofFileNames{options_vars["proof-files-basename"].as<string>()},
-                  true, options_vars.count("full-proof-encoding"))
+            ? make_optional(options_vars.count("full-proof-encoding")
+                      ? ProofOptions{ProofFileNames{options_vars["proof-files-basename"].as<string>()}}.enable_full_encoding()
+                      : ProofOptions{ProofFileNames{options_vars["proof-files-basename"].as<string>()}})
             : nullopt);
 
     if (options_vars.contains("stats"))
--- a/examples/crystal_maze/crystal_maze.cc
+++ b/examples/crystal_maze/crystal_maze.cc
@@ -8,6 +8,7 @@
 
 #include <cstdlib>
 #include <iostream>
+#include <optional>
 #include <string>
 #include <vector>
 
@@ -49,12 +50,13 @@ auto main(int argc, char * argv[]) -> int
     cxxopts::ParseResult options_vars;
 
     try {
-        options.add_options("Program options")                                                       //
-            ("help", "Display help information")                                                     //
-            ("prove", "Create a proof")                                                              //
-            ("proof-files-basename", "Basename for the .opb and .pbp files",                         //
-                cxxopts::value<string>()->default_value("crystal_maze"))                             //
-            ("stats", "Print solve statistics")                                                      //
+        options.add_options("Program options")                               //
+            ("help", "Display help information")                             //
+            ("prove", "Create a proof")                                      //
+            ("proof-files-basename", "Basename for the .opb and .pbp files", //
+                cxxopts::value<string>()->default_value("crystal_maze"))     //
+            ("stats", "Print solve statistics")                              //
+            ("assert", "Use annotated assertions")                           //
             ;
 
         options.add_options()             //
@@ -101,6 +103,12 @@ auto main(int argc, char * argv[]) -> int
         p.post(LinearEquality{WeightedSum{} + 1_i * xs[x1] + -1_i * xs[x2] + -1_i * diffs.back(), 0_i, options_vars.contains("gac")});
     }
 
+    auto proof_options = options_vars.contains("prove")
+        ? make_optional(options_vars.contains("assert")
+                  ? ProofOptions{options_vars["proof-files-basename"].as<string>()}.enable_assertions()
+                  : ProofOptions{options_vars["proof-files-basename"].as<string>()})
+        : nullopt;
+
     auto stats = solve_with(p,
         SolveCallbacks{
             .solution = [&](const CurrentState & s) -> bool {
@@ -111,9 +119,7 @@ auto main(int argc, char * argv[]) -> int
                 return true;
             },
             .branch = branch_with(variable_order::dom_then_deg(xs), value_order::smallest_first())},
-        options_vars.contains("prove")
-            ? make_optional<ProofOptions>(options_vars["proof-files-basename"].as<string>())
-            : nullopt);
+        proof_options);
 
     if (options_vars.contains("stats"))
         print("{}", stats);
--- a/examples/crystal_maze/crystal_maze.cc
+++ b/examples/crystal_maze/crystal_maze.cc
@@ -3,6 +3,7 @@
 #include <gcs/constraints/equals.hh>
 #include <gcs/constraints/linear.hh>
 #include <gcs/problem.hh>
+#include <gcs/proof.hh>
 #include <gcs/search_heuristics.hh>
 #include <gcs/solve.hh>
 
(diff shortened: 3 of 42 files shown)


gcs/proof.hh
```diff
@@ -38,12 +38,32 @@ namespace gcs
...
+        explicit ProofOptions(const ProofFileNames &);
         ProofOptions(const ProofOptions &) = default;
 
         ProofFileNames proof_file_names;       ///< Filenames for OPB, proof, and mapping files
         bool verbose_names = true;             ///< Use verbose names in proofs?
         bool always_use_full_encoding = false; ///< Always write the full variable encoding to the OPB file
+        bool use_annotated_assertions = false; ///< Only write annotated assertions instead of full justifications.
+
+        /// Write annotated assertions instead of full justifications.
+        ProofOptions & enable_assertions()
```
ciaranm: Minor: if you make these take a bool argument that's defaulted to true, it'll save you some conditional mess when doing command line handling.

mmcilree: Yeah I'm going to rethink this. I've decided it's worth implementing at least 3 assertion modes:
1. Only emit backtracking justifications as assertions, no inferences, no literals. This matches the proof scaffolding approach from Flippo et al. CP2025
2. Emit backtracking justifications as RUP steps, emit ALL inferences as assertions and emit literal linking constraints as assertions, but no literal definition red steps. I believe this should still verify, and would allow the trimmer/justifier to decide which BinEnc definitions are actually needed
3. Emit backtracking justifications as RUP steps, emit full literal linking and definitions as normal, and emit potentially some inferences as assertions and fully justify others. This has the advantage that inferences that are easy/one-line to justify can be done eagerly, while other ones can be asserted.

I'm thinking an env flag with 3 options (4 if you count OFF) is probably the way to go on this.

ciaranm:
Heads-up @mmcilree — I've opened #317, which adds an explicit `ConstraintID` first argument to `Propagators::install(...)` (the "stable constraint → propagator association" enabler from #315). It's mechanical but touches every constraint's install site, so it will conflict with this branch once it lands.

I deliberately **did not** push the fix to your branch, in case you have uncommitted local work here. The overlap with this PR is exactly five files:

- `gcs/constraints/abs.cc`
- `gcs/constraints/equals.cc`
- `gcs/constraints/all_different/gac_all_different.cc`
- `gcs/constraints/linear/linear_equality.cc`
- `gcs/constraints/innards/reified_dispatcher.hh`

Resolving (after #317 merges) is purely mechanical:

- Every `propagators.install([lambda], triggers)` becomes `propagators.install(constraint_id(), [lambda], triggers)` — i.e. add the owning constraint's `constraint_id()` as the new **first** argument (it's already available as a member).
- The shared helper call `install_reified_dispatcher(propagators, ...)` becomes `install_reified_dispatcher(propagators, constraint_id(), ...)`.
- Where a conflict hunk shows both my `constraint_id()` addition and your `AssertionAnnotation{...}`, **keep both** — they're on different arguments (the `install` first arg vs. the `inference.infer*` trailing arg), so there's no real clash.

Your `AssertionAnnotation` edits are to `inference.infer*` calls inside the propagation lambdas; mine are to the enclosing `propagators.install(...)` call — so in practice most of the five files should merge with little or no hand-editing.

Happy to do the rebase for you whenever you'd like — just say the word and I'll merge `main` in and push.

mmcilree:
> Happy to do the rebase for you whenever you'd like — just say the word and I'll merge `main` in and push.

Have rebased. Only difficulty was the clang-format strangeness that I've now fixed on main #319

ciaranm:
> **🤖 AI-generated audit — scoping note, not a merge blocker.**
> Ciaran here: I had an independent AI code-quality auditor look over this branch **purely to scope how much work the combined #302 + #347 merge (#378, per the #377 gate) will be** — not to gate or grade @mmcilree's work, which is explicitly a demonstration branch. Posting it so we have it in front of us; **let's talk it through at Tuesday's meeting** rather than actioning anything off it now. Findings are an auditor's read against `b342bac5` and may include false positives — treat as discussion input.

---

## Audit — `annotated-assertions` (#302), scoping read

**Scope:** 42 files, +864/−285 (merge-base diff vs `main`), at `b342bac5`. Feature: `AssertionLevel` enum, `AssertionAnnotation`, hint builders, `GCS_ASSERTION_LEVEL` env var, threading `optional<AssertionAnnotation>` through `infer*`/`emit*`, and assertion-level guarding of proof-logging sites.

### Verdict
A **prototype/demonstration branch**, and a clean one architecturally. The infrastructure — the `AssertionLevel` enum, the `ProofOptions` fluent-setter refactor (a real improvement over the old positional `bool, bool`), the env override, and the annotation param-threading — is idiomatic mechanical work. What's *unfinished* is the part that delivers the value: the hint content is **one** worked demo (`gac_all_different`), per-constraint coverage is uneven, there are **no tests**, and there's **one real bug**. Mechanical sweep was otherwise clean (no debug prints, commented-out code, `#if 0`, TODO/FIXME). Findings tagged **[would-ship-broken]** vs **[prototype-expected]** to separate "fix regardless" from "expected of a demo".

### 🔴 Bug — would-ship-broken
- **`symmetric_all_different.cc:140`** — `if (! logger || logger->get_assertion_level() >= AssertionLevel::Off) return;`. Since `Off = 0` is the minimum enum value, `>= Off` is **always true**, so the AM1-cache initialiser never runs — at any level, including the default `Off`. The five sibling guards all use `> Off` (`inverse.cc:153`, `disjunctive.cc:198/594`, `sort.cc:985`, `disjunctive_2d.cc:326`). This regresses symmetric-all-different's normal proof path (leaves `value_am1s` empty for the GAC propagator's proof logging). Likely just `>=` → `>`; worth a proof-logging test over `SymmetricAllDifferent` to confirm the blast radius. *(Independent of the assertion feature — affects the default path.)*

### 🟠 Serious
- **No tests for the feature** — the only added file is `assertion_hints.hh`; nothing runs VeriPB against asserted proofs at any level, and the env parser / `bit_assignment_string_for` are untested. The 3 test-file edits are just `ProofOptions(ProofFileNames)` constructor adaptations. [prototype-expected, but the gating item for "finished"]
- **`names_and_ids_tracker.cc:1179`** — `auto will_define = true;` feeding a ternary whose `: make_pair(ProofLine{}, ProofLine{})` branch is dead (always-true condition; copy-paste artifact from `define_plain_invar`). Confusing dead code in proof-critical machinery. [would-ship-broken]

### 🟡 Moderate
- **`names_and_ids_tracker.cc:969`** — `// Struggling to get clang-format to behave here...` shipped, with a formatter-defeating trailing-`//` hack. [would-ship-broken]
- **`proof_logger.cc:445-448` & `502-507`** — tab indentation (incl. mixed `\t \t`) injected into a space-indented file, closing crammed onto `}}`; same tab/space mangling in the `need_gevar` BoundLink lambdas. Not clang-formatted. [would-ship-broken]
- **Scattered assertion-level policy** — the `assert_or_rup = level==Links ? Assert : RUP` ternary is duplicated 4×, and the `level > AssertionLevel::Links → return` guard **15×**, all as bare magic comparisons. In proof-critical code a wrong threshold is a soundness risk — candidate for named predicates. [prototype-expected]
- **`gac_all_different.cc:264-275` vs `393-404`** — the only rich `hint_fields` block (constraint_id / hall_vars / hall_vals / justifier + the term-building loop) is copy-pasted verbatim between `prove_matching_is_too_small` and `prove_deletion_using_sccs`. [prototype-expected]
- **Demo-scoped taxonomy** — `AssertionHintName` + `AssertionHintIdentifier` enums, each with a hand-written `operator<<`/`as_string` switch, sized to crystal_maze's constraints. **NB:** #347 (on #378) already replaces this enum scheme with decentralised typed witnesses — so this churn is partly superseded by work sitting on top. [prototype-expected]
- **Uneven coverage** — abs/equals/linear_equality annotate every inference with *empty* payloads; gac_all_different richly; circuit_scc/regular/smart_table/extensional_utils/disjunctive*/knapsack/auto_table got only `Off`-guards (bare `a …;` with no hint above Off); most constraints untouched. At `Inferences` a verifier gets meaningful hints for crystal_maze's handful only. [prototype-expected]

### 🟢 Mild
- Global-scope `using std::format;` / `using fmt::format;` in `assertion_hints.hh` (a header) — now transitively pulled in via `proof_logger.hh`, leaking `format` into many TUs.
- Typos in shipped comments: `proof_logger.cc:455` "runtime-emitted"→"runtime-**mitted**" (corrupted an existing comment); `names_and_ids_tracker.cc:633` "emmitted".
- `proof.cc` env parser doc says "case-insensitive" but only accepts two exact spellings per level.
- `proof.hh` `AssertionLevel` enumerator docs use `///` (attaches to the next member) instead of `///<`.
- `assertion_hints.hh` `hint_sexpr(SExpr expr)` by-value identity overload copies every already-`SExpr` arg in the `hint_list` forwarding builder.
- `AssertionAnnotation{.hint_name = ::X}` bare-tag repetition (abs ×11, equals ×11, linear_equality ×9), constant-per-file.
- `circuit_scc.cc:1101` stray blank line; the `NoJustificationNeeded`(Off)-vs-`JustifyUsingRUP`(else) inference fork duplicated ad hoc (circuit_scc, smart_table, extensional_utils).

### Clean (good craft / mechanical threading, no findings)
`linear/propagate.{cc,hh}` (incl. all 6 explicit instantiations), `reified_dispatcher.hh`, among/knapsack/auto_table/inverse/sort/arg_sort/all_different_except/circuit_base/circuit_prevent, `recover_am1.cc`, the `ProofOptions` fluent-setter migration across example/scp/verified_encodings, and the `proof.cc` env-var parser.

### Effort to production-ready (the scoping question)
The architecture is done; what remains is breadth + rigour:
1. **~1 hour** — the `symmetric_all_different` guard bug + clang-format + typos.
2. **Days → weeks** — decide whether the hint taxonomy is the real contract; if so, every ~30 constraints needs a hall-set-style worked hint. *(Partly mooted by #347's decentralised-witness rework.)*
3. **~1 day** — factor the duplicated rich-hint helper, the Off-vs-else inference forks, the threshold predicates.
4. **Several days (the real gate)** — a VeriPB-gated test suite proving assertions verify at each `AssertionLevel`.

**Bottom line for #378/#377:** one bug-fix from "safe to merge behind the default-`Off` flag"; many days from "feature finished and trusted." Since #378 already builds the next layer on top and they graduate as one unit, some demo-taxonomy churn here is already superseded.

<sub>Auditor-generated, independent code-quality pass against `annotated-assertions` @ `b342bac5` vs `main`. Discussion input for Tuesday's meeting; not a merge gate. Posted per the repo's AI-contribution policy.</sub>

ciaranm:
Landed via #394 (merged). The annotated-assertions machinery is on `main` as commit `4a433958`, **authored by @mmcilree** — the #302 work reconciled onto the post-#393 restructure (the original 22 commits couldn't replay verbatim because #384 had rewritten `disjunctive{,_2d}.cc` in the meantime, so it's a faithful squash of them with authorship preserved).

Closing here since it merged as part of the integrated #394 rather than on its own branch. Thanks @mmcilree! The one outstanding piece is assertion-mode end-to-end verification, which still needs the VeriPB dev branch (stock 3.0.2 caps at Definitions).

PR #303, opened 2026-06-11
Range literals: design spec + reimplementation (always-covered partitions)

ciaranm:
## What this is

The range-literal **reimplementation**, built against the design spec it ships with (`dev_docs/range_literals_spec.md`). The first two commits are the curated foundation + spec (reviewed 2026-06-11); the remaining commits implement the spec in full. This supersedes the approach on #281: that branch stays open, unmerged, as the archaeological record.

## Background, briefly

The #281 work concluded that interval *inference* "does not compose across the equality family" and that a global cross-variable covering would be needed. The 2026-06-10/11 sessions refuted that: every reproducible failure traced to unit-propagation linking gaps in the interval-literal layer — width-1 doppelgänger flags, and reason flags that nothing can falsify (including union flags) — each fixable with one RUP line, and each now pinned by a deterministic 2-3 variable witness instance that fails veripb in milliseconds. The "bound can't cross a bit-sum equality" limit is real but was never the binding constraint.

## The implementation (spec §§2–4, commits 3+)

- **Always-covered partitions in `need_invar`** (spec §3): each variable with interval literals carries a partition of its initial range into cells; a request's endpoints split cells (each split emits its covering), the requested literal gets a covering over the cells it spans, containment edges link immediate neighbours, and the first request emits the root covering. The whole-variable literal is never materialised. Everything is state-independent, at `ProofLevel::Top`. Requests clamp to the definition bounds (same solver fact, keeps the partition invariant).
- **Width-1 ≡ eq atom** end to end: `need_invar(v,v)` returns the direct-encoding literal (return type is now `ProofLiteralOrFlag`); `infer_not_in_range` concludes width-1 removals as `var != v`; `need_direct_encoding_for` keeps the other direction honest by singleton-splitting the partition for new eq atoms (and partition setup gives every pre-existing eq atom its singleton cell).
- **First-class interval reasons** (spec §4): `VariableNotInRange` reason elements, resolved by `NamesAndIDsTracker::resolve_reason` at all four `proof_logger.cc` sites (spec §9.2); per-value fallback for views, constants, and direct-only variables. `generic_reason` states hole runs as interval elements; the three `short_reasons` consumers resolve before building their conjunction flags.
- **Consumers**: the Equals holes path is interval-based unconditionally for simple variables (`GCS_RANGE_INFERENCES` retired), keeping the Theorem-2.9 bridge lemmas; `In` over constants batches initial-domain gaps into interval conclusions (spec §9.3) — the per-value eq-atom factory is gone.

## The witness suite (spec §8)

`range_witness_w{1,2,3,5}_test`, veripb-gated; W4 is `range_branch_test` + suite-wide interval-reject branching. Each was validated by temporary ablation before commit:

- **W1** fails if width-1 mints flags — *even with coverings and containment intact*, which the pre-implementation UP analysis did not predict (a live instance of the spec §1 trap).
- **W3** fails within milliseconds if either the request covering or containment is removed.
- **W2 / W5** are structural under the new design (no single-family ablation breaks them); kept as composed regression nets. The root covering's apparent forward-UP redundancy is recorded in spec §8.1 as an observation, deliberately not acted on, per the change protocol.
- Known gap (spec §8.1): no deterministic witness yet for split coverings composing across refinements; the W4 net covers that shape randomly.

## Validation

- full suite 329/329, caps **on** and caps **off**, interval-reject branching default (multiple randomized repetitions)
- seed soak: equals/count/among/inverse × 5 seeds, element × 4 modes × 2 seeds, all veripb green
- `invar_test` (partition + adversarial cut order + width-1 unification), `range_infer_test`, `range_branch_test` veripb-verified
- W3's proof inspected by hand: root covering, unit interval conclusions, 2.9 bridge lemmas, exact-match interval reason, containment edges, and the union covering all present and minimal (a reason-duplication wart in the old mutable-lambda pattern was found and fixed along the way)


mmcilree:
@ciaranm Are you looking for a general review, or something specific? Does Claude claim this is fully working now?

ciaranm:
Yeah, the claim is that this is fully working. It's not fully done in that there are currently some awkward bits of the code because (intentionally for now) intervals aren't yet exposed to IntegerVariableCondition, and not everywhere that makes use of intervals actually does. I'm going over the code carefully myself, but I'd appreciate another check of the design.

mmcilree:
Cool, I'll have a look over it tomorrow morning.

ciaranm:
## Remaining work to mergeable

**Done (2026-06-11/12):**
- [x] **Fold range conditions into `IntegerVariableCondition`** (`InRange`/`NotInRange` operators), replacing the `ReasonElement` variant-of-variants and the `VariableNotInRange` bolt-on. Ranges then ride the standard `need_proof_name` → `xliteral_for` path (the §9.2 "four resolution sites" collapse into the existing machinery), reasons revert to `small_vector<ProofLiteralOrFlag, 2>`, and `BranchGuess`/`IntegerVariableRangeGuess` fold back into plain `Literal` branching. No change to the §3 partition design — vocabulary plumbing only.
- [x] Re-validate after the refactor: full suite caps on/off, witness suite, CI matrix.

**Done before review (results in PR comments below):**
- [x] **Proof-size/verify-time benchmarks vs main**, focused on the pathology that matters: benchmarks doing simple things over *large ranges* must not produce huge proofs. Also measure whether *small* intervals cost more to verify than per-value lines — if so, a width threshold for enabling interval literals is on the table.
- [x] **Overnight randomized soak** at the prior evidence bar (≥1M instances, veripb-on, gate-free), refreshing the `range_compose` probe harness; runs against the post-refactor shape.

**After review:**
- [ ] L1/L2 lemma proof sketches as a spec appendix (timing per review — see questions below).
- [ ] Deterministic split-covering witness (W6): branch A excludes an interval coarsely (minting it over the then-cells), branch B re-excludes it piecewise after splits, so the replay must compose through the split covering. Closes the §8.1 witness gap.
- [ ] Promote ratified status notes into normative spec text (bounds clamping; width-1 return type); fold W6 into §8.
- [ ] File follow-up issues for deferred scope: OPB-side initial-domain gaps (CakePB conformance), folding views into interval *conclusions*, `In`-over-variables interval batching (needs selector bridge lemmas), root-covering question if deferred.
- [ ] Merge latest main, final suite run, un-draft, merge; close #281 as superseded.

**Deliberately out of scope:** In bounds-only mode; any OPB encoding changes; per-equality global coverings (refuted, Appendix B6).

## Questions for @mmcilree

1. **§8.1 root-covering observation**: the root covering looks forward-UP-redundant given the bound-axiom units + reverse reifications + Inv1 chain. Recommendation is to *keep it* and merge (W1's ablation just showed these UP analyses under-predict), but it's your call whether to pursue removal via the change protocol now or as a follow-up.
2. **Bounds clamping**: `need_invar` clamps requests to the definition bounds (same solver fact given the bound axioms; required so requests stay unions of cells). Currently a status note — OK to promote into §3's normative text?
3. **L1/L2**: merge gate or fast-follow? The empirical evidence (suite + soak + witnesses) doesn't substitute for the lemma write-ups, but they also don't block the code.
4. **CakePB surface**: all interval machinery is proof-phase only (`need_invar` still throws during model writing), so the OPB encoding the verified-encoding pipeline conforms against is unchanged. Please confirm you're happy that's the right boundary for now (the §9.3 open decision would change it later).
5. **Range conditions as `IntegerVariableCondition` ops** (underway): any proof-layer concerns with ranges being constructible wherever conditions are — in particular, reification-condition contexts will throw `UnimplementedException` until model-phase `need_invar` exists (§9.3)?
6. **W2/W5 being structural** (no single-family ablation breaks them — multiple mechanisms each suffice): comfortable keeping them as composed regression nets, with W6 as the targeted addition?

ciaranm:
## Benchmark results: proof sizes, main vs this branch

Answering the checklist's performance question (do proofs go huge on benchmarks doing simple things with large ranges, and do small intervals cost enough to justify a width threshold). Identical instances both sides (seeded where the example is randomised); OPB files byte-identical throughout, confirming the encoding surface is unchanged. Proof **size** is the primary metric; wall times for the two largest runs were contaminated by concurrent jobs and will be re-measured if they matter.

**The large-range direction is a win, and it widens with range size.** A dedicated probe (Equals chain over `0..n` with wide initial-domain gaps, default branching, identical search both sides):

| n | main pbp | branch pbp | ratio |
|---|---:|---:|---:|
| 1,000 | 638,733 | 273,145 | 0.43× |
| 10,000 | 7,590,118 | 2,860,612 | 0.38× |
| 100,000 | 85,948,358 | 30,150,829 | 0.35× |

**Most examples engage the machinery not at all** — proofs byte-identical: cake, money, crystal_maze, cumulative, multiply_random ×2, random_polynomial ×2, circuit_random ×2, regular_random (1 of 2 seeds). Zero overhead when unused.

**Where proofs changed:**

| example | main pbp | branch pbp | Δ |
|---|---:|---:|---|
| odd_even_sum | 43,069,854 | 43,076,045 | +0.014% |
| langford 7 | 998,336 | 1,002,838 | +0.45% |
| knapsack | 403,276 | 409,223 | +1.5% |
| sudoku | 1,067,202 | 1,088,479 | +2.0% |
| regular_random (seed 43) | 75,977 | 74,460 | −2.0% |
| skyscrapers | 8,394,612,977 | 8,361,743,061 | −0.4% |
| ortho_latin 5 | 2,174,526 | 1,789,950 | −17.7% |

**Width-threshold assessment**: the worst regression is sudoku at +2.0% (veripb 0.063s → 0.075s on a quiet machine) — alldiff eliminations make many width-2/3 hole runs, each minting a flag + reification + linking at Top. My read: this does not currently justify a magic constant; per the spec's cost model ("measure, don't pre-optimise") I'd record the option — the natural site would be `generic_reason` / the Equals holes path emitting per-value below width W — and revisit if a real workload shows worse. Happy to add the knob now if you'd rather have it from the start.

Full data and method notes in `~/claude/tmp/range_bench/` (results.tsv, NOTES.md) on the dev box. The ≥1M-instance overnight soak (full machinery, veripb-on, deterministic per-index repro) is running; status lands in `range_compose/STATUS_v2.md` by morning.

---

**Addendum — verification wall time at scale** (bigrange n=10,000): veripb takes **84 min on main vs 66 min on this branch** (−21%, and the branch run shared the machine with the soak, so understated if anything). The interesting part: a 7.6MB proof taking 84 minutes means wide-domain verification time is dominated by per-RUP-check cost over the large OPB (13.8MB of bit-encoded constraints), not by proof bytes. The branch improves the constant, not the shape — relevant context for any plan to routinely verify wide-domain benchmarks, but not a regression.

dev_docs/range_literals_spec.md
```diff
@@ -0,0 +1,435 @@
...
+
+The trap: **the checker's RUP search is strictly stronger than unit
+propagation, so every P2 clause looks deletable under P1 testing.** Any
+linking clause can be removed and a local test stays green, because RUP
+re-finds the derivation that UP would have needed the clause for. "This
+clause was never load-bearing in my tests" is *always* observable for P2
+clauses on small tests. It is not evidence. The only evidence that a P2
+clause is unnecessary is passing the witness suite of §8, which was built so
+that each clause family's removal fails within milliseconds.
+
```
mmcilree: This is the kind of thing that is maybe useful for Claude but not particularly good if it's expected to double as docs for the solver for humans. In particular "replay-completeness" is a term Claude made up that shouldn't really be classed as a "correctness" property. The basic point is just "the root cause of a proof failure isn't necessarily going to come directly from the line that VeriPB takes issue with".

dev_docs/range_literals_spec.md
```diff
@@ -0,0 +1,435 @@
...
+   at-least-one clause, and gives wipeout detection (Theorem 3.2 analogue)
+   without descending to the bits.
+
+   *Decided (2026-06-11 review):* conceptually this is the covering of the
+   whole-variable literal `[x in lb..ub]`, so the covering rule is uniform
+   with no special root case — but that literal is **never materialised**.
+   The clause over the top-level partition is emitted directly: it has the
+   same propagation power given the root bound units, positively asserts the
+   last surviving piece without a detour through a root reification (which is
+   what feeds Theorem 2.8 value-crossing), and avoids minting a literal
```
mmcilree: "minting" is kind of a cute term but "defining" or "introducing" would be more standard, and probably clearer overall.

dev_docs/range_literals_spec.md
```diff
@@ -0,0 +1,435 @@
...
+clauses), where the path's own inference clauses are live and fire over the
+vocabulary above. **A lone bound still cannot cross a bit-sum equality — the
+thesis Example 2.14 limit is real — but no replay obligation requires it once
+the vocabulary is complete.** The 2026-06 experiments that seemed to show
+otherwise were P2 gaps in disguise (Appendix B).
+
+**Lemma obligations (to be formalised, not assumed):**
+
+- **L1 (intra-variable).** With §3 maintained, the {gevar, eqvar, invar}
+  layer is forward-UP-complete with respect to the variable's possible
```
mmcilree: Claude seems to like saying "forward-UP", presumably to distinguish from "reverse UP". I don't like this - if it means plan UP from the empty assignment it should say that. If it means something else it should define precisely what it means by forward-UP.

gcs/constraints/circuit/circuit_scc.cc
```diff
@@ -42,6 +42,7 @@ using std::string;
 using std::stringstream;
 using std::unique_ptr;
 using std::vector;
+using std::visit;
```
mmcilree: Why?

gcs/constraints/regular/regular.cc
```diff
@@ -44,6 +44,7 @@ using std::unique_ptr;
 using std::unordered_map;
 using std::unordered_set;
 using std::vector;
+using std::visit;
```
mmcilree: Again, why? Is this to silence a warning or something?

gcs/constraints/smart_table.cc
```diff
@@ -40,6 +40,7 @@ using std::tuple;
 using std::unique_ptr;
 using std::unordered_map;
 using std::vector;
+using std::visit;
```
mmcilree: Same again

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -74,6 +75,20 @@ struct NamesAndIDsTracker::Imp
...
+    // entered here: a width-1 interval IS the eq atom, in eqvars_that_exist.
+    map<SimpleOrProofOnlyIntegerVariableID, map<pair<Integer, Integer>, pair<ProofFlag, pair<ProofLine, ProofLine>>>> invars_that_exist;
+
+    // The always-covered partition (spec §3) for each variable with interval
+    // literals: the sorted cell start points, always containing the definition
+    // lower bound and ub+1 as a sentinel, so the cells are the intervals between
+    // consecutive boundaries. Invariants: every cell has a literal (an eq atom if
+    // width 1, a range flag otherwise); every eq atom on a partitioned variable is
+    // a singleton cell; every requested interval is a union of adjacent cells.
+    // Absent until the variable's first interval request.
```
mmcilree: It feels like there's a creep of enormous verbose comments appearing all over the solver. Obviously it's somewhat useful when we're trying to understand what it's done, but would it be worth a sweep to trim down to only the needed comments as concise as possible before merge?

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -259,6 +274,10 @@ auto NamesAndIDsTracker::need_pol_item_defining_literal(const IntegerVariableCon
             case NotEqual:
                 need_direct_encoding_for(var, cond.value);
                 return _imp->eqvars_that_exist.at(var).at(cond.value).second;
+            case InRange:
+            case NotInRange:
+                // no pol consumer references a range literal's reification lines yet
+                throw UnimplementedException{};
```
mmcilree: But we could easily make use of this (e.g. to try replacing some of the RUP linking with pol linking instead), so may as well add it.

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -301,6 +323,9 @@ auto NamesAndIDsTracker::need_pol_item_defining_literal(const IntegerVariableCon
                 throw UnimplementedException{};
             case NotEqual:
                 throw UnimplementedException{};
+            case InRange:
+            case NotInRange:
+                throw UnimplementedException{};
```
mmcilree, line 336: Likewise

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -328,6 +353,30 @@ auto NamesAndIDsTracker::need_proof_name(const VariableConditionFrom<SimpleOrPro
     case GreaterEqual:
         need_gevar(cond.var, cond.value);
         break;
+    case InRange:
```
mmcilree, line 367: Flagging here for the function above: create_literals_for_introduced_variable_value seems to be another way that eqvars can appear. Looks like this is only used for tabulation but we should double check we aren't missing rangevar invariants there.

ciaranm:
**Overnight soak result**: 4,736,000 randomized instances over 10 hours, **zero veripb failures, zero errors** — gate-free, on the merged vocabulary (partitions + coverings + condition-based interval reasons + `In` batching + interval-reject branching), across the chain/star equality-family shapes, domains 0..7 through 0..31, all hole patterns (mid/wide/rand/multiwide), with and without NotEquals drivers, random and fixed variable order. Every instance is reproducible from its index (`tmp/range_compose/overnight_v2.sh`). For calibration: the pre-reimplementation evidence bar was 3.2M gated inference-only runs; this is 1.5× that, on the full machinery.

That completes everything in the "before review" half of the checklist above.

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -74,6 +75,20 @@ struct NamesAndIDsTracker::Imp
     map<SimpleOrProofOnlyIntegerVariableID, pair<Integer, Integer>> integer_variable_definition_bounds;
     map<SimpleOrProofOnlyIntegerVariableID, map<Integer, pair<variant<ProofLine, XLiteral>, variant<ProofLine, XLiteral>>>> gevars_that_exist;
     map<SimpleOrProofOnlyIntegerVariableID, map<Integer, pair<variant<ProofLine, XLiteral>, variant<ProofLine, XLiteral>>>> eqvars_that_exist;
+    // Range ("in") literals [lo, hi]: a proof-only flag meaning lo <= var <= hi,
+    // keyed by (lo, hi). Defined as a "wide equality literal" reified against the
+    // variable's own two order cuts; see need_invar. Width-1 intervals are never
+    // entered here: a width-1 interval IS the eq atom, in eqvars_that_exist.
+    map<SimpleOrProofOnlyIntegerVariableID, map<pair<Integer, Integer>, pair<ProofFlag, pair<ProofLine, ProofLine>>>> invars_that_exist;
```
mmcilree: Having the range literals stored as ProofFlags seems like an unnecessary layer of indirection. What's the reason for not just having them stored as XLiterals like the geqvars and eqvars?

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -751,6 +816,246 @@ auto NamesAndIDsTracker::need_gevar(SimpleOrProofOnlyIntegerVariableID id, Integ
...
+    if (bounds_it == _imp->integer_variable_definition_bounds.end())
+        throw ProofError{"need_invar: variable's bounds are not registered"};
+    auto [lb, ub] = bounds_it->second;
+    lo = max(lo, lb);
+    hi = min(hi, ub);
+    if (lo > hi)
+        throw ProofError{"need_invar: range lies entirely outside the variable's bounds"};
+
+    // A width-1 interval IS the eq atom (spec §2, witness W1): a separate flag would be
+    // an unlinked doppelganger that per-value reasoning can never unit-propagate through.
```
mmcilree: To me the fact that width-1 intervals and eqvars need to be treated as the same thing is obvious; but this keeps getting emphasised again and again. Potentially a case of Claude over-documenting something simply because it found it hard / went down the wrong path?

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -751,6 +816,246 @@ auto NamesAndIDsTracker::need_gevar(SimpleOrProofOnlyIntegerVariableID id, Integ
...
+    // axioms, so the clamped literal is the same solver fact, and clamping keeps the
+    // partition invariant (cells partition [lb, ub]) intact.
+    auto bounds_it = _imp->integer_variable_definition_bounds.find(id);
+    if (bounds_it == _imp->integer_variable_definition_bounds.end())
+        throw ProofError{"need_invar: variable's bounds are not registered"};
+    auto [lb, ub] = bounds_it->second;
+    lo = max(lo, lb);
+    hi = min(hi, ub);
+    if (lo > hi)
+        throw ProofError{"need_invar: range lies entirely outside the variable's bounds"};
```
mmcilree: Feels like it'd be better to define a trivially false literal than throw an error here?

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -751,6 +816,246 @@ auto NamesAndIDsTracker::need_gevar(SimpleOrProofOnlyIntegerVariableID id, Integ
...
+        need_direct_encoding_for(id, lo);
+        return visit([&](const auto & id) -> ProofLiteralOrFlag { return id == lo; }, id);
+    }
+
+    auto & for_this_var = _imp->invars_that_exist[id];
+    if (auto it = for_this_var.find(pair{lo, hi}); it != for_this_var.end())
+        return it->second.first;
+
+    if (! _imp->logger)
+        throw UnimplementedException{"range literals during model writing are not yet supported"};
```
mmcilree, line 1008: Why not? It feels like they'd be useful e.g. for defining In constraints more concisely. Would just need to use `emit_proof_line_now_or_at_start()` for the containment/covering links

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -751,6 +816,246 @@ auto NamesAndIDsTracker::need_gevar(SimpleOrProofOnlyIntegerVariableID id, Integ
...
+                continue;
+            auto is_maximal = none_of(existing.begin(), existing.end(), [&](const ExistingLit & m) {
+                return strictly_contains(lo, hi, m.lo, m.hi) && strictly_contains(m.lo, m.hi, c.lo, c.hi);
+            });
+            if (is_maximal)
+                emit_edge(c.term, *self_flag);
+        }
+}
+
+auto NamesAndIDsTracker::mint_plain_invar(SimpleOrProofOnlyIntegerVariableID id, Integer lo, Integer hi) -> ProofFlag
```
mmcilree: Yeah I'm really not sure about "mint"  as the name for this.

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -751,6 +816,246 @@ auto NamesAndIDsTracker::need_gevar(SimpleOrProofOnlyIntegerVariableID id, Integ
...
+        else
+            static_cast<void>(mint_plain_invar(id, cell_lo, cell_hi));
+    };
+    mint_cell(a, p - 1_i);
+    mint_cell(p, b);
+
+    // The split covering (P2, spec §3.3): the split cell is no longer a leaf, so to
+    // falsify it, unit propagation falsifies its two halves. Coverings compose through
+    // UP across refinements, so the split cell's own appearances in earlier coverings
+    // are never revisited.
```
mmcilree: Nice. This was my inuition, but we do probably want to prove this is the case.

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -751,6 +816,246 @@ auto NamesAndIDsTracker::need_gevar(SimpleOrProofOnlyIntegerVariableID id, Integ
...
+    }
+    else {
+        // An endpoint strictly inside an existing cell splits that cell (spec §3.2).
+        ensure_partition_cut(id, lo);
+        ensure_partition_cut(id, hi + 1_i);
+    }
+
+    // The request may have been minted just now, as a cell.
+    if (auto it = for_this_var.find(pair{lo, hi}); it != for_this_var.end())
+        return it->second.first;
```
mmcilree: Could `ensure_partition_cut`  not return information that would tell us whether this is the case, rather than having to do a search over all the defined range literals? At the very least we should only have to search over the partition boundaries and not the complete set of defined range literals

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -751,6 +816,246 @@ auto NamesAndIDsTracker::need_gevar(SimpleOrProofOnlyIntegerVariableID id, Integ
...
+    for (const auto & p : existing) {
+        if (! strictly_contains(p.lo, p.hi, lo, hi))
+            continue;
+        auto is_minimal = none_of(existing.begin(), existing.end(), [&](const ExistingLit & q) {
+            return strictly_contains(p.lo, p.hi, q.lo, q.hi) && strictly_contains(q.lo, q.hi, lo, hi);
+        });
+        if (is_minimal)
+            emit_edge(self_term, std::get<ProofFlag>(p.term));
+    }
+
```
mmcilree: This is an O(n^2) traversal over all defined range literals. Maybe fine but I was imagining we would be able to use a smarter data structure involving the containment partial order that would avoid us having to check every literal? `https://en.wikipedia.org/wiki/Interval_tree` ?

mmcilree (changes requested):
Okay, overall this is looking pretty good to me. It basically seems to match what I was proposing now, with a couple of differences 
1. root partition is dealt with by a top-level AL1 as opposed to a root range literal. I think Claude is probably right that this is fine, and avoids an unnecessary sentinel.
2. I was hoping we'd have a more efficient way of outputting the needed containment links. The current implementation looks to be doing a brute force search for minimal/maximal containing ranges, whereas I was hoping we could have some kind of tree structure (interval tree?) that would speed this up. Of course, it's only going to matter in practice if the number of range literals gets big, and maybe if that happens the more important concern will be deleted unused ones rather than improving the containment search. 

There's various stylistic things I don't like. I think Claude could cut down on the verbosity of comments while still clarifying things that need clarified (and I don't think it should section numbers in the dev docs). And I don't particularly like "mint" for defining literals, we've never used that term before.

Re. the questions
1. Yes let's keep it, and only remove if we can prove on paper/in Lean it's not needed.
2. This seems unnecessarily restrictive. I think we can allow defining range literals for out of bounds ranges, don't bother with an range-specific links constraints. I think we can rely on the normal geq links to make them propagate to false.
3. If it appears to be working, then I don't see why we can't merge, don't necessarily have to have a paper writeup first. But this is ultimately @ciaranm 's call. 
4. Yeah it would be good to allow them in the model as well but don't need it necessarily right now, particularly if the CakePB encoder doesn't support them yet. I'd just create an issue for it and defer to later.
5. Same as above, I think this will be what we want eventually but not necessarily right now.
6. Yeah keep these as tests in case a future Claude decides once again that we don't need the links.

ciaranm:
Regarding the verbose comments: yes, but I think that's one for Grumpy Claude to work on rather than as part of this.

dev_docs/range_literals_spec.md
ciaranm: Reworded: §1 no longer bills this as a correctness property, states the point in your terms (the root cause of a proof failure isn't necessarily the line VeriPB flags), and defines "UP" once as unit propagation from the empty assignment.

dev_docs/range_literals_spec.md
ciaranm: Renamed throughout, code and spec — "define".

dev_docs/range_literals_spec.md
ciaranm: Now defined once in §1 (plain unit propagation from the empty assignment, no case-splitting — what a RUP check runs after asserting the negation) and referred to as plain "UP" everywhere.

gcs/constraints/circuit/circuit_scc.cc
ciaranm: Stale leftover from a reverted edit — the loop that used `visit` was rewritten and I missed the using-declaration. Removed, and same for the other two you found.

gcs/constraints/regular/regular.cc
ciaranm: Same — stale from a reverted edit; removed.

gcs/constraints/smart_table.cc
ciaranm: Same — stale from a reverted edit; removed.

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: Done — a sweep over everything this PR touches; the partition machinery, equals, In, reason and justify comments are all substantially shorter now. Shout if any specific one still reads too long.

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: Implemented: range conditions now return their reification lines here, and with the xliteral change below those lines are tracked exactly like the eq/ge def lines, so pol linking experiments can use them.

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: The simple-variable branch is implemented; the registered-view branch deliberately still throws, because range conditions on views are blocked at the literal layer until the view story is done (#311).

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: Checked: these literals do bypass eqvars_that_exist and the containment structures, but an introduced variable never gets a bits encoding, and need_invar refuses variables without bits, so no range literal can ever exist over one and there is no invariant to violate. Now documented at the function.

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: Done — range literals are now xliterals registered under their InRange/NotInRange condition pair, exactly like the eq and order atoms; the ProofFlag layer is gone (proof names are i[x][in5_10] rather than f[n][in_x_5_10], and the varmap entry carries the operator and both values).

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: Guilty as charged. Trimmed to one statement for the public helpers plus one line where the storage map is defined.

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: Done — and with your question-2 answer adopted, need_invar no longer clamps at all: out-of-bounds ranges are defined over their own cuts and the bound axioms falsify the outside part (an entirely-outside literal needs no covering; a partially-outside one keeps a covering over its in-bounds cells, which is what later unit propagation needs when the exclusion arrives piecewise). Empty ranges return a trivially false literal.

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: Agreed it's mostly emit_proof_line_now_or_at_start away, but the partition setup also consults the eq atoms that exist at definition time, and model-phase literals change the OPB surface the CakePB conformance work checks against — so deferred to #310 per your answer to question 4.

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: Renamed: define_plain_invar.

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: Agreed — that's the L1/L2 write-up, which per your answer to question 3 (and Ciaran's call) is queued as a follow-up rather than a merge gate.

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: It was a log-time map lookup rather than a scan over the literals, but the point stands: it now reads the partition boundary set directly (one find plus next), no literal-map probe. The broader neighbour search is the interval tree commit.

ciaranm:
All review feedback addressed, in four commits (`90e1474..979d0eb`):

- **Range literals are xliterals** (no ProofFlag indirection), registered under their InRange/NotInRange condition pair like the eq and order atoms, and `need_pol_item_defining_literal` returns their reification lines.
- **No more clamping** (your Q2, Ciaran adopted): literals are defined over their own cuts even out of bounds, with the bound axioms falsifying the outside part; partially-outside literals keep a covering over their in-bounds cells (that part is load-bearing for later UP when an exclusion arrives piecewise); entirely-outside literals need no link constraints at all. Empty ranges give a trivially false literal. invar_test grows out-of-bounds coverage.
- **Interval tree** for containment neighbours (treap augmented with subtree min/max ends, fixed-seed so emission stays deterministic), with a randomised-against-brute-force Catch2 test; the minimal/maximal antichains fall out of one sweep over the query results.
- **Style sweep**: comments trimmed across everything the PR touches, "mint" → "define", spec section numbers out of code comments, §1 reworded per your comment, "forward-UP" replaced by a one-time definition of UP, the width-1 point stated once, and the three stale `using std::visit` removed (leftovers from a reverted edit — good catch).

On the body points: the root-AL1-rather-than-root-literal stays as endorsed; your Q1 answer (keep the root covering pending a paper/Lean argument) is recorded in the spec; L1/L2 is queued post-merge per Q3 and Ciaran's call; Q4 and Q5 are filed as #310 and #311, and literal deletion (your body point 2's second half) as #312.

Revalidation: full suite 330/330 caps on and off, witness suite green. Since dropping the clamping is a real encoding change, the ≥1M-instance overnight soak is being rerun against this head as well; result will be posted here.

gcs/innards/proofs/names_and_ids_tracker.cc
ciaranm: (Human) Deferring this one because it mucks around with CakePB, and getting a working formally verified encoding setup needs to be high up the priority list.

mmcilree: Hah is it worth setting a guideline for Claude that it has to prepend all it's comments with [from Claude] or something? Rather than you having to specify your human-ness

ciaranm:
**Revalidation soak after the round-2 changes**: 5,676,000 randomized instances over 12 hours on the post-review head (`b536afc`), **zero veripb failures, zero errors**. This was the change-protocol rerun for dropping the bounds clamping (the round-2 change that touched the encoding) — it confirms out-of-bounds literals defined over their own cuts, with the bound axioms falsifying the outside part, hold up across the same chain/star equality-family shapes, domains 0..7 through 0..31, all hole patterns, with and without drivers, random and fixed orderings. Combined with the pre-review 4.7M-run soak, that's ~10.4M instances clean across the two vocabulary designs.

CI is green across the matrix on this head. From my side everything in the review is addressed; over to you for re-review whenever convenient.

Issue #315, opened 2026-06-12
Search-heuristics RFC: dom/wdeg variable ordering with dynamic constraint weighting (+ restarts & nogood recording)


ciaranm:
This is a design RFC for adding **dom/wdeg** as a *properly dynamic*, weighted variable-ordering
heuristic (with dynamic weights, not a static degree proxy). It is also the planning home for the
interlocking **restarts** and **nogood-recording** work — see *Companion work* at the end — because
they share two seams with the weighting design: weights persist across restarts (CHS smoothing), and
a future share-nothing parallel search periodically syncs *both* nogoods and weights through the same
extract/merge/inject machinery. dom/wdeg is implementable and useful on its own; restarts/nogoods are
scoped here for discussion, not yet designed.

## Why this is more than a one-file brancher

1. **The "what actually failed" signal does not exist today.** A domain wipeout is thrown as a
   payload-less `TrackedPropagationFailed{}` and the failing propagator id is discarded
   (`propagators.cc:235`). Weighting needs to know *which constraint* failed, and ideally *which part*
   of it (the `Reason` attached to the contradiction).
2. **Forward-looking requirements shape the data model** (must not be precluded):
   - **Reason-based attribution** — every inference carries a `Reason` that can identify "which bit" of
     a global constraint caused the failure; we want to experiment with weighting only the implicated
     variables.
   - **Share-nothing parallel search** (Archibald, Dunlop, Hoffmann, McCreesh, Prosser, Trimble,
     *Sequential and Parallel Solution-Biased Search…*, CPAIOR 2019): each thread owns an independent
     `State`+`Propagators`; the only communication is *periodic, lock-free* syncing of restart-nogoods
     and (maybe) weights. So we must get weights out of a thread and put new weights in cheaply.
   - **Proof-mined weights** — later, seed a fresh search with weighting data parsed offline from a
     proof log, to test whether that is a better heuristic.

Unifying consequence: weighting state must be a **standalone, copyable value object keyed by a stable
identity a proof log can also name** — i.e. by **`ConstraintID`** — never mutable state inside the
(potentially shared) constraint model. Snapshot / merge / inject are first-class.

## Literature survey (the dom/wdeg variant landscape)

No single "best" heuristic — claims otherwise don't survive per-family analysis. The space factors into
**three orthogonal axes**, and *every* variant reduces to **"branch on the variable minimising
[...]

PR #318, opened 2026-06-12
Stage 0 (cont.): conflict-observation seam for dom/wdeg (#315)

ciaranm:
## What

Completes **Stage 0** of the dom/wdeg search-heuristics work (issue #315) — the engine
seam shared by both the weighting and the (later) nogood/restart work. PR #317 landed the
first checkbox (an explicit `ConstraintID` on `Propagators::install`); this adds the rest
of the conflict-observation half. Nothing user-visible yet: with no observer attached,
behaviour and proofs are unchanged.

> Stacked on #317 (`stable-constraint-propagator-id`). The base will retarget to `main`
> once #317 merges; if `main` is rebased in the process, this branch wants a quick rebase
> so the diff shows only its four commits.

## Changes (four self-contained commits)

1. **Propagator scope + variable→constraint adjacency.** `Propagators` records, per
   installed propagator, its scope — the trigger variables across all trigger kinds, with
   views resolved to the underlying simple variable and duplicates removed — and the
   transpose aggregated by constraint (per variable, the deduplicated dense indices of the
   constraints it is in). Exposed via `scope_of_propagator()` and
   `constraint_indices_of_variable()`.
2. **Stash the contradiction's reason.** Before throwing `TrackedPropagationFailed`, the
   inference tracker records the `ReasonFunction`, in all three paths (the explicit
   `contradiction()` and the `Contradiction` case of both `track_impl` overloads),
   independent of the logger. Reachable at the catch site via `last_contradiction_reason()`;
   cleared by `reset()`. Cost is one `std::function` copy per failed propagate; the reason
   is evaluated only if a consumer asks.
3. **ConflictObserver + notification.** A minimal `ConflictObserver` interface
   (`gcs/innards/conflict_observer.hh`) with one `note_conflict(constraint index, scope,
   reason, state)` hook. `Propagators` carries a borrowed `ConflictObserver*`
   (`set_conflict_observer()`). When a propagator wipes out a domain,
   `Propagators::propagate` notifies the observer exactly once. Non-propagator contradiction
   paths (initialisers, the objective bound) don't reach the notification, so they are not
   attributed to any constraint. The observer lives on `Propagators` (not `State`): a
   conflict is a `Propagators` event and the index/scope/reason it carries all originate
   here, so the whole conflict-observation seam stays in one place and `State` stays about
   domains.
4. **Crafted-wipeout test** (`gcs/innards/conflict_observer_test.cc`): a recording observer
   confirms a wipeout is blamed on the failing constraint's `ConstraintID` and scope, never
   on an innocent bystander; that both contradiction paths fire and carry the reason
   (engaged, empty vs non-empty as supplied); that propagation with no observer still
   detects the contradiction; and that the scope/adjacency accessors resolve views,
   deduplicate, and give the right per-variable adjacency.

## Scope / forward-looking

These accessors and the observer have **no consumer yet** — they are the scaffolding the
Stage 1 weighting object will build on (`VariableWeighting` will implement
`ConflictObserver`). Deliberately left for Stage 1: the value object, the schemes, the
`dom_wdeg` brancher, and the `solve_with` wiring. One open question recorded for Stage 1 —
the brancher's read path (a `CurrentState` view vs. passing the weighting to the brancher
directly); the write side added here precludes neither.

## No proof impact

No constraint identity or proof model changes, so OPB/proof output is byte-identical. Full
`ctest` passes (including the VeriPB proof-verifying tests). Built under **both GCC 15.2 and
clang 21**.

### DIFF
--- a/gcs/CMakeLists.txt
+++ b/gcs/CMakeLists.txt
@@ -430,6 +430,10 @@ if(GCS_BUILD_TESTS)
     target_link_libraries(state_test PRIVATE glasgow_constraint_solver Catch2::Catch2WithMain)
     add_test(NAME state_test COMMAND $<TARGET_FILE:state_test>)
 
+    add_executable(conflict_observer_test innards/conflict_observer_test.cc)
+    target_link_libraries(conflict_observer_test PRIVATE glasgow_constraint_solver Catch2::Catch2WithMain)
+    add_test(NAME conflict_observer_test COMMAND $<TARGET_FILE:conflict_observer_test>)
+
     add_executable(linear_utils_test constraints/linear/utils_test.cc)
     target_link_libraries(linear_utils_test PRIVATE glasgow_constraint_solver Catch2::Catch2WithMain)
     add_test(NAME linear_utils_test COMMAND $<TARGET_FILE:linear_utils_test>)
--- a/gcs/innards/conflict_observer.hh
+++ b/gcs/innards/conflict_observer.hh
@@ -1,0 +1,55 @@
+#ifndef GLASGOW_CONSTRAINT_SOLVER_GUARD_GCS_INNARDS_CONFLICT_OBSERVER_HH
+#define GLASGOW_CONSTRAINT_SOLVER_GUARD_GCS_INNARDS_CONFLICT_OBSERVER_HH
+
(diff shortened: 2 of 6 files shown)


ciaranm:

PR #321, opened 2026-06-13
Stage 1 foundation: dom/wdeg weighting value object + classic scheme (#315)

ciaranm:
## What

The **Stage 1 foundation** for dom/wdeg (issue #315): the weighting value object, the live
weighting interface, and the classic scheme — unit-tested in isolation. **No brancher and no
`solve_with` wiring yet** (agreed scope); the new types have no consumer, so nothing about
search or proofs changes.

> Stacked on #318 (`conflict-observation-seam`), which is itself stacked on #317. Bases will
> auto-retarget as each lands; rebase onto `main` if `main` is rebased in the process.

## Changes (three commits)

1. **`Propagators::scope_of_constraint` + a shared `std::hash<ConstraintID>`.**
   - `scope_of_constraint(constraint_index)` returns the union of a constraint's propagators'
     scopes (views resolved, deduped), built in `install()` next to the Stage-0 adjacency. The
     weighted-degree filter needs it to count a constraint's unassigned variables.
   - `std::hash<gcs::ConstraintID>` (hashing `as_string`) moves into `constraint.hh`, so
     `ConstraintID` can key an `unordered_map` without a bespoke functor; `Propagators` drops
     its local `ConstraintIDHash` and uses it.
2. **`gcs/variable_weighting.{hh,cc}` — the value object, interface, and classic scheme.**
   - `WeightingState` — a portable, copyable `ConstraintID → weight` value object (the seam for
     share-nothing parallel sync and proof-mined seeding): `weight_of` / `set_weight`, and
     `merge(other, policy)` over `Sum` / `Max` / `Average`.
   - `VariableWeighting` — the live per-search interface. It **is** an `innards::ConflictObserver`
     (the Stage-0 write side, `note_conflict` on a wipeout) plus `weighted_degree_of` (the read
     side the brancher will sum), `on_restart` (inert until restarts), and `snapshot()` / `load()`
     bridging dense-by-constraint-index storage to the `ConstraintID`-keyed `WeightingState`.
   - `ClassicDomWDeg` — one weight per constraint, **initialised to 1** and bumped by 1 on each
     wipeout. `weighted_degree_of(x)` sums `w(c)` over `x`'s constraints with at least two
     unassigned variables (the `|fut|>1` filter), so at the root it equals the variable's degree
     and the heuristic starts out like `dom_then_deg`.
3. **`gcs/variable_weighting_test.cc`** — degree-at-root, per-constraint increment + accumulation,
   the `|fut|>1` filter, `snapshot → load` round-trip, and the three `merge` policies.

## Decisions baked in (from the design pass on #315)

- First PR is **foundation only** — brancher and config surface come later.
- `WeightingState` is **global-per-constraint only** for now; the per-`(constraint, variable)`
  AbsCon-local map is a later additive extension (chosen to keep the serialisation format minimal).
- Initial weight **1** (so initial weighted degree = static degree — same thing, per-constraint).
- `weighted_degree_of` sums **distinct constraints** via the adjacency (filtered to `|fut|>1`),
  not `degree_of` (which counts trigger occurrences); `degree_of` will be the brancher's tie-break.

## Deferred / notes

- The `dom_wdeg` brancher and the `solve_with` config surface (how you select the scheme, inject an
  initial `WeightingState`) are the next, more user-facing step — held for review.
- `variable_weighting.hh` is intentionally **not** yet in `gcs.hh`; it gains that once there's a
  wired consumer.
- When `ca.cd` / CHS land (Stage 2), the dense-storage + `weighted_degree_of` + `snapshot`/`load`
  common to all global schemes can be factored into a shared base; left concrete for now with one
  scheme in hand.

## Verification

Built under **GCC 15.2 and clang 21**; the new unit tests and the full `ctest` suite pass; **zero
proof impact** (no constraint identity or proof model changes, no consumer wired).

### DIFF
--- a/gcs/CMakeLists.txt
+++ b/gcs/CMakeLists.txt
@@ -435,6 +435,10 @@ if(GCS_BUILD_TESTS)
     target_link_libraries(conflict_observer_test PRIVATE glasgow_constraint_solver Catch2::Catch2WithMain)
     add_test(NAME conflict_observer_test COMMAND $<TARGET_FILE:conflict_observer_test>)
 
+    add_executable(variable_weighting_test variable_weighting_test.cc)
+    target_link_libraries(variable_weighting_test PRIVATE glasgow_constraint_solver Catch2::Catch2WithMain)
+    add_test(NAME variable_weighting_test COMMAND $<TARGET_FILE:variable_weighting_test>)
+
     add_executable(linear_utils_test constraints/linear/utils_test.cc)
     target_link_libraries(linear_utils_test PRIVATE glasgow_constraint_solver Catch2::Catch2WithMain)
     add_test(NAME linear_utils_test COMMAND $<TARGET_FILE:linear_utils_test>)
--- a/gcs/CMakeLists.txt
+++ b/gcs/CMakeLists.txt
@@ -11,6 +11,7 @@ add_library(glasgow_constraint_solver
         stats.cc
         variable_condition.cc
         variable_id.cc
+        variable_weighting.cc
         constraints/abs.cc
         constraints/abs/justify.cc
         constraints/all_different/all_different_except.cc
--- a/gcs/innards/propagators.cc
+++ b/gcs/innards/propagators.cc
@@ -31,7 +31,7 @@ using std::swap;
 using std::to_underlying;
 using std::vector;
 using std::visit;
-using std::ranges::find;
+using std::ranges::contains;
 
 namespace
 {
@@ -135,14 +135,15 @@ auto Propagators::install(const ConstraintID & constraint_id, PropagationFunctio
     // Record this propagator's scope (its trigger variables, views resolved and
     // deduplicated) and add it to each variable's constraint adjacency.
     auto & scope = _imp->propagator_scope.emplace_back();
+    scope.reserve(triggers.on_change.size() + triggers.on_bounds.size() + triggers.on_instantiated.size());
     auto add_to_scope = [&](IntegerVariableID var) {
         overloaded{
             [&](const SimpleIntegerVariableID & v) {
-                if (find(scope, v) == scope.end())
+                if (! contains(scope, v))
                     scope.push_back(v);
             },
             [&](const ViewOfIntegerVariableID & v) {
-                if (find(scope, v.actual_variable) == scope.end())
+                if (! contains(scope, v.actual_variable))
                     scope.push_back(v.actual_variable);
             },
             [&](const ConstantIntegerVariableID &) {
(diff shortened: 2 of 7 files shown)


gcs/innards/propagators.cc
```diff
@@ -173,6 +164,13 @@ auto Propagators::install(const ConstraintID & constraint_id, PropagationFunctio
             indices.push_back(constraint_index);
     }
 
+    if (_imp->constraint_scope.size() <= static_cast<std::size_t>(constraint_index))
```
ciaranm, line 168: Isn't there a `std::less_than` or something like that now for this kind of thing? Also possibly an "ensure capacity" in some containers?

gcs/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ add_library(glasgow_constraint_solver
         stats.cc
         variable_condition.cc
         variable_id.cc
+        variable_weighting.cc
```
ciaranm, line 14: Is it worth making a search_heuristics directory and splitting things up yet?

gcs/innards/propagators.cc
ciaranm: Good catch — `std::ranges::contains` (C++23, already used in among.cc) is the direct replacement for the `find(…) == end()` membership checks here, and builds under both GCC 15 and clang 21. I'll switch both dedup loops to `if (! contains(…)) push_back(…)`.

On ensure-capacity: most of these grow once at install time and amortise (the `resize(index+1)` / `push_back` patterns), and the final counts aren't known up front since `install()` runs per-constraint — so I'd lean to leaving the dense vectors as-is. The one obviously-knowable reserve is the per-propagator `scope` vector (`on_change + on_bounds + on_instantiated` sizes), which I'll `reserve()`. Happy to go further if you'd like.

gcs/CMakeLists.txt
ciaranm: I think it's tipping into worth-it: there's already `search_heuristics.{hh,cc}`, `variable_weighting.{hh,cc}` (plus `innards/conflict_observer.hh`), and Stage 2 adds the schemes / Stage 3 will add restart machinery, so it'll keep growing. I'd propose a `gcs/search_heuristics/` directory for the weighting + scheme files (`variable_weighting.{hh,cc}` and friends), keeping `search_heuristics.{hh,cc}` as the brancher-API entry point — mirroring how `gcs/constraints/` is laid out.

The one catch is that it moves *public* headers, so the `#include <gcs/variable_weighting.hh>` paths change for every includer (and it ripples across the stacked PRs). Given that, I'd do it as one deliberate move — happy to fold it into the rebase pass when addressing the review, or keep it flat for now if you'd rather not churn include paths yet. Your call.

gcs/innards/propagators.cc
ciaranm: Done — all four dedup checks now use `std::ranges::contains`, and the per-propagator scope vector is `reserve()`d to its trigger count. (Left the dense `Imp` vectors growing/amortising as-is, since their final sizes aren't known at install time.)

gcs/CMakeLists.txt
ciaranm: Holding this one for your explicit call, since it moves *public* headers and ripples include paths across the whole stack — didn't want to churn that as part of the review pass without a clear yes. Happy to do the `gcs/search_heuristics/` move (or leave it flat) whenever you decide.

ciaranm:

PR #322, opened 2026-06-13
dom/wdeg brancher: variable_order::dom_wdeg (#315)

ciaranm:
## What

The **dom/wdeg brancher** for issue #315: `variable_order::dom_wdeg`, built on the Stage 1
weighting foundation (#321).

> Stacked on #321 → #318 → `main`. Bases auto-retarget as each lands.

## Changes (two commits)

1. **`variable_order::dom_wdeg` + a new `BranchVariableHeuristic` type.**
   - `BranchVariableHeuristic = function<BranchVariableSelector(const Problem&, State&, Propagators&)>`
     — the "do per-search setup, then return the selector" shape a *stateful* heuristic needs.
     Called once per search; in a parallel search, once per thread with that thread's own
     `State`/`Propagators`, each owning its own weighting.
   - `dom_wdeg(...)`'s setup constructs a `ClassicDomWDeg` (optionally seeded from an initial
     `WeightingState`), attaches it as the `Propagators`' conflict observer, and returns a
     selector that branches on the variable minimising `dom(x)/W(x)` — cross-multiplied
     (`dom(a)·W(b) < dom(b)·W(a)`) so there's no division and a `W=0` variable sorts last for
     free, with a highest-degree tie-break matching `dom_then_deg`. Weights start uniform, so
     at the root it orders by `dom(x)/degree(x)` and specialises as conflicts accumulate.
2. **Test** (`gcs/search_heuristics_test.cc`) — drives the selector directly: root ordering by
   `dom/W` with a zero-weight variable last; a seeded `WeightingState` shifting the choice to
   the now-heaviest variable; and the degree tie-break.

## Deliberately not done here

- The other `variable_order::*` orderings still return a bare `BranchVariableSelector`. The
  per-node branch hook (`BranchCallback` / `SolveCallbacks::branch`) is typed
  `(const CurrentState&, const Propagators&)` — **const** `Propagators`, no `Problem` — so it
  can't host a stateful heuristic's setup; migrating cleanly means lifting setup into
  `solve_with` and changing `SolveCallbacks` + `branch_with` + the example call sites. That
  migration (and the build-once `solve_with` restructure, and the public way to *select*
  `dom_wdeg`) is a deliberate later step, once the interface has settled — kept out of this PR
  to avoid churning the public API and the ~7 examples while it might still move. So `dom_wdeg`
  is **not yet wired into `solve_with`'s defaults**; it is exercised directly for now.

## Verification

Built under **GCC 15.2 and clang 21**; the new test and the full `ctest` suite pass. No
behaviour or proof change (no consumer wired).

### DIFF
--- a/gcs/CMakeLists.txt
+++ b/gcs/CMakeLists.txt
@@ -439,6 +439,10 @@ if(GCS_BUILD_TESTS)
     target_link_libraries(variable_weighting_test PRIVATE glasgow_constraint_solver Catch2::Catch2WithMain)
     add_test(NAME variable_weighting_test COMMAND $<TARGET_FILE:variable_weighting_test>)
 
+    add_executable(search_heuristics_test search_heuristics_test.cc)
+    target_link_libraries(search_heuristics_test PRIVATE glasgow_constraint_solver Catch2::Catch2WithMain)
+    add_test(NAME search_heuristics_test COMMAND $<TARGET_FILE:search_heuristics_test>)
+
     add_executable(linear_utils_test constraints/linear/utils_test.cc)
     target_link_libraries(linear_utils_test PRIVATE glasgow_constraint_solver Catch2::Catch2WithMain)
     add_test(NAME linear_utils_test COMMAND $<TARGET_FILE:linear_utils_test>)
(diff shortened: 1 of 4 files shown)


ciaranm:

PR #323, opened 2026-06-13
Migrate variable orderings to setup-then-select; wire dom_wdeg into solve (#315)

ciaranm:
## What

The interface migration that makes **dom/wdeg a first-class, selectable variable ordering** —
and, with it, the `solve_with` build-once restructure. This is the deferred wiring step from the
dom/wdeg brancher PR (#322); after this, `dom_wdeg` actually runs in a solve.

> Stacked on #322 → #321 → #318 → `main`. @mmcilree — this touches the solve path (`SolveCallbacks`,
> `branch_with`, `solve.cc`) that #302 also touches; happy to coordinate the rebase order.

## Changes (two commits)

1. **Migrate variable-ordering heuristics to the setup→select shape.**
   - Every `gcs::variable_order::*` ordering now returns a `BranchVariableHeuristic`
     (`function<BranchVariableSelector(const Problem&, State&, Propagators&)>`) instead of a bare
     `BranchVariableSelector`. A stateful ordering (dom_wdeg) does its per-search setup here; a
     stateless one is wrapped via a small `as_heuristic()` helper.
   - `branch_with` returns a `BranchHeuristic` (`function<BranchCallback(const Problem&, State&,
     Propagators&)>`), `branch_sequence` chains `BranchHeuristic`s, and `SolveCallbacks::branch` is
     a `BranchHeuristic`.
   - **`solve_with` runs the heuristic's setup exactly once**, after propagators (and any
     presolver-added constraints) are final, and reuses the resulting per-node `BranchCallback` at
     every node. Previously the brancher was rebuilt every node — which would have reconstructed a
     stateful weighting (losing its accumulated weights) on each recursion. The autotabulation
     presolver gets the same once-then-reuse treatment, and the MiniZinc frontend's
     search-annotation branchers move to the new types.
2. **End-to-end test** — an all-different triangle solved under `dom_wdeg` enumerates all six
   permutations, exercising selection via `callbacks.branch`, the once-per-search setup, and the
   conflict observer driving the weights.

## Behaviour-preserving

Stateless orderings wrap their existing selector, so the default search and every example/test
that builds a brancher with `branch_with(variable_order::X, value_order::Y)` are unchanged — those
call sites compile as-is (the input and output types moved together). Full `ctest` is unchanged at
329/329.

`dom_wdeg` is now selectable like any other ordering:
```cpp
solve_with(problem, SolveCallbacks{
    .branch = branch_with(variable_order::dom_wdeg(problem), value_order::smallest_first())});
```

## Deliberately not here

- The MiniZinc `dom_w_deg` search annotation still maps to `dom_then_deg` (as before): wiring it to
  real `dom_wdeg` raises a multi-observer question (several stateful branchers in a `seq_search`
  would share one `Propagators` observer slot), which is its own follow-up.
- Per-thread RNG independence for `variable_order::random` under the new per-thread-call model is
  unchanged (still a shared RNG, as before) — a separate, minor concern.

## Verification

Built under **GCC 15.2 and clang 21** (all targets — library, examples, MiniZinc frontend, tests);
`ctest` 329/329; the new end-to-end test passes.

### DIFF
--- a/gcs/innards/proofs/range_witness_w1_test.cc
+++ b/gcs/innards/proofs/range_witness_w1_test.cc
@@ -22,15 +22,17 @@ namespace
     // Scripted first decision: where `var` is undecided and still has value 0,
     // branch var != 0 then var == 0 (only the root qualifies here); elsewhere fall
     // through to the default heuristic via branch_sequence.
-    auto scripted_neq_then_eq_zero(IntegerVariableID var) -> BranchCallback
+    auto scripted_neq_then_eq_zero(IntegerVariableID var) -> BranchHeuristic
     {
-        return [var](const CurrentState & state, const innards::Propagators &) -> std::generator<IntegerVariableCondition> {
-            return [](const CurrentState & state, IntegerVariableID var) -> std::generator<IntegerVariableCondition> {
-                if (state.domain_size(var) >= 2_i && state.in_domain(var, 0_i)) {
-                    co_yield var != 0_i;
-                    co_yield var == 0_i;
-                }
-            }(state, var);
+        return [var](const Problem &, innards::State &, innards::Propagators &) -> BranchCallback {
+            return [var](const CurrentState & state, const innards::Propagators &) -> std::generator<IntegerVariableCondition> {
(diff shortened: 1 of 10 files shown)


ciaranm:

PR #324, opened 2026-06-13
Stage 2 (part 1): Conflict-History Search + scheme selector (#315)

ciaranm:
## What

**Stage 2, part 1** of dom/wdeg (issue #315): the **Conflict-History Search (CHS)** scheme and a
**`WeightingScheme` selector** on `dom_wdeg`, behind a shared dense-weighting base. CHS is one of
the RFC's "two to beat".

> Stacked on #323 → #322 → #321 → #318 → `main`.

## Changes (two commits)

1. **`DenseConstraintWeighting` base + `ConflictHistorySearch`.**
   - Factor the dense per-constraint storage common to global schemes into a
     `DenseConstraintWeighting` base — the weight vector, the `|fut|>1` `weighted_degree_of` loop,
     and `snapshot`/`load` — with a virtual `contribution_of` hook for how a stored weight maps to a
     variable's weighted degree. `ClassicDomWDeg` becomes a thin derived class (init 1, `+1` on
     conflict).
   - `ConflictHistorySearch` (Habet & Terrioux): a recency-weighted per-constraint score `q(c)`,
     updated by an exponential recency-weighted average on each conflict (`alpha` from 0.4 decaying
     to 0.06), contributing `q(c)+delta` to weighted degree. `on_restart` smooths the scores and
     resets `alpha` — **wired but inert** until restarts land. The CHS tunables are isolated
     constants, to be auto-tuned rather than chosen by design.
   - Unit tests: the ERWA update, recency ordering, snapshot/load.
2. **Scheme selector.** `variable_order::dom_wdeg` gains a `WeightingScheme` parameter (default
   `Classic`) that picks the live weighting at setup. So the scheme is a knob behind a default,
   not hard-coded. The end-to-end test runs under both schemes and still enumerates every solution.

```cpp
branch_with(variable_order::dom_wdeg(problem, WeightingScheme::Chs), value_order::smallest_first())
```

## Behaviour / scope

`ClassicDomWDeg` is unchanged behaviourally (the refactor is mechanical; its tests are untouched and
pass). `dom_wdeg`'s default stays `Classic`. No proof impact.

**Next (Stage 2, part 2):** `ca.cd` (`RefinedWeighting`), which is *local* (per-`(constraint,
variable)`) and needs the `(ConstraintID, var)` weight extension deferred from Stage 1; it becomes
`dom_wdeg`'s default once it lands. The RefinedWeighting `ia/ca/id/cd` ablation variants and the
benchmark `--branch dom-wdeg:variant` hook follow.

## Verification

Built under **GCC 15.2 and clang 21**; the new unit tests and full `ctest` pass.

### DIFF
--- a/gcs/variable_weighting.hh
+++ b/gcs/variable_weighting.hh
@@ -227,8 +227,8 @@ namespace gcs
      */
     enum class WeightingScheme
     {
-        Classic, ///< ClassicDomWDeg
-        Chs      ///< ConflictHistorySearch
+        Classic,              ///< ClassicDomWDeg
+        ConflictHistorySearch ///< ConflictHistorySearch (recency-weighted)
     };
 }
 
--- a/gcs/variable_weighting.hh
+++ b/gcs/variable_weighting.hh
@@ -123,40 +123,112 @@ namespace gcs
     };
 
     /**
+     * \brief Common storage for weighting schemes that keep one weight per
+     * constraint, indexed densely by constraint index.
+     *
+     * Provides the dense vector and the read / snapshot / load machinery;
+     * concrete schemes supply note_conflict (the update) and on_restart, and may
+     * override contribution_of to map a stored weight to a variable's weighted
+     * degree (the default is the weight itself).
+     *
+     * \ingroup SearchHeuristics
+     */
+    class DenseConstraintWeighting : public VariableWeighting
+    {
+    public:
+        [[nodiscard]] auto weighted_degree_of(const CurrentState & state,
+            const innards::Propagators & propagators, IntegerVariableID var) const -> double override;
+
+        [[nodiscard]] auto snapshot(const innards::Propagators & propagators) const -> WeightingState override;
+
+        auto load(const WeightingState & state, const innards::Propagators & propagators) -> void override;
+
+    protected:
+        /**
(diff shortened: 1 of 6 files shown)


gcs/variable_weighting.hh
```diff
@@ -122,41 +122,113 @@ namespace gcs
...
+
+    /**
+     * \brief Selects which weighting scheme gcs::variable_order::dom_wdeg uses.
+     *
+     * \ingroup SearchHeuristics
+     */
+    enum class WeightingScheme
+    {
+        Classic, ///< ClassicDomWDeg
+        Chs      ///< ConflictHistorySearch
```
ciaranm: I think I have a mild preference for `ConflictHistorySearch` over `Chs`: I don't think the acronym is so widely known, and the capitalisation is weird otherwise.

ciaranm: Done — renamed to `ConflictHistorySearch`. One wrinkle: the enumerator then shadows the same-named class in the dispatch's `make_shared` under `using enum WeightingScheme`, so that one call is qualified `gcs::ConflictHistorySearch` (with a comment).

ciaranm:

PR #325, opened 2026-06-13
Stage 2 (part 2): ca.cd (RefinedWeighting) with local weights (#315)

ciaranm:
## What

**Stage 2, part 2** of dom/wdeg (issue #315): the **`ca.cd` scheme (`RefinedWeighting`)** — the
RFC's default-to-be and the other of the "two to beat" — built on a per-`(constraint, variable)`
**local weight** extension to `WeightingState`.

> Stacked on #324 → #323 → #322 → #321 → #318 → `main`.

## Changes (two commits)

1. **`WeightingState` local weights + `RefinedWeighting`.**
   - `WeightingState` gains a per-`(constraint, variable)` local weight map alongside the global
     one (`merge` and the accessors handle both) — the portable form of the local weights a scheme
     like `ca.cd` needs.
   - `RefinedWeighting` (Wattez, Lecoutre, Paparrizou, Tabary, ICTAI 2019): a local weight
     `w(c)[x]` per constraint variable, initialised to 1, incremented on a conflict over the
     still-unassigned variables of the failing constraint by a `Variant`-chosen amount —
     `ia` (`1/|scp|`), `ca` (`1/|fut|`), `id` (`1/dom_init`), `cd` (`1/(1+dom)`), and the default
     **`ca.cd` (`1/(|fut|*(1+dom))`)**. `weighted_degree_of` sums `w(c)[x]` over `x`'s constraints
     with `|fut|>1`, so at the root (all weights 1) it equals the degree — same starting point as
     the global schemes. Initial domain sizes are captured at construction for the `id` variant.
   - Unit tests: the `ca.cd` increment, a variant difference, and the local snapshot/load round-trip.
2. **Scheme selection.** `WeightingScheme` gains `Ia`/`Ca`/`Id`/`Cd`/`CaCd`, and **`CaCd` becomes the
   `dom_wdeg` default** (the RFC's intended default). `dom_wdeg` dispatches each scheme at setup. The
   end-to-end test runs under `Classic`, `CaCd`, and `Chs` and still enumerates every solution.

## Notes

- **Default change:** `dom_wdeg(problem)` now uses `ca.cd` rather than classic. Since at the root
  every local weight is 1, the root ordering is still degree-based (identical to classic at the
  root), so the existing selector tests are unaffected; the schemes diverge only as conflicts
  accumulate. `dom_wdeg` remains opt-in (the default brancher is still `dom_then_deg`).
- `note_conflict`'s `|fut|`/arity for the increment use the *failing propagator's* scope (what the
  observer is handed); `weighted_degree_of`'s `|fut|>1` filter uses the full constraint scope. They
  coincide for single-propagator constraints.

## Verification

Built under **GCC 15.2 and clang 21**; the new unit tests and full `ctest` pass.

**Next (Stage 2, part 3):** the example/benchmark `--branch dom-wdeg:variant` hook with `--stats`
search-shape columns, so the schemes (incl. the `ia/ca/id/cd` ablations) are benchmarkable.

### DIFF
--- a/gcs/variable_weighting.hh
+++ b/gcs/variable_weighting.hh
@@ -6,8 +6,10 @@
 #include <gcs/innards/propagators-fwd.hh>
 #include <gcs/variable_id.hh>
 
+#include <map>
 #include <optional>
 #include <unordered_map>
+#include <utility>
 #include <vector>
 
 namespace gcs
@@ -25,8 +27,10 @@ namespace gcs
      * search/thread and entering another: the basis for share-nothing parallel
      * sync and for seeding a search with proof-mined weights.
      *
-     * v1 holds only global per-constraint weights; a per-(constraint, variable)
-     * map for AbsCon-local schemes is a later additive extension.
+     * It holds a global per-constraint weight map (for whole-constraint schemes)
+     * and a per-(constraint, variable) local weight map (for schemes that weight
+     * variables within a constraint, such as ca.cd); a scheme uses whichever it
+     * needs.
      *
      * \ingroup SearchHeuristics
      */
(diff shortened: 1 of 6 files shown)


gcs/variable_weighting.hh
```diff
@@ -220,6 +243,63 @@ namespace gcs
...
+     */
+    class RefinedWeighting final : public VariableWeighting
+    {
+    public:
+        /**
+         * \brief Which of the ia / ca / id / cd / ca.cd increments to use.
+         */
+        enum class Variant
+        {
+            Ia,
```
ciaranm: I think I want these spelled out as long names.

gcs/variable_weighting.hh
```diff
@@ -228,6 +308,11 @@ namespace gcs
     enum class WeightingScheme
     {
         Classic, ///< ClassicDomWDeg
+        Ia,      ///< RefinedWeighting, initial-arity increment
```
ciaranm: Likewise.

gcs/variable_weighting.hh
ciaranm: Done — spelled out to InitialArity / CurrentArity / InitialDomain / CurrentDomain / CurrentArityCurrentDomain, in both the RefinedWeighting::Variant and WeightingScheme enums (and the --branch CLI strings stay short).

gcs/variable_weighting.hh
ciaranm: Done (same rename, both enums).

ciaranm:

PR #326, opened 2026-06-13
Stage 2 (part 3): --branch hook to benchmark the dom/wdeg schemes (#315)

ciaranm:
## What

**Stage 2, part 3** of dom/wdeg (issue #315) — and the last of Stage 2: a `--branch` hook on an
example so the weighting schemes can actually be **measured**. With this, "the three schemes are
selectable and benchmarkable" (the Stage 2 done-criterion) is met.

> Stacked on #325 → #324 → #323 → #322 → #321 → #318 → `main`.

## Change

The `langford` example (one of the RFC's named benchmark families) gains a `--branch` flag and
switches from `solve()` to `solve_with()` so it can use it:

- `--branch dom-then-deg` (default — identical to the previous behaviour), or
- `--branch dom-wdeg[:VARIANT]` with `VARIANT` ∈ `classic` / `ia` / `ca` / `id` / `cd` / `ca.cd` /
  `chs` (bare `dom-wdeg` means `ca.cd`).

The existing `--stats` flag already prints the `Stats` (recursions, failures, propagations, …), so
the search shape per scheme is reported without further change. Example:

```
$ langford 5 --stats --branch dom-then-deg     # recursions: 25, failures: 24
$ langford 5 --stats --branch dom-wdeg:ca.cd    # recursions: 41
$ langford 5 --stats --branch dom-wdeg:chs      # recursions: 28
```

(langford 5 is unsatisfiable, so this is a clean full-tree comparison — and a reminder that dom/wdeg
isn't a free win on small instances, especially before restarts, which is the sort of thing the hook
now lets you check.)

## Notes

- Default behaviour is unchanged: `--branch dom-then-deg` is exactly the brancher `solve()` used.
- The same `brancher_from_string` pattern copies trivially to other examples; one is enough to make
  the schemes benchmarkable, so this keeps the change to a single example.

This closes out **Stage 2** (all three schemes built, selectable, and benchmarkable). **Stage 3** —
the restart loop — is the next phase per the #315 checklist; `on_restart` is already wired (inert)
in the weighting schemes, waiting for it.

## Verification

Built under **GCC 15.2 and clang 21**; full `ctest` passes; the flag exercised across schemes.

### DIFF
--- a/examples/langford/langford.cc
+++ b/examples/langford/langford.cc
@@ -2,10 +2,12 @@
 #include <gcs/constraints/arithmetic.hh>
 #include <gcs/constraints/element.hh>
 #include <gcs/problem.hh>
+#include <gcs/search_heuristics.hh>
 #include <gcs/solve.hh>
 
 #include <cstdlib>
 #include <iostream>
+#include <optional>
 #include <vector>
 
 #include <cxxopts.hpp>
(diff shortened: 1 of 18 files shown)


ciaranm:

PR #327, opened 2026-06-13
Return a SearchResult from the recursive search, not a bool

ciaranm:
Stage 3 of #315 (restarts) needs the recursive search to be able to signal a third outcome — "abandon this tree and unwind to the root, but we are *not* finished" — distinct from both "subtree fully explored, carry on" and "stop the whole search". Today `solve_with_state` returns a `bool` that only has room for the latter two.

This PR is the **behaviour-preserving** groundwork: it swaps that `bool` for a small `SearchResult` enum, mirroring the [Glasgow subgraph solver's searcher](https://github.com/ciaranm/glasgow-subgraph-solver/blob/master/gss/innards/homomorphism_searcher.cc#L75).

- `Complete` ← old `true` (subtree fully explored)
- `Stop` ← old `false` (abort flag, or a callback returned false; proof abandoned, returns immediately)
- `RestartCutoffHit` — defined but **never produced yet**; it arrives with the restart loop, where (unlike `Stop`) it must unwind through the normal backtrack/forget logging so the proof stays balanced for the restart that follows.

No behavioural or proof change. Full `ctest` 329/329 green under GCC 15 and clang 21.

Stacked on #326 (base `dom-wdeg-benchmark`); the restart loop itself follows in the next PR.


ciaranm:

PR #328, opened 2026-06-13
Add a restart loop around the depth-first search

ciaranm:
Stage 3 of #315: a restart loop around the depth-first search — the foundation dom/wdeg needs, since weights persist across restarts so a later pass searches differently. Stacked on #327 (the `SearchResult` refactor).

**Opt-in, default-identical.** Restarts are enabled by a new `RestartSchedule` field on `SolveCallbacks` (Luby only, scale isolated behind a default per [the tunables guidance]). With none set the cutoff is infinite and search is a single exhaustive pass — byte-for-byte today's behaviour, and the full suite confirms it.

**Control flow** (mirrors the GSS searcher, as suggested): `solve_with_state` produces `SearchResult::RestartCutoffHit` once a run has spent its conflict cutoff; unlike `Stop` it unwinds through the normal backtrack/forget logging so the proof stays balanced. The driver loops while the result is `RestartCutoffHit`, calls `on_restart()` on the weighting (reached via the `ConflictObserver` already on the propagators), and advances the schedule. One recursive function still serves decision / enumeration / optimisation — no duplication. `stats` gains a `restarts` counter.

**Proofs** (the subtle part, and the bit @ciaranm's one-level hint cracked):
- A restart abandons a subtree *unrefuted*, so it asserts **no** backtrack lemma (the negation-of-guesses isn't RUP-derivable) — it only deletes what the pass derived, which is always sound.
- But the **root node's guess-independent propagation lives at proof level 1**, and the next pass starts from the same root fixpoint so `propagate()` re-emits nothing there. Deleting it would leave the next pass's backtracks non-RUP. So the root (depth 0) **keeps level 1**; deeper guess-dependent levels are re-derived each pass.

**Soundness guard.** Without nogoods a restart re-explores and so re-finds solutions, so a schedule is only sound for finding one solution or optimising. Continuing past a solution with restarts on and no objective now throws `UnimplementedException` rather than miscounting; Stage 4 (nogoods) lifts this.

**Testing.** `solve_test` covers unsat, optimisation, the enumeration guard, and an unsatisfiable Langford instance whose AllDifferent/Element root propagation exercises the level-1 fix (it fails without it). `langford` gains a `--restart` hook to benchmark restarts on a real instance (e.g. langford-5 unsat: a couple of restarts, then the growing cutoff exhausts and proves unsat — proof verifies). Full `ctest` 330/330; proofs verify under GCC 15 and clang 21.

Next: Stage 4, the `Nogoods` constraint.

### DIFF
--- a/examples/langford/langford.cc
+++ b/examples/langford/langford.cc
@@ -98,6 +98,9 @@ auto main(int argc, char * argv[]) -> int
             ("branch", "Branching heuristic: dom-then-deg, or dom-wdeg[:VARIANT] " //
                        "(VARIANT = classic / ia / ca / id / cd / ca.cd / chs)",    //
                 cxxopts::value<string>()->default_value("dom-then-deg"))           //
+            ("restart", "Restart on a Luby schedule with the given conflict scale " //
+                        "(finds one solution only, since restarts cannot enumerate)", //
+                cxxopts::value<unsigned long long>()->implicit_value("100"))       //
             ;
 
         options.add_options()                                                                   //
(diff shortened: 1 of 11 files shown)


ciaranm:

PR #329, opened 2026-06-13
Add the Nogoods constraint (propagation core)

ciaranm:
Stage 4 of #315: the **propagation core** of the `Nogoods` constraint, fed a *fixed* set (extraction at restart boundaries is Stage 5; restart wiring is Stage 6). Stacked on #328.

A `Nogoods` constraint stores a set of nogoods — each a conjunction of decision literals (`=`/`≠`/`≥`/`≤`) that must not all hold — and enforces **per-clause unit propagation**, *not* GAC over the conjunction (which is co-NP-hard).

**Mechanism.** It generalises `NegativeTable`'s two-watched-literal scheme from `var==value` tuples to arbitrary order/equality literals. Watching is **entailment-based**, riding `State::test_literal`: a literal `x ≥ 10` counts as held once `lower(x) ≥ 10`, so an `x ≥ 12` bound makes it hold too. Watches are non-backtrackable (a moved watch stays sound as the state relaxes). `define_proof_model` emits each nogood as a CNF clause of negated literals; inferences are `JustifyUsingRUP` — the order-encoding axioms already in the model let RUP bridge literal strengths, so the order-literal units verify with no explicit steps.

**Testing — the two-sided net** (companion plan §6): the solution set is checked against an independent oracle and `--prove` catches over-pruning; a **per-node unit-propagation reference**, run at every search node via the trace callback (deliberately *not* `solve_for_tests_checking_gac`, which would wrongly flag a correct UP propagator), catches *under*-pruning and mis-computed entailment. The reference asserts the 2WL invariant — every unsatisfied clause has ≥2 non-entailed literals — using its *own* bound-based entailment, so it fires if the propagator fails to see `x≥12 ⊨ x≥10`. I confirmed it's non-vacuous by weakening the propagator to skip unit inferences and watching the reference catch them while the solution-set check passed. Cases cover equality / order / not-equal / mixed literals, cascades, a unit nogood, the entailment case, and root unsat.

Full `ctest` 331/331; proofs verify under GCC 15 and clang 21.

Next: Stage 5 — reduced-nld extraction on restart unroll + the log-on-unroll / delete-on-purge proof lifecycle, feeding this constraint.

### DIFF
--- a/gcs/CMakeLists.txt
+++ b/gcs/CMakeLists.txt
@@ -60,6 +60,7 @@ add_library(glasgow_constraint_solver
         constraints/minus.cc
         constraints/mult_bc.cc
         constraints/n_value.cc
+        constraints/nogoods.cc
         constraints/parity.cc
         constraints/plus.cc
         constraints/regular/regular.cc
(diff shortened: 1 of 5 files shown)


ciaranm:

PR #330, opened 2026-06-13
Learn nogoods from restarts and feed them back (end-to-end)

ciaranm:
Stage 5a of #315: closes the restart-learning loop end-to-end. Stacked on #329.

Per the plan-order discussion, this does the **thin end-to-end slice first** — get the proof lifecycle and the feed-back plumbing working with a simple (unreduced) extraction — and leaves the reduced-nld minimisation + d-way edge cases for Stage 5b. The point was to de-risk the novel proof machinery before investing in extraction sophistication, and it held.

**What it does.** When a restart abandons the search, each sibling a frame refuted *before* the cutoff becomes a nogood — the path to that frame plus the refuted decision — fed to a `Nogoods` constraint so later passes prune the refuted region instead of re-exploring it.

**Proof lifecycle (the keystone).** A learned nogood's clause is RUP from that sibling's refutation, which is still in scope at level `depth+1` during the unwind. `ProofLogger::emit_learned_nogood()` derives it there as a `ProofLevel::Top` line, so it survives the restart's `forget` and the next pass can cite it by RUP. No disjunctive reification — it stays within the integer-encoding RUP-closure invariant, exactly as the companion plan predicted.

**Plumbing.** `Nogoods` now backs onto a growable `NogoodStore` shared with the engine; the propagator initialises a newly learned nogood's watches lazily on its next fire (the non-backtrackable watch design already allowed this). `solve_with`, when restarts are on, installs an initially empty `Nogoods` subscribed to every variable, over a store it owns and grows at the unwind — **engine-owned, never user-posted** (which keeps per-thread parallel sane). Decisions that aren't plain conditions are skipped. New `stats.learned_nogoods`.

**Effect** (langford, `--prove`, all verify):

| instance | restarts before | restarts with learning |
|---|---|---|
| size 5 (`--restart=2`) | 14 | 5 |
| size 6 (`--restart=2`) | 126 | 14 |

`solve_test` gains a `learned_nogoods` assertion; the existing restart proof tests now exercise learning and still verify — an unsound learned clause would fail RUP. Full `ctest` 331/331, GCC 15 + clang 21.

Next (Stage 5b): reduced-nld minimisation (smaller nogoods), d-way correctness edge cases, interval/reject-branch handling, then measure on the fast-propagator families.


ciaranm:
Added a follow-on commit: **enumeration + restarts now works**, which the learning enables for free.

Stage 3 guarded against continuing past a solution while restarting (without nogoods, the next pass re-finds and re-counts solutions). The restart nogoods lift that with no new machinery — just dropping the guard — and it's sound without *solution-blocking* nogoods:

- Complete enumeration already logs each solution with `solx`, which **excludes** it (adds its negation to the proof database).
- So once a subtree is fully explored and its solutions logged, every leaf of it is a conflict (a dead end or an excluded solution), and the plain nld nogood we record for it is RUP-derivable — exactly like a genuine refutation's.
- That nogood then stops a later pass re-entering the region, so each solution is found / counted / `solx`'d exactly once.

A solution-blocking nogood would just re-derive what `solx` + the region nogood already give. New test enumerates a 6-solution instance under random branching (so it conflicts and a `luby(1)` schedule restarts mid-enumeration): each solution reported exactly once and VeriPB verifies `complete enumeration of 6`. The old "refuse to enumerate" guard test is removed.

ciaranm:

PR #331, opened 2026-06-13
Allow several conflict observers; wire MiniZinc dom_w_deg to real dom/wdeg

ciaranm:

## The gap

The conflict-observation seam (#318) put a **single** borrowed `ConflictObserver*` slot on `Propagators`, attached once at search start. That is enough for one stateful brancher, but:

- a `seq_search` can chain several stateful branchers, each wanting its own weighting; and
- MiniZinc's `dom_w_deg` was stubbed to `dom_then_deg` precisely because two `int_search(..., dom_w_deg, ...)` annotations would each call `set_conflict_observer` and the second would **silently clobber** the first, freezing the earlier brancher's weights at their initial values.

## The fix (commit 1)

Hold a `vector<ConflictObserver*>` instead of one slot: `add_conflict_observer()` appends, `conflict_observers()` returns them all, `propagate()` notifies each exactly once on a wipeout, and the restart hook fans out to every one. Attribution is unchanged (constraint index, scope, reason); every observer simply gets told.

**Design note (flagging for your call).** I took the *fan-out* option — each brancher owns its own weighting and all observers see all conflicts — because it is the minimal change and matches the stated design ("each [brancher] carry their own weighting"). Since weights are per-constraint and global, several same-scheme weightings just track *identical* counts (correct, but redundant); different schemes each keep their own state. The alternative is a single *shared* weighting that multiple selectors read over different variable subsets (less memory, but needs a new injection API and forces one scheme). Happy to switch to shared-weighting if you'd prefer it.

## MiniZinc wiring (commit 2)

`dom_w_deg` (and the unknown-heuristic fallback, whose warning already promised `dom_w_deg`) now maps to `variable_order::dom_wdeg` with the library default scheme. New differential test `seqwdeg`: a `seq_search` of two `dom_w_deg` groups over coupled permutations — full solution set matches the reference solver, and the proof **verifies complete enumeration of 216 solutions**.

This unblocks running real `dom_w_deg` benchmark suites (XCSP / MiniZinc challenge) — which is what proper search-gain validation needs (see the #315 comment).

## Testing

- GCC 15 + clang 21 builds clean.
- Full ctest 332/332 (was 331; +`minizinc-seqwdeg`), including proof verification.
- New unit test: "Every attached conflict observer is notified of a wipeout".

### DIFF
--- a/gcs/innards/conflict_observer.hh
+++ b/gcs/innards/conflict_observer.hh
@@ -15,13 +15,14 @@ namespace gcs::innards
      * happen, so that a search heuristic can attribute each wipeout to the
      * constraint --- and, later, the specific Reason --- that caused it.
      *
-     * A borrowed observer is attached to the Propagators for the duration of a
-     * search (Propagators::set_conflict_observer). Whenever a propagator raises
-     * a contradiction, Propagators::propagate notifies the observer exactly once
-     * with the failing constraint and its scope. This is the write side of the
-     * dom/wdeg weighting seam (issue #315, Stage 0): the weighting value object
-     * will implement this interface. It is deliberately minimal and carries no
-     * query API of its own --- the brancher reads back weights by another path.
+     * Borrowed observers are attached to the Propagators for the duration of a
+     * search (Propagators::add_conflict_observer). Whenever a propagator raises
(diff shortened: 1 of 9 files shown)


ciaranm:

PR #332, opened 2026-06-13
colour: dom-wdeg / restart / decision-mode benchmark hooks

ciaranm:
Part of the dom/wdeg work (issue #315), stacked on #330 (`restart-nogood-learning`).

Mirrors the langford benchmark hooks (#326) on the graph-colouring example, so dom/wdeg + restarts + nogood learning can be exercised on a classic testbed:

- `--branch dom-then-deg | dom-wdeg[:VARIANT]` (VARIANT one of `classic`, `ia`, `ca`, `id`, `cd`, `ca.cd`, `chs`), branching over the vertex variables;
- `--restarts[=SCALE]` (Luby);
- `--colours K` — a **decision** variant ("is the graph K-colourable?", no objective, find-first) alongside the default minimise-colours mode;
- `--first` to stop at the first solution.

The decision variant matters for dom/wdeg: in the minimise-colours mode the objective machinery dominates the conflicts and skews the weights away from the graph structure, so dom/wdeg looks bad; the decision variant keeps the weighting focused on the graph.

This is the example used for the search-gain validation in the #315 comment (it's how the backdoor-instance result — baseline DNF vs dom/wdeg+restarts solving in ~0.007s — was produced).

## Testing

- GCC 15 + clang 21 builds clean; existing behaviour (no flags) unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #333, opened 2026-06-13
xcsp: --branch / --restarts flags for benchmarking dom/wdeg

ciaranm:

The XCSP frontend had no way to choose a search heuristic — it always used `solve_with`'s default (`dom_then_deg` over all variables) — so dom/wdeg couldn't be benchmarked on XCSP instances at all. This adds, mirroring the langford (#326) and colour (#332) example hooks:

- `--branch dom-then-deg | dom-wdeg[:VARIANT]` (VARIANT one of `classic`, `ia`, `ca`, `id`, `cd`, `ca.cd`, `chs`), branching over all variables as the default brancher does;
- `--restarts[=SCALE]` (Luby) — so the full restart + nogood-learning stack is reachable from the XCSP frontend;
- three extra stat lines: `d RECURSIONS`, `d RESTARTS`, `d LEARNED NOGOODS` (the frontend already printed wrong-decisions / propagations / solve-time, but not the node count).

Default behaviour (no flags) is unchanged: `--branch` defaults to `dom-then-deg`, which is exactly the brancher `solve_with` picks when none is given.

This is what made the suite-style search-gain validation in the #315 comment possible on real XCSP instances (k-colourability decision) — including the finding that the default weighting scheme `ca.cd` is fragile where `classic`/`chs` win.

## Testing

- GCC 15 + clang 21 build clean; existing XCSP behaviour unchanged.

### DIFF
--- a/xcsp/xcsp_glasgow_constraint_solver.cc
+++ b/xcsp/xcsp_glasgow_constraint_solver.cc
@@ -1618,6 +1618,45 @@ namespace
             return result;
         }
     };
+
+    // Map a dom-wdeg variant string to a weighting scheme, for the --branch flag.
+    auto scheme_from_string(const string & name) -> optional<WeightingScheme>
+    {
+        using enum WeightingScheme;
+        if (name == "classic")
+            return Classic;
+        if (name == "ia")
+            return InitialArity;
+        if (name == "ca")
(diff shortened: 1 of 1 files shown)


ciaranm:

PR #334, opened 2026-06-13
Default the dom/wdeg weighting scheme to CHS

ciaranm:

Changes the default weighting scheme for `variable_order::dom_wdeg` from `ca.cd` (CurrentArityCurrentDomain) to `chs` (ConflictHistorySearch).

The search-gain validation ([#315 comment](https://github.com/ciaranm/glasgow-constraint-solver/issues/315#issuecomment-4700060234)) found `ca.cd` to be **fragile**: on a fair (randomly-relabelled) backdoor k-colourability instance it fails to solve even with restarts (574k nodes / 738 restarts in 45 s), whereas `classic` and `chs` solve it in ~14k nodes and the full stack (`+ restarts`) in 412. `chs` was the most robust scheme everywhere measured — also the strongest on uniform random instances in decision mode (3.2× fewer nodes than `dom_then_deg`). So `chs` is a much better out-of-the-box default. This also fixes what MiniZinc `dom_w_deg` inherits (via #331).

The three selector unit tests that exercised the dom/W ratio + degree tie-break used the default scheme implicitly; they're now pinned to `WeightingScheme::Classic`, whose uniform root weights are what those tests actually describe — so they test the comparator, not whichever scheme is the default.

Other tunables (broadening, decay, scale) remain SMAC questions; this is just the scheme default.

## Testing

- GCC 15 + clang 21 build clean; full ctest 331/331.

### DIFF
--- a/examples/langford/langford.cc
+++ b/examples/langford/langford.cc
@@ -66,15 +66,16 @@ namespace
     }
 
     // Build the brancher named by --branch: "dom-then-deg", or "dom-wdeg"
-    // optionally suffixed with a scheme, e.g. "dom-wdeg:chs" (default ca.cd).
+    // (the library default scheme) optionally suffixed with a scheme,
+    // e.g. "dom-wdeg:classic".
     auto brancher_from_string(const string & spec, const Problem & problem) -> optional<BranchHeuristic>
     {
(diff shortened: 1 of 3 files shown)


ciaranm:

Issue #335, opened 2026-06-14
Refined per-literal propagation triggers (watched-literal scheduling at the queue stage)


ciaranm:
## Summary

Today a constraint propagator is the unit of scheduling, and triggers are coarse: a
per-variable bitmask wakes the *whole* propagator on any change / bounds-change /
instantiation of a subscribed variable. "Watched-literal" constraints (NegativeTable,
the new Nogoods) therefore subscribe to *every* variable and re-establish their watch
invariant by *scanning their whole store on every wake* — the 2WL bookkeeping saves
inference work but not wake work, and the wake work dominates.

This issue proposes a **refined trigger mechanism** that lets a propagator watch
individual literals (`X ≥ k`, `X = k`, their negations, later intervals), be woken
*only* when a watched literal is decided, and be handed *exactly which* of its watches
fired — so it does work proportional to breakage, not to store size. Propagators can
refine / replace / remove their watches over time, choosing per-watch whether the
change is restored on backtrack. The immediate motivator is to make Nogoods efficient;
the mechanism is general (the worked second client is bounds-consistent linear).

It is the trigger-redesign stage of the search-heuristics work in #315, but it is an
engine subsystem in its own right.

## Evidence / problem statement

Measured on `restart-nogood-learning`, plain vs `--restarts` (Luby, no weighting), the
slowdown is *purely* a rec/s slowdown (recursions within ~1% of baseline) that scales
with nogood-store size: tsp luby×1e4 = +13.1M total propagations, 99.93% of them
non-effectful — the entire +143s. Two compounding degeneracies in
`gcs/constraints/nogoods.cc`:

1. `install_propagators` triggers `on_change` for **every** variable mentioned in any
   nogood (and `solve.cc` installs the learned-nogoods constraint over
   `problem.all_normal_variables()`), so it is woken on essentially every domain change.
2. The propagator body loops `for ni in 0..store.size()` testing both watches of
   **every** nogood per wake — O(store) per wake, defeating the point of 2WL.

The current trigger machinery (`gcs/innards/propagators.{hh,cc}`): the queue is a
permutation of propagator ids partitioned into enqueued / idle / disabled zones;
`iv_triggers[v.index]` is a `vector<pair<propagator_id, mask>>`; on a change to `v` the
engine ORs the inference bit against each mask and re-enqueues the matching propagators.
The mask only distinguishes change / bounds / instantiated. There is no way to say "wake
[...]

PR #336, opened 2026-06-14
Refined triggers (Stage A): inert RefinedWatchContext seam

ciaranm:
First step of the refined per-literal trigger mechanism (#335). This lands **only the calling-convention seam**, inert and behaviour-preserving — directly analogous to the conflict-observation seam (#318).

## What this does

- Adds `RefinedWatchContext` (`gcs/innards/propagators.hh`): the per-call object a propagator will use to learn which of its refined watches fired since it last ran, via the opaque payloads it chose at registration time. Today it exposes only `fired_payloads()`, which is always empty.
- Threads the context through the `PropagationFunction` wrapper layer. Crucially, `PropagationFunctionImpl` forwards it to the wrapped function **only `if constexpr`** that function is invocable with it; otherwise it calls the old 3-argument form. So **all 47 existing propagators are untouched** — they neither see nor mention the context.
- `Propagators::propagate()` constructs an empty context and passes it at the single call site.

## Why a separate seam PR

The watch index that populates the context, the registration / watch-editing API, and the first client (a bounds-consistent mini-linear) land in later stages (#335, Stages B1 onward). Isolating the wrapper-layer signature change here keeps it tiny and behaviour-preserving, and in particular isolates the one piece that is sensitive to the compiler — the `if constexpr` / `std::is_invocable_v` forwarding — so it gets its own clean check under both GCC and clang.

## Validation

Inert by construction: no refined watches exist, so every propagator receives an empty payload set and ignores it. Gate is byte-identical behaviour.

- GCC 15: `ctest` 331/331.
- clang 21: `ctest` 331/331.
- clang-format clean.

No proof impact (no inference is added, removed, or reordered).

### DIFF
--- a/gcs/innards/propagators.cc
+++ b/gcs/innards/propagators.cc
@@ -295,7 +295,10 @@ auto Propagators::propagate(const optional<Literal> & lit, State & state, ProofL
         int propagator_id = _imp->queue[--_imp->enqueued_end];
         try {
             ++_imp->total_propagations;
-            auto propagator_state = _imp->propagation_functions[propagator_id](state, tracker, logger);
+            // No refined watches are registered yet, so every propagator sees an
+            // empty fired-payload set (see RefinedWatchContext); this is inert.
+            RefinedWatchContext watches{};
+            auto propagator_state = _imp->propagation_functions[propagator_id](state, tracker, logger, watches);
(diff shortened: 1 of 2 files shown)


ciaranm:

PR #337, opened 2026-06-14
Refined triggers (B1): per-literal watch index + mini-linear client

ciaranm:
Stage B1 of the refined per-literal trigger mechanism (#335), on top of the Stage-A seam (#336). This is the first stage with real semantics: it builds the watch engine and exercises it end-to-end with a small client.

## Engine (`gcs/innards/propagators`)

- **Refined per-literal watch index**, parallel to `iv_triggers` (which is untouched — coarse-triggered propagators see no change). A watch is `(literal, payload, owner)`; when the literal becomes entailed (`test_literal == DefinitelyTrue`) it appends `payload` to the owner's fired-watch inbox, wakes it, and is **consumed**.
- Propagators arm watches two ways: at install via `Triggers::refined` (a persistent baseline) or dynamically via `RefinedWatchContext::watch` while propagating.
- **Restore-on-backtrack** via a per-`propagate()` edit trail replayed in reverse by a single `on_backtrack` callback. This is necessary because `State` runs an epoch's callbacks in *registration order*, so naive per-edit undo closures would compose incorrectly (e.g. an add-then-consume of the same watch). The trail snapshot is taken **before** the guess first pass, which can itself fire and consume watches.
- The fired-watch inbox is transient within a `propagate()`: cleared after a propagator runs, with anything left undelivered by a contradiction cleared at exit.

## Client + validation (`gcs/constraints/mini_linear_test.cc`)

`MiniLinearGreaterEqual` is a test-only `sum c_i x_i ≥ K` (positive coefficients, non-negative variables) bounds-consistency propagator with **no coarse triggers** — it arms one upper-bound watch per variable and re-arms on fire. It's a deliberately small vehicle to exercise the engine, and is expected to be retired when refined triggers fold into the real `Linear` (stage E).

Validated three independent ways over hand-picked + 200 random instances:
1. **Per-node BC consistency** (`solve_for_tests_checking_consistency`) — catches under-propagation, including at post-backtrack nodes (so it tests restore-on-backtrack directly).
2. **Solution-count vs a brute-force oracle** — catches duplicate solutions, which a solution-*set* comparison silently hides.
3. **Refined-vs-coarse** (same propagation body, two trigger modes) — asserts an identical search tree (equal recursions) and solution count, with refined never waking more. Measured ~1.3–1.5× fewer wakes, from skipping the lower-bound-change and self-retrigger wakes coarse `on_bounds` incurs.

Every machinery point was **mutation-tested** — disabling firing, re-arm, restore-on-backtrack, trailing the consume, or correct snapshot ordering each makes a check fail, confirming the path is load-bearing and the checks are non-vacuous.

## A bug the methodology caught

During development the trail snapshot was taken *after* the guess first pass, so that pass's consumes were never undone on backtrack — producing **duplicate solutions**. This **passed the set-based consistency check** (a set dedups) and was caught only by the count check added for the laziness measurement — exactly the "differential testing is sampling / green ≠ no-op" gap from the design discussion, caught in practice. The snapshot was moved before the first pass, and the count check now guards the regression.

## Gate

- GCC 15 and clang 21: `ctest` **332/332**.
- clang-format clean.
- No proof impact: the trigger change is semantics-preserving, and mini-linear runs without proof (the real `Linear` certifies the bounds logic).

### DIFF
--- a/gcs/CMakeLists.txt
+++ b/gcs/CMakeLists.txt
@@ -165,6 +165,7 @@ if(GCS_BUILD_TESTS)
     add_executable(linear_constant_test constraints/linear/linear_constant_test.cc)
     add_executable(logical_test constraints/logical_test.cc)
     add_executable(min_max_test constraints/min_max_test.cc)
+    add_executable(mini_linear_test constraints/mini_linear_test.cc)
     add_executable(mult_bc_test constraints/mult_bc_test.cc)
     add_executable(n_value_test constraints/n_value_test.cc)
     add_executable(nogoods_test constraints/nogoods_test.cc)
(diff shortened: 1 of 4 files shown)


ciaranm:

PR #338, opened 2026-06-14
Refined triggers (B2): PB watched-set laziness for mini-linear

ciaranm:
Stage B2 of the refined per-literal trigger mechanism (#335), on top of B1 (#337). This is the **real laziness**: mini-linear now watches only a subset of its variables, so variables that don't matter never wake it.

## Engine (`gcs/innards/propagators`)

One small addition: `ctx.is_watching(var)` — a query over the (already backtrack-consistent) watch index telling a propagator whether it currently watches a variable. That lets a propagator recompute its watched set on each run **without keeping its own state**, which would otherwise have to be kept consistent with backtracking. No new storage, no new trail — it reads what B1 already maintains.

## The PB watched-set (`gcs/constraints/mini_linear_test.cc`)

`MiniLinearGreaterEqual` now has three trigger modes sharing one propagation body — `Coarse` (on_bounds), `RefinedAll` (B1: a refined watch on every variable), and `WatchedSet` (B2). The watched-set keeps armed only enough upper-bound watches to cover `K + W` (`W = maxⱼ cⱼ·(ubⱼ−lbⱼ)`). Since `max_sum ≥ Σ over the watched set`, that certifies global dormancy, so **variables outside the set are never watched**. Correctness rests on two monotonicity facts:

- `W` only ever *decreases* under tightening, so a stale `W` is conservative — never a missed wake.
- After BC pruning, `slack ≥ W`, so arming all variables always achieves coverage (the grow loop terminates).

It arms all variables at install (so scope/degree — and hence the search tree — match the other modes), **sheds** a fired variable by not re-arming it when it's not needed for coverage, and **grows** (largest contribution first) when coverage drops. Shedding needs no explicit watch removal: a shed variable is just one whose consumed watch wasn't re-armed, and backtrack-restore (B1) handles the rest.

## Validation

- Correctness over hand-picked + 200 random instances in **both** refined modes: per-node BC consistency (under-pruning, incl. post-backtrack) and a raw solution-count check vs a brute-force oracle.
- A three-way laziness comparison asserts an **identical search tree** (recursions) and solution count across coarse / refined-all / watched-set, with `watched-set ≤ refined-all ≤ coarse` wakes and **strictly fewer** for watched-set on dominant instances. Measured, `x,y,z ∈ [0,15]`, `100x+y+z ≥ 100`: **7680 / 3840 / 1112** propagations (watched-set is 3.5× fewer than refined-all, 7× fewer than coarse) — same tree, same 3840 solutions.
- **Mutation-tested**: breaking the grow step (under-coverage) is caught by the BC/count checks; degenerating the watched-set to watch-all is caught by the strict-win check. Honest note: breaking `is_watching` only produces duplicate watches on the already-covering variables — benign bloat, with no correctness or headline-laziness effect — so no behavioural check flags it, correctly.

## Gate

GCC 15 and clang 21: `ctest` **332/332**. clang-format clean. No proof impact (semantics-preserving; mini-linear runs without proof).

### DIFF
--- a/gcs/constraints/mini_linear_test.cc
+++ b/gcs/constraints/mini_linear_test.cc
@@ -1,17 +1,22 @@
-/* Test vehicle for the refined per-literal trigger mechanism (issue #335, stage
- * B1). MiniLinearGreaterEqual is a deliberately small `sum c_i x_i >= K`
- * bounds-consistency propagator (all c_i > 0, all x_i >= 0) that drives itself
- * ENTIRELY from refined watches: it has no coarse Triggers, arms one upper-bound
- * watch per variable at install, and re-arms a variable's watch when it fires.
+/* Test vehicle for the refined per-literal trigger mechanism (issue #335, stages
+ * B1 and B2). MiniLinearGreaterEqual is a deliberately small `sum c_i x_i >= K`
+ * bounds-consistency propagator (all c_i > 0, all x_i >= 0). It runs in three
+ * trigger modes with an identical propagation body, so they can be compared
+ * directly: coarse on_bounds; a refined upper-bound watch on every variable,
+ * re-armed on fire (B1); and the PB watched-set that keeps armed only enough
+ * upper-bound watches to cover K + W, so variables outside the set are never
+ * watched (B2).
  *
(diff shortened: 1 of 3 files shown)


ciaranm:

PR #339, opened 2026-06-14
Refined triggers (C-1): Nogoods via refined per-literal watches

ciaranm:
Stage C-1 of the refined per-literal trigger mechanism (#335), on top of B2 (#338). This is the first *real* client of the refined-watch engine: the `Nogoods` constraint's fixed-nogoods path moves off the coarse "wake on every nogood variable and re-scan the whole store" scheme onto refined watches, so it wakes only when a clause loses a non-entailed literal and visits only the clauses that fired. That whole-store scan is the source of the ~13.1M wasted propagations measured in #315. The growable restart-learning store stays coarse for now (stage C-2).

## Scheme: watch every literal, restore-on-backtrack

Each nogood arms a refined watch on **every** literal (payload = the clause index). A literal's watch is consumed when it becomes entailed and restored when it un-entails on backtrack, so the armed set is always exactly the clause's non-entailed literals; the propagator visits a fired clause and does the unit/contradiction check (0 non-entailed → contradiction, 1 → force its negation, ≥2 → nothing). This reuses the B1/B2 restore mechanism directly and is **stateless** — there is no per-clause "which positions are watched" bookkeeping to hold consistent with backtracking.

I started on the two-watched-literal scheme we'd planned (no-change-on-backtrack) and the differential below shot it down, for two reasons worth recording:

- **Restore, not leave-in-place, is essential.** A watch can fire and be consumed inside a `propagate()` that a *sibling* clause's contradiction ends *before this propagator runs* to process the fire. Restore puts that abandoned consume back on backtrack; leaving it removed (classic non-backtrackable watches) loses the watch for good and the clause later misses a unit/contradiction. Classic SAT 2WL never hits this because it doesn't consume-then-abandon a watch — it's specific to this engine's consume-on-fire. A true 2WL scheme (fewer watches) would need an extra engine primitive to keep its watched-pair bookkeeping backtrack-consistent; deferred to C-2 if measurement shows the extra wakes matter.

- **`initialise()` does not fire refined watches** (only `propagate()`'s requeue does), so a clause made unit by an initialise-time entailment would be missed. The propagator does a one-time full scan on its first (root) run to cover this, mirroring the coarse path's root re-scan; thereafter fired-only suffices. (Non-backtrackable flag — restarts, in C-2, will need to re-scan at each root.)

`Nogoods` gains a `refined` flag: the fixed-nogoods constructor defaults to refined, the growable-store constructor stays coarse, so `solve.cc`'s learned-nogood path is unchanged. Proof model and inferences are untouched (`JustifyUsingRUP` + `generic_reason`), so the conversion is semantics-preserving and proofs verify.

## Validation

`gcs/constraints/nogoods_test.cc` gains a **scan-vs-refined differential**: solve each instance both ways with a deterministic, degree-independent brancher (`in_order` — degree-independent matters, since the coarse path adds an `on_change` trigger per variable and the refined path adds none, so `dom_then_deg` would diverge for non-propagation reasons) and assert **identical recursions and solution count**. A missed inference under-propagates (more recursions); an unsound one prunes a real solution (fewer). Run over the hand-picked instances and 300 random backtrack-heavy ones. The differential is what caught both subtleties above during development.

Each mode is also run through the existing brute-force oracle, the per-node unit-propagation reference, and VeriPB proof verification. Three mutation tests confirm the checks are load-bearing: disabling the first-run scan is caught via recursions (same solutions, more nodes); disabling restore brings the abandoned-fire bug back; skipping the unit inference is caught.

### DIFF
--- a/gcs/constraints/nogoods.cc
+++ b/gcs/constraints/nogoods.cc
@@ -10,6 +10,7 @@
 
 #include <algorithm>
 #include <cstddef>
+#include <cstdint>
 #include <limits>
 #include <memory>
 #include <optional>
@@ -37,8 +38,10 @@ using std::shared_ptr;
 using std::size_t;
 using std::string;
 using std::stringstream;
+using std::unique;
 using std::unique_ptr;
 using std::vector;
+using std::ranges::sort;
 
 #if defined(__cpp_lib_print) && defined(__cpp_lib_format)
 using std::print;
@@ -69,8 +72,9 @@ auto NogoodStore::size() const -> size_t
     return _nogoods->size();
 }
 
-Nogoods::Nogoods(vector<Nogood> nogoods) :
-    _store(make_shared<NogoodStore>())
+Nogoods::Nogoods(vector<Nogood> nogoods, bool refined) :
+    _store(make_shared<NogoodStore>()),
+    _refined(refined)
 {
     for (auto & nogood : nogoods) {
         for (const auto & lit : nogood)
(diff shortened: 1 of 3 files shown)


ciaranm:

PR #340, opened 2026-06-14
Refined triggers (C-2): refined watches for the learned-nogood store

ciaranm:
Stage C-2 of the refined per-literal trigger mechanism (#335), on top of C-1 (#339). This is the actual payoff: the restart loop's engine-owned **learned-nogood store** moves off the coarse wake-on-every-variable-and-scan-the-whole-store path onto refined per-literal watches — the ~13.1M-wasted-propagation cost measured in #315.

## Growable catch-up

C-1 handled a fixed nogood set (all known at install, armed via `Triggers`). The learned-nogood store is empty at install and grows during search, so its watches can't be armed up front. The key observation from `solve.cc`: the refined propagator runs *un-fired* only at a root re-propagation (`propagate(nullopt)`, depth 0), which the restart loop re-enters once per pass, and the root epoch persists across passes (only child epochs are backtracked between passes). So a non-backtrackable "armed high-water" stays in step with the watches armed there.

On each root re-propagation the propagator now:

- **catches up** newly-learned nogoods — arms a watch on every literal of each clause learned since the last pass; and
- **re-scans** every clause for a unit/contradiction (`initialise()` and root propagation can entail literals without firing refined watches, and a just-armed clause may already be unit).

Away from the root, every entailment goes through `requeue` and fires a watch, so fired-only processing suffices — the laziness win. This replaces C-1's one-time `did_root_scan` flag with a per-restart-root re-scan keyed off `fired_payloads().empty()`.

`solve.cc` selects the refined path by default; `GCS_LEARNED_NOGOODS_SCAN` selects the legacy scan path (for the differential and the perf comparison below). The scan path stays in the tree as the differential oracle.

## Validation

- **Scan-vs-refined differential** (`solve_test`): a pigeonhole UNSAT (6 into 5) under `luby(1)` — many restarts, a growing store, the catch-up exercised at every restart root — under deterministic, domain-based (degree-independent) branching. Asserts byte-identical recursions, failures, restarts, learned-nogood count and solutions between the two paths. Two mutations confirm it is load-bearing: skipping the catch-up arming diverges (194 vs 62 recursions), as does skipping the per-root re-scan (105 vs 62).
- All existing restart tests pass on the refined path (now the default) and **VeriPB** verifies their proofs.
- **Semantics-preserving and faster**, measured scan vs refined (identical search each time):
  - langford-8 `luby(10)`: identical tree, fewer propagations;
  - qap-12 `luby(1000)` (471 learned nogoods): 7.50s → **6.83s**;
  - qap-12 `luby(100)` (1896 learned nogoods): 10.02s → **7.69s**.

  The win grows with the nogood-store size — it is the whole-store-per-wake scan being removed.

A true two-watched-literal scheme would arm fewer watches still, but in this consume-on-fire engine it needs an extra primitive to keep its watched-pair bookkeeping backtrack-consistent; left as a later optimisation if the per-literal wake count shows up in profiling.

### DIFF
--- a/gcs/constraints/nogoods.cc
+++ b/gcs/constraints/nogoods.cc
@@ -265,24 +265,24 @@ namespace
     // rather than losing the watch for good and later missing a unit/contradiction.
     // (A true two-watched-literal scheme would arm fewer watches, but in this
     // consume-on-fire engine it needs an extra primitive to keep its watched-pair
-    // bookkeeping backtrack-consistent; left to stage C-2 if measurement shows the
-    // extra wakes matter.)
+    // bookkeeping backtrack-consistent; deferred as a later optimisation if
+    // measurement shows the extra wakes matter.)
     auto install_refined_nogoods(Propagators & propagators, const ConstraintID & id,
         shared_ptr<vector<Nogood>> nogoods, shared_ptr<vector<vector<IntegerVariableID>>> nogood_vars) -> void
     {
-        // Arm a refined watch on every literal of every nogood at install (the
-        // permanent base set; the firing consume is what is trailed and undone on
-        // backtrack). payload = the nogood index, so a fire says which clause to
-        // revisit.
+        // Arm a refined watch on every literal of each nogood present at install
+        // (the permanent base set, via Triggers; the firing consume is what is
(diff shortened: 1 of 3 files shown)


ciaranm:

PR #341, opened 2026-06-14
Refined triggers: per-watch trigger masks to skip dead literal tests

ciaranm:
A perf-driven optimisation on top of C-2 (#340). Profiling the refined nogood firing on the worst case (`tsp --restarts=10000`) showed the dominating cost was **`test_literal`**: `requeue` tested *every* refined watch armed on a variable on *every* change to it, and `test_literal` (variant dispatch + `in_domain` + `has_single_value`) was ~25% of the run — roughly half the nogood overhead. The clause scan, by contrast, was negligible (1.4%); the cost was the firing.

But a watch's literal can only *become* entailed on certain kinds of change: `x == v` only when `x` is instantiated; `x >= k` / `x < k` only on a bound move (an interior removal never moves a bound); `x != v` on any removal. That's exactly the `Inference` granularity the coarse `iv_triggers` already key off (`on_instantiated` / `on_bounds` / `on_change`).

So each `RefinedWatch` now carries a `trigger_mask` derived from its literal's operator — the *same* masks the coarse triggers use — and the firing loop calls `test_literal` only when the current `Inference` is in the mask. Skipping a watch outside its mask is sound (the literal cannot have changed status), so it's purely **semantics-preserving**: same wakes, same inferences, same search tree. The mask is derived from the literal alone, so every refined client benefits with no API change.

## Effect

`tsp --restarts=10000`, refined, identical search either way:

| | before | after |
|---|---|---|
| solve time | 105.3s | **76.4s** (−27%) |
| `in_domain` (profile) | 15% | **4.3%** |
| nogood overhead vs plain (55.6s) | +49.7s | **+20.8s** |

`test_literal` and `has_single_value` drop out of the hot set entirely; the profile now looks like plain tsp. Propagation counts are unchanged (the mask gates the *test*, not propagator wakes). The residual firing cost is now the loop *iteration* (`O(watches/var)`), which a sorted/indexed structure could shave later if it ever matters.

## Validation

The existing scan-vs-refined differentials all still pass **identically** — `nogoods_test` across seeds, the `solve_test` restart differential, and `mini_linear_test` — so the change provably doesn't alter the search. Two mutations confirm the masks are load-bearing: making `Equal` exclude `Instantiated`, and `GreaterEqual`/`Less` exclude `BoundsChanged`, each make the nogoods differential diverge.

### DIFF
--- a/gcs/innards/propagators.cc
+++ b/gcs/innards/propagators.cc
@@ -45,13 +45,17 @@ namespace
 
     // A refined per-literal watch armed on a variable: when `literal` becomes
     // entailed, `payload` is delivered to propagator `owner`. `id` is a stable
-    // identity used to find and undo the watch on backtrack.
+    // identity used to find and undo the watch on backtrack. `trigger_mask` is the
+    // set of Inference granularities that could make `literal` newly entailed (see
+    // refined_watch_trigger_mask); a change outside it cannot fire this watch, so
+    // the firing loop skips the test_literal for it.
     struct RefinedWatch
     {
         Literal literal;
         std::uint32_t payload;
         int owner;
         std::uint64_t id;
+        unsigned trigger_mask;
     };
 
     enum class WatchEditOp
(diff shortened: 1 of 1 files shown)


ciaranm:

PR #343, opened 2026-06-15
Refined triggers: two-watched-literal nogood propagation

ciaranm:
The optimisation deferred at stage C-1, now unblocked: a true **two-watched-literal** scheme for the refined Nogoods path, on top of #341. It's driven by the nogood dump from the earlier probe — the learned nogoods are *fans* of `x=v` clauses sharing a common prefix (e.g. `{v0=0,v1=1,v2=2,v3=6,v4=k}` for many `k`), so an early-decision variable is watched in nearly every clause. Watch-all fires those prefix literals across the whole fan on each instantiation; 2WL watches each clause at only two literals, so an entailed prefix literal isn't one of a clause's watches and doesn't fire there. That's exactly the fan's sweet spot.

## The C-1 obstacle, and the small primitive that clears it

2WL needs per-clause "which two positions are watched" bookkeeping that stays consistent with backtracking — which the `const State &` propagator can't maintain, and which `is_watching` is too coarse to reconstruct. This adds one small engine primitive: a **per-propagator backtrackable scratch** (`ctx.watch_state` / `set_watch_state`), restored by the *same* per-`propagate()` `on_backtrack` that already restores the watch edits, so the bookkeeping moves in lockstep with the watches. The two watched positions pack into one `uint64` per clause.

Using the existing **RESTORE** watches (not non-backtrackable no-change) keeps it clean:

- The unit case just infers and lets the consumed watch be restored on backtrack — no need to keep a watch on the falsified literal as classic SAT 2WL does.
- **Abandoned fires** — a watch consumed in a `propagate()` that a sibling clause's contradiction ends before this propagator runs, the bug that killed no-change 2WL at C-1 — are undone for free by the following backtrack. (Confirmed by mutation.)

Each clause arms two watches at its first root re-propagation (catch-up, shared by the fixed and growable-store paths), or emits a unit/contradiction if already so; on a fire it moves the fired watch to another non-entailed literal, or infers/contradicts when none remains.

## Effect (`tsp --restarts=10000`, identical search)

| | solve time | overhead vs no-nogoods floor (55.6s) |
|---|---|---|
| scan | 190.6s | +135s |
| watch-all + mask (#341) | 76.4s | +20.8s |
| **2WL** | **58.4s** | **+2.8s** |

The learned-nogood machinery is now essentially free on this instance.

## Validation

The scan-vs-refined differentials all still pass **identically** — `nogoods_test` across seeds and the `solve_test` restart differential — so 2WL is semantics-preserving. Two mutations confirm the new mechanisms are load-bearing: skipping the `watch_state` restore, and not trailing the consume, each make the nogoods differential diverge.

### DIFF
--- a/gcs/constraints/nogoods.cc
+++ b/gcs/constraints/nogoods.cc
@@ -248,38 +248,24 @@ namespace
             triggers);
     }
 
-    // The refined path: no coarse triggers. Instead, watch *every* literal of every
-    // nogood, restore-on-backtrack. A literal's watch is consumed when the literal
-    // becomes entailed and restored when it un-entails on backtrack, so the armed
-    // set is always exactly the clause's non-entailed literals. The propagator
-    // therefore wakes precisely when a clause loses a non-entailed literal, and
-    // visits only the clauses that fired -- never the whole store.
+    // The refined path: a two-watched-literal scheme. Each nogood arms exactly two
+    // watches, on two non-entailed literals; when one fires (its literal becomes
+    // entailed) the propagator moves it to another non-entailed literal, and if there
+    // is none the clause is unit/violated, so it forces the survivor's negation or
+    // contradicts. The two watched positions live in per-clause backtrackable scratch
+    // (ctx.watch_state, packed), restored in lockstep with the watches, so this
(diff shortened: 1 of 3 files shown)


ciaranm:

PR #344, opened 2026-06-15
Document the refined per-literal trigger mechanism

ciaranm:
Documentation for the refined-trigger arc (#339–#343), as a stacked docs-only PR so the code PRs stay focused.

Adds **`dev_docs/refined-triggers.md`**, covering:

- the `RefinedWatchContext` propagator API (`fired_payloads`, `watch`, `is_watching`, `watch_state`) and `Triggers::refined`;
- the engine: the per-variable watch index, fire → consume → inbox → wake, restore-on-backtrack, and *why* watches are restore-not-no-change (the abandoned-fire reasoning);
- the per-literal **trigger masks** and why they're sound (the same granularity contract the coarse triggers rely on);
- the backtrackable **`watch_state`** scratch, and why it is deliberately *not* `state.add_constraint_state` (per-node full-copy + wrong granularity);
- the **two-watched-literal `Nogoods`** client: catch-up, the move/unit/contradiction logic, the fan structure that motivates 2WL, and the measured perf arc (190.6s → 58.4s on `tsp --restarts=10000`);
- the correctness invariants and pitfalls a future change must preserve, and the test/mutation methodology.

Also tidies the cross-references: the now-resolved "trigger caveat" in `restarts-nogoods-weighting.md` points to the new doc, `constraints.md`'s Triggers section notes the refined alternative, and the dev_docs README indexes it.

No code changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #345, opened 2026-06-15
Benchmark CLI flags: --branch, --restarts, --timeout

ciaranm:
Adds uniform command-line options to the benchmark / example programs so the
restarts + nogoods + heuristic configurations can be measured against a
baseline. This is the benchmarking infrastructure behind the dom/wdeg + restarts
+ nogoods work (#315), and produced the review dataset for the refined-triggers
stack (#336–#344).

### Flags

- `--branch default | dom-wdeg[:VARIANT]` — keeps each program's existing value
  order; the minicp programs' default stays `variable_order::dom` + `smallest_in`
  so the same-search-tree comparison against MiniCP is unaffected.
  (`VARIANT = classic / ia / ca / id / cd / ca.cd / chs`; bare `dom-wdeg` = chs.)
- `--restarts[=SCALE]` — Luby restart schedule (added where missing).
- `--timeout SECONDS` — graceful abort that still prints the statistics reached,
  so a capped run is not lost.

### Shared helper

A header-only `examples/benchmark_cli.hh` holds the shared bits: `scheme_from_string`
for `--branch`, and `solve_with_timeout`, which wraps `solve_with` with an
interruptible timer that sets the solver's abort flag at the deadline.

`langford` already had `--branch` / `--restarts` and only gains `--timeout`;
`ortho_latin` moves from `solve()` to `solve_with` to gain the options.

**No behaviour change with default flags** — baseline invocations are unchanged.

Based on #344. Logically independent of the triggers stack; sits at the top of
it only so it can exercise the whole chain.

### DIFF
--- a/examples/benchmark_cli.hh
+++ b/examples/benchmark_cli.hh
@@ -1,0 +1,72 @@
+#ifndef GLASGOW_CONSTRAINT_SOLVER_GUARD_EXAMPLES_BENCHMARK_CLI_HH
+#define GLASGOW_CONSTRAINT_SOLVER_GUARD_EXAMPLES_BENCHMARK_CLI_HH
+
+// Shared helpers for the benchmark / example command-line programs: a dom-wdeg
+// weighting-scheme name parser (for a `--branch` flag) and a graceful
+// `--timeout` wrapper around solve_with. Header-only; included by the benchmark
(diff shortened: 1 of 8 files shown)


ciaranm:

Issue #354, opened 2026-06-17
Two-level aux-naming primitive for verified encodings (x[name][i][role] / b[name][role])


ciaranm:
Part of #353.

The single biggest blocker for the verified-encoding (workflow-2) counting/extensional family.

## Problem

The solver names every proof flag/selector `f[{index}][{stem}]` (`names_and_ids_tracker.cc`, the `allocate_xliteral_meaning(ProofFlag)` render `format("f[{}][{}]", flag.index, name_of(flag))`). `cake_pb_cp` instead uses a **two-level** naming:

- per-position / per-pair **selectors** `x[name][i][role]` (e.g. `x[_1][0][eq]`, `x[_1][i_j]`), and
- fully-reified **flags** `b[name][role]` (e.g. `b[_1][gt]`, `b[_1][lt]`).

Because VeriPB binds variables by name, the solver's proof referencing `f[idx][stem]` (or reusing an operand's `i[Y][eq j]` literal) where cake's OPB has `x[name][i][role]` is a hard mismatch — the chain sees them as different variables. (The spec's planned `@b[name][role]` family was never built.)

After #351, *count* differences are tolerated, so this is purely a **naming** problem: where cake defines a selector we need to reference, we emit it under cake's name (and definition).

## What it unblocks

- **count** — per-position `x[name][i][eq]` selectors.
- **element** — per-position guarded selectors (vs. the solver reusing the index var's `i[Y][eq j]`).
- **`in`** — per-value selectors (the arg-order half is trivial and separate).
- **all_different** — byte-match (`x[name][i_j]` naming + polarity; already chain-verifies).
- **reified equality** (`equals_iff`/`lin_equals_iff`) — fully-reified `b[name][role]` flags (cake already shortened its side, 4→2 lines).
- **Tier-3 selector users**: among, nvalue, table.

## Scope

1. A naming entry point in `NamesAndIDsTracker` (parallel to `create_proof_flag`) that mints a flag/selector rendered as `x[constraint_id][position][role]` / `b[constraint_id][role]`, threaded the owning `ConstraintID` + position + role at creation (see #119, [feedback: explicit args over stateful "current X"]).
2. Re-encode each consuming propagator to use it — and **re-verify each one's explicit (`JustifyExplicitly`) proof steps**. This is the real cost, not the naming.
3. Land per-constraint on top, with a `scp_cases` chain test each (and an `opbdiff --match-labels strict` byte-match where the encoding now lines up).

[...]

Issue #353, opened 2026-06-17
Verified-encoding (workflow-2) constraint coverage: conform the solver's OPB to cake_pb_cp


ciaranm:
Tracking issue for getting more constraints through **workflow-2** of the verified-encoding chain: `cake_pb_cp` re-derives the OPB from the `.scp`, VeriPB elaborates the solver's proof against *cake's* OPB, and the verified `cake_pb_cp` re-checks the elaborated core.

## Framework

After the relative-line work (#351), the chain needs the two OPBs to agree on **variable names**, **not** on constraint **count**. Every divergence is one of two axes:

- **Naming** — the solver's proof references a variable cake doesn't have by that name (`f[idx][stem]` vs cake's `x[name][i][role]`/`b[name][role]`). **Hard blocker** (VeriPB binds by name).
- **Shape/count** — cake emits more constraints than we do. **No longer a blocker** (relative refs + no `f` rule).

**Conform-vs-request:** default to **conform** (cake's two-level naming is the project's adopted convention, and cake's encoders are formally verified so cake-side changes are expensive). Request upstream only where conforming is genuinely wasteful *and* cake's authors prefer it. Track record: `in` / reified-shorter / nested brackets all landed upstream; binary `{0,1}` we conformed instead.

## Current state

| Tier | Constraints | Status |
|---|---|---|
| **1 — chain-verified, in harness** | abs, comparison (4 ops incl. negative domains), equals, not_equals, linear (eq/le/ge/ne, neg coeffs, multi-term), all_different (`none`), binary `{0,1}` vars, multi-constraint | ✅ works end-to-end |
| **1 — new (`none` mode)** | plus, minus (unsat), and, or, inverse | ✅ #352 |
| **2 — format-ready, encoding diverges** | element, count, `in` (selectors), all_different byte-match, reified equality (`equals_iff`/`lin_equals_iff`) | needs the keystone |
| **3 — cake + solver have it, no round-trip** | among, nvalue, table (need keystone); lex, circuit (flag-heavy, need keystone); min_max/parity (1 flag, verify) | mixed |
| **4 — cake can't encode** | cumulative, diffn, knapsack, regular, multiplication, nonlinear arith, sort, smart_table, gcc, value_precede | needs upstream cake encoders |

## Done

[...]

PR #359, opened 2026-06-17
Keystone: scalar b[id][role] aux-naming overload; not_equals byte-matches cake (#354)

ciaranm:
Part of #354 (umbrella #353). Extends the verified-encoding (workflow-2) keystone with cake_pb_cp's **second** naming pattern.

## What

`create_proof_flag(ConstraintID, annotation)` mints a **scalar** flag named `b[id][annotation]`, mirroring cake's `Flag annotation` constructor (no index list) — the `b[...]` sibling of the position-indexed `x[id][...]` overload landed in #355. ProofModel gets the matching forwarder.

## First consumer: not_equals

cake encodes `V1 != V2` with a single per-constraint selector `nev` = `b[id][ne]`: the `gt` half (`V1 > V2`) half-reified on the flag, the `lt` half (`V1 < V2`) on its negation. The solver's encoding (both the integer `ReifiedEquals` and linear `ReifiedLinearEquality` `MustNotHold` paths) was already identical **modulo the flag name** (`f[index][gt]` / `f[index][linne]`) and **same polarity** — the two cases previously chain-verified in `aux` mode. The propagators prune by RUP / domain-disjointness and never reference the selector by name, so this is a **pure OPB-definition rename** — the `b[...]` analogue of all_different for `x[...]`.

## Tests

- `not_equals_sat`, `lin_not_equals_sat`: `aux` → **`strict`** (now byte-match cake).
- `not_equals_unsat`, `lin_not_equals_unsat`: new, SAT/UNSAT parity (refutation proof shape). Operands forced equal via `1..1` singletons (clean bit-encode, not the `{0,1}` binary case).
- `scp_chain` 37 → 39, all green. All `equals` / `linear` constraint tests pass with proof verification (the rename touches no proof step).

## Note

The scalar overload's `std::string` parameter is ambiguous against a braced index list (`{i, j}` also matches `std::string`'s `(count, char)` ctor), so all_different's call site now passes an explicit `std::vector<long long>`.


gcs/constraints/equals.cc
```diff
@@ -182,12 +182,14 @@ auto ReifiedEquals::define_proof_model(ProofModel & model) -> void
                 WPBSum{} + (1_i * _v1) + (-1_i * _v2) == 0_i);
         },
         [&](const reif::MustNotHold &) {
-            // cake_pb_cp: V1 != V2 split into gt (V1 > V2) and lt (V1 < V2).
-            auto gtflag = model.create_proof_flag("gt");
+            // cake_pb_cp: V1 != V2 split into gt (V1 > V2) and lt (V1 < V2) on a
+            // single per-constraint selector b[id][ne] (cake's `nev`): the flag
+            // true selects the gt half, false the lt half.
+            auto neflag = model.create_proof_flag(_constraint_id, "ne");
             model.add_labelled_constraint(as_string(_constraint_id), "gt", "ReifiedEquals", "greater option",
```
ciaranm: I guess the "gt" here is wrong now (same for the "lt" below). These are only used for comments in the OPB file, I think, but if we're going to have the comments they should probably make sense.

Issue #358, opened 2026-06-17
Conform the solver's variable order/eq-literal encoding to cake_pb_cp (gates element/among/table byte-match)


ciaranm:
Part of #353. Surfaced by element (#357), the keystone's third consumer.

## The divergence

When a constraint reifies on a variable's `==` literal (element on the index, and the eq-literal parts of among/table), workflow-2 byte-match is blocked not by the constraint's own encoding but by how the solver encodes the **variable's order/eq literals** vs how `cake_pb_cp` does.

For an index `I ∈ 0..2`:

| | solver (lazy, lean) | cake (eager, full) |
|---|---|---|
| `ge` ladder | omits trivial `ge0` (=true) / `ge(n+1)` (=false) | emits the whole ladder, incl. boundaries |
| eq definition | `eq0 ⇔ ¬ge1` (2 clauses, unlabelled) | `eq0 ⇔ ge0 ∧ ¬ge1` (`@i[I][eq0][f]`/`[r]`) |
| labels | unlabelled | every line `@i[I][…]` labelled |
| timing | introduced lazily, on first reference | eager, up front |

The two are **semantically equivalent** (the solver's literals are a subset of cake's, and `ge0`/`ge(n+1)` are trivially true/false), which is why element **chain-verifies** in workflow 2 today (the solver's proof checks fine against cake's OPB). It's purely the OPB *shape* that diverges, so `opbdiff` can't byte-match.

## Scope

This is the #303 range/order-literal machinery in `names_and_ids_tracker`, shared by every constraint. Conforming it is a substantial change to core proof infrastructure and risks perturbing many existing proofs, so it wants its own design pass — not a per-constraint fix.

## Conform-vs-request decision

Genuinely open, unlike the earlier consumers:
- **Conform**: make the solver emit cake's eager, full, labelled ladder (with `eq ⇔ ge ∧ ¬ge'`). Big, cross-cutting, but unblocks the whole eq-literal/extensional family's byte-match at once.
- **Request**: ask the cake side whether it can accept (or normalise to) the lean form — the solver's encoding is arguably the tidier one, and the chain already verifies, so this may be the better trade.

Either way the chain already verifies; this is strictly about the `opbdiff` byte-match oracle for element / among / table. Flagging for a decision rather than defaulting to conform.

## Unblocks (strict byte-match)

- element (#357) — currently `none`-mode, chain-verified
- among, table — their eq-literal parts

ciaranm:
[...]

PR #361, opened 2026-06-17
Proof names: render negative values with '-', not "minus"

ciaranm:
## What

The verbose-names variable encoder spelled negative values as `minusN`
(`i[X][eqminus1]`, `i[X][inminus2_0]`) on the belief that a raw `-` is
invalid in a VeriPB variable name. **That belief is wrong.** The VeriPB
3.0 grammar's variable token (from `veripb-parser`) is:

```
[a-zA-Z_][_a-zA-Z0-9\-\^\[\]\{\}]+
```

so `-` is legal **anywhere except the first character** — and the
surrounding `i[...]` / `p[...]` prefix always supplies a leading letter.
This drops the `minus` workaround in both `allocate_xliteral_meaning`
overloads (eq/ge order literals and `in[lo_hi]` interval literals);
`Integer::to_string()` already renders the sign.

## Why

cake_pb_cp renders negatives with `-` (`int_to_string #"-"`), so the old
`minus` spelling diverged from cake on *every* negative value name — a
self-inflicted gap in the workflow-2 `opbdiff` byte-match oracle. With
this change negative-domain encodings byte-match cake directly.

## Verification

- veripb 3.0.2 accepts `-` mid-name (regex + a hand-built `s VERIFIED`
  refutation referencing a dash-named variable).
- 62 negative-value order-literal names (`i[_][eq-N]`/`[ge-N]`) in
  `n_value_test`'s generated proof verify under veripb.
- A negative-domain `count` instance passes the full chain in **strict**
  mode (`opbdiff` byte-match against cake).
- 37/37 `scp_chain` + 111/111 order-encoding constraint tests
  (n_value, element, count, all_different, among, gcc, in, equals,
  linear, abs, table, smart_table, minizinc variants) pass with proof.
- No `minus`/`eqminus`/`geminus` references remain anywhere in the repo.

### DIFF
--- a/gcs/innards/proofs/names_and_ids_tracker.cc
+++ b/gcs/innards/proofs/names_and_ids_tracker.cc
@@ -1338,11 +1338,11 @@ auto NamesAndIDsTracker::allocate_xliteral_meaning(SimpleOrProofOnlyIntegerVaria
     auto result = XLiteral{++_imp->next_xliteral_nr, false};
 
     if (_imp->verbose_names) {
-        string value_name;
-        if (value < 0_i)
-            value_name = "minus" + abs(value).to_string();
-        else
-            value_name = value.to_string();
+        // Negative values render with a leading '-' (e.g. i[X][eq-1]). The VeriPB
+        // 3.0 grammar permits '-' anywhere except the first character of a name
+        // (which the surrounding i[...] / p[...] always supplies), so this is a
+        // legal name and matches cake_pb_cp's rendering directly.
+        auto value_name = value.to_string();
 
         overloaded{
             [&](const SimpleIntegerVariableID & id) -> void {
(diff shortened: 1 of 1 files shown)


ciaranm:
Closing — the premise is wrong. Dashes are legal in VeriPB 3.0 *variable names* but **not in labels**:

- Var:   `[a-zA-Z_][_a-zA-Z0-9\-\^\[\]\{\}]+`  (`-` allowed)
- Label: `@[a-zA-Z0-9_^\[\]\{\}]+`             (no `-`)

Confirmed against veripb 3.0.2: `@foo-bar` fails (`Expected literal!`), `@foo_bar` verifies. The solver derives a literal's definition **label** from its name, so rendering a negative value as `-N` produces e.g. `@i[X][ge-1]` / `@v[id][-1]` — which veripb rejects. (`v[...]` occurrence flags and `ge` literals are labelled; `eq` literals happen to be unlabelled, so only those would have been safe.)

So the existing `minus` convention is correct and necessary — just for the *label* reason, not the *name* reason. This PR has no upside either: no strict scp_chain case puts a value in a name/label (comparisons/linear bit-encode; count's sum labels are non-negative), and every case that *would* (element/nvalue/among, order-encoded) is already `none`-gated by #358 — and cake itself emits dash-labels there (`@i[A][eq-1]`), which are *also* veripb-invalid, so byte-match is impossible regardless.

(Aside, for upstream CakePB: cake_pb_cp emits dash-bearing OPB labels for negative values, which the reference veripb won't parse. It stays latent because cake treats labels as opaque raw strings and the chain rewrites them during `--elaborate` before cake re-checks the core.)

PR #362, opened 2026-06-17
Among: conform to cake's encode_among; reuse the x[] aux-naming API (#354)

ciaranm:
## What

cake_pb_cp encodes `among` with, per position *i*, a flag `x[id][i][al1] ⇔ (var_i takes a value in the set)` (= `Σ_{v∈iS}(var_i = v) ≥ 1`), then bit-sums those flags to `how_many` as `c[id][le]`/`c[id][ge]` — the same counting-family shape as `count`. The solver previously emitted a single flat `Σ(var = v)` sum plus per-variable at-most-one lines in the propagator's `pol` steps. That structure has no counterpart in cake's OPB, so among couldn't chain-verify at all.

This conforms `Among::define_proof_model` to cake's `encode_among` (one `create_proof_flag_fully_reifying` per position with the `al1` annotation, summed as cake's `c[id]` halves) and reworks the propagator's four explicit justifications to **pin the flags via RUP** instead:

- f = 1 for must-match vars (at-least-one-value axiom + the flag's reverse half),
- f = 0 for must-not-match vars (domain-disjointness + the forward half),

letting the `how_many` bounds follow by RUP from the `c[id]` sum. The whole at-most-one apparatus (`recover_am1`, the initialiser, the `PolBuilder` steps, and the `_sum_line` member) is **deleted** — a single `al1` Boolean per position is inherently ≤ 1. `among` is also wired into the scp reader.

## Result

- among's **own** encoding byte-matches cake exactly: the `@x[_1][i][al1][r|f]` flags and `@c[_1][le|ge]` sum are identical between solver and cake.
- Like element/nvalue, among references the per-value equality literals, so byte-match is gated by the lazy-vs-eager **variable**-encoding divergence (#358, the `@i[...]` ge/eq ladder) — the only differing lines. So the scp_chain cases are registered `none` (chain-verify only).

## Verification

- `among_test` self-verifies: 21/21 `s VERIFIED`, exit 0 (propagation strength unchanged — GAC tests pass).
- `among_sat` / `among_unsat` chain-verify against cake (`s VERIFIED COMPLETE ENUMERATION` / `s VERIFIED UNSATISFIABLE`).
- Full `scp_chain` suite + `among_constraint*` + `minizinc-among`: 43/43, no regressions.


gcs/constraints/among.cc
```diff
@@ -104,89 +110,66 @@ auto Among::install_propagators(Propagators & propagators) -> void
...
-                    return logger->emit(RUPProofRule{}, WPBSum{} + 1_i * a + 1_i * b >= 1_i, ProofLevel::Temporary);
-                }));
-            }
-        }
-    });
-
     propagators.install(
         constraint_id(),
-        [vars = _vars, values_of_interest = _values_of_interest, how_many = _how_many, sum_line = _sum_line, am1_lines = am1_lines](
+        [vars = _vars, values_of_interest = _values_of_interest, how_many = _how_many, flags = _flags](
```
ciaranm, line 115: I'm not so sure about this one: it looks like you're replacing a stateful propagator with one that does full recomputation? Can you benchmark on both performance and proof size?

ciaranm (commented):
I think this needs a rethink: even if the stateful propagator doesn't buy us much here, in general we want to be able to support this kind of pattern, so we'll need to think about how we keep track of the relevant information.

ciaranm:
Reverting per review discussion. Benchmark (7 vars, value set {0,1}, enumerate + prove):

| | old (flat) | new (cake flag) |
|---|---|---|
| among OPB lines | 2 | 16 (7×al1[r/f] + c[le/ge]) |
| aux variables | 0 | 7 |
| proof (pbp) lines | 28,937 | 39,672 (1.37×) |
| proof bytes | 1.43 MB | 3.49 MB (2.43×) |
| veripb verify | 0.14s | 0.20s (1.43×) |

among's value set is a **constant**, so the count is directly `Σ_{v∈S}(X_i = v) == how_many` over fixed eq literals -- no per-position flag needed. cake's uniform `x[id][i][al1]` flag shape is pure overhead here (unlike count, whose value is a variable, and nvalue, which counts distinct values -- both of which genuinely need flags and whose conforms were size-neutral).

Since we own both ends, the right move is to keep our leaner flat encoding and ask CakePB to implement it, rather than adopt cake's larger model + proof. Closing; will re-do once CakePB supports the flat among encoding.

PR #376, opened 2026-06-19
In: dedicated conjunction encoding for the constant case

mmcilree:
## What

For an `In` constraint defined purely over constants (`In{X, <constants>}`,
including the variant `create_integer_variable(vector)` posts), the PB model
now encodes the constraint as the **conjunction of the literals the value set
excludes**, emitted as a single *at-least-n* constraint:

```
var >= min  ∧  var <= max  ∧  ⋀_{h a hole} var != h
```

e.g. `Y in {3,5,8}` becomes one OPB line

```
1 i[Y][ge3] 1 ~i[Y][ge9] 1 ~i[Y][eq4] 1 ~i[Y][eq6] 1 ~i[Y][eq7] >= 5;
```

Every literal has coefficient 1 and the right-hand side equals the literal
count, so unit propagation forces them all true — logically the conjunction,
but in a single line rather than one unit clause per conjunct, keeping the OPB
compact. The mixed / genuine-variable case is unchanged and keeps the existing
selector + at-least-one encoding.

## Why

Against this encoding every `In` inference for the constant case is RUP: a run
of unsupported values below `min` or above `max` contradicts a bound literal
directly, and an interior run of holes walks the order chain through the
`var != h` literals. The constant-case propagator justifications are therefore
plain `JustifyUsingRUP` with no explicit proof steps.

`emit_inequality_to` already simplifies each literal, so no domain-range guard
is needed: a literally-true conjunct (e.g. an upper bound past the variable's
own maximum) folds into the right-hand side, and a literally-false one drops
from the sum but not the count — leaving an unsatisfiable `... >= n`, which is
the correct UNSAT for a set disjoint from the variable's domain.

Per a follow-up clarification, the encoding deliberately uses only order and
equality literals (`>=`, `<=`, `!=`) — no `InRange`/`NotInRange` range literals
for now.

## Testing

Built with the linux-release configuration (capped `-j` to stay within the
container's memory) and run with VeriPB:

- `in_constraint` ctest passes — 45/45 proof cases `VERIFIED`, covering integer
  values, constant vars, mixed, out-of-domain, and UNSAT cases.
- Other consumers of `In` over constants — `scp_reader_test`, `all_equal`,
  `exception`, `circuit_disconnected` (and their view-mixed / xcsp variants):
  7/7 pass.
- `clang-format` applied.

---

## Original request (verbatim from @mmcilree )

>The current PB encoding for the In constraint using an at least one constraint. When an In constraint is defined over constant values In{X, <constants>} we should use an encoding which is the conjunction of literals excluded by the in set. For example X in {3,4} should be [xge3] /\ [xle4] and for Y in {3, 5, 8} we should use [yge3] /\ ~[ye4] /\ ~[ye6] /\ ~[ye7] /\ ~[yge9]. Using this encoding all In inferences for the constant case will be RUP. Create this special case for constants and use the better encoding, and switch the corresponding justifications to use RUP. Make sure it builds, and passes the VeriPB tests, and use clang-format. Be careful when building - you are in a docker container and the build/ is my local machines build (macOS). Ensure you don't run out memory in the container and use the right build configuration (don't just blindly use the linux preset).

### DIFF
--- a/gcs/constraints/in.cc
+++ b/gcs/constraints/in.cc
@@ -178,25 +178,19 @@ auto In::install_propagators(Propagators & propagators) -> void
         return;
     }
 
-    Triggers triggers;
-    triggers.on_change.emplace_back(_var);
-    for (const auto & V : _var_vals)
-        triggers.on_change.emplace_back(V);
-
-    propagators.install(
-        constraint_id(),
-        [var = _var, var_vals = _var_vals, val_vals = _val_vals, selectors = _selectors](
-            const State & state, auto & inference, ProofLogger * const logger) -> PropagatorState {
-            // Step 1: filter dom(var) — drop any value that no source supports.
-            if (var_vals.empty()) {
-                // The pure-constant shape (e.g. create_integer_variable(vector) posts
-                // In over constants): each contiguous run of unsupported values is one
-                // interval conclusion, RUP with no reason. The conjunction encoding
-                // (define_proof_model) forces every var >= min / var <= max / var != h
-                // literal true by unit propagation, so this is immediate: a run below
-                // min or above max contradicts a bound literal directly, and an interior
-                // run of holes walks the order chain through the var != h literals.
-                // Collect the runs first so the domain is not mutated mid-iteration.
+    // Pure-constant In: the conjunction model encoding (define_proof_model)
+    // already forces var into the value set, so the domain filtering follows
+    // immediately from the definition and is established once, at the root, with
+    // no justification needed — the model does all the proof work, so no per-run
+    // proof line is emitted. Domains only ever shrink afterwards, so there is
+    // nothing left to propagate: the initialiser does the filtering and the
+    // propagator disables itself on its first call.
+    if (_var_vals.empty()) {
+        propagators.install_initialiser(
+            [var = _var, val_vals = _val_vals](
+                State & state, auto & inference, ProofLogger * const logger) -> void {
+                // Collect the unsupported runs first so the domain is not mutated
+                // mid-iteration, then drop each as one interval, no justification needed.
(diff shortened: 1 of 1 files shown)


ciaranm:
Isn't this basically "support intervals in the OPB file"?

ciaranm:
I think I don't like this because it's putting actual code into the encoding part. If we're doing intervals here, they should be coming from the user, not consolidating within the encoding. The MiniZinc interface does this by an `Or` constraint currently, not `In`.

mmcilree:
> Isn't this basically "support intervals in the OPB file"?

No, that would be a step after this.

mmcilree:
> I think I don't like this because it's putting actual code into the encoding part. If we're doing intervals here, they should be coming from the user, not consolidating within the encoding. The MiniZinc interface does this by an `Or` constraint currently, not `In`.

Hmm... but if we can verify this encoding then this is ideal right? At the moment there's no range literals used, so if an `In{}` constraint is posted then this gives you
`>= min /\ <= max /\ holes`

After which we can install an initialiser with `infer_not_in_range` and use `NoJustificationNeeded{}`.

Allowing range literals in the encoding/verified encoding would be a separate concern, I agree.

ciaranm:
A different proposal, then: give `In` an `IntervalSet` overload, and use your encoding only on that (no consolidation of constants). Then figure out how we do s_expr for this, and ask Singapore nicely to extend In with more arguments.

mmcilree:
> A different proposal, then: give `In` an `IntervalSet` overload, and use your encoding only on that (no consolidation of constants). Then figure out how we do s_expr for this, and ask Singapore nicely to extend In with more arguments.

Okay yeah that's fair. And then this would be the candidate for using range literals in the OPB file when we get to that.

Issue #377, opened 2026-06-19
Implement the Reasons + `infer` rework (#347) on a 302-then-347 integration branch


ciaranm:
Execution / tracking issue for the Reasons + `infer` rework. The **design** lives in #347 — Plan A (reasons: the `ReasonFunction`-thunk → declarative `Reason` variant) and Plan B (the whole `infer` triple + typed justification hints) are the two long comments there. This issue is the concrete **implementation plan**; #347 stays the design discussion.

Builds on @mmcilree's annotated-assertions work in #302, which is the foundation.

## Strategy: one all-or-nothing integration branch

#302 + the hint-typing refactor are only valuable if the external-justifier vision pans out, so none of it goes to main until the vision is proven end-to-end.

- Branch **`302-then-347`** forks off **review-final #302**, so @mmcilree's commits are the base with his authorship intact. Our refactor goes on top; commits that build on his hint code carry `Co-Authored-By: Matthew McIlree`.
- **All-or-nothing**: the foundation graduates to main as a single unit once validated, or the branch is deleted wholesale if the approach doesn't work out. We deliberately do **not** split the reasons rework out for an early independent merge — it rides with the rest. (#302's own PR/branch survives independently, so binning ours loses nothing of Matthew's.)
- **Graduation gate**: the branch carries {#302 + reasons rework + typed-hint infrastructure + 1–2 constraints fully converted}, verified end-to-end against the VeriPB temp branch. Passing that gate = "the vision works" → graduate the whole foundation to main as one unit. The remaining constraint roll-out (Phase 3) is then ordinary incremental on-main work, no longer throwaway-risk. This bounds the branch's lifetime — no months-long drift against a rebasing main.
- **Do not merge #302 to main on its own.**

## Dependencies / gates

- **#302 must be review-final first** (Ciaran reviewing; minor nitpicks outstanding). The branch forks from the settled foundation, not a moving one.
- **VeriPB**: assertion levels above `Definitions` need @mmcilree's temp branch `temp/annotated-assertions-and-sol-fix` (VeriPB-dev MRs 187/188 — solx semantics). Stock VeriPB fails above `Definitions`. Track landing those upstream; until then, proof-test assertion modes against the temp branch. Default `AssertionLevel::Off` (= current behaviour) verifies on stock VeriPB.

### Phase 0 — Foundation on the branch
- [ ] Fork `302-then-347` off review-final #302.
[...]

Issue #347, opened 2026-06-16
Figure out Reasons (and possibly also `infer`)


ciaranm:
Splitting out a design discussion from the reason/justification rework we've been circling. Two plans below as comments, least to most ambitious:

- **Plan A — Reasons: incremental improvement** (first comment). Replace the `ReasonFunction` thunk — which eagerly materialises a reason vector on *every* `infer()` and is then discarded when proofs are off (~20% of runtime on some benchmarks) — with a cheap, declarative `Reason` variant materialised on demand. Reasons only; doesn't touch justifications; shippable on its own.

- **Plan B — Redesigning `infer()`** (second comment). The ambitious version: treat the whole inference triple (literal / reason / justification) as one declarative, mode-driven mechanism, and generalise the annotated-assertions work so each justification is a per-rule **tag + data** with pure `emit`/`gather` functions, instead of a global `hint_name` enum + flat field vector. Plan A is the reason layer of this, so it's a strict stepping stone — you can do A and stop.

Context / overlap:
- **#302 (annotated assertions)** is the symmetric move on the justification half — it rewrites the same `infer()` seam, so the merge point is one agreed signature. Plan B is essentially "generalise #302's hints into per-rule tags."
- The **#315 / #335** stack (conflict observer + nogood learning) sets the lifetime constraint a contradiction's reason must satisfy; none of it edits `reason.*`/`justification.hh`, so no textual conflict, but Plan A's `_last_contradiction_reason` change has to keep it working.
- Sequencing: I'd land nothing here until **#303** is in.

@mmcilree — the part most relevant to you is in Plan B: the tag-based justification dispatch (replacing the global enum), the worked `plus` / `all_different` examples against the real code, and the observation that `all_different`'s per-constraint `def` includes a *mutable*, lazily-built at-most-one line cache reached via `ConstraintID`. Two things I'd most like your read on: (1) does the tag scheme fit what the external justifier needs, and (2) is it worth adopting the tag mechanism *inside #302 now*, before a flat-field decoder hardens?

Both plans also exist as drafts under `dev_docs/` locally — happy to turn them into a PR instead if you'd rather review them as files.

ciaranm:
# Reasons: incremental improvement (DRAFT for discussion)

**Scope:** improve *only* the reason mechanism. This deliberately does **not**
touch justifications. The justification side — and a unified rethink of the whole
`infer()` triple — is explored separately in Plan B and in PR #302
[...]

Issue #254, opened 2026-06-01
Audit all constraints for empty / all-constant argument edge cases, and add tests


ciaranm:
## Motivation

#248 was an unsound wrong-UNSAT caused by `LinearEquality` mishandling an
**all-constant** term list: `tidy_up_linear` folds constants into a `modifier`,
leaving the variable-term list **empty**, and the empty-sum shortcut had a sign
error (`modifier != _value` instead of `modifier != -_value`). It flagged the
tautology `1*1 == 1` as a contradiction.

That bug was dormant for a long time and only surfaced once #247 let the FlatZinc
frontend pass `ConstantIntegerVariableID`s into linear constraints (a `true`
literal in `bool_lin_eq([1],[true],1)`). MiniZinc models routinely produce
constant/near-constant constraint arguments after flattening, so this class of
edge case is exactly what a MiniZinc Challenge entry will hit — and it's exactly
where our existing data-driven tests are weakest (they almost always build
constraints over *genuine* variables with non-trivial domains).

This is also high-value for us specifically as a **certifying** solver: VeriPB
caught #248 (the bogus contradiction wasn't RUP-derivable), so turning `--prove`
on while fuzzing these cases gives a strong oracle for free.

## What to do

For **every** constraint in `gcs/constraints/`, audit and add test coverage for
two degenerate argument shapes:

1. **All-constant arguments** — every variable replaced by a
   `ConstantIntegerVariableID` (or a singleton-domain variable). The constraint
   should reduce to a simple true/false check: satisfiable iff the constants
   actually satisfy it. Both the tautology and the genuine-contradiction
   direction must be checked (the #248 sign error got the contradiction case
   right by luck).
2. **Empty / minimal collections** — for constraints taking arrays/collections
   (linear sums, `AllDifferent`, `Table`, `Count`, `Element`, `Among`,
   `Circuit`, `Min`/`Max`, `Cumulative`, `Knapsack`, the lex family, ...): the
   empty array, the single-element array, and (where the constant case folds
   terms away) the "all terms cancel / fold out" case.

Mixed cases (some constant, some variable — e.g. one real variable plus a
constant that folds into a modifier) are also worth a row, since they exercise
the normal propagation path with a non-zero modifier.

### Testing approach

- Prefer extending each constraint's existing data-driven test with constant /
  singleton-domain / empty rows, run with proofs on so VeriPB validates them.
  Where that doesn't fit, add a small focused test in the style of the new
[...]

PR #378, opened 2026-06-19
Reasons + infer rework (Plan A + Plan B) on the 302-then-347 integration branch [#377]

ciaranm:
**Draft — review-as-you-go.** The Reasons + `infer` rework. Implementation plan: #377. Design: #347 (Plan A = `reasons-improvement.md`, Plan B = `infer-redesign.md`; both docs now carry an **IMPLEMENTED** note describing the as-built result).

This is the **all-or-nothing `302-then-347` integration branch**: it forks off @mmcilree's #302 (annotated assertions) and carries the whole reason/justification rework on top. **Please don't merge this to `main` on its own** — per #377 the foundation graduates as one unit at the validation gate, or the branch is binned. The base is set to `annotated-assertions` so this PR's diff shows **only the rework commits**; I'll retarget to `main` once #302 is review-final and we reach the gate.

The inference triple — *what* is inferred, its *reason*, its *justification* — is now one declarative, mode-driven mechanism. Three layers:

## 1. Reason layer (Plan A) — the G1 perf win

The eager `ReasonFunction = std::function<()->ReasonLiterals>` is gone. A reason is now a declarative `Reason` variant (`NoReason` / `ExplicitReason` / `{Generic,BothBounds,Lazy}ReasonOver` + `Narrowable*`), cheap to build; `materialise(Reason, State)` is the single place a reason walks domains, deferred into the inference tracker.

- `snapshot_reason` in the tracker pins the eager-vs-lazy *timing*: eager reasons snapshot pre-inference into an `ExplicitReason`, lazy ones materialise post-inference. This is why every step is **byte-identical**.
- **G1:** proofs-off never materialises (no logger → no walk); and the eager per-inference `std::function` is gone — `ReasonFunction`, its `LegacyReasonFunction` bridge, and the `Reason(F&&)` ctor are deleted. (One opt-in `LazyReasonFn = std::function` escape hatch remains, used only by the genuinely-lazy `LazyReasonOver` reasons — e.g. the gcc flow-cut and all_different's Hall justification.)
- Unified signature: `infer(lit, Justification, Reason, optional<AssertionAnnotation>)` — the merge point with #302's annotation param.

## 2. Typed assertion hints (Plan B, hint half) — enums retired, names decentralised

- The two prototype enums (`AssertionHintName`, `AssertionHintIdentifier`) are gone. The second annotation field (`hint_name`) is now a coarse, model-level `std::string_view`; field keys are plain `SExpr::atom` literals.
- Coarse names are **decentralised**, not gathered into one header: each typed witness carries its own `static constexpr std::string_view hint_name` (e.g. `AllDifferentHall::hint_name = "all_different"`), and inline-closure call sites wrap a local `constexpr` name in `hints::ModelName{...}` (the minimal name-only hint tier). The fine, per-shape `justifier` keyword likewise stays decentralised on each witness (e.g. `AllDifferentHall::justifier = "hall_set_or_violator"`), reached by ADL — `gcs::innards::hints` (reopened per constraint) *is* the enumerable justifier-interface surface.
- `all_different`'s Hall witness is a typed `AllDifferentHall` struct owning its wire schema via one `hint_sexpr` overload.

## 3. Justification dispatch (Plan B, core) — one vocabulary, strategy by mode

- The explicit-justification form is `JustifyExplicitly<Emit, Hint>`, which deliberately **mirrors** `JustifyUsingRUP<Hint>`: both are parameterised on an optional typed `Hint`, and the hint means the same thing in both — a typed assertion annotation, **orthogonal** to *how* the inference is proved (RUP vs explicit steps). This replaced and **deleted** `JustifyByData` / `JustifyExplicitlyOnly` / `JustifyExplicitlyThenRUP`; all ~89 existing explicit closures moved over unchanged.
- **`emit`** is either an inline `(const ReasonLiterals &) -> void` closure (the common case) or a named **fat-witness struct** providing an `emit_justification` overload by ADL — the latter for a reification verdict, whose return type must *name* the justification (a lambda type can't be named). `emit_explicitly()` dispatches the two at compile time (`if constexpr (requires { emit(reason); })`).
- **`then_rup`** is now a mandatory `ThenRUP{Yes, No}` enum, not a bare `bool`: `Yes` RUPs the inferred literal after the explicit steps, `No` lets the steps conclude it themselves. Mandatory and named at every call site, so the proof shape is legible rather than hidden behind an unlabelled `true`/`false`.
- `JustifyExplicitly` is **not** a `Justification`-variant alternative — a templated type can't be type-erased into the variant, and reification verdicts need to name it — so it is dispatched by **overload resolution**. The variant (`JustifyUsingRUP<NoHint>` / `AssertRatherThanJustifying` / `NoJustificationNeeded`) carries only the plain, nameless forms; a `JustifyUsingRUP<Hint>` carrying a typed hint is likewise its own templated overload.
- `ProofLogger::infer` and the explicit back end `infer_explicitly()` dispatch by proof mode: **eager** runs `emit`, then RUPs the inference iff `ThenRUP::Yes` (byte-identical to the old closures); **assert** resolves the annotation by precedence (the explicit `hint` if it advertises anything → else the witness's own `hint_sexpr`/`hint_name` → else the annotation passed to `infer()`) and never runs `emit`.
- **Two worked examples**, the design's opposite ends: `all_different` GAC (both Hall shapes; `emit` ignores the reason, captures per-constraint state) and `plus` (`emit` reads the reason positionally; `PlusBound` + `emit_justification` is the witness-driven-emit example).
- **No logger-side store.** An earlier draft added a `ConstraintID`-keyed `constraint_proof_data` store so `emit` could reach a constraint's mutable proof state (all_different's at-most-one cache) by owner; it was removed as over-engineering. That cache is a `shared_ptr` member of the `Constraint`, which outlives the whole search, so `emit` simply **captures** it — valid even for future backtrack-time replay. The witness stays the small, serialisable thing `hint_sexpr` turns into the wire form.

## Verification

- **Byte-identical** proofs against a deterministic probe at every commit (covers all_different GAC Hall, linear, count, element, plus, abs, comparison).
- Full suite **361/361** (Release) and **261/261** (Sanitize, ASan+UBSan clean) on stock VeriPB 3.0.2.
- **End-to-end under assertions:** at `GCS_ASSERTION_LEVEL=Definitions`, crystal_maze / sudoku / langford verify (`s UNDER ASSERTIONS …`) with 118 / 651 / 463 typed Hall hints, on the `temp/annotated-assertions-and-sol-fix` VeriPB branch (stock VeriPB can't go above `Definitions`, so this isn't wired into ctest yet).

## Deliberately deferred

Per @mmcilree's "the fewer hints, the better — expand when the Justifier needs them":

- Adding *typed* witnesses (with `hint_sexpr`) to more constraints. Every constraint is now on `JustifyExplicitly`, so this is purely additive per constraint as the external Justifier dictates what it needs.
- The `gather`-deferral so the witness/closures aren't built in proofs-off mode (the remaining proofs-off cost is closure/witness construction, not a domain walk).
- The defer modes themselves — `AssertExternal` (serialise the witness to the wire) and `LazyInternal` (park the typed witness + replay at backtrack), with the `Deferrable`-concept tier-3 fallback. The witnesses are *built for* these (serialisable via `hint_sexpr`, replayable via constraint-owned captures), but the modes are future work.
- Owner-derived coarse `hint_name`; capability tiers beyond witness-complete.

## Credit

The branch base is @mmcilree's #302 commits, with his authorship intact; commits building on his hint code carry his co-author line. cc @mmcilree


ciaranm:
@mmcilree I'm not sure whether this is totally perfect or not but I think it's in a plausible state. In particular, 4c85dfe8 shows you what to do to convert over one constraint to use assertions. This might be a "discuss on Tuesday" thing.

Currently this branch is on top of yours, so it's a "merge them both or not" decision.

**Update:** since posting I've done a cleanup pass, and the branch now builds warning-free on both GCC and clang. The compiler warnings the rework introduced are fixed:
- **23 unused-parameter warnings** (`ae40a69a`) — the reified-dispatch lambdas and justification builders no longer touch `state`/`logger` (they return a typed witness + `Reason` now), so those interface-mandated params are anonymised.
- **6 GCC `-O3 -Wmaybe-uninitialized` false positives** in `element` (`a2a32f87`) — SRA picking apart `ReasonLiterals`' small-buffer storage; there's no real uninitialised read, so it's a narrow GCC-guarded `#pragma` (clang doesn't warn).

Also folded in the cheap, definitely-correct findings from the auditor pass (`763ed667`) and brought `infer-redesign.md` up to date (`fb314eb7`). Everything stays byte-identical to golden.

ciaranm:
## Auditor pre-merge checklist — Reasons + `infer` rework

Independent code-quality pass over the **rework-only** diff (`annotated-assertions..302-then-347`, 63 files, +3580/−1538) at `4c85dfe`. #302's foundation is out of scope. Mechanical sweep clean — no debug prints, commented-out code, `#if 0`, or env/ablation knobs (`ProofLevel::Temporary` is a legitimate VeriPB level). **No `bug`-class defects — all `code cleanup`.** Tiered worst-first; `[scope]` = local / cross-cutting.

### 🔴 Must fix before merge
- [ ] **`tmp/issue254_audit.md`** — accidentally-committed dev journal for the *unrelated* issue-#254 audit (swept in via `d018f0f8`); not gitignored, referenced nowhere. Delete. `[local]`

### 🟠 Serious — shipped misinformation (cheap)
- [ ] **`abs.cc:58`** — "names mirror the `justify_abs_*` helpers" is false (7 justifier strings vs 6 helpers, no 1:1; `v1_no_preimage` maps to nothing). Fix the comment or align the names. `[local]`
- [ ] **`equals.cc:81-82`** — tautological "Replaces `JustifyUsingRUP{hints::EqualsRUP{owner}}` with `JustifyUsingRUP{hints::EqualsRUP{owner}}`" (before == after). `[local]`

### 🟡 Moderate
- [ ] **`plus.cc` / `minus.cc`** — twin divergence: `plus` on typed witness `hints::PlusBound`, `minus` left inline; logically identical bodies incl. a verbatim-copied #166 comment block. `[cross-cutting]`
- [ ] **`reason.hh` `Narrowable{Generic,BothBounds,Lazy}ReasonOver`** — zero construction sites; 3 structs + 3 ctors + 3 duplicated `materialise()` arms, all inert. Speculative; drop until the deferred tracker lands. `[cross-cutting]`
- [ ] **`generic_reason` / `bounds_reason`** (reason.hh/.cc) — unused `const State &` ("retained only for call-site compatibility while the migration is in progress"); 39 sites pass `state` to a function that discards it. `[cross-cutting]`
- [ ] **`inference_tracker.hh`** — literal-recording visitor triple-duplicated; `record_firing_inference()` exists but both `track_impl` bodies re-inline it instead of calling it. `[local]`
- [ ] **`all_different/justify.hh`** — `AllDifferentForcedValue` emits a structured `hint_sexpr` but carries no `justifier`; sibling `AllDifferentHall` does. Irregular wire schema for the Justifier. `[local]`
- [ ] **abs justifier strings** — 7 discriminator literals (several repeated across 11 call sites), no enum/constant source of truth (the deleted enum world prevented this drift). `[local]`
- [ ] **`knapsack.cc:254-535`** — `eager_reason(generic_reason(state, reason_variables), state)` written out 27× in one function; hoist it. `[local]`
- [ ] **`equals.cc:48`** — `// PROTOTYPE:` marker on a shipped `EqualsRUP` witness used at ~8 call sites. `[local]`

### 🟢 Mild
- [ ] **`sort.cc` / `arg_sort.cc`** — `reason_fn` parameter name stale (type is now `const ReasonLiterals &`, not a function). `[local]`
- [ ] **`equals.cc:56`, `lex.cc:284`, `linear_inequality.cc:81`** — "Was an inline `JustifyByData` closure" comments reference a deleted type. `[cross-cutting]`
- [ ] **`equals.cc`** — `"equals"` hard-coded 3× despite the `equals_hint` constant; copy-and-append IIFE `ExplicitReason{[&]{auto r=reason; r.emplace_back(…); return r;}()}` ×7 (wants a `reason_plus()` helper). `[local]`
- [ ] **`equals.cc:173`** — dead unnamed `ProofLogger*` param in `no_overlap_justification`; caller still passes `logger`. `[local]`
- [ ] **`hint_sexpr` bodies** — `("constraint_id" owner)` / `("justifier" …)` fragments copy-pasted across 4+ witnesses; no shared `hint_constraint_id()` / `hint_justifier()`. `[cross-cutting]`
- [ ] **`reason.hh` `Reason`** — 10 hand-written converting ctors where a constrained template + the `ReasonLiterals` overload would do. `[local]`
- [ ] **`AssertRatherThanJustifying`** — pointless `explicit` nullary ctor with `{}` body; inconsistent with sibling empty-tag structs. `[local]`
- [ ] **`inference_tracker.hh`** — same call-site-annotation arg named `assertion_hints` in plain overloads, `fallback` in hint-carrying ones; `assertion_hints` is plural for a single `optional`. `[local]`
- [ ] **`eager_reason()`** — one-line pass-through to `materialise()`; two public names for one operation. `[local]`
- [ ] **PR body** — "no `std::function` on the reason path at all" is contradicted by `LazyReasonFn = std::function<…>` (LazyReasonOver, gcc flow-cut, all_different Hall). `[doc]`
- [ ] **`infer-redesign.md`** — title still "(DRAFT — ambitious, exploratory)" though it now leads with an authoritative IMPLEMENTED section. `[doc]`
- [ ] **`encoding.cc`** — includes `assertion_hints.hh` but uses nothing from it. `[local]`
- [ ] **Half-migration** — typed hints in ~7/40 constraints; the new assertion-hint path has no ctest coverage (stock VeriPB caps at `Definitions`), so behaviour preservation rests entirely on the byte-identical proofs. `[cross-cutting, by design]`

---
<sub>Auditor-generated — independent code-quality pass against `302-then-347` @ `4c85dfe871c5` (rework-only diff vs `annotated-assertions`). Quality findings only (`code cleanup`); no correctness defects found. Posted per the repo's AI-contribution policy.</sub>

CLAUDE.md
```diff
@@ -168,8 +168,8 @@ Each constraint in `gcs/constraints/` follows this pattern:
 
 Every inference must be accompanied by a justification:
 - `NoJustificationNeeded` — assertion without proof
```
mmcilree, line 170: "assertion without a proof" is maybe a confusing description now we have annotated assertions. We don't even bother to assert this in the proof if NoJustificationNeeded is used.

gcs/constraints/abs.cc
```diff
@@ -45,6 +45,42 @@ using fmt::format;
 using fmt::print;
 #endif
 
+namespace gcs::innards::hints
+{
+    // The typed assertion-hint witness for *every* Abs inference -- this is the
```
mmcilree, line 50: Tooo looooooooooooong

gcs/constraints/abs.cc
```diff
@@ -45,6 +45,42 @@ using fmt::format;
...
+    // *cannot* re-derive a value adds a typed field here and serialises it in
+    // hint_sexpr -- see all_different's AllDifferentHall for the multi-field shape.
+    //
+    // One witness serves both proof methods: it rides JustifyUsingRUP for abs's
+    // pure-RUP pruning and JustifyExplicitly for the explicit-steps bounds, because
+    // the assertion-hint axis is orthogonal to *how* the inference is proved.
+    struct AbsJustification
+    {
+        ConstraintID owner;
+        std::string_view justifier;
```
mmcilree, line 74: On the one hand it's good to have this here for demonstration purposes, on the other hand, I'm worried claude will assume this more granular hint will always be needed, which it shouldn't be in general. I think (without thinking too much about it) that there's actually a good chance all the abs justifications can be dealt with by one justification function that just reads the literals and dispatches that way without needing any more additional hints.

gcs/constraints/abs.cc
```diff
@@ -45,6 +45,42 @@ using fmt::format;
...
+    // No operand data travels in the hint: each abs bound is re-derivable from the
+    // asserted literal + the reason + the definition lines, so carrying it would be
+    // redundant ("the fewer hints, the better"). A constraint whose Justifier genuinely
+    // *cannot* re-derive a value adds a typed field here and serialises it in
+    // hint_sexpr -- see all_different's AllDifferentHall for the multi-field shape.
+    //
+    // One witness serves both proof methods: it rides JustifyUsingRUP for abs's
+    // pure-RUP pruning and JustifyExplicitly for the explicit-steps bounds, because
+    // the assertion-hint axis is orthogonal to *how* the inference is proved.
+    struct AbsJustification
```
mmcilree, line 71: I would prefer a naming convention for these like <Constraint>AssertionAnnotation or <Constraint>AssertionHints. I wouldn't call them "Justification" - we already have Justification types, and I would avoid any and all references to "witnesses" (claude seems to like calling them these), both to avoid confusion with redundance witnesses, and to avoid implying that these always have to hold the same info as the internal justification functions.
___
...coming back to this comment after reading more of the code, I now realise the typed hint structs and the internal justification parameters are somewhat  intertwined, so I think this is a wider design problem

gcs/constraints/abs.cc
```diff
@@ -117,7 +153,7 @@ auto Abs::install_propagators(Propagators & propagators) -> void
     // is constant we bail out -- the propagator's per-value loop will
     // discover any UNSAT.
     propagators.install_initialiser(
-        [v1 = _v1, v2 = _v2,
+        [v1 = _v1, v2 = _v2, owner = constraint_id(),
```
mmcilree, line 156: (opinionated) Is "owner" the right name for this? I'm not sure the constraint "owns" the inferences in any obvious way? Maybe "originator"?

gcs/constraints/all_different/justify.hh
```diff
@@ -16,6 +21,58 @@ namespace gcs::innards
...
+    namespace hints
+    {
+        /**
+         * \brief Witness for the GAC all_different "Hall set or violator"
+         * justification: the implicated variables, the values they collectively
+         * saturate, and the owning constraint. Shared by the matching-too-small
+         * contradiction and the SCC Hall-set deletion (same shape, different
+         * owner and call site).
+         *
+         * The two trailing pointers are the per-constraint emit context (the full
```
mmcilree, line 38: loooooooooooong

gcs/constraints/all_different/justify.hh
```diff
@@ -16,6 +21,58 @@ namespace gcs::innards
...
+        auto emit_justification(ProofLogger & logger, const AllDifferentHall & hall, const ReasonLiterals & reason) -> void;
+
+        /**
+         * \brief Witness for the trivial SCC deletion: a value is deleted because a
+         * single other variable is forced to take it. The justification is plain
+         * RUP (no explicit steps), so this hint has no emit_justification — it is
+         * the pure-RUP capability tier, carrying only an assertion hint with the
+         * owning constraint.
+         */
+        struct AllDifferentForcedValue
```
mmcilree, line 68: I guess this is fine for now, but I'm not a massive fan of this layout. a) There's nothing specific about ForcedValue for this hint so I don't think it should be named like that b) the struct's fields are a strict subset of the larger AllDifferent hall. Could we do hints by composition? E.g. AllDifferentBasicAnnotation is contained within AllDifferentHallAnnotation?

gcs/constraints/among.cc
```diff
@@ -153,54 +153,53 @@ auto Among::install_propagators(Propagators & propagators) -> void
...
-                        if (holds_alternative<ConstantIntegerVariableID>(m))
-                            continue;
-                        b.add(logger->names_and_ids_tracker().need_constraint_saying_variable_takes_at_least_one_value(m));
-                    }
-                    b.emit(*logger, ProofLevel::Temporary);
-                }
-            }},
+            auto vars_reason = eager_reason(generic_reason(state, vars), state);
+            inference.infer(logger, how_many >= must_match_count, JustifyExplicitly{[&](const ReasonLiterals &) -> void {
+                                                                                        // Combine the (sum <= how_many) half of the Among encoding with the
```
mmcilree, line 158: X-rated clang-format gore strikes again...

gcs/constraints/equals.cc
```diff
@@ -42,18 +42,67 @@ using std::print;
...
+                logger.emit_rup_proof_line(WPBSum{} + 1_i * (w.v1 != val) + 1_i * (w.v2 == val) + 1_i * ! w.cond >= 1_i, ProofLevel::Temporary);
+            else
+                logger.emit_rup_proof_line(WPBSum{} + 1_i * (w.v2 != val) + 1_i * (w.v1 == val) + 1_i * ! w.cond >= 1_i, ProofLevel::Temporary);
+    }
+
+    // The pure-RUP hint for "a value is forced out of the other operand because this
+    // one is fixed to it" (the bare-NotEquals propagator). RUP-derivable, so no
+    // emit_justification; carries the owning constraint id so assertion mode can name
+    // the source.
+    struct EqualsRUP
```
mmcilree, line 82: Same comment as for AllDifferent, this hint is a subset of the other one.

gcs/constraints/linear/linear_equality.cc
```diff
@@ -64,6 +64,31 @@ ReifiedLinearEquality::ReifiedLinearEquality(WeightedSum coeff_vars, Integer val
 
 namespace
 {
+    // Coarse model-level hint name for linear_equality's inferences, carried on the
+    // raw annotation passed into propagate_linear. The pure-RUP reified inferences
+    // use hints::LinearEqualityRUP instead.
+    constexpr std::string_view linear_equality_hint = "linear_equality";
```
mmcilree, line 70: This isn't in namespace hints... :-(

gcs/constraints/minus.cc
```diff
@@ -61,31 +61,31 @@ namespace
...
-                        if (holds_alternative<ConstantIntegerVariableID>(lit.var))
-                            continue;
-                        b.add_for_literal(logger->names_and_ids_tracker(), lit);
-                    }
-
-                    b.emit(*logger, ProofLevel::Temporary);
-                }};
+        auto justify = [&](Conclude c) {
+            return JustifyExplicitly{[c, sum_line, logger](const ReasonLiterals & reason) {
+                                         auto sum_line_value = (c == Conclude::LE ? sum_line.first : sum_line.second);
```
mmcilree, line 66: Not as bad, maybe 12A-rated clang-format

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -714,7 +714,7 @@ auto NamesAndIDsTracker::need_gevar(SimpleOrProofOnlyIntegerVariableID id, Integ
                 return;
 
             ProofRule assert_or_rup = _imp->logger->get_assertion_level() == AssertionLevel::Links ? ProofRule(AssertProofRule{}) : ProofRule(RUPProofRule{});
-            auto annotation = AssertionAnnotation{.hint_name = AssertionHintName::InitialBound};
+            auto annotation = AssertionAnnotation{.hint_name = "initial_bound"};
```
mmcilree, line 717: Hmm... I was promised that the full justifier API would at least live in one namespace, but these are again magic strings not contained anywhere else.

gcs/innards/proofs/proof_logger.cc
```diff
@@ -222,7 +222,7 @@ auto ProofLogger::solution(const vector<pair<IntegerVariableID, Integer>> & all_
             optional_minimise_variable_and_value->first);
     else if (_imp->assertion_level > AssertionLevel::Definitions) {
         // solx and no links => have to assert the blocking constraint
-        emit(AssertProofRule{}, blocking_sum >= 1_i, ProofLevel::Top, AssertionAnnotation{.hint_name = AssertionHintName::SolxBlock});
+        emit(AssertProofRule{}, blocking_sum >= 1_i, ProofLevel::Top, AssertionAnnotation{.hint_name = "solx_block"});
```
mmcilree, line 225: Likewise, more magic strings

gcs/innards/assertion_hints.hh
```diff
@@ -175,7 +114,12 @@ namespace gcs::innards
     struct AssertionAnnotation
     {
         std::vector<ProofLineLabel> derivable_from = {};
-        AssertionHintName hint_name = AssertionHintName::None;
+        // A coarse, model-level constraint-type name (e.g. "all_different",
+        // "abs"), carried in the second annotation field for vanilla-VeriPB
+        // statistics. The fine, per-justification-shape discriminator (the
```
mmcilree, line 119: This comment isn't quite right: the vanilla-VeriPB statistics possibility is why the "coarse" hint gets it's own colon-separated field, but I'm not thinking of it as the sole use for this field. I'm expecting the name to be the first port-of-call for justification dispatch, with the additional hints being read only when constraint-family alone is not enough to decide which justifier function to use.

gcs/constraints/equals.cc
```diff
@@ -42,18 +42,67 @@ using std::print;
...
+}
+
+namespace gcs::innards::hints
+{
+    // Witness for the "domains don't overlap, so the reified equality cannot hold"
+    // justification, carried in a reified verdict (built by no_overlap_justification).
+    // Holds the operands, the bounds
+    // walked over, and the condition; the State pointer is valid because the verdict
+    // is consumed synchronously while the constraint is live.
+    struct EqualsNoOverlap
```
mmcilree, line 60: I'm not a fan of the hard-coupling of assertion hints and internal justifier parameters here. The hints namespace at the moment doesn't allow me to cleanly extract or define a justifier API, since it's mixed in which stuff needed for the internal justification functions, which may or may not end up being the same thing.

gcs/constraints/equals.cc
```diff
@@ -42,18 +42,67 @@ using std::print;
 using fmt::print;
 #endif
 
+namespace
+{
+    // Coarse model-level hint name for equals's explicit-steps (JustifyExplicitly,
+    // including the EqualsNoOverlap fat witness) and the reified-verdict fallback
+    // annotation. The pure-RUP inferences use hints::EqualsRUP instead.
+    constexpr std::string_view equals_hint = "equals";
```
mmcilree, line 50: Not sure why this is separate.

gcs/constraints/equals.cc
```diff
@@ -42,18 +42,67 @@ using std::print;
...
+    // Holds the operands, the bounds
+    // walked over, and the condition; the State pointer is valid because the verdict
+    // is consumed synchronously while the constraint is live.
+    struct EqualsNoOverlap
+    {
+        const State * state;
+        IntegerVariableID v1, v2;
+        std::pair<Integer, Integer> v1_bounds;
+        Literal cond;
+        static constexpr std::string_view hint_name = "equals";
```
mmcilree, line 66: There's currently nothing constraining multiple hints on the same constraint type to have the same `hint_name`. It'd be very easy for EqualsNoOverlap to use "equals" and EqualsRUP to use "equal", for example. 

This maybe links to the struct containment point I made elsewhere.

gcs/constraints/element.cc
```diff
@@ -215,6 +215,16 @@ auto NDimensionalElement<EntryType_, dimensions_>::define_proof_model(ProofModel
...
 }
 
+// GCC's -O3 scalar-replacement of aggregates picks apart the small-buffer storage
+// of ReasonLiterals (gch::small_vector<ProofLiteralOrFlag, 2>) used for the reason
+// literals below, then cannot prove one of the resulting temporaries initialised,
+// emitting a -Wmaybe-uninitialized false positive (attributed to the first
+// install() lambda). The reason locals are all default-constructed before use --
+// there is no real uninitialised read -- and clang's analysis does not warn.
+#if defined(__GNUC__) && ! defined(__clang__)
+#pragma GCC diagnostic push
```
mmcilree, line 225: Not sure what's going here but I'll assume this is fine? Seems a little hacky?

gcs/constraints/linear/linear_equality.cc
```diff
@@ -310,11 +335,11 @@ auto ReifiedLinearEquality::install_propagators(Propagators & propagators) -> vo
...
+                    propagators.install(constraint_id(), [sanitised_cv = sanitised_cv, value = _value + modifier, cond = _reif_cond, proof_line = _proof_line, all_vars = move(all_vars), owner = constraint_id()](const State & state, auto & inference, ProofLogger * const logger) -> PropagatorState { // This comment is needed to stop clang-format exploding
                         return overloaded{
                             [&](const evaluated_reif::MustHold & reif) {
                                 // we now know the condition definitely holds, so it's a linear equality
-                                return propagate_linear(sanitised_cv, value, state, inference, logger, true, proof_line, reif.cond, AssertionAnnotation{.hint_name = AssertionHintName::ReifiedLinearEquality});
+                                return propagate_linear(sanitised_cv, value, state, inference, logger, true, proof_line, reif.cond, AssertionAnnotation{.hint_name = linear_equality_hint});
```
mmcilree, line 342: How will this one work if we need to pass finer grained hints in here? Doesn't this still rely on eager materialisation of the hints?

mmcilree (changes requested):
Okay, overall I like the idea of having typed hints per constraint and I'm sure this is improving on the Reasons allocation front, but I'm entirely convinced by this implementation yet:
1. The string view names are scattered and repeated in multiple places, and some of them don't live in the hints namespace 
2. There's various bits of repetition in the typed hint structs, and I'm worried that the hint specifications and the internal justification parameters are too tightly coupled, in a way that will make it awkward to try to use a different justifier method externally than what is used internally. 
3. Some long comment and clang-format nitpicks

ciaranm:
Superseded by #394, now merged to `main`. #394 carries this whole reasons/infer rework (Plan A + Plan B), reconciled onto the post-#393 restructure — `63fbe2cd` (declarative `Reason` + unified `infer` + `JustifyExplicitly`/`JustifyUsingRUP`) and `e4593e89` (typed per-constraint hints) — on top of @mmcilree's #302, which landed as its own authored commit. Closing in favour of #394.

Issue #384, opened 2026-06-22
Disjunctive: support variable durations (follow-up to #146)


ciaranm:
## Background

The 1D `Disjunctive` constraint currently requires **constant** durations
(`std::vector<Integer> lengths`). Its sibling constraints already accept variable
sizes: `Cumulative` (variable lengths/heights/capacity) and `Disjunctive2D`
(variable widths/heights, added in #260). This issue covers giving 1D
`Disjunctive` the same variable-duration capability, implemented **natively in
`Disjunctive`** (not by routing to `Cumulative`, so the strict zero-length
semantics and the leaner bridge/`recover_am1` proof are preserved).

Follow-up to #146 (which shipped the constant-length propagator and explicitly
parked variable lengths). The XCSP frontend stub at
`xcsp/xcsp_glasgow_constraint_solver.cc:489` currently reports
`variable-length Disjunctive not yet supported (#146)` and should be re-pointed
here and then lifted.

## Why it's well-bounded

1D `Disjunctive` is structurally `Cumulative` specialised to heights = 1,
capacity = 1, and is one axis of `Disjunctive2D`. Both already solve the only
hard part — reifying `after` on a proof-only `end = s + l` and cancelling `end`
back to `s + l` in the precedence-flag `pol`. 2D's `Lpol` helper
(`disjunctive_2d.cc:469`) is literally 1D's `pair_ne` step plus one extra
`.add(*end_le)`. Estimated ~340-430 net core LOC + tests + frontend, comparable
to #260.

## Design decisions

- Mirror the `Cumulative`/`Disjunctive2D` snapshot idiom: `_lengths` becomes
  `vector<IntegerVariableID>`; `_length_vals[i]` holds the constant (0 placeholder
  for a variable), `_length_ub[i]` the root upper bound; resolved in `prepare()`.
- `after` reifies on a proof-only `end = s + l` when the length varies (single
  variable keeps the pin RUP-friendly), capturing both `end >= s+l` and
  `end <= s+l` (the latter cancels `end` in the `pair_ne` pol). Constant lengths
  keep the current `s >= t-l+1` form byte-for-byte.
- Strict vs non-strict diverge only in the OPB clause (as in 2D): strict needs no
  zero flag (the pairwise before-clause already enforces zero-length-inside);
  non-strict adds a reified `size <= 0` escape disjunct.
- `lb >= 0` non-negativity guard for variable durations at `prepare()` time.

## PR stack

[...]

PR #394, opened 2026-06-23
Reasons/infer rework + annotated assertions + typed per-constraint hints

ciaranm:
**Draft — review-ready.** Reasons/infer rework + annotated assertions (#302) + typed per-constraint hints, integrated onto `main` (after the per-constraint directory restructure, #393).

## What this is

The whole reasons/justification body of work as one branch:
- the declarative `Reason` rework + unified `infer` signature + `JustifyExplicitly` / `JustifyUsingRUP` (the core machinery);
- Matthew's annotated-assertions machinery (#302), carried here with his authorship;
- the new **typed per-constraint assertion hints** answering the #378 review.

It is `302-then-347` reconciled onto restructured `main` via a single `git merge` (rename detection followed the #393 moves; the only manual conflicts were `equals.hh` and `disjunctive{,_2d}.cc`).

## The hint design

A typed assertion hint is a small struct in `gcs::innards::hints`, one per constraint, living in `gcs/constraints/<foo>/hints.hh`. It **defaults to emitting only its identity** — `(constraint_id <originator>)`, plus `(subhint <name>)` where the struct names one — via a single generic `hint_sexpr`. It carries **no operand data**: anything re-derivable stays off the wire (the internal proof writer holds it for its own emission). A constraint opts into serialising more by defining its own `hint_sexpr` next to its struct — and that boundary, not the struct's fields, is where an external justifier's data is allowed to be a (possibly strict) subset of the internal data.

Base + derived by inheritance: a base struct carries `originator` + `hint_name` (single source, inherited so siblings can't drift); derived structs add a `subhint_name` and/or held data only where the internal writer distinguishes shapes.

**Reference conversions to look at first:** `abs` (collapses to a single base `hints::Abs{originator}` — the 7 justifier strings gone) and `all_different` (base `AllDifferent` for the forced-value deletions + `AllDifferentHall : AllDifferent` for the GAC Hall shape).

## Scope

- **Every constraint with its own inferences carries a hint** (~32). The minimal ones are just `{originator}` + `hint_name`, threaded into every justification (and through helper functions where the inference lives in one — e.g. circuit's SCC explore chain, sort's Mehlhorn-Thiel propagator, mult_bc's recursive `filter_quotient`).
- The infer family is uniformly hintable: `infer` / `infer_equal` / `infer_not_equal` / `infer_less_than` / `infer_greater_than_or_equal` / `infer_not_in_range` / `infer_all` / `contradiction`, in both `JustifyUsingRUP<Hint>` and `JustifyExplicitly<Emit,Hint>` forms (two overloads — `contradiction` and `infer_not_in_range` — were added to close gaps the rollout exposed).
- Proof-machinery names (`initial_bound`, `bound_link`, `backtrack`, `solx_block`, `soli_improve`) and the s-expr field keys (`constraint_id`, `subhint`) live in `gcs/innards/proofs/hints.hh` — so the whole justifier-facing wire vocabulary is extractable from `gcs::innards::hints` rather than scattered as string literals.

### Deliberate exceptions
- `arithmetic` and `seq_precede_chain` post no inferences of their own (they delegate to Table / ValuePrecede and inherit those hints).
- `table`'s `Table` hint is **infrastructure-only**: Table's propagation runs through the shared *innards* helper `propagate_extensional()`, which can't name a constraint-specific hint without an innards→constraints layering inversion (documented on the struct). When that util is made hint-aware, Table is ready.
- Reification-verdict `JustifyUsingRUP{}` stores and `install_initial_contradiction` sites stay bare — they hold the plain `Justification` variant.

## Verification

Normal-mode (proofs-off) output is **byte-identical** throughout — hints emit only in assertion mode. Release **396/396** + Sanitize (ASan/UBSan) **293/293** at each step; assertion-mode wire spot-checked across the set (`:foo:((constraint_id _N))`, plus `(subhint …)` where applicable).

## History

Rewritten to a clean, logical four-commit series (was the original 62-commit history + the merge + the hint commits). Each commit builds independently (bisectable), and the final tree is byte-identical to the reviewed state:

1. **Annotated assertions** (#302) — *authored by @mmcilree* — `AssertionLevel`, `AssertionAnnotation`, assertion-mode proof logging.
2. **Reasons/infer rework** — declarative `Reason`, unified `infer`, `JustifyExplicitly` / `JustifyUsingRUP`.
3. **Typed per-constraint assertion hints**.
4. **Review cleanup** — lambda hoisting, dead-code removal, hint-header re-split.

@mmcilree's 22 original #302 commits couldn't replay verbatim onto restructured `main` (they collide on the #384-rewritten `disjunctive{,_2d}.cc`), so commit 1 is a faithful squash of them onto restructured `main`, keeping his authorship; the few superseded disjunctive annotations are resolved to the base side and the correct ones land in the later commits.

## Not yet done (intentionally)
- **Assertion-mode e2e** verification needs your VeriPB branch (stock 3.0.2 caps at Definitions), so it isn't in ctest.

### DIFF
--- a/examples/cake/cake.cc
+++ b/examples/cake/cake.cc
@@ -92,8 +92,9 @@ auto main(int argc, char * argv[]) -> int
                 variable_order::dom_then_deg(vector<IntegerVariableID>{banana, chocolate}),
                 value_order::largest_first())},
         options_vars.contains("prove")
-            ? make_optional<ProofOptions>(ProofFileNames{options_vars["proof-files-basename"].as<string>()},
-                  true, options_vars.count("full-proof-encoding"))
+            ? make_optional(options_vars.count("full-proof-encoding")
+                      ? ProofOptions{ProofFileNames{options_vars["proof-files-basename"].as<string>()}}.enable_full_encoding()
+                      : ProofOptions{ProofFileNames{options_vars["proof-files-basename"].as<string>()}})
             : nullopt);
 
     if (options_vars.contains("stats"))
--- a/examples/cake/cake.cc
+++ b/examples/cake/cake.cc
@@ -92,9 +92,7 @@ auto main(int argc, char * argv[]) -> int
                 variable_order::dom_then_deg(vector<IntegerVariableID>{banana, chocolate}),
                 value_order::largest_first())},
         options_vars.contains("prove")
-            ? make_optional(options_vars.count("full-proof-encoding")
-                      ? ProofOptions{ProofFileNames{options_vars["proof-files-basename"].as<string>()}}.enable_full_encoding()
-                      : ProofOptions{ProofFileNames{options_vars["proof-files-basename"].as<string>()}})
+            ? make_optional(ProofOptions{ProofFileNames{options_vars["proof-files-basename"].as<string>()}}.enable_full_encoding(options_vars.contains("full-proof-encoding")))
             : nullopt);
 
     if (options_vars.contains("stats"))
--- a/examples/crystal_maze/crystal_maze.cc
+++ b/examples/crystal_maze/crystal_maze.cc
@@ -28,8 +28,7 @@ using namespace gcs;
 
 using std::cerr;
 using std::cout;
-using std::make_optional;
-using std::nullopt;
+using std::optional;
 using std::pair;
 using std::string;
 using std::to_string;
@@ -104,11 +103,14 @@ auto main(int argc, char * argv[]) -> int
         p.post(LinearEquality{WeightedSum{} + 1_i * xs[x1] + -1_i * xs[x2] + -1_i * diffs.back(), 0_i, options_vars.contains("gac")});
     }
 
-    auto proof_options = options_vars.contains("prove")
-        ? make_optional(options_vars.contains("assert")
-                  ? ProofOptions{options_vars["proof-files-basename"].as<string>()}.set_assertion_level(AssertionLevel::Inferences)
-                  : ProofOptions{options_vars["proof-files-basename"].as<string>()})
-        : nullopt;
+    optional<ProofOptions> proof_options;
+    if (options_vars.contains("prove")) {
+        proof_options = ProofOptions{options_vars["proof-files-basename"].as<string>()};
+        // Only mark the level explicitly when --assert is given, so the
+        // GCS_ASSERTION_LEVEL env var still applies in the plain --prove case.
+        if (options_vars.contains("assert"))
+            proof_options->set_assertion_level(AssertionLevel::Inferences);
+    }
 
     auto stats = solve_with(p,
         SolveCallbacks{
--- a/examples/crystal_maze/crystal_maze.cc
+++ b/examples/crystal_maze/crystal_maze.cc
@@ -3,11 +3,13 @@
 #include <gcs/constraints/equals.hh>
 #include <gcs/constraints/linear.hh>
 #include <gcs/problem.hh>
+#include <gcs/proof.hh>
 #include <gcs/search_heuristics.hh>
 #include <gcs/solve.hh>
 
 #include <cstdlib>
 #include <iostream>
+#include <optional>
 #include <string>
 #include <vector>
 
@@ -49,12 +51,13 @@ auto main(int argc, char * argv[]) -> int
     cxxopts::ParseResult options_vars;
 
     try {
-        options.add_options("Program options")                                                       //
-            ("help", "Display help information")                                                     //
-            ("prove", "Create a proof")                                                              //
-            ("proof-files-basename", "Basename for the .opb and .pbp files",                         //
-                cxxopts::value<string>()->default_value("crystal_maze"))                             //
-            ("stats", "Print solve statistics")                                                      //
+        options.add_options("Program options")                               //
+            ("help", "Display help information")                             //
+            ("prove", "Create a proof")                                      //
+            ("proof-files-basename", "Basename for the .opb and .pbp files", //
+                cxxopts::value<string>()->default_value("crystal_maze"))     //
+            ("stats", "Print solve statistics")                              //
+            ("assert", "Use annotated assertions")                           //
             ;
 
         options.add_options()             //
@@ -101,6 +104,12 @@ auto main(int argc, char * argv[]) -> int
         p.post(LinearEquality{WeightedSum{} + 1_i * xs[x1] + -1_i * xs[x2] + -1_i * diffs.back(), 0_i, options_vars.contains("gac")});
     }
 
+    auto proof_options = options_vars.contains("prove")
+        ? make_optional(options_vars.contains("assert")
+                  ? ProofOptions{options_vars["proof-files-basename"].as<string>()}.set_assertion_level(AssertionLevel::Inferences)
+                  : ProofOptions{options_vars["proof-files-basename"].as<string>()})
+        : nullopt;
+
     auto stats = solve_with(p,
         SolveCallbacks{
             .solution = [&](const CurrentState & s) -> bool {
@@ -111,9 +120,7 @@ auto main(int argc, char * argv[]) -> int
                 return true;
             },
             .branch = branch_with(variable_order::dom_then_deg(xs), value_order::smallest_first())},
-        options_vars.contains("prove")
-            ? make_optional<ProofOptions>(options_vars["proof-files-basename"].as<string>())
-            : nullopt);
+        proof_options);
 
     if (options_vars.contains("stats"))
         print("{}", stats);
--- a/gcs/constraints/abs/hints.hh
+++ b/gcs/constraints/abs/hints.hh
@@ -8,16 +8,7 @@
 namespace gcs::innards::hints
 {
     /**
-     * \brief abs's assertion hint: just the owning constraint.
-     *
-     * abs needs no per-shape discriminator. One justification function reads the
-     * asserted literal and the reason and dispatches on them, so there is nothing
-     * for the wire to disambiguate -- every abs bound is re-derivable from the
-     * asserted literal, the reason and the definition lines. The same hint rides
-     * both JustifyUsingRUP (the pure-RUP pruning) and JustifyExplicitly (the
-     * explicit-steps bounds): the assertion-hint axis is orthogonal to how the
-     * inference is proved. With no own hint_sexpr it takes the default wire form,
-     * `(constraint_id <originator>)`.
+     * \brief abs's assertion hint: just the owning constraint, no subhint.
      *
      * \ingroup Innards
      */
--- a/gcs/constraints/abs/hints.hh
+++ b/gcs/constraints/abs/hints.hh
@@ -1,0 +1,31 @@
+#ifndef GLASGOW_CONSTRAINT_SOLVER_GUARD_GCS_CONSTRAINTS_ABS_HINTS_HH
+#define GLASGOW_CONSTRAINT_SOLVER_GUARD_GCS_CONSTRAINTS_ABS_HINTS_HH
+
+#include <gcs/constraint_id.hh>
+
+#include <string_view>
+
+namespace gcs::innards::hints
+{
+    /**
+     * \brief abs's assertion hint: just the owning constraint.
+     *
+     * abs needs no per-shape discriminator. One justification function reads the
+     * asserted literal and the reason and dispatches on them, so there is nothing
+     * for the wire to disambiguate -- every abs bound is re-derivable from the
+     * asserted literal, the reason and the definition lines. The same hint rides
+     * both JustifyUsingRUP (the pure-RUP pruning) and JustifyExplicitly (the
+     * explicit-steps bounds): the assertion-hint axis is orthogonal to how the
+     * inference is proved. With no own hint_sexpr it takes the default wire form,
+     * `(constraint_id <originator>)`.
+     *
+     * \ingroup Innards
+     */
+    struct Abs
+    {
+        ConstraintID originator;
+        static constexpr std::string_view hint_name = "abs";
+    };
+}
+
+#endif
--- a/gcs/constraints/all_different/gac_all_different.cc
+++ b/gcs/constraints/all_different/gac_all_different.cc
@@ -205,31 +205,6 @@ namespace
         }
     }
 
-    // Build the typed Hall hint shared by both GAC all_different Hall shapes (the
-    // matching-too-small contradiction and the SCC Hall-set deletion). The hall set
-    // and values, the variable scope and the at-most-one cache are all emit context
-    // for emit_justification -- the pointers reference constraint-owned data, valid
-    // for the whole solve including any later replay. None of it is serialised: the
-    // hint takes the default identity-plus-subhint wire form, so the assertion names
-    // only the constraint and the "hall" sub-rule.
-    auto hall_hint(
-        const vector<IntegerVariableID> & vars,
-        const vector<IntegerVariableID> & hall_variable_ids,
-        const vector<Integer> & hall_value_nrs,
-        const ConstraintID & constraint_id,
-        map<Integer, ProofLine> & value_am1_constraint_numbers) -> hints::AllDifferentHall
-    {
-        // Aggregate init: the AllDifferent base (originator) first, then the Hall
-        // emit context in declaration order (hall_vars, hall_vals, all_vars,
-        // value_am1_constraint_numbers).
-        return hints::AllDifferentHall{
-            {constraint_id},
-            hall_variable_ids,
-            hall_value_nrs,
-            &vars,
-            &value_am1_constraint_numbers};
-    }
-
     auto prove_matching_is_too_small(
         const ConstraintID & constraint_id,
         const vector<IntegerVariableID> & vars,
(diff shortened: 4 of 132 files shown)


examples/cake/cake.cc
```diff
@@ -92,8 +92,9 @@ auto main(int argc, char * argv[]) -> int
                 variable_order::dom_then_deg(vector<IntegerVariableID>{banana, chocolate}),
                 value_order::largest_first())},
         options_vars.contains("prove")
-            ? make_optional<ProofOptions>(ProofFileNames{options_vars["proof-files-basename"].as<string>()},
-                  true, options_vars.count("full-proof-encoding"))
+            ? make_optional(options_vars.count("full-proof-encoding")
+                      ? ProofOptions{ProofFileNames{options_vars["proof-files-basename"].as<string>()}}.enable_full_encoding()
```
ciaranm: There's a nicer way of doing this with the fluent interface now, I think (and if it doesn't do this exact pattern, let's make it do so)

examples/crystal_maze/crystal_maze.cc
```diff
@@ -101,6 +104,12 @@ auto main(int argc, char * argv[]) -> int
         p.post(LinearEquality{WeightedSum{} + 1_i * xs[x1] + -1_i * xs[x2] + -1_i * diffs.back(), 0_i, options_vars.contains("gac")});
     }
 
+    auto proof_options = options_vars.contains("prove")
+        ? make_optional(options_vars.contains("assert")
+                  ? ProofOptions{options_vars["proof-files-basename"].as<string>()}.set_assertion_level(AssertionLevel::Inferences)
```
ciaranm: Likewise (and for every other place that does this)

gcs/constraints/all_different/justify.hh
```diff
@@ -16,6 +16,7 @@ namespace gcs::innards
         const std::vector<IntegerVariableID> & hall_variables,
         const std::vector<Integer> & hall_values,
         std::map<Integer, ProofLine> & constraint_numbers) -> void;
+
```
ciaranm: No blank line

gcs/constraints/among/among.cc
```diff
@@ -222,24 +223,25 @@ auto Among::install_propagators(Propagators & propagators) -> void
...
-                                for (const auto & m : must_match_vars) {
-                                    if (holds_alternative<ConstantIntegerVariableID>(m))
-                                        continue;
-                                    b.add(logger->names_and_ids_tracker().need_constraint_saying_variable_takes_at_least_one_value(m));
-                                }
-                                b.emit(*logger, ProofLevel::Temporary);
-                            }
-                        }},
+                        inference.infer_all(logger, inferences, JustifyExplicitly{[&](const ReasonLiterals &) -> void {
+                                                                                      // We need to bound the sum from BELOW: must_match vars each
```
ciaranm: Holy indenting, Batman

gcs/constraints/among/among.cc
```diff
@@ -272,26 +274,27 @@ auto Among::install_propagators(Propagators & propagators) -> void
...
-                                            for (const auto & m : must_match_vars)
-                                                b.add(am1_lines->at(m));
-                                            for (const auto & other_var : can_be_either_vars)
-                                                if (var != other_var)
-                                                    b.add(am1_lines->at(other_var));
-                                            b.emit(*logger, ProofLevel::Temporary);
-                                        }
-                                    }},
+                                    inference.infer(logger, var != val, JustifyExplicitly{[&](const ReasonLiterals &) {
+                                                                                              // need to point out that if var == val then var != voi for each voi
```
ciaranm: Spaaaaaaaaaaaaaaace

gcs/constraints/count/count.cc
```diff
@@ -191,47 +192,47 @@ auto Count::install_propagators(Propagators & propagators) -> void
...
-                        // Per-voi unconditional lines: emit after all the
-                        // conditionals so they have the full set of pairwise
-                        // facts to RUP-derive against.
-                        for (const auto & voi : voi_set.each())
-                            logger->emit_rup_proof_line_under_reason(reason,
-                                WPBSum{} + 1_i * (value_of_interest != voi) + 1_i * (how_many < *highest_how_many_might + 1_i) >= 1_i,
-                                ProofLevel::Temporary);
-                    }};
+                auto just = JustifyExplicitly{[&](const ReasonLiterals & reason) -> void {
+                                                  // Per-(voi, var) conditional pairs: emit when voi is
```
ciaranm: More annoying indenting from clang-format. Is there a way we can change the settings to get it to not do this?

ciaranm:
@mmcilree I think this is ready for review, and possibly then rewrite-history-and-merge?

gcs/constraints/abs/hints.hh
```diff
@@ -0,0 +1,31 @@
...
+#include <gcs/constraint_id.hh>
+
+#include <string_view>
+
+namespace gcs::innards::hints
+{
+    /**
+     * \brief abs's assertion hint: just the owning constraint.
+     *
+     * abs needs no per-shape discriminator. One justification function reads the
```
mmcilree: Definitely don't need this paragraph

gcs/constraints/all_different/gac_all_different.cc
```diff
@@ -190,17 +205,43 @@ namespace
...
     }
 
+    // Build the typed Hall hint shared by both GAC all_different Hall shapes (the
+    // matching-too-small contradiction and the SCC Hall-set deletion). The hall set
+    // and values, the variable scope and the at-most-one cache are all emit context
+    // for emit_justification -- the pointers reference constraint-owned data, valid
+    // for the whole solve including any later replay. None of it is serialised: the
+    // hint takes the default identity-plus-subhint wire form, so the assertion names
+    // only the constraint and the "hall" sub-rule.
+    auto hall_hint(
```
mmcilree: Is this function serving any purpose? It looks like it's just a wrapper - why not just construct the hint directly?

gcs/constraints/element/element.cc
```diff
@@ -541,12 +552,15 @@ auto NDimensionalElement<EntryType_, dimensions_>::install_propagators(Propagato
...
                 auto array_var = get_array_var<dimensions_>(elem, *array);
-                enforce_equality(logger, result_var, array_var, state, inference, index_reason);
+                enforce_equality(logger, result_var, array_var, state, inference, index_reason, owner);
             }
 
             return PropagatorState::Enable; }, equality_triggers);
     }
 }
+#if defined(__GNUC__) && ! defined(__clang__)
+#pragma GCC diagnostic pop
```
mmcilree, line 562: What's this for?

gcs/constraints/linear/propagate.cc
```diff
@@ -245,33 +251,36 @@ auto gcs::innards::propagate_linear(const auto & coeff_vars, Integer value, cons
...
+// for the equality propagator, NoHint for the reified-inequality propagators (which
+// pass no hint). The hint rides the JustifyExplicitly, so it is materialised lazily
+// in assertion mode rather than eagerly at the call site.
+#define GCS_INSTANTIATE_PROPAGATE_LINEAR(CoeffVars, Tracker, Hint)                                  \
+    template auto gcs::innards::propagate_linear(const CoeffVars & coeff_vars,                      \
+        Integer value, const State & state, Tracker &, ProofLogger * const logger,                  \
+        bool equality, const optional<pair<optional<ProofLine>, optional<ProofLine>>> & proof_line, \
+        const optional<Literal> & add_to_reason, const Hint & hint) -> PropagatorState
+
+GCS_INSTANTIATE_PROPAGATE_LINEAR(SumOf<Weighted<SimpleIntegerVariableID>>, SimpleInferenceTracker, hints::LinearEquality);
```
mmcilree, line 264: Just checking, are these macros definitely a good idea? I know they save a bit of space, but might interfere with linting etc.

gcs/constraints/plus_minus/hints.hh
```diff
@@ -0,0 +1,57 @@
...
+     * \ingroup Innards
+     */
+    struct Minus
+    {
+        ConstraintID originator;
+        std::optional<ProofLine> pol_line;
+        static constexpr std::string_view hint_name = "minus";
+    };
+
+    auto emit_justification(ProofLogger & logger, const Minus & minus, const ReasonLiterals & reason) -> void;
```
mmcilree, line 54: Again, should `emit_justification` be in the hints namespace?

gcs/innards/justification.hh
```diff
@@ -1,72 +1,91 @@
...
+         *
+         * The minimal assertion hint: in assertion mode it annotates the inference
+         * with just this name (no \c hint_sexpr), telling the justifier which
+         * constraint family the step came from without serialising any operands. The
+         * tier between NoHint (annotate with nothing, fall back to the call site's
+         * annotation) and a per-constraint typed hint (name plus structured fields).
+         *
+         * \ingroup Innards
+         */
+        struct ModelName
```
mmcilree: Don't think this is used anywhere anymore?

gcs/innards/proofs/hints.hh
```diff
@@ -0,0 +1,63 @@
+#ifndef GLASGOW_CONSTRAINT_SOLVER_GUARD_GCS_INNARDS_PROOFS_HINTS_HH
+#define GLASGOW_CONSTRAINT_SOLVER_GUARD_GCS_INNARDS_PROOFS_HINTS_HH
+
+#include <string_view>
+
+namespace gcs::innards::hints
```
mmcilree: Much prefer this, but could/should this all live in `assertion_hints.hh` ?

gcs/innards/proofs/infer_explicitly.hh
```diff
@@ -0,0 +1,162 @@
...
+     * \c emit's own advertisement is used (a named fat witness that carries its own
+     * \c hint_sexpr / \c hint_name); otherwise the call site's \c fallback. A bare
+     * emit closure advertises nothing, so a closure with no hint resolves straight to
+     * the fallback. This reproduces the old witness-carried annotation exactly while
+     * letting an inline closure carry a separate ModelName / structured hint.
+     *
+     * \ingroup Innards
+     */
+    template <typename Hint_, typename Emit_>
+    [[nodiscard]] auto resolve_annotation(ProofLogger & logger, const Hint_ & hint, const Emit_ & emit,
```
mmcilree: Is this function still necessary? The name makes it sound like it does more than it does, and it's only called in one place as far as I can see.

gcs/innards/proofs/names_and_ids_tracker.cc
```diff
@@ -1088,14 +1176,21 @@ auto NamesAndIDsTracker::need_view(const ViewOfIntegerVariableID & view) -> Proo
         v_lo, v_hi, name, IntegerVariableProofRepresentation::Bits);
 
     Integer s_coeff = view.negate_first ? -1_i : 1_i;
-    auto [link_le, link_ge] = _imp->model->add_constraint(StringLiteral{"view link"}, StringLiteral{"definitional"},
-        WPBSum{} + 1_i * v_id + (-s_coeff) * view.actual_variable == view.then_add);
+
+    auto will_define = true;              // We do need to define views properly in the model!
```
mmcilree: This is from my branch and I think Claude pointed out that it is stupid. I agree

gcs/innards/reason.cc
```diff
@@ -1,78 +1,129 @@
...
+        [&](const BothBoundsReasonOver & r) { materialise_bounds(state, *r.vars, r.extra, result); },
+        [&](const ExactSingleValue & r) { materialise_exact_single_value(state, *r.vars, r.extra, result); },
+        [&](const LazyReasonOver & r) { r.fn(state, result); },
+        [&](const NarrowableGenericReasonOver & r) { materialise_generic(state, *r.vars, r.extra, result); },
+        [&](const NarrowableBothBoundsReasonOver & r) { materialise_bounds(state, *r.vars, r.extra, result); },
+        [&](const NarrowableLazyReasonOver & r) { r.fn(state, result); }});
+    return result;
+}
+
+auto gcs::innards::generic_reason(const State &, const std::vector<IntegerVariableID> & vars,
```
mmcilree: Do these functions still need to take state? Seems like not?

mmcilree (commented):
Cool, yes I like this a lot better. Couple of minor nitpicks but I'd support a squash/cleanup and then merge.

gcs/constraints/element/element.cc
ciaranm: Current GCC on Ubuntu has a spurious warnings bug in its vectorisation code that spams pages of annoyingly distracting output about uninitialised variables that aren't.

gcs/constraints/linear/propagate.cc
ciaranm: I'm kind of mixed on these. On balance, probably they're better at this point, if we're going to need a lot of explicit instantiations.

gcs/constraints/plus_minus/hints.hh
ciaranm: There's probably a right answer here involving ADL (argument-dependent lookup) and I'll let Claude answer this.

ciaranm: Keeping it in `hints`, and it turns out the reason is ADL — argument-dependent lookup (sometimes "Koenig lookup"). The short version: when you call a free function unqualified, e.g. `emit_justification(logger, hint, reason)`, the compiler doesn't only look in the current scope — it *also* searches the namespaces where the **argument types** are defined. It's the same mechanism that lets `cout << x` find an `operator<<` you defined next to your own type, or that finds the right `swap`/`begin`/`end`.

That's exactly how the dispatch in `infer_explicitly.hh` works — `emit_justification(...)` is called unqualified, and the compiler finds the right overload because the hint type (`Plus`, `Minus`, `AllDifferentHall`, …) lives in `gcs::innards::hints`. So the function *has* to be in a namespace associated with one of its arguments, and `hints` (home of the hint struct) is the natural one. It's the emit-side twin of `hint_sexpr`, which is *also* an unqualified ADL call and *also* lives in `hints`: the two are the wire-serialisation half and the proof-emission half of the same per-hint interface, so they sit next to the struct — the `operator<<` idiom.

(Strictly, ADL would *also* find it in plain `gcs::innards`, since `ProofLogger`/`ReasonLiterals` live there — so it's not forced. But splitting one type's `hint_sexpr` + `emit_justification` across two namespaces would be arbitrary, so they stay together.)

I've added a doc comment to the `hints` namespace making this explicit: it's the typed-hint *customization surface* (struct + its `hint_sexpr` + its `emit_justification`), not just a bag of wire-vocabulary constants.

gcs/innards/proofs/hints.hh
ciaranm: Kept separate — and digging in, the dependency actually runs the other way, so I've re-split rather than merged. `proofs/hints.hh` is the lightweight leaf (`#include <string_view>` only) and it's included *by* `assertion_hints.hh` (which needs `s_expr.hh` / `proof_line.hh` / `<format>`). `NoHint` is used ~38× including in lightweight propagator headers (`linear/propagate.hh`, `extensional_utils.hh`, …) that shouldn't have to drag in the s-expression machinery for a hint they never serialise — so folding `hints.hh` *up* into `assertion_hints.hh` would tax all of those.

The real fragmentation was wider: `NoHint` and `ModelName` were marooned over in `justification.hh`, and `ModelName` was dead (your other comment — confirmed, deleted). So:

- moved `NoHint` *down* into `proofs/hints.hh` next to the other bare hint tags;
- deleted `ModelName`;
- left `assertion_hints.hh` as pure machinery.

The split now has a principle behind it — **vocabulary vs machinery**: `proofs/hints.hh` = the hint tags + wire field-keys (`string_view` only), `assertion_hints.hh` = the `hint_sexpr` builders + `AssertionAnnotation` (needs `s_expr`). Build's green and proofs are byte-identical (header moves only).

ciaranm:
@mmcilree Thanks for the review — cleanup pass pushed. Rundown of the nitpicks:

**Formatting**
- `among.cc` (×2) and `count.cc`: the hyper-indented lambda bodies are gone — each inline lambda is now hoisted into a named `auto emit = [&]{…}` local, so the body sits at normal indentation rather than aligning under the far-right `[`.
- `all_different/justify.hh`: stray blank line removed.
- On *"can we change the clang-format settings to stop this?"* — yes, `LambdaBodyIndentation: OuterScope` is the knob, but it reflows ~110 files (the codebase is lambda-heavy), so I've left it as a separate repo-wide decision rather than folding it in here. The hoist above is the in-PR fix. (Aside: clang-format 21 already disagrees with the tree on ~52 files, so there's a clang-format *version* drift independent of any of this, and no CI format check to pin it.)

**Cleanups**
- `abs/hints.hh`: paragraph trimmed to the one-line brief.
- `gac_all_different.cc` `hall_hint`: inlined into its two call sites and deleted — the field order is documented on the `AllDifferentHall` struct, so the wrapper wasn't earning its keep.
- `ModelName`: confirmed dead, deleted (stale doc reference fixed).
- `resolve_annotation`: inlined into its single call site (with the precedence comment moved alongside) and deleted.
- `generic_reason` / `bounds_reason`: dropped the unused `State &` param — updated all ~101 call sites and anonymised one now-unused `state` in a `circuit_base` propagator lambda.
- `will_define`: removed the always-true variable and its dead branch.

**Design questions** (replied inline on the two threads):
- `emit_justification` stays in `gcs::innards::hints` — it's an ADL customization point, the emit-side twin of `hint_sexpr`; added a namespace doc making that explicit.
- `proofs/hints.hh` vs `assertion_hints.hh`: kept separate, re-split on a vocabulary-vs-machinery line — `NoHint` moved down into the lightweight `proofs/hints.hh`, `assertion_hints.hh` left as the s-expression machinery.

**Kept as-is** per your/my replies: the `element.cc` GCC vectoriser pragma, and the `propagate_linear` instantiation macros.

**Verification**: Release 396/396 + ASan/UBSan green, warning-free on GCC, every edited file clang-format-stable. All edits are semantics-preserving, so proofs are unchanged.

I'll do the squash/history-cleanup next so this is ready to land.

PR #396, opened 2026-06-25
Reformat: 150-column limit + lambda indentation cleanup, with CI check

ciaranm:
Tree-wide clang-format cleanup, aimed mainly at stopping justification and
propagator lambdas from being heavily indented.

## The config change (commit 1)

The tree was formatted at `ColumnLimit: 0`, leaving all line wrapping to the
author. Keys changed in `.clang-format`:

| Key | Old | New |
|---|---|---|
| `ColumnLimit` | `0` | `150` |
| `LambdaBodyIndentation` | *(default `Signature`)* | `Signature` (pinned) |
| `BreakTemplateDeclarations` | *(default)* | `Yes` |
| `AllowShortIfStatementsOnASingleLine` | `WithoutElse` | `Never` |
| `ReflowComments` | *(default)* | `false` |

`ColumnLimit: 150` is the lever: at `0`, a lambda introducer was hand-broken
onto its own deeply-nested line and the body (`Signature`, the default) sat one
level past that far-right column, so it drifted rightward with every level of
call nesting. At `150`, clang-format joins the introducer onto its call line, so
the signature starts much further left and the body lands one level past it —
shallow regardless of nesting. The other keys pin layout that `ColumnLimit: 0`
had left to the author (template heads on their own line, guard-clause bodies on
their own line, hand-wrapped narrow comments preserved).

Before / after, a representative propagator inference:

```cpp
// before (ColumnLimit 0)
            inference.infer(logger, v2 >= 0_i,
                JustifyExplicitly{[logger, v1, v2, abs_nonneg_ge](const ReasonLiterals &) -> void {
                                      justify_abs_v2_ge_zero(*logger, v1, v2, *abs_nonneg_ge);
                                  },
                    ThenRUP::Yes, hints::Abs{originator}},
                NoReason{});

// after (ColumnLimit 150)
            inference.infer(logger, v2 >= 0_i,
                JustifyExplicitly{
                    [logger, v1, v2, abs_nonneg_ge](const ReasonLiterals &) -> void { justify_abs_v2_ge_zero(*logger, v1, v2, *abs_nonneg_ge); },
                    ThenRUP::Yes, hints::Abs{originator}},
                NoReason{});
```

## The reformat (commit 2)

`clang-format -i` over every tracked `.cc`/`.hh` except the vendored
`XCSP3-CPP-Parser`. Pure mechanical change, no semantics touched — 255 files.

## Keep hand-laid-out layouts multi-line (commit 3)

The mechanical reformat collapsed some deliberate author layouts onto single
long lines: the embedded instance data (puzzle grids, the among / element /
knapsack data tables, the qap / tsp distance matrices), `solve_with(...)` call
heads, and designated-initialiser / `overloaded{...}` visitor blocks. These are
restored to one entry per line and locked with a trailing `//` on each row so
clang-format won't re-merge them. For the data tables it's whitespace and `//`
markers only — stripping them reproduces the previous content byte-for-byte, so
no data value changes.

## CI check (commit 4) + docs (commit 5)

A new `clang-format` workflow runs `clang-format --dry-run --Werror` over the
same file set on every push/PR, pinned to **21.1.8** (the version the tree was
formatted with — output drifts between releases, so an unpinned binary would
fail spuriously). CONTRIBUTING / CLAUDE.md / README are updated to point at the
formatting requirement.

Verified on this branch: clean release build and **all 396 ctests pass**; the
tree is clang-format-stable so the new CI check is green.


ciaranm:
@mmcilree I'm going over this and checking it manually for oddities, but just to warn you that I'm about to do a tree-wide "change the clang-format settings now that it's better with lambdas, and reformat everything for the greater good".

PR #398, opened 2026-06-26
reason: retire the per-struct `extra` slot; guard linear_bounds_reason

ciaranm:
Stacked on the ConcatReason PR.

**Retire `extra`:** `GenericReasonOver`/`BothBoundsReasonOver`/`ExactSingleValue` (+ Narrowable twins) each carried an `optional<Literal> extra` — a one-literal concat bolted onto every reason type, with `materialise_*` hand-coding first-vs-last ordering. That's exactly what `ConcatReason` generalises. The structs are now pure "facts over these vars"; `generic_reason`/`bounds_reason` keep their signatures but route through `with_extra`; `ExactSingleValue`'s leading-condition becomes explicit via a new `concat(a,b)`. Byte-identical by construction — **all 81 scp_cases proofs unchanged**; equals_test 284 / arg_sort_test 23 / sort_test 29, zero failures.

**linear guard:** `linear_bounds_reason` built an O(coeff_vars) reason on every firing, handed straight to `infer_*`, unguarded. Guarded on `want_reasons()`. **magic_series 300: 7.06s → 6.25s (−11.5%)**, search bit-identical, proofs still byte-identical.

### DIFF
--- a/gcs/constraints/equals/equals.cc
+++ b/gcs/constraints/equals/equals.cc
@@ -298,12 +298,14 @@ auto ReifiedEquals::install_propagators(Propagators & propagators) -> void
                                                 ProofLogger * const logger, const Literal & cond) -> PropagatorState {
         auto value1 = state.optional_single_value(v1);
         if (value1) {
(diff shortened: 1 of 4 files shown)


ciaranm:
Content landed on `main` via #407 (the stack had been merged onto its parent feature branch rather than `main`; #407 re-lands #398–#406 cleanly). Closing as included.

PR #399, opened 2026-06-26
min_max, in: with_extra + want_reasons guard

ciaranm:
Stacked on the retire-`extra` PR. The last two eager "materialise(generic_reason(...)) then push a loop" concat sites, onto the same pattern as Element.

- **in.cc**: the O(`var_vals` × `dom(var)`) cross-product support reason is now declarative + guarded (and its small sibling reason too).
- **min_max.cc**: the O(`vars` × `dom(result)`) base reason (reused across removed intervals) becomes a declarative `with_extra` base, extended per-interval via `with_extra(reason, …)`, under the guard.

Byte-identical (all 81 scp_cases unchanged, incl. array_min/array_max/in_*); min_max_test 88 + in_test 53, zero failures. No benchmark exercises In/MinMax, so the win is by-construction (the O(n²) builds are skipped proofs-off).

### DIFF
--- a/gcs/constraints/element/element.cc
+++ b/gcs/constraints/element/element.cc
@@ -10,6 +10,7 @@
 #include <gcs/innards/propagators.hh>
 #include <gcs/innards/s_expr.hh>
 #include <gcs/innards/state.hh>
+#include <gcs/innards/variable_id_utils.hh>
 #include <gcs/integer.hh>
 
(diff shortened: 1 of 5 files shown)


ciaranm:
Content landed on `main` via #407 (re-lands #398–#406 cleanly onto `main`). Closing as included.

PR #430, opened 2026-06-30
Cumulative: chain-verify variable durations via in-proof end + bridge lemma

ciaranm:
Stacked on #428 (uses its `introduce_bits_of` / `create_proof_only_integer_variable_in_proof`). Will retarget to `main` once #428 merges.

## What
Makes **variable-duration** cumulative (variable start *and* length) chain-verify against `cake_pb_cp`. Constant-length / variable-height / variable-capacity already chained; variable duration did not, and no test exercised it.

## Why it didn't chain
cake reifies the after-flag `ca ⟺ s+l ≥ t+1` **directly** (no `end` variable). The solver introduced a proof-only `end = s+l` defined by **OPB axioms** (`@c[id][endge_i]`) to get a single-variable, RUP-closable `after` pin — but cake's OPB has no such axiom, so the chain died on `pol @c[_1][endge_0]`.

## The fix (an in-proof reformulation)
Keep `after` reified on `s+l` (matching cake), but introduce `end = s+l` **inside the proof** as a conservative extension (`introduce_bits_of`, no OPB encoding) in the install initialiser, which also emits a per-`(i,t)` **bridge lemma** `end ≥ t+1 → after`:

```
pol  @v[id][i_t][ca][f] (¬after → s+l ≤ t)  +  end_le (end ≤ s+l)   =   M·after − end + t ≥ 0
```

The `s+l` bits cancel exactly, leaving a single-variable-in-`end` handle. The propagator's after pin is **unchanged in shape** — materialise `end ≥ s_lo + lb(l)` over `end`'s in-proof `end_ge`, then RUP `after=1` single-variable in `end` against the bridge lemma — so `pin_contributor` / `pin_pushed` (including the `ext_lit` disjunct) ride along unchanged. Because both `end`'s definition and the lemma live in the proof, nothing in cake's OPB needs to match.

## Verification
- New `scp_chain_cumulative_var_duration_sat` passes the full workflow-2 chain (cake re-derives the OPB, VeriPB elaborates, cake re-checks).
- `cumulative_test` self-verifies (incl. the wide bound-push cases that exposed the two-variable-RUP trap of a naive "drop end" attempt).
- Full suite **406/406**, no regressions; existing cumulative chain tests unchanged.

## Follow-up (documented in `dev_docs/cumulative-proof-logging.md`)
Variable **heights** still diverge (cake's binary `cc` flags vs the solver's bits-integer `contrib`), so the new case keeps a constant height. Same flavour of fix; separate work.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/dev_docs/cumulative-proof-logging.md
+++ b/dev_docs/cumulative-proof-logging.md
@@ -302,17 +302,32 @@ and the pol differently:
 - **Variable durations** rewrite `after_{i,t} ⇔ s_i + l_i ≥ t+1`. The
   pinning `after = 1` then needs the *cross-variable* fact
   `s_i + l_i ≥ B`, which RUP cannot derive from the operands' bounds
-  alone (the VeriPB linear-combination limit). When **both** `s_i` and
-  `l_i` vary, `after` is reified instead on a proof-only
-  `end_i = s_i + l_i` (single variable), and the pin first materialises
-  `end_i ≥ s_lo + lb(l_i)` with a `pol` over the captured `end ≥ s + l`
-  definition line plus the two operand order-literal defining lines
-  (the `plus_minus/plus.cc` pattern). The `after = 1` RUP is then
-  single-variable in `end_i`, exactly like the constant-duration case.
-  `s_lo` is the chain running bound (lb-push), `t − lb(l_j) + 1`
-  (ub-push, i.e. `¬ext_lit`), or `lb(s_i)` (a mandatory task). If either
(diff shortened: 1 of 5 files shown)


ciaranm:
Auto-closed when base branch `disjunctive-var-chain` was deleted on #428's merge; GitHub won't reopen it. Continued as #433 (same branch, rebased onto `main`).

Issue #444, opened 2026-07-01
Tracker: arithmetic constraints redesign — semantic API, real propagators, no more OPB pseudo-tables


ciaranm:
The "turn it into a table" arithmetic constraints date back to the earliest
iteration of the solver, and were always a temporary solution. Today `Times`,
`Div`, `Mod`, and `Power` (plus the `PlusGAC` / `MinusGAC` test variants) are
all one template, `innards::GACArithmetic<op>`: its `prepare()` materialises
the full `(v1, v2, result)` tuple cross-product and forwards everything to a
`Table`. That is O(|dom(v1)|·|dom(v2)|) memory at post time, and
`Table::define_proof_model` writes a selector variable plus one reified
constraint per tuple into the OPB file. Meanwhile the good implementations we
do have (`Plus` / `Minus` bounds-consistent, `MultBC`) are unreachable from
the MiniZinc and XCSP3 frontends, and the constant-folding logic that ought to
live in the solver is duplicated ad-hoc in the XCSP3 frontend
(`post_product`, the #153 workaround) and absent from the FlatZinc one.

This tracker supersedes #126 and #153, and implements the first slice of #299.

## Design goals

- **G1**: users write constraints semantically, without needing to know how
  this maps onto propagation algorithms internally — for the direct API and
  for MiniZinc / XCSP3 alike. We select a good default.
- **G2**: there are still ways to select particular propagators when the user
  knows what they want (likewise for testing).
- **G3**: we no longer use "big pseudo-table in the OPB file" as an encoding.
  Single exception: `Power` with a variable exponent, which is rarely used and
  has no sensible PB encoding.
- **G4**: each individual scp constraint type maps onto exactly one encoding
  procedure. The OPB encoding is a deterministic function of the constraint
  type and its argument *structure* (which operands are constants — visible in
  the scp form), and never of arbitrary parameters like domain sizes.
  Domain-size-dependent choices are legal only where they don't touch the OPB
  or the scp term.

### Names and selection API

One public class per operation: **`Multiply`**, **`Divide`**, **`Modulus`**,
`Power`, `Plus`, `Minus` (spelled out in full, matching the Plus/Minus style;
FlatZinc says `int_times` and XCSP3 says `mul`, so neither of those wins). The
public `Times` / `Div` / `Mod` names are deleted — hard break, no deprecated
aliases; if anyone hates this enough to complain we can add a `using` alias
later.

Selection follows the #299 design: each constructor takes an optional
[...]

Issue #299, opened 2026-06-10
Constraints: unify how callers select consistency / propagation algorithm (three inconsistent mechanisms today)


ciaranm:
**Classification:** cross-cutting design, public constraint API. Behaviour-preserving once done, but it needs a design decision first — so this is a discussion/design issue, not a turnkey change.

## Problem

The choice of *how* a constraint should propagate — its consistency level / propagation algorithm, plus proof-output knobs like reason aliasing — is currently expressed **three different and mutually inconsistent ways** across the constraint API. Where it is a bare `bool`, the call site is unreadable.

### Pattern A — opaque trailing `bool`

The caller writes `true`/`false` with no hint of meaning:

```cpp
// examples/sudoku/sudoku.cc — the trailing `true` means "use the GAC propagator"
p.post(LinearEquality{WeightedSum{} + 1_i * grid[r][c] + 1_i * grid[r + 1][c], 5_i, true});

// examples/tour/tour.cc — the `false` is gac_all_different
p.post(Circuit{succ, false, options});
```

Sites:
- **`gac`** on **seven** constructors in `gcs/constraints/linear/linear_equality.hh`: `LinearEquality`, `LinearEqualityIf`, `LinearEqualityIff`, `LinearNotEquals`, `LinearNotEqualsIf`, `LinearNotEqualsIff`, and `ReifiedLinearEquality` (which has *two* adjacent bools: `bool gac, bool flipped_cond`). Reader-facing call sites in `sudoku.cc` (×6), `odd_even_sum.cc` (×2), `knapsack.cc`, `skyscrapers.cc`.
- **`bounds_only`** on the Element family (`gcs/constraints/element.hh`): `NDimensionalElement` and the `Element` / `Element2D` / `ElementConstantArray` / `Element2DConstantArray` wrappers. Extra footgun: the default **flips** between siblings — `false` for `Element`/`Element2D`, `true` for the `*ConstantArray` variants — so the same-looking call propagates differently depending on the type.
- **`gac_all_different`** on `CircuitBase` (`circuit_base.hh:82`) and `CircuitSCC` (`circuit_scc.hh:30`).

### Pattern B — separate types, with a `using` alias picking a default

The algorithm choice is encoded as the *type*, and the friendly public name is a typedef to one of them:

```cpp
using GlobalCardinality = BoundsGlobalCardinality;   // global_cardinality.hh — GAC variant is a separate type
[...]

Issue #232, opened 2026-05-25
MultBC: support aliased operands (squaring, fixed points) in proof encoding


ciaranm:
Follow-up to PR #233 (Bucket B of the duplicate-variable audit).

## Context

`MultBC(v1, v2, v3)` with any two slots aliased to the same underlying `SimpleIntegerVariableID` currently throws `InvalidProblemDefinitionException` at construction. The semantics are well-defined (`x*x=z` is squaring; `x*y=x` says "x is a fixed point of multiplying by y", etc.) but the bit-product proof encoding doesn't tolerate them today.

## What the audit suggested

In `gcs/constraints/mult_bc.cc:1143`, `make_magnitude_representation` should reuse an already-emitted magnitude variable when one exists for the underlying var, rather than emitting a duplicate batch under a fresh mag variable while `channelling_constraints.insert` silently keeps only the first batch.

I tried this on a branch and it isn't sufficient. The proof still fails at `prove_product_bounds` because of additional duplications downstream.

## What else needs to change

1. **`make_magnitude_representation` (line 1143)** — reuse the existing magnitude variable when called twice with the same `SimpleIntegerVariableID`. This is the audit's suggested fix and is the prerequisite for the rest.

2. **`prove_product_bounds` (line 705)** has four `for (const auto & var : {x, y})` loops. With `x == y`, `conditional_bounds[var].emplace_back(...)` accumulates 2× the entries because both iterations write to the same map key. The nested loop `for x_bound in conditional_bounds[x]: for y_bound in conditional_bounds[y]:` then explores `8x8 = 64` pairs instead of `4x4 = 16`, emitting redundant proof lines and breaking the RUP chain at the final fusion-resolution step. Dedup on `{x, y}` so the loop iterates over unique vars only.

3. **`define_proof_model` sign_lines emission (lines 1226-1234)** uses `mag_var.contains(_v1)` / `mag_var.contains(_v2)` to decide which of the 4 sign-product constraints to emit. After the `make_magnitude_representation` reuse fix, both checks return `true` for aliased `_v1 == _v2`, so all 4 lines emit — but lines 1228 and 1231 have triggers that fold to `false ∧ true = false`, making them vacuous. The propagator side captures `sign_lines` by index/count, so adding vacuous entries shifts later indexed lookups. Either suppress the vacuous emissions or update the propagator to look up by structural key.

[...]

Issue #153, opened 2026-05-08
Friendly C++ API for `Times`, `Div`, `Mod`, `Power` (constant-folding + propagator selection)


ciaranm:
## Background

The `Times`, `Div`, `Mod`, `Power` propagators in `gcs/constraints/arithmetic.*`
share a `GACArithmetic` base that materialises every tuple in the cross-product
of operand domains. Memory blows up the moment any operand domain is wide —
this is documented as a known sharp edge (see the auto-memory note "GACArithmetic
memory blowup"). For multiplication specifically, `MultBC`
(`gcs/constraints/mult_bc.hh`) provides a bounds-consistent alternative without
that explosion, but at weaker propagation strength.

Today, every frontend that emits one of these constraints (XCSP3 binding,
MiniZinc binding, future CPMpy backend) has to:

- Detect the constant-operand cases (`a` constant ⇒ `Times` collapses to a
  linear equality `a * x = r`, `Div` by a constant ⇒ different
  consideration, `Power` with constant exponent ⇒ a chain of `Times`, etc.)
  and rewrite them upstream of the propagator.
- Pick `Times` vs `MultBC` (or accept the GAC version's memory cost) based on
  domain widths it has visibility into.

That logic currently lives at every binding site — XCSP3's `walk_intension`
hand-rolls the constant case for `mul`; FlatZinc presumably does similar; a
CPMpy backend would have to repeat it again. It is logic that belongs **in
the user-facing constraint**, not in every caller.

## What "friendly" looks like

User-facing C++ classes `Times`, `Div`, `Mod`, `Power` that inspect their
operands at construction (or at `install` time) and route to the right
underlying machinery:

- One operand is a constant or constant view ⇒ post a linear equality
  (`WeightedSum`), no propagator at all.
- Both operands variable, both domains narrow ⇒ a GAC propagator (the
  current implementation, but only entered when memory is bounded).
- Both operands variable, at least one wide ⇒ `MultBC` for `Times`, BC
  variants for the others (which may need to be written).
- `Power` with constant non-negative exponent ⇒ unfold to a chain of `Times`.
- `Power` with constant exponent of 2 ⇒ specialise to `Sqr` (or to `Times{x,x,r}`).
- Where the propagator we'd want doesn't yet exist, fall back honestly
  (e.g. `Power` with variable exponent → unsupported, with a clear message).

This means callers — including the XCSP3, MiniZinc, and CPMpy frontends —
[...]

PR #445, opened 2026-07-02
Arithmetic constraints redesign: semantic API, real propagators, no more OPB pseudo-tables (#444)

ciaranm:
Implements the arithmetic constraints redesign from #444: the "turn it into a table" arithmetic family (`Times`/`Div`/`Mod`/`Power`/`PlusGAC`/`MinusGAC`, all `innards::GACArithmetic<op>`) is retired, and every operation gets a semantically-named constraint that picks its own implementation. This also lands the first slice of #299, and the reworks agreed during review: fused compound encodings (#448), the SeqPrecedeChain proof-flag fix (#449), and a set of tabulation optimisations (branching order, functionally determined variables, reifications via wildcards).

## What users see

- **`Multiply`, `Divide`, `Modulus`, `Power`** (plus the existing `Plus`/`Minus`) are the public constraints; each takes an optional `std::variant<consistency::Auto, consistency::BC, consistency::Tabulated>` defaulting to `Auto`. `gcs/consistency.hh` is new (the #299 first slice; the linear family's `bool gac` constructors migrate to `variant<consistency::BC, consistency::Tabulated>` in the same breath). `consistency::Tabulated` is deliberately a different tag from `consistency::GAC`: the latter names a genuine propagation algorithm (AllDifferent and friends, when they migrate), while `Tabulated` says the level is reached by enumerating every satisfying assignment, whose set-up cost and proof size grow with the enumeration — a cost model the signature should not hide.
- Structural dispatch is automatic: constant operands fold to linear equalities, views fold through their affine structure, and aliased operands (`x * x = y`, `x mod y = x`, `x ^ 2 = x`, ...) are decomposed — these were previously only expressible through the tables, or threw.
- `consistency::Tabulated` (always) and `Auto` (when the enumeration is small enough) tabulate the relation **in-proof**: the OPB encoding never changes with the choice — the structural encoding is a deterministic function of the constraint and its argument shapes, and the table is derived via RED selector introductions plus RUP backtrack lines (the mechanism extracted from the linear GAC path). The `Auto` budget (default 100, `GCS_TABULATION_THRESHOLD` overrides) counts the enumeration tree, leaving out a functionally determined variable's level since that level is skipped — so, say, a 10×10×100 multiply tabulates by default.
- Wide-domain products/divisions now get bounds consistent propagation instead of exhausting memory on a materialised cross-product. The one deliberate table survivor is `innards::PowerTable` for a variable exponent.
- `ReifiedLinearEquality` under `consistency::Tabulated` now supports `reif::If`/`NotIf`/`Iff` (previously `UnimplementedException`), giving GAC over the terms and the condition variable together.

## Semantics

Validated against the MiniZinc spec, XCSP3-core Remark 7, Gecode, Choco, OR-Tools CP-SAT, ACE, and CPMpy (details in #444):

- Divide/Modulus: truncation towards zero, remainder takes the sign of the dividend (unchanged, now enforced by a decomposition whose sign conditions pin truncation — the identity plus `|r| < |y|` alone does not determine the quotient for negative operands).
- Division by zero stays relational: the divisor's zero has no support; a constant zero divisor is unsatisfiable, not an error.
- `0^0 = 1` (unchanged). **Behaviour change**: negative exponents now follow MiniZinc 2.7.1's `x^y = 1 div x^|y|` rule, so `2^-3 = 0` rather than having no support (`1^-n = 1`, `(-1)^-n` by parity, `0^-n` unsupported). FlatZinc `int_pow` now matches its spec; previously we silently disagreed.
- `Power{10, 20, z}` with a small result domain is UNSAT rather than an install-time overflow: the chain auxiliaries' ranges are clamped by the result's, since any intermediate power divides the result or has magnitude one.

## Frontends

FlatZinc (`int_times`/`int_div`/`int_mod`/`int_pow`), XCSP3 (`mul`/`sqr`/`div`/`mod`/`pow`), and the Python bindings all post the new constraints. The XCSP3 `post_product` constant-folding workaround for #153 is deleted, and `pow` with a constant exponent now posts `Power` directly, including negative exponents (previously rejected; the XCSP3 spec is silent there, so it follows the MiniZinc rule like everything else).

## Notable internals

- **Fused encodings (#448)**: every operation emits ONE flat `@c[id][role]` OPB block and runs ONE propagator, rather than installing child constraints. `mult_bc::define_encoding`/`mult_bc::propagate` expose the multiplication machinery (`MultiplyBC` is now a thin wrapper); the shared `innards::linear_stages` run the linear pieces, half-reified on their gates, with an explicitly-justified contrapositive when a gated stage is violated in bounds (bare RUP cannot combine the stage line with per-term bounds — this also fixes a pre-existing deterministic verification failure on the wide-domain divide/modulus tests).
- `innards::build_table_in_proof` + `install_tabulation<Hint>` + `want_tabulation` (`gcs/constraints/innards/tabulation.{hh,cc}`): the shared tabulation machinery; the linear GAC path, Multiply, Divide/Modulus, Power, Plus, and Minus all use it. The derivation branches smallest-domain-first (a line per tree node, so order matters); a caller can claim `DeterminedVariable{var, value()}` slots — the largest-domained claim is branched last and its whole level replaced by one `value()` call, sound because the encoding pins the variable by unit propagation once the others are assigned (a wrong `value()` cannot inject a tuple, and a missed one fails verification at the parent backtrack line). On a 25×25×625 multiply forced to tabulate, the proof drops from 3.2 GB / 5m37s of veripb to 24 MB / 7.2s.
- **Reified tabulation, via wildcards**: an optional `TabulationReification{var, branch()}` branches the condition variable first; values that release the constraint become one tuple that is wildcard everywhere else (its reverse implication alone makes the enclosing backtrack line RUP), enforcing values enumerate with the determined skip, negating values enumerate in full. `reify_tabulation()` maps any `ReificationCondition` onto this, so other constraints can add reified tabulation with one call.
- The Divide/Modulus decomposition multiplies into `w = x - r` with a view divisor channelled through a plain copy first. This shape is load-bearing for **soundness**, not just strength: Modulus's quotient is auxiliary and unbranched, so once the user's variables are fixed the multiplication must determine it (or fail) by propagation alone — the quotient filtering does exactly that for plain arguments, but a view fold leaves it coupled through a two-unknowns linear that bounds propagation cannot finish (caught by the test suite as spurious solutions).
- `|r| < |y|` is sign-split half-reified linear stages within the fused block (an earlier draft used `Abs` children; the crash that prompted the switch was #446, since fixed by #447).
- The tabulation enumeration includes the compound's auxiliary slot, so every rejected leaf in the in-proof derivation refutes against the constraint's own flat encoding by unit propagation; the go/no-go prototype (`tabulation_test`) established that this also verifies over the non-linear bit-product encoding.
- Unnamed proof flags dedupe by identity: two `SeqPrecedeChain`s over overlapping variables used to share `pos` flags and emit a wrong OPB (#449); the chain's flags now carry the constraint's identity.
- scp forms: `multiply` (the term `MultBC` already wrote), `divide`, `modulus`, `power`, all with reader branches and write→read→write round-trips. A constant-operand `Multiply` records its resolved `lin_equals` form.

## Testing

- Full suite green with test caps on **and off** (456 tests each), proofs verified by veripb throughout: new `multiply_test`, `divide_modulus_test`, `power_test`, `tabulation_test`, extended `plus_minus_test` and `linear_test` (including tabulated-GAC reified runs that check GAC on the condition variable), covering all three tags, aliased/duplicate operands, views, constants in every slot, zero-spanning divisor domains, constant zero divisors, negative-exponent pinned cases, and the 9^19 / 10^20 canaries via both Power paths.
- The linear GAC path's tabulation produces the same table as before the extraction — the derivation's line order has since deliberately changed (smallest-domain-first, determined-skip, wildcards), so equivalence is table-identity plus veripb, not byte-identity.
- `ortho_latin` size 5: OPB 670K → 285K (no more baked tables); recursions 371 → 515 because its 25-value domains exceed the Auto tabulation threshold, so Divide/Modulus run at decomposition-BC strength — solution counts identical. This is direct input for the threshold benchmarking follow-up.

## Follow-ups (tracked in #444)

Tabulation-threshold benchmarking (the 100 is a guess; ortho_latin suggests ~1000 would recover GAC there); #232 (true aliased bit-product encoding, for full-strength BC on `x*x`); repeated squaring for large constant exponents; cake_pb_cp encoders for the new scp forms; the scp reader's general inability to read view terms (pre-existing, applies to all constraints); chain-peel tabulation (determined-given-earlier would collapse Power's enumeration to the base's domain); dispatcher-scaffolding dedup across Multiply/Divide/Modulus/Power.

Addresses #444.

https://claude.ai/code/session_01DXLnAkxbdS8izwV2jZ4LGg


ciaranm:
The Abs justification crash mentioned in the description is now filed as #446 — it turns out to be pre-existing and entirely independent of this branch: a two-variable model (Abs over [0,1] domains) reproduces it against main with proofs on. Nothing in-tree currently triggers it since this PR's decomposition avoids Abs.

ciaranm:
Review feedback applied in b0bb03ab: `consistency::GAC` is now reserved for genuine GAC algorithms, and everything that reaches the level by enumerating a table — the linear equality family included, since its gac path was always a tabulation — takes the new `consistency::Tabulated` tag instead, making the very different cost model visible in the signature. Nothing currently takes `consistency::GAC`; it waits for the AllDifferent/GlobalCardinality/Element migrations under #299. PR description updated to match.

ciaranm:
Design discussion outcome: the child-constraint composition in Divide/Modulus/Multiply reviewed here is planned to be reworked onto exposed define/propagate functions with one flat OPB block and one fused propagator per constraint — filed as #448 with the full comparison. Not done in this PR: it gates on refactoring mult_bc into a define+propagate pair, and the current composition is correct and fully tested in the meantime.

ciaranm:
Added in 20450499: `GCS_TABULATION_THRESHOLD` overrides the Auto tabulation cut-off (getenv-once, mirroring `GCS_LINEAR_INCREMENTAL_THRESHOLD`), for autotuner sweeps. First data point: at 1000, ortho_latin size 5 goes 515 → 341 recursions — better than the old table constraints' 371, because the tabulated relation includes the auxiliary slot — with proofs verifying.

ciaranm:
#448 and #449 are now implemented on this branch, per review discussion:

**#448 (fused compounds)** — commits 2a202347..ff427fe9:
- `mult_bc` split into `define_encoding` (role-prefixed, byte-identical output) + single-pass `propagate`; MultiplyBC is a thin wrapper.
- Multiply, Divide, Modulus, and Power each emit **one flat `@c[id][role]` OPB block** and install **one propagator**: gated `propagate_linear` stages (shared `linear_stages` innards, with contrapositive gate inference) around `mult_bc::propagate` links. A constant divisor/quotient now has no multiplication at all; Power's chain is role-prefixed mult blocks under its own id; only the variable-exponent PowerTable delegation remains.
- Two composition lessons encoded in the commits: `propagate_linear` signals failure via the tracker's non-throwing path, so fused callers check `contradicted()` between stages; and the proof search-tree comparison confirmed the fused constraints propagate at least as much node-for-node (ortho_latin size 6: 1.34M vs 1.93M recursions in the fused version's favour; size-5 delta was heuristic noise).

**#449** — SeqPrecedeChain passes its identity to its single ValuePrecede child (whose position flags are id-keyed and previously deduped across chains, corrupting the model); regression test posts two chains in one proofs-on problem and aborts without the fix.

**`child_constraint_id` is deleted** — the #448 litmus test: with everything fused and SeqPrecedeChain using its own identity, the derived-child-identity mechanism has no users left.

Full suite green with caps on and off (456/456 each), proofs verified throughout.

ciaranm:
Two more commits from today's review requests:

**`99e2c758` — fixes the red CI** (the macos `divide_modulus_constraint` failure, e.g. run 28605016973). It was pre-existing and deterministic, not random: the wide-domain `divide`/`modulus` `(-20,20)(-6,6)(-20,20)` fixed test cases fail verification at the contrapositive gate inference that came in with the fused stages. Inferring ¬gate from "the gated inequality is violated in bounds" combines the half-reified stage line with a bound on every term, and RUP can't combine bounds across variables; the small-domain cases just happened to unit-propagate. The fix mirrors `justify_linear_bounds`: `justify_linear_contrapositive` materialises the stage line plus the per-term bound axioms in a pol, and the ¬gate conclusion is then genuinely RUP.

**`c38ee204` — the two `build_table_in_proof` optimisations.** Smallest-domain-first branching (the derivation emits a line per tree node, and only the leaf level is order-invariant), and skipping the last level for a functionally determined variable. A caller claims one with `DeterminedVariable{var, value()}`; the largest-domained claim goes last, its level becomes a single `value()` call per parent, and the parent backtrack lines stay RUP because the encoding pins the variable by unit propagation — that pinning is the contract, over and above semantic determination, and it's checkable: a wrong `value()` can't inject a tuple (still domain- and accept-checked), and a missed tuple fails verification at the parent line. There is no domain iteration at the skipped level at all.

Claims: Multiply/Power the result; Plus/Minus any nonzero-net-coefficient variable; Divide/Modulus the remainder slot and the dividend — not the quotient, which would need UP *backwards* through the multiplication and demonstrably fails; LinearEquality (MustHold) every term.

Benchmark, Multiply 25×25×625 forced to tabulate (`GCS_TABULATION_THRESHOLD=500000`): proof 555k lines / 3.2 GB / 14.2 s to produce → 166k lines / 24 MB / 0.2 s; veripb 5m37s → 7.2 s. Both verify as complete enumerations of the same 625 solutions.

Both suites re-run green at the head: 456/456 caps-on and caps-off. Note `want_tabulation` still budgets for the full tree; letting `Auto` account for a skippable level is folded into the #444 threshold-benchmarking follow-up, along with a possible chain-peel generalisation (determined-given-*earlier*, which would collapse Power's enumeration to just the base's domain).

ciaranm:
One more from review: `d08bff91` — `want_tabulation` now takes the determined-variable claims and leaves the largest-domained one out of the Auto budget, matching what the derivation actually costs. So a 10×10×100 multiply now tabulates at the default threshold of 100 (verified: the table comment appears in the proof and it checks as a complete enumeration), while the wide-domain divide/modulus test cases still correctly fall back to BC. Multiply's constant-operand delegation to LinearEquality applies the same rule over the sanitised terms. Suites re-run green: 456/456 caps-on and caps-off.

ciaranm:
\`80328e0d\` — reified tabulation, via wildcards. \`build_table_in_proof\` takes an optional \`TabulationReification{var, branch()}\`: the condition variable is branched first, and each of its values classifies the whole subtree — **Free** collapses to one tuple that is wildcard everywhere else (\`propagate_extensional\` already understood wildcards; the selector RED implications just skip wildcard slots, and the tuple's reverse implication alone makes the enclosing backtrack line RUP, so a released condition value costs one tuple instead of the term product), **Holds** enumerates with the determined-variable skip, **Negated** enumerates in full. \`reify_tabulation\` maps any \`ReificationCondition\` onto this plus the wrapped acceptance test, and filters the determined claims (they survive only where some branch enforces the base relation, never on the condition variable) — so wiring reified tabulation into another constraint is one call on top of its existing accept/determined pieces.

\`ReifiedLinearEquality\` now uses it: \`reif::If\`/\`NotIf\`/\`Iff\` tabulate instead of throwing \`UnimplementedException\`, giving GAC over the terms and the condition variable together, and MustHold/MustNotHold enumerate exactly as before. The linear tests grow a tabulated flavour of the reified runner (eq_if/eq_iff/ne_if/ne_iff × the view sweep × cond domains {0}, {1}, {0,1}), checking GAC on all four variables, proofs verified. Suites green at the head: 456/456 caps-on and caps-off.

One deliberate scope note: the determined-skip contract on Holds branches reads "given the condition value and every other variable, UP pins the term through the half-reified equality" — the single-free-variable case again, and veripb agrees across the test sweep.

ciaranm:
## Independent code-quality audit — reviewed at `80328e0d`

An independent auditor pass over the whole diff (`6d8e2c06..80328e0d`, +4342/−1064, 73 files), split across five focused reviewers: the `#299` consistency surface, the tabulation + linear proof machinery, `Multiply`, `Divide`/`Modulus`/`Power`, and the frontends/scp/examples/docs. No build or VeriPB run here, so the soundness observations are static reasoning about the emitted RED/RUP structure (which the in-tree tests exercise under VeriPB).

**Verdict: strong work — no soundness or correctness bug found.** Overflow is handled (checked-throw or saturate), the structural dispatch is exhaustive across every operand shape (both-constant / one-constant / views / all four alias forms), the extracted tabulation proof machinery is behaviourally faithful (table-identical; determined-skip and wildcard enumeration trace clean; labels don't collide), and the tests genuinely test — independent C++ oracles, GAC checks where `Auto` is guaranteed to tabulate, VeriPB on every proof. The `#299` consistency design (a `std::variant` of tag types in the constructor, so an unsupported level is a *compile error* and the signature documents the supported set) is a clean, self-documenting mechanism. And `Constraint::s_expr` being pure-virtual (`= 0`) now makes "forgot to override `s_expr`" a compile error rather than an infinite recursion.

Everything below is craft / dead-code / doc-drift / family-consistency. **Nothing is merge-blocking.**

### Already handled — please don't redo
- **#446** (Abs `bad_variant_access`) — closed, pre-existing, regression-tested (`abs_test.cc:136-145`). Verified independently: `abs/justify.cc` is unchanged on this branch, resolves atoms through a `PolItem` variant, and has zero remaining `get<ProofLine>` sites. The PR description's "possibly a latent bug worth its own issue" is **stale** — that path is currently sound, and the divide/modulus code comments correctly do not repeat the claim.
- **#448** (fuse compound arithmetic constraints) — already implemented here; `child_constraint_id` is gone at head. The PR description still describes that removed machinery.
- **#449** — the `seq_precede_chain` change is its fix (nicely tested with a two-chain cross-product).

---

### Theme A — the PR *description* has drifted from the branch (description-only; no code change)
Three claims went stale as the branch evolved:
1. Describes the removed `child_constraint_id` machinery (see #448).
2. Repeats the fixed-Abs rationale (see #446) as the reason for using sign-split half-reified linears instead of `Abs` children.
3. "linear GAC proof output byte-identical before and after" is no longer true at head — commits `c38ee204` (order enumeration smallest-domain-first) and `d08bff91`/`80328e0d` (skip the determined level; reifications via wildcards) deliberately reorder the RED/RUP lines. The **table is provably the same set of tuples** and VeriPB still passes — so it is *behaviourally faithful / table-identical*, just not byte-identical.

*Recommendation:* refresh the description before merge. Trivial.

### Theme B — dead `using`/include residue in every new/rewritten arithmetic `.cc` (the single most-actionable sweep)
- `using std::make_shared;` was **added by this PR but is unused** in all five: `plus.cc:32`, `minus.cc:32`, `multiply.cc:32`, `divide_modulus.cc:37`, `power.cc:31` (`install_tabulation` owns the `shared_ptr` now).
- `plus.cc` / `minus.cc` left the old `s_exprify` support dead after the rewrite: unused `using std::string;` / `std::stringstream;` / `std::print;`(`fmt::print;`) + `#include <sstream>`.
- `multiply_bc.cc` carried through the move: duplicate `#include <sstream>` (`:14` **and** `:26`), unused `#include <iostream>`, unused `using std::stringstream;`, dead `print`.
- `divide_modulus.cc` also: unused `using std::pair;` (`:45`), `using std::size_t;` (`:46`).

*Recommendation:* one trivial sweep across the five files.

### Theme C — overflow-robustness asymmetry across the arithmetic family (latent; unreachable under `Auto`)
- `divide_modulus.cc:188-189` computes the `w = q·y` product bound with raw `Integer::operator*`, which **throws** on `long long` overflow at install time. The twin `power.cc:211-214` **saturates** (`saturating_product`) for exactly this reason; `multiply.cc:146-147` also throws but that is documented/intended (`multiply.hh:43-46`). `w` equals `x − r`, which is representable whenever `x` is — bound it from there instead.
- `minus.cc:192` negates first — `__builtin_add_overflow(av, -bv.raw_value, &sum)` — leaving `-bv.raw_value` itself unguarded (UB at `LLONG_MIN`), whereas `plus.cc` is fully guarded. `__builtin_sub_overflow(av, bv, &sum)` is cleaner and complete.
- Broader: the plus/minus `accept` lambdas guard with `__builtin_*_overflow` (return `false`), but the `determined` `value()` lambdas and linear `base_accept` use throwing `Integer` arithmetic — two policies for one table build.

Cannot bite under `Auto` (threshold 100 keeps values tiny); only forced `Tabulated` on pathological domains. *Recommendation:* one policy, ideally a `product_bounds(lo1,hi1,lo2,hi2, optional cap)` helper in the new `arithmetic_utils.hh`; `__builtin_sub_overflow` in minus. ~1–2h.

### Theme D — migration strength-consistency (one user-facing)
The **linear** GAC sites were migrated faithfully (`LinearEquality{..., true}` → `consistency::Tabulated{}`). Two **arithmetic** explicit-GAC sites were instead relaxed to the `Auto` default:
- **`examples/shikaku/shikaku.cc:166`** — `MultBC{w,h,area}` (explicitly BC) → `Multiply{...}` (`Auto` → tabulates to GAC for any grid ≤ 10×10). That makes three things **false**: the code comment (`:165` "deliberately weak bounds-consistent product"), the printed help banner (`:124` "bounds-consistent multiplication -- --autotable strengthens it"), and the whole `--autotable` opt-in (`:174`). **The one worth fixing before merge** — not a correctness bug (GAC is sound), but a documented-contract change that leaves the example contradicting itself. Pin `consistency::BC{}` (or rewrite the comment + banner if `Auto` is intended).
- `examples/langford/langford.cc:92` — `PlusGAC` → `Plus` (`Auto`) is the same class but benign (operand size `2k` tabulates for realistic `k`). Pin `Tabulated{}` for consistency, or confirm `Auto` is intended.

### Theme E — duplication across the three top-level dispatchers (related to #448, but distinct)
`multiply.cc`, `divide_modulus.cc`, `power.cc` share structurally-identical dispatcher scaffolding (affine fold → aux vars → `mult_bc::define_encoding`/`propagate` → `want_tabulation`/`install_tabulation`), a **byte-identical `watch` lambda** (`divide_modulus.cc:246-249` == `power.cc:235-238`), and the box-corner product-bound (`min`/`max` of the four corner products) in three places. `arithmetic_utils.hh` is the natural home (fold Theme C's cap logic in here too). Distinct from #448 (children → define/propagate); this is the top-level dispatchers.

Also: `multiply.cc:80-108` re-computes the exact `Auto` budget rule that `want_tabulation` / `default_tabulation_threshold` own, because `LinearEqualityConsistency = variant<BC, Tabulated>` has no `Auto` for `Multiply` to delegate to. Two homes for one rule → silent-drift risk (worst case a suboptimal BC-vs-Tabulated pick, not a correctness bug). An `Auto`-accepting `LinearEquality` entry point, or a shared budget helper, would give the rule one home.

### The nit pile (trivial)
- **`minus.hh:15-25`** — the new `MinusConsistency` alias+doc was inserted *between* the `Minus` `\brief` and `class Minus`, stranding the brief; **`Minus` now renders with no documentation** (`plus.hh` gets the order right). Move the class brief down to sit immediately above `class Minus`.
- **`power.cc:222-223`** — `if (is_last && t != (result_plain ? *a3.var : t)) ; // unreachable; kept for clarity` — an `if` guarding an empty statement whose own comment admits it is unreachable. Delete.
- **`consistency.hh:99`** — `consistency::VC` is accepted by no constraint and referenced nowhere, so a user who writes `consistency::VC{}` gets a compile error with no hint that nothing will ever accept it. (`GAC` at `:58` is defensibly reserved for genuine GAC algorithms; `VC` is pure speculative surface.) Drop `VC`, or document it as forward-looking.
- **`power.cc:51-54`** — `clamp_to(v, m) = min(max(v,-m), m)` reinvents `std::clamp(v, -m, m)` (`Integer` has `<=>`).
- **`propagators.hh:313-318`** — empty `/** \name Support variables. */ ///@{ ///@}` group left behind by `create_auxilliary_integer_variable`'s removal. Delete.
- **`multiply.cc:139`** — `… && u1 != m2` is always true here; it misleads the reader into thinking `u1 == m2` is reachable. Drop it (or note it's defensive).
- **`multiply_bc.cc:925,1159`** — `filter_quotient`'s `vector<IntegerVariableID> & all_vars` is never read (only threaded into its own recursion); `propagate` rebuilds it solely to pass it. Remove the parameter and the local.
- **`python/gcspy.cc:277-298`** — the `WRITE_API_CALLS` debug strings emit uncompilable source (missing the `result` operand, unbalanced parens: `p.post(Multiply(v1 , v2);`), and the `pow` branch emits `Pow(` for a `Power(...)` call. Fix or delete the dead emission.
- **`dev_docs/xcsp.md:135-137`** and **`xcsp_glasgow_constraint_solver.cc:1107-1108`** — both still list `OPOW` as a `post_product` caller; `OPOW` now posts `Power` directly. And **`dev_docs/xcsp.md:293-301`** omits `Power` from the "posts `Multiply`/`Divide`/`Modulus`" list.
- **`tabulation.cc:239-259`** (`want_tabulation`) — takes no reification and always drops the largest determined level, but under a `Negated` branch that level is enumerated in full, so a future `Auto`+reified constraint could exceed the budget. Currently unreachable (linear reif is `Tabulated`-only; all `Auto` consumers are non-reified). Thread the reification in, or document the precondition.
- **`tabulation.cc:129,162`** — `current_branch` is function-scope mutable state set at `depth==0` and read at the skip level; correct, but a reader has to prove the non-interleaving invariant. Passing `branch` as a `search()` parameter would make it self-evident.
- **`scp_reader.cc:812-846`** — the `power` / `multiply` reader branches repeat the `terms.size()!=3` → `resolve_variable_list` → post-ternary skeleton (`divide`/`modulus` already share one copy). Optional `resolve_ternary` helper.
- **`tabulation.cc:52-56,70-74`** — a couple of hand-rolled "scan for a matching `IntegerVariableID`, remember the index" loops read more clearly as `std::ranges::find` + `std::distance`.
- **`CMakeLists.txt:349`** — `add_test(NAME tabulation ...)` is off-pattern vs peers (`plus_minus_constraint`, `multiply_bc_constraint`); rename for grep-ability.

### Done well (worth keeping as-is)
- `Constraint::s_expr` pure-virtual + `s_exprify()` removed — the override-neither footgun is now a compile error.
- The consistency design: per-constraint `variant` over empty tags, unsupported levels = compile error, `Auto`/`BC`/`Tabulated` consistent across the six siblings, the `GAC`-vs-`Tabulated` cost-model split well-explained. No ad-hoc trailing `bool` survives, and no `innards::` type leaks into any public signature/return/callback alias.
- Structural dispatch is exhaustive and correct (every operand shape traced); the `mult_bc/` → `multiply/` move is a clean, behaviour-preserving extraction.
- `install_tabulation<Hint_>` is a justified template (the hint type must be known at compile time), backed by real instantiations; the proof-soundness contract is documented *for the reader* ("a wrong `value()` can't inject a bad tuple — `accept` re-checks and a missed one fails at the parent backtrack line"). The threshold lives behind `default_tabulation_threshold()` + a `GCS_TABULATION_THRESHOLD` env override, and the node-product is overflow-safe.
- `AffineForm`/`affine_of` in `arithmetic_utils.hh` is genuine consolidation, not reinvention.
- Tests genuinely test: independent oracles (including the full MiniZinc power rule), real structural equality on scp round-trips (`scp_a == scp_b` **and** non-empty), aliases/views/constants-in-every-slot/wide domains/`9^19`/`10^20`/`0^0`/negative exponents via both Power paths, VeriPB throughout. `checked_integer_power` is overflow-safe; div-by-zero relational semantics and truncation-sign pinning are correct and tested.

---

*Independent quality audit by a separate "grumpy auditor" Claude Code instance (Opus 4.8), isolated from the developer instance's context, over commit `80328e0d`. Static review only — not built or VeriPB-checked here; treat the soundness observations as pointers to verify, not verdicts. Per this repo's AI-agents policy, a qualified human should check each item before acting.*

ciaranm:
Triage of the audit (five commits, `cc9481e1..c46461ee`, plus a PR-description refresh; suites re-run green 456/456 caps-on and caps-off):

**Taken.**
- *Theme A*: the description is rewritten at head — fused #448 encodings instead of `child_constraint_id`, the Abs rationale now points at #446/#447 as fixed, and "byte-identical" is corrected to table-identical-plus-veripb (the derivation order changed deliberately with the tabulation optimisations, which the description now also covers).
- *Theme B*: full sweep, all confirmed dead (`cc9481e1`).
- *Theme C, the two real items* (`db5a9276`): minus's `add_overflow(av, -bv)` → `sub_overflow` (the negation was genuine UB at `LLONG_MIN`); Divide/Modulus's `w` auxiliary now bounded from `w = x − r` instead of the throwing `q·y` corner products — reachable for large-but-sane dividends (x≈10¹², y≈10⁷), and usually tighter anyway, a sum of ranges rather than a product.
- *Theme D* (`8f464ed9`): shikaku pinned to `consistency::BC{}` with a comment saying why (the whole `--autotable` story depends on the product staying weak); langford pinned to `Tabulated{}` (it was `PlusGAC`, and `Auto` would drop to BC for k > 5).
- *Nit pile* (`c46461ee` + `2cb74d2d`): minus.hh doc order, the unreachable `if`, `std::clamp`, the empty doxygen group, the always-true `u1 != m2`, `filter_quotient`'s unread `all_vars`, `consistency::VC` documented as reserved for the #299 AllDifferent migration, the ctest entry renamed `tabulation_test`, gcspy's `WRITE_API_CALLS` strings made compilable, the OPOW/post_product doc drift in both places, `want_tabulation`'s unreified-budget precondition documented, `search()` takes the branch as a parameter, and the position scans are `find()`.

**Declined, with reasons.**
- *Theme C's "one overflow policy"*: the remaining throwing paths (`value()` lambdas, linear's accept) need domain *values* near 2⁶³ under explicitly-forced `Tabulated` — at which point a loud `Integer` throw is the failure mode we want, and the plus/minus `accept` guards stay because accept must classify every enumerated tuple cheaply. Guarding five more lambdas buys noise, not safety.
- *Theme E*: real, but a structural refactor of three files mid-review would cost more churn than it saves; added to the #444 follow-up list (along with an `Auto`-accepting `LinearEquality` entry point as the fix for the budget rule's two homes — the multiply copy carries a cross-reference comment meanwhile).
- *scp_reader `resolve_ternary`*: two nearly-identical branches don't yet justify the indirection.

One factual correction to the audit: `divide_modulus.cc`'s corner-product install-time throw was described as "unreachable under Auto; only forced Tabulated on pathological domains" — it is independent of the consistency level (the auxiliary is created on every non-constant path), which is why it got fixed rather than noted.

PR #463, opened 2026-07-03
TEMP: C++23 feature probe (do not merge)

ciaranm:
Compile-only CI probe to confirm `std::optional` monadic ops (and_then/transform/or_else) and `std::vector::append_range` build on every supported toolchain (Ubuntu 24.04/26.04 GCC, macOS libc++). Will be reverted once CI records the answer. **Do not merge.**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,6 +248,9 @@ install(FILES
         ${CMAKE_CURRENT_BINARY_DIR}/GlasgowConstraintSolverConfigVersion.cmake
         DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/GlasgowConstraintSolver)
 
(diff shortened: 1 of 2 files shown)


ciaranm:
Probe complete. Results across the CI matrix:
- **std::optional monadic ops** (`and_then`/`transform`/`or_else`): available on **all** supported toolchains, including Ubuntu 24.04's GCC 13 (the static_assert on line 11 compiled cleanly there) and macOS libc++ (both lanes passed the full build).
- **std::vector::append_range**: still **unavailable** on Ubuntu 24.04's GCC 13 (`__cpp_lib_containers_ranges` undefined). Available on GCC 15 and macOS libc++.

So we can adopt the monadic optional ops now; append_range must wait until we drop Ubuntu 24.04 / GCC 13. Closing without merging.

PR #471, opened 2026-07-05
Arithmetic clean rewrite: Multiply / Divide / Modulus / Power

ciaranm:
Clean from-scratch rewrite of Multiply, Divide, Modulus and Power, replacing the internals in place: same public classes, `.with_consistency()` API, frontends, `s_expr` shapes, and the same cake OPB encodings (byte-identical for plain shapes, checked against goldens throughout). Proofs are done cleanly — no in-proof reformulation, no fusion-resolution engine, no hint scavenging — and views and repeated variables are supported natively.

**Do not merge without explicit go-ahead.** All eight stages are complete; every commit leaves the tree green.

### What changed

- **`product_bounds.hh`** — the pure interval arithmetic (JaCoP quotient filter, exact square/sqrt hulls), exhaustively tested against defining inequalities and enumeration.
- **`product_encoding.{hh,cc}`** — cake's encoding as slot-keyed emitters (never keyed by variable id — the root cause of the old #232 alias throw).
- **`product_justify.{hh,cc}`** — the justification core: thesis Ch. 7 procedures 7.1/7.2 over the grid, sign-case channelling, and `conclude_by_sign_cases`, a fixed-shape nested-cut driver replacing the old searching resolution engine. Every helper was verified in isolation in a fragment harness (`product_justify_test`, 20+ veripb-checked fragments) before any consumer used it.
- **`signed_multiply.{hh,cc}`** — the shared relation layer for Multiply and Power's chain links, with native squares (exact square/sqrt filtering on the true relation).
- **Divide/Modulus** — single route; the product `w = |q||y|` exists nowhere (not in the OPB, State, or proof); `introduce_bits_of` and the index-coupled stage fusing are gone; views taken directly.
- **Deleted**: `multiply_bc.{hh,cc}` (~1,300 lines: fusion resolution, sign-bit channelling with the `keep_sign_atom` fudge, `DerivedPBConstraint`, scheme flags, 14 defensive throws), the aux copy/fold layer in Multiply, the legacy divmod route, `aux_power_base`/`aux_power_copy`.

### New capabilities

- `(multiply X X Z)` and `(multiply X Y X)` are legal, propagate on the true relation (`x*x=z` infers `z ≥ 0` and sqrt hulls at BC), and **chain-verify against cake_pb_cp** — three new `scp_cases`. cake's soundness theorems have no distinctness hypotheses, verified by probe in S0.
- View shapes flow through every row natively (self-verify; cake's grammar has no view terms — documented). `add_view_tests` registered and green for multiply, divide/modulus, and power.

### Benchmarks vs main (fixed instances)

Search identical everywhere (recursions/propagations match; node parity was enforced per stage against baselines). Proof trade-offs, honestly:

| instance | pbp size | emit | veripb check |
|---|---|---|---|
| multiply (prime, unsat-ish) | **−9%** | par | par |
| multiply (wide) | **−15%** | −33% | **par** (0.9–1.1×) |
| divide (wide) | **−35%** | −8% | **−39%** |
| modulus (wide) | **−96%** | −93% | **−36%** |

(Updated after two follow-up commits. **Complete RUP hints**: operand bounds are single-antecedent `ia` steps citing the atom's defining row; result-bound claims carry complete hint kits, including pol-derived order-ladder bridge clauses — unit propagation can't cross between two order atoms of the same variable through bit definitions alone, and the tracker's own ladder lines aren't citable. The multiply wide check went 343s → ~97–117s vs main's 109s. **Chain-line cache**: divide/modulus re-derived the same grid chains and operand bounds for every inference; each is now derived once and cited — magnitude-corner chains, filter grid lines and operand bounds in constraint-lifetime maps at Top (their value keys repeat heavily), x-side chains in a per-pass cache at Current (their keys churn, and standing copies would make every hint-free RUP step pay for them — a full-Top experiment cost 9× check time on modulus). Divide wide: 349M → 60M vs main's 91M, check 201s → 123s; modulus wide 56M vs main's 1516M. Multiply's proofs are byte-identical through both follow-ups apart from the hints themselves.)

### Acceptance

No `AssertRatherThanJustifying` / `introduce_bits_of` in the arithmetic code; no legacy symbols; public headers changed in comments only; tree-wide clang-format clean; 490/490 ctest; all 26 arithmetic chain cases pass full workflow-2 with `CAKE_PB_CP` set; ASan+UBSan clean on all four arithmetic tests plus the fragment harness; capped and caps-off soaks green (5× each for divide/modulus, 3× for multiply). Design doc: `dev_docs/arithmetic-proofs.md`.

https://claude.ai/code/session_01QnzpNyXwHCxuPaBoCFM5kn


ciaranm:
@mmcilree would be interested in your thoughts. I asked Claude to redo these constraints from scratch + your thesis, and to match the new Cake encodings cleanly (rather than through reformulation, which is how it's currently done). Claude is currently working on putting hints back in.

gcs/constraints/innards/product_justify.cc
```diff
@@ -0,0 +1,424 @@
...
+    // V-form, so the line cancels against the V-form channel rows. One `ia`
+    // citing the bound atom's definition, rather than a database-wide RUP.
+    auto sum = lower ? WPBSum{} + 1_i * v : WPBSum{} + -1_i * v;
+    auto rhs = lower ? bound : -bound;
+    // The `ia` shape only lines up when the cited bound is itself carried by
+    // the reason (the drivers always arrange this); otherwise fall back to RUP.
+    auto def = reason.empty() ? std::nullopt : def_line_for(logger, lower ? v >= bound : v < bound + 1_i);
+    auto line = def ? logger.emit_under_reason(ImpliesProofRule{*def}, logger.reify(sum >= rhs, HalfReifyOnConjunctionOf{}), result_level, reason)
+                    : logger.emit_rup_proof_line_under_reason(reason, sum >= rhs, result_level);
+    return ConditionalBound{sum, rhs, HalfReifyOnConjunctionOf{}, line};
```
mmcilree, line 133: I don't think either RUP or IA is actually necessary here. The bound id should be obtainable by doing `need_pol_item_defining_literal` for the corresponding bound lit.

mmcilree (changes requested):
So in terms of code style, I'd say this is cleaner / more readable than what I had, and it does look like it's following the thesis procedures pretty faithfully. However, I did a bit of performance probing and I'm worried it's a verification time regression, even with the RUP hints. I ran `multiply_random` to find some of kind of seeds I was plotting in the thesis (large number of propagations/recursions for a trivial one-constraint problem) and benchmarked this branch against main for some of the bad seeds.

It seems the opb and pbp files are a little bit bigger, which could be a tolerable constant factor, particularly given the proof logging overhead is about the same. But verification time appears to be blowing up, starts at 2-3x for easier seeds and is climbing 5-6x and worse for harder ones. Given that mult_bc is already by far the worst for verification vs solve time without proof logging overhead we might want to avoid this. Alternatively we could say it's a price worth paying for cake conformance for now, considering the trimmer justifier should hopefully address this sort of gap in the future. 

I'm not entirely sure from looking what's causing the slowdown, but my first guess would be the handful of remaining unhinted IA/RUP steps + the unnecessary pol link rederivations are probably the culprit. 

Can probably just ask Claude to test the seeds below itself and see if it can figure it out though.

## N = 5000

| seed | impl | solve no-proof (s) | solve w/ proof (s) | .pbp (MB) | .opb (KB) | verify (s) |
|---:|:--|---:|---:|---:|---:|---:|
| 103 | branch | 0.008 | 1.235 | 45.8 | 38.8 | 4.396 |
| 103 | main | 0.008 | 1.237 | 41.7 | 33.2 | 1.672 |
| 220 | branch | 0.007 | 1.501 | 51.0 | 42.1 | 6.627 |
| 220 | main | 0.008 | 1.291 | 44.1 | 33.6 | 2.710 |
| 429 | branch | 0.006 | 1.184 | 45.7 | 45.6 | 4.834 |
| 429 | main | 0.007 | 1.132 | 39.6 | 36.4 | 2.241 |
| 843 | branch | 0.006 | 1.290 | 47.0 | 45.9 | 5.670 |
| 843 | main | 0.007 | 1.663 | 40.8 | 36.5 | 3.786 |
| 987 | branch | 0.006 | 2.299 | 40.9 | 45.7 | 5.962 |
| 987 | main | 0.007 | 1.019 | 35.5 | 36.4 | 1.898 |

## N = 15000

| seed | impl | solve no-proof (s) | solve w/ proof (s) | .pbp (MB) | .opb (KB) | verify (s) |
|---:|:--|---:|---:|---:|---:|---:|
| 232 | branch | 0.011 | 2.832 | 97.4 | 53.3 | 17.995 |
| 232 | main | 0.011 | 2.498 | 87.5 | 45.0 | 3.949 |
| 457 | branch | 0.015 | 4.088 | 159.8 | 57.1 | 45.895 |
| 457 | main | 0.018 | 4.256 | 143.1 | 48.0 | 7.960 |
| 707 | branch | 0.013 | 3.358 | 130.6 | 53.0 | 28.425 |
| 707 | main | 0.016 | 4.025 | 118.6 | 44.8 | 6.075 |
| 1030 | branch | 0.012 | 3.465 | 138.2 | 61.2 | 25.703 |
| 1030 | main | 0.014 | 3.399 | 122.0 | 51.2 | 7.835 |
| 220 | branch | 0.019 | 4.730 | 188.8 | 53.1 | 92.160 |
| 220 | main | 0.022 | 4.524 | 162.3 | 42.3 | 18.706 |

ciaranm:
Claude dug into this as suggested, with your seeds and an instrumented veripb (per-rule-type timing added to `execute_rule`). Verdict: it wasn't the pol rederivations or the long hint lists — the branch was accidentally **minting a few fresh atoms per search node**, and everything else follows from that.

**The mechanism.** `channel_grid_bound_to_result` built its claim hints by calling `add_condition_def_hints` on `z >= a_lb*b_lb` — the *raw grid product value*, which is fresh at every inference and often far outside z's domain (the live database at proof end contains atoms like `i[z][ge-10003950]`). `need_pol_item_defining_literal` creates atoms that don't exist yet, so each justification permanently added ~4 atoms ≈ 19 constraints (two bit-composition defining rows plus ladder links each): on N=15000 seed 220 (2084 nodes) the checker ends with 54,906 live constraints against main's flat 15,321. A hint-free `rup` costs O(live database), and there are ~11 of those per node (4 top-level, 7 inside the sign-case `red` subproofs), so checking went quadratic in node count — the marginal cost per 200 nodes grows 0.40s → 9.88s through the proof, where main stays flat at ~0.3–0.7s. That's also why the ratio worsens with N: more nodes, and every node pays for all its predecessors.

Measured breakdown of that seed's 44.7s (this machine): 37.4s in hint-free rup steps, whose average cost grows 13× from the start of the proof to the end. The ~170-hint rup lines cost 0.067s *in total* (hinted rup really is O(|hints|)), the hinted `ia` steps 0.027s, all pol 0.5s. So your unhinted-RUP guess named the vehicle, but the leak was the driver: with it plugged, the same steps cost ~4s. Amusingly, the old code's habit of re-deriving identical link lines is why main never had this problem — veripb dedups syntactically identical constraints, so re-derivations don't grow the database, whereas per-node atom definitions are all distinct.

**The fix** (`de4727e9`): drop the claimed-bound def hints. The claim is stated over z's *bits*, so the minted atom appears in no constraint the checker ever propagates — those hints were dead weight; the reason/case-literal hints (kept) do the actual bridging. Search and the OPB are unchanged (byte-identical `.scp`), proofs get slightly smaller, and multiply / divide_modulus / power / product_justify / product_bounds all still verify.

Your tables re-run on this machine, fixed branch vs main (verify seconds; before-the-fix, branch was 2.57s on N=5000 seed 220 and 44.71s on N=15000 seed 220):

| N=5000 seed | fixed | main |
|---:|---:|---:|
| 103 | 0.96 | 0.82 |
| 220 | 1.04 | 1.02 |
| 429 | 0.87 | 0.86 |
| 843 | 0.89 | 0.90 |
| 987 | 0.75 | 0.77 |

| N=15000 seed | fixed | main |
|---:|---:|---:|
| 232 | 2.21 | 1.85 |
| 457 | 4.38 | 3.46 |
| 707 | 3.37 | 2.72 |
| 1030 | 3.41 | 3.63 |
| 220 | 6.44 | 5.08 |

Checker memory is now below main's (48MB vs 55MB on the worst seed; it was 114MB). The residual ≤1.3× at N=15000 is a constant factor: the remaining ~1/node top-level and 7/node subproof hint-free rups each propagate a fatter big-M claim than main's short clauses (840µs vs 72µs per step, database sizes now equal). If we want that gone too, the next step is hinting the sign-case subproof closures in `conclude_by_sign_cases` — can do here or as a follow-up.

gcs/constraints/innards/product_justify.cc
ciaranm: Plausibly, yes — citing the def row's forward half directly in the pol chain would save the restatement. Two notes from the perf dig (details in the main thread): this isn't where the time went — the hinted `ia` steps measure 0.027s of the worst seed's 44.7s, so it's a proof-economy cleanup rather than the regression fix — and citing the def row directly leaves its big-M `~[v>=b]` term riding through the later additions, which needs probing against the exact-value accounting in the grid sum bounds before switching. Queued as a follow-up.

ciaranm:
Addendum: `287862d3` also takes up the residual constant factor mentioned above, by hinting the sign-case subproof RUP steps in `conclude_by_sign_cases` (the five dead-pattern clauses and two closing contradictions per inference, ~160µs each hint-free → under 2µs hinted). Updated numbers, same machine (verify seconds):

| N=5000 seed | branch | main | | N=15000 seed | branch | main |
|---:|---:|---:|---|---:|---:|---:|
| 103 | 0.69 | 0.82 | | 232 | 1.55 | 1.85 |
| 220 | 0.76 | 1.02 | | 457 | 2.87 | 3.46 |
| 429 | 0.66 | 0.86 | | 707 | 2.30 | 2.72 |
| 843 | 0.67 | 0.90 | | 1030 | 2.25 | 3.63 |
| 987 | 0.58 | 0.77 | | 220 | ~4.0 | 5.08 |

So every seed from the review now verifies 20–40% faster than main (the worst seed started this morning at 44.7s). The only hot hint-free RUP left is the one-per-node search-skeleton backtrack clause, which costs main the identical ~1.7s on the worst seed — that's generic solver machinery, not this PR's constraints.

One instructive wrinkle: the hints are only complete when the premises' sum terms cancel exactly against the negated goal, which product bounds arrange but divide/modulus premises don't (they drag view bits and aux-atom rows into the cut result, where only a database-wide RUP closes — `divide_modulus_test` caught this immediately on a view-typed instance). So the hinting is opt-in per call site (`SubproofRUPHints::Assemble`); multiply, squares and power engage it, divide/modulus stay on the hint-free path they were already winning on. Validated by all five arithmetic suites with verification plus a 300-instance `multiply_random` sweep over n ∈ {5, 20, 100, 1000}, all proofs checked; proof files grow ~2%.

gcs/constraints/innards/product_justify.cc
ciaranm: Tried it — you're right, and it's in as `4606915b`: `derive_operand_bound` now hands back the definition line directly, no `ia`, no RUP. The gate term rides through the pol chains as one more reason-shaped rider and dies where the reason's units pin the atom; proofs shrink ~3% on your worst seed, everything verifies (all five arithmetic suites + a 300-instance sweep).

Interesting counterexample though: the identical swap in `derive_assumed_operand_bound` fails verification. Its `ia` restatement turns out to be load-bearing — it renormalises the definition's gate big-M to reify's choice, and the syntactic `ia` row implications in `grid_sum_lower_bound` need that exact coefficient when an assumed bound composes with view channel rows. Power's factor paths hit exactly that (`power_test` fails "not syntactically implied"); plain multiply happens to line up, so it only shows on views. Documented in a comment on that function.

ciaranm: Correction to the above: CI falsified it. `4606915b` passed the full local battery (all five suites + a 300-instance sweep) and then failed power and multiply with "not syntactically implied" on three of five CI lanes — the same signature the assumed-bound variant showed locally. The restatement is load-bearing in *both* flavours: the downstream syntactic `ia` row implications in the grid procedures are coefficient-exact, the definition's gate big-M differs from reify's choice, and whether the dominance check happens to pass depends on the instance's bound values — which vary with search order, hence memory layout, hence lane. Reverted; both functions keep their `ia`, with the reason documented in the code and in dev_docs/arithmetic-proofs.md (including the general lesson: local batteries validate hint completeness, not restatement removal under `ia` implications).

Issue #416, opened 2026-06-28
Explore per-propagator idempotence (skip self-requeue) to avoid redundant propagator runs


ciaranm:
## Idea

Many propagators are *idempotent*: they reach their own fixpoint in a single call (they
process all cascades internally), so running them again immediately would infer nothing. Yet
the propagation queue still re-wakes a propagator whenever a variable it watches changes —
**including changes the propagator made itself**. For an idempotent propagator that
self-requeue is pure waste: a full re-run that does nothing.

Avoiding it (don't re-wake a propagator from its own inferences) is the classic "idempotent
propagator" optimisation (e.g. Gecode). It does not change the search tree when the propagator
really is idempotent — it only removes wasted no-op runs.

Note: ~50% of all propagator firings on our benchmarks are no-ops today (e.g. tsp 91.0M total /
45.1M effectful; magic_square 94.2M / 40.4M), so the headroom is large in principle.

## Mechanism in our engine

`Propagators::propagate` (gcs/innards/propagators.cc) is a staged single queue (Schulte &
Stuckey). Requeue is **batched at round boundaries**: when the ready band drains, the round's
accumulated inferences (`tracker.each_inference()`) are replayed to wake the next round's
propagators. A propagator that changed vars it watches is re-woken there by its own inferences.

To skip *only* self-requeue you need to attribute each inference to the propagator that
produced it. Inferences are appended to a `std::deque` in order, and each propagator's run
appends a contiguous range — so recording `(propagator_id, inference_count_before_run)` per run
gives the attribution cheaply, and the round-boundary requeue can then skip waking a propagator
from inferences it produced itself (when that propagator is flagged idempotent).

## Experiment + findings (2026-06-28)

Prototyped as a global, env-var-gated "treat every propagator as idempotent, skip all
self-requeue" (with per-run source attribution), measured against the current engine
(Release `-O3`):

| benchmark | recursions (off → on) | result |
|---|---|---|
| **ortho_latin** (vc all-diff, UNSAT) | 501,387 → **501,387** (identical) | **−24% wall time, identical tree** — the ideal case |
| magic_square (vc all-diff) | 6,042,079 → 7,702,401 (**grew**) | +15% slower (under-propagation; sound but more search) |
| tsp (circuit/prevent) | 4,285,765 → 2,223,279 (changed) | **UNSOUND** — reports tour length 2014 < true gr17 optimum 2085 |

**Conclusions:**
[...]

PR #474, opened 2026-07-08
Per-run propagator idempotence claims (EnableButIdempotent) + wave-1 adopters

ciaranm:
Implements the revised plan on #416: a per-run idempotence claim, `PropagatorState::EnableButIdempotent`, the wave-1 adopters (VC and GAC AllDifferent, Table/`propagate_extensional`, LinearLessThanEqual/GreaterThanEqual), the wave-2 Element claims, and the plan's Phase D refinement (skip re-waking a claimant from anything it had already seen). One commit for the mechanism (zero behaviour change on its own), one per adopter with its audit note in the code, and one for the already-seen refinement, so anything can be reverted in isolation.

## Mechanism

A propagator whose run reached its own fixpoint returns `EnableButIdempotent`: re-running it immediately, against the domains exactly as the run left them, would infer nothing and not contradict. Because the store applies inferences immediately and runs are serial, the run has seen every inference recorded up to its end — so the queue skips re-waking it from any of those (its own included), and only changes recorded after its run ended wake it. This subsumes the classic "skip self-requeue" and also stops the very common wasted wake where a propagator is re-run because of a change it had already acted on; a claiming run that inferred nothing earns the same protection.

Safety nets, per the plan:

* **Install-time aliasing downgrade.** Most algorithms are only idempotent when no two scope positions alias the same underlying variable, and only the engine sees the posted scope. `install()` canonicalises trigger variables through views and silently ignores claims from any propagator with a repeat (`x` twice, or `x` and `-x + 3`), with a `Stats` counter (`idempotence_downgrades`) so it is observable. Claiming propagators must register triggers 1:1 with their scope — or de-alias themselves: Element's propagators deliberately do not watch the variables they write, so Element performs its own whole-scope aliasing check at install (the `*_dup_xx` tests exercise it).
* **Claim checker.** `GCS_CHECK_IDEMPOTENT_CLAIMS` re-runs every honoured claim immediately and aborts the solve if it infers anything or contradicts. `solve_for_tests_with_callbacks` turns it on for every data-driven constraint test, so all existing lanes double as idempotence nets; a passing re-run emits nothing, so proof lanes are unaffected.
* **Directed unit tests** (`propagators_test`, `idempotent_claim_checker_test`): self-wake skipped; already-seen foreign wake skipped; post-run foreign wake preserved (including for no-op claiming runs); repeated-variable and view-aliasing downgrades; non-aliasing views not downgraded; lying claimant caught; downgraded claims neither honoured nor checked.

## Audit summary per adopter

* **VCAllDifferent**: internal worklist re-checks `optional_single_value` after every removal and cascades to quiescence in one call. Claimed at the AllDifferent call site only — circuit wraps the same helper in propagators that do more per run and must not claim.
* **GACAllDifferent** (and ArgSort's permutation propagator): one call prunes to the union of maximum matchings; every surviving edge is in some maximum matching, so a re-run deletes nothing. SymmetricAllDifferent, Inverse, and AllDifferentExcept do channelling or phantom-edge work in the same runs and stay unclaimed pending their own audits.
* **Table / `propagate_extensional`** (covers Table, AutoTable, and the tabulation propagators, including tabulated arithmetic): each surviving value is witnessed by a still-selectable tuple whose own entries all survive. Repeated variables break exactly that self-witnessing and are downgraded via the 1:1 triggers.
* **LinearLessThanEqual/GreaterThanEqual**: the inequality pass writes only max-side bounds and reads only min-side contributions (and vice versa), disjoint per variable, robust to hole-snapping. Equality's inverted second pass reads what the first pass wrote — the closed loop that under-propagated magic_square in the prototype — so equalities never claim. Prerequisite included: triggers built from the sanitised terms, and the reified dispatcher's runtime path strips claims (the install-time decided paths forward them).
* **Element**: the index propagators' support tests never read the index variable they prune; the union result propagator's supported-set never reads result; the endgame equality pass leaves both sides equal. The bounds-only result propagator does **not** claim — it reads and writes result's bounds, and snapping past a hole can licence a strictly tighter inference on a re-run (counterexample in the code).
* **Circuit (prevent)**: the propagator that made the global-skip prototype unsound. Both of its phases already run internal worklists to quiescence, but a small-cycle prevention or forced closing can fix a successor whose alldifferent consequences only the self-requeue used to catch; the fix is alternating the two existing passes until the prevention pass stops fixing successors (the common run still makes exactly one pass of each). A full tsp solve with the checker on verifies every claim by immediate re-run. The SCC propagator stays unclaimed.

## Deviations from the plan

* **No byte-identical `.pbp`.** The plan expected skipped runs to leave the proof byte-identical; that is unattainable in principle. A re-woken run in the unclaimed engine free-rides on changes applied after the wake that requeued it, so skipping shifts work between rounds. Because propagators are monotone the per-node fixpoint is unique — the search tree is identical, which the gates below verify — but inference attribution (which constraint prunes what, citing which reason) legitimately shifts, and with it proof line order and the propagation-count split. The claims themselves were confirmed honest by running the checker over full solves; the proof gate is therefore **veripb verifies** (confirmed on ortho_latin vc and gac, langford, and tables --prove runs, all complete enumerations).
* **Element claims alone move nothing.** Element never had self-wakes to skip (it does not watch what it writes), which is why the plain wave-2 claim left langford and qap propagation counts unchanged to the last digit, and why the Phase D already-seen refinement is included: that is where Element-style propagators actually pay.

## Benchmarks

Release build, 5 interleaved runs per configuration per build, medians; baseline is current `main`. **Every row has identical recursions, failures, contradicting propagations, and solutions across all runs of both builds**, and tsp reports the true gr17 optimum 2085.

| benchmark | claiming propagators | propagations | wall time |
|---|---|---:|---:|
| ortho_latin 6 gac (UNSAT) | GAC alldiff, tabulated div/mod | **−22.1%** | **−20.9%** (20.05s → 15.86s) |
| ortho_latin 6 vc (UNSAT) | VC alldiff, tabulated div/mod | **−22.0%** | **−18.2%** (18.81s → 15.40s) |
| langford 10 (UNSAT) | GAC alldiff, Element, tabulated Plus | +1.7% | **−11.7%** (1.38s → 1.22s) |
| sudoku --all | GAC alldiff | −26.4% | −14.6% (12ms scale) |
| crystal_maze | alldiff + linear≤ | −8.5% | −8.7% (4ms scale) |
| money | alldiff + linear | 0.0% | (3ms, noise scale) |
| magic_square 5 vc | VC alldiff among linear eq | +0.1% | −0.3% |
| magic_square 5 (not-equals) | none — control | 0.0% (identical) | −0.1% |
| minicp n_queens 12 --all | none — control | 0.0% (identical) | +0.4% |
| minicp tsp (gr17 = 2085) | Circuit prevent, objective bound | −1.2% | +0.9% |
| minicp qap | Element (2d, const arrays) | −0.9% | +2.4% |

Notes:

* langford's wall time drops 11.7% despite slightly more runs: the runs that disappear are the expensive ones (GAC matching, Element support scans), and the extra ones are cheap no-ops from the one-round work shift. Its wave-1-only state was a +21%-propagations regression, which decomposes as the GAC alldifferent claim (+19pp) plus the tabulated-Plus tables (+1.9pp); the already-seen refinement recovers it.
* tsp (+0.9%) and qap (+2.5%) are the residual cost of the claims-path bookkeeping on models where claims fire but yield little; both are propagation-bound at ~65–140ns/run, the pure controls (identical propagation counts) sit at −0.3% to +0.4%, and rounds whose claimants all come up empty before the first inference stay on the claim-free fast path. tsp's byte-identical search (the entire 104-incumbent sequence ending at the true optimum 2085) is preserved by the prevent claim; its remaining propagation volume is dominated by the 34 Element propagators and by branching wakes carrying genuinely new information, which no idempotence mechanism can soundly skip.

## Phase C assessment against the plan's bar

The bar was "two gain benchmarks ≥10% wall with identical trees, controls within ~2%". **Met**: three wall-time wins over 10% at real scale (ortho_latin gac −20.9%, ortho_latin vc −18.2%, langford −11.7%), every pure control within noise, all proofs verify, and the checker-on caps-off suite is green after every commit. The residual +2.4% on qap is the one number over the noise bar, and it is a claim-active model rather than a control; frontend XCSP/MiniZinc table- and inequality-heavy instances remain the natural next measurement, now from a position of strength.

Also on record: one caps-off suite run during wave 1 had `linear_constraint_ne_if_stateless` and `linear_constraint_ne_iff_incremental` fail; those lanes exercise `propagate_linear_not_equals` and linear_equality's own dispatch, which this branch does not touch, the failures did not reproduce in 25+ fresh-process runs of the exact lane configurations, and every subsequent full suite run passed 494/494 — consistent with the known seed-dependent flakiness of those tests.

Part of #416.

https://claude.ai/code/session_011KhPJ5EmGUS6S77me9ib3Z

### DIFF
--- a/dev_docs/constraints.md
+++ b/dev_docs/constraints.md
@@ -214,6 +214,23 @@ on backtrack to a level above where it was disabled. Don't return
 *right now* — the propagator will be triggered again as soon as a
 domain changes.
 
+In between the two sits `PropagatorState::EnableButIdempotent`: the
+propagator stays registered, but this run reached its own fixpoint —
+re-running it immediately, against the domains exactly as this run
+left them, would infer nothing and not contradict — so the propagation
+queue skips re-waking it from its own inferences (changes made by
+other propagators wake it as usual). The claim is per-run, and only
+correct for algorithms that process all their internal cascades in one
+call *and* whose scope has no two positions aliasing the same
+underlying variable; `Propagators::install` detects aliased trigger
+scopes (including through views) and silently ignores claims from such
+propagators, which relies on claiming propagators registering triggers
+1:1 with their scope positions. A wrong claim silently under-propagates
+or is unsound, so every adoption needs an audit note, and the test
+harness sets `GCS_CHECK_IDEMPOTENT_CLAIMS` to re-run every honoured
+claim and abort if it infers anything. When in doubt, return `Enable`:
+the only cost is a possible wasted no-op run.
+
 ### Triggers
 
 Three trigger kinds, applied to specific variables:
--- a/gcs/CMakeLists.txt
+++ b/gcs/CMakeLists.txt
@@ -464,6 +464,14 @@ if(GCS_BUILD_TESTS)
     target_link_libraries(state_test PRIVATE glasgow_constraint_solver Catch2::Catch2WithMain)
     add_test(NAME state_test COMMAND $<TARGET_FILE:state_test>)
 
+    add_executable(propagators_test innards/propagators_test.cc)
+    target_link_libraries(propagators_test PRIVATE glasgow_constraint_solver Catch2::Catch2WithMain)
+    add_test(NAME propagators_test COMMAND $<TARGET_FILE:propagators_test>)
+
+    add_executable(idempotent_claim_checker_test innards/idempotent_claim_checker_test.cc)
+    target_link_libraries(idempotent_claim_checker_test PRIVATE glasgow_constraint_solver Catch2::Catch2WithMain)
+    add_test(NAME idempotent_claim_checker_test COMMAND $<TARGET_FILE:idempotent_claim_checker_test>)
+
     add_executable(linear_utils_test constraints/linear/utils_test.cc)
     target_link_libraries(linear_utils_test PRIVATE glasgow_constraint_solver Catch2::Catch2WithMain)
     add_test(NAME linear_utils_test COMMAND $<TARGET_FILE:linear_utils_test>)
--- a/gcs/constraints/all_different/all_different.cc
+++ b/gcs/constraints/all_different/all_different.cc
@@ -151,7 +151,17 @@ auto AllDifferent::install_propagators(Propagators & propagators) -> void
                     if (! propagate_non_gac_alldifferent(
                             unassigned_handle, state, tracker, logger, owner, reasons.empty() ? nullptr : &reasons, reason_base))
                         return PropagatorState::Enable; // contradiction: loop sees tracker.contradicted()
-                    return PropagatorState::Enable;
+                    // Idempotent: the to_propagate worklist re-checks
+                    // optional_single_value after every removal and processes the
+                    // whole cascade in this call, so at return no variable left in
+                    // unassigned is single-valued and a re-run collects nothing.
+                    // Duplicate scope variables never reach here (prepare rejects
+                    // them), and triggers are 1:1 with the scope, so view aliasing
+                    // is caught by the install-time downgrade. The claim is this
+                    // call site's, deliberately not the shared helper's: circuit
+                    // wraps the same helper in propagators that do more work in
+                    // the same run and must not claim.
+                    return PropagatorState::EnableButIdempotent;
                 },
                 triggers);
         }}
--- a/gcs/constraints/all_different/all_different.cc
+++ b/gcs/constraints/all_different/all_different.cc
@@ -113,7 +113,16 @@ auto AllDifferent::install_propagators(Propagators & propagators) -> void
                     constraint_id = constraint_id()](const State & state, auto & inference, ProofLogger * const logger) -> PropagatorState {
                     propagate_gac_all_different(
                         constraint_id, vars, vals, vector<Integer>{}, *value_am1_constraint_numbers.get(), state, inference, logger);
-                    return PropagatorState::Enable;
+                    // Idempotent: one call prunes to the GAC closure. The
+                    // matching and SCCs are built from an entry snapshot, and
+                    // what survives is exactly the union of maximum matchings,
+                    // so a re-run deletes nothing (every remaining edge is in
+                    // some maximum matching) and a matching still exists (no
+                    // contradiction). Duplicate scope variables never reach here
+                    // (prepare rejects them), and triggers are 1:1 with the
+                    // scope, so view aliasing is caught by the install-time
+                    // downgrade.
+                    return PropagatorState::EnableButIdempotent;
                 },
                 triggers);
         },
--- a/gcs/constraints/circuit/circuit_prevent.cc
+++ b/gcs/constraints/circuit/circuit_prevent.cc
@@ -39,9 +39,13 @@ namespace
     // (succ[end] != start), force the full chain to close (succ[end] == start), and
     // contradict if a sub-cycle has actually closed. Each edge is processed once, in
     // O(1); the outer loop re-runs only because forcing an edge fixes a new one.
+    // Returns whether any successor became single-valued along the way -- the one
+    // event whose alldifferent consequences this pass does not handle itself, so the
+    // caller knows whether another alldifferent pass is needed.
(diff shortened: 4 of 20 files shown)


ciaranm:
Updated with wave 2. Two commits on top: Element claims (audit in the code; on their own they move nothing, because Element never watches what it writes and so never had a self-wake to skip — that finding is why the next commit exists), and the plan's Phase D already-seen refinement: a claiming run has seen every inference recorded before its run ended, so none of those should re-wake it, including for claiming runs that inferred nothing. That refinement is where the headroom was: ortho_latin gac goes from −13% to −20.9% wall, ortho_latin vc to −18.2%, and langford flips from the wave-1 +3.6% regression to −11.7%, all still on byte-identical search trees with proofs verifying. The PR body's table and Phase C assessment are updated: the plan's keep bar (two ≥10% wins, controls in noise) is now met, with qap's +2.4% bookkeeping cost the one residual over the line.

ciaranm:
One more adopter: Circuit's prevent propagator — the one that made the global-skip prototype unsound on tsp. It turned out not to need a queue-in-a-propagator: both phases already run internal worklists to quiescence, and the only escape was the cross-feed where a prevention inference fixes a successor whose alldifferent consequences the self-requeue used to catch. Alternating the two existing passes until the prevention pass stops fixing successors closes the cascade in ~20 lines; a naive alternate-on-any-inference version cost 3.7% on tsp, re-looping only on actual fixings gets the common run back to exactly one pass of each. Gates: tsp keeps a byte-identical search (all 104 incumbents, optimum 2085, recursions/failures/contradictions identical), propagations −1.2%, and a full tsp solve with GCS_CHECK_IDEMPOTENT_CLAIMS=1 re-verifies every claim. Honest outcome for the tsp target: wall goes from +1.8% to +0.9% against main — the claim recoups most of the branch's bookkeeping there, but tsp's volume is Element- and branching-dominated, so this was never going to be a big win on its own.

ciaranm:
Made sum-equality idempotent, and it turned into the best lever yet — including finally winning tsp.

**The mechanism** is small and it is *not* a queue-in-a-propagator: propagate_linear / propagate_linear_incremental now alternate their forward and inverse sweeps internally until neither infers anything, so one call reaches the equality's own fixpoint and it claims EnableButIdempotent. The loop breaks the instant a sweep is clean, so the common case is one or two sweeps.

**On your ping-pong:** the pain was the *queue* mediating every fwd↔inv exchange at full re-dispatch (re-snapshotting all n bounds each time). The internal loop keeps the snapshot warm and hits the identical fixpoint (monotone → unique). Measured: no pathological iteration — on tsp's 18-term sum it internalises ~2.3M confirming self-wake runs with the tree byte-identical. One trap worth noting: I detect per-sweep change with the non-destructive count_inferences(), *not* did_anything_since_last_call_inside_propagator() — that flag is shared with the propagate_stages callers (multiply/divide/power loop on it across all stages), and clearing it inside propagate_linear would silently corrupt their fixpoint loop.

**Results vs main** (all tree-identical, all proofs verify, 494/494 caps-off with checker on):
- tsp: props −3.9%, **wall −1.6% — now faster than main** (was +0.9%). The wide sum uses the incremental path, so I had to make that idempotent too, not just the stateless one.
- magic_square (not-equals): props −12.1%, **wall −4.9%** — this was a documented *control* ("no gain expected"); its 5-term row/col/diag sums hit the stateless path and it's now a real gain, still tree-identical.
- magic_square vc −4.6%, crystal_maze −47.9% props, money −12% props.
- n_queens (no linear equality) stays a clean control at +0.0%.

So idempotence turned out to be the right lever for the *equality-heavy* models and enough to cross tsp from losing to winning. It is emphatically **not** the 2×-to-Choco lever — that's still the delta/advisor filtering of the 94%-waste foreign dist-bound wakes I measured earlier, which idempotence structurally cannot touch. Pushed as one commit on the branch.

PR #492, opened 2026-07-10
Replace per-row proof model comments with per-constraint block headers

ciaranm:
### What

The `StringLiteral` `constraint_name` / `rule` arguments on `ProofModel::add_constraint`, `add_labelled_constraint`, `add_two_way_reified_constraint`, and `create_proof_flag_fully_reifying` only ever fed `* constraint <name> <rule>` OPB comments. A survey showed nothing consumes them (not veripb, not the test harness, not any tooling), roughly half the call sites went through the uncommented overloads — emitting `* constraint ? ?` — and the newer cake-conform APIs (#354-era flag reification, the `@i` encoding labels) had already dropped them.

This PR replaces the per-row comments with one block header per constraint, emitted by `Problem::create_propagators` before each `install()`:

```
* constraint AllDifferent _3
@c[_3][0lt1] ... ;
@c[_3][0gt1] ... ;
```

A constraint's rows follow the header (near enough) contiguously — reification helper rows land inside their owning block — so every row now gets provenance, and a row's finer identity, where it matters, is its `@c[id][role]` label. The type name comes from demangling the constraint's `typeid` at the single driver call site (both supported toolchains use the Itanium ABI), rather than a per-constraint virtual that could go stale.

### API changes

- `StringLiteral` is gone, along with the parameters threading it through `Propagators::define_bound`, the `product_encoding` helpers, `linear_stages`, and `divide_modulus`.
- The `c[id][role]` `add_labelled_constraint` overloads take `const ConstraintID &` directly instead of a pre-stringified id, removing the `as_string(_constraint_id)` boilerplate at every call site and making the id argument impossible to confuse with the role strings that follow it.
- The tracker's view-link definition, whose invented label is not `c[id][role]`-shaped, uses a new raw label-pair overload for equalities; its emitted labels are byte-identical to before.

### What doesn't change

Only OPB comment text. Constraint rows, labels, the constraint counter, and proofs are all unaffected.

### Testing

Full test suite (494 tests, including veripb verification) passes on three runs; spot-checked `magic_square --prove` output and verified complete enumeration with veripb 3.0.2.

https://claude.ai/code/session_01Tc9UsttXJpDviZddAKdPQx

### DIFF
--- a/dev_docs/constraints.md
+++ b/dev_docs/constraints.md
@@ -511,13 +511,12 @@ inequality holds", encode it as a full equivalence — both
 
 ```cpp
 auto gt = optional_model->create_proof_flag_fully_reifying(
-    "gt", "Foo", "var greater than",
-    WPBSum{} + 1_i * v1 + -1_i * v2 >= 1_i);
+    "gt", WPBSum{} + 1_i * v1 + -1_i * v2 >= 1_i);
 ```
 
 This creates `gt` and emits both halves of `gt ⇔ (v1 > v2)`. The
 equivalent two-step form is
-`add_two_way_reified_constraint(name, rule, ineq, flag)` if you've
+`add_two_way_reified_constraint(ineq, flag)` if you've
 already created the flag elsewhere. Both compute the reverse direction
 by integer-negating the supplied inequality.
 
--- a/gcs/constraint.hh
+++ b/gcs/constraint.hh
@@ -66,6 +66,16 @@ namespace gcs
         {
             _constraint_id = std::move(constraint_id);
         }
+
+        /**
+         * \brief The constraint's type, as the stem of the s_expr() term's head:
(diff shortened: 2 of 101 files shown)


ciaranm:
Third commit replaces the RTTI: the block header's type name now comes from a pure virtual `Constraint::constraint_type()` instead of demangling `typeid`. It returns the stem of the `s_expr()` head — the `.scp` keyword without any reification suffix (`abs`, `lin_less_equal`, `array_min`) — so the header speaks the `.scp` vocabulary rather than C++ class names, e.g. `* constraint not_equals _9`. Each `s_expr()` now composes its head from `constraint_type()` plus its own reification suffix, so the two cannot drift, and the pure virtual makes the compiler enforce coverage (including the two test-local constraints). `s_expr()` keeps its throw-on-unrepresentable-form behaviour; `constraint_type()` is total, which is what a comment emitter needs.

PR #534, opened 2026-07-16
Speed up proof writing (2.6-3x on cheap-constraint problems)

ciaranm:
Performance work on `gcs/innards/proofs/`, one measured commit per change. Rebased onto `3db5786e`; every figure and byte-comparison below was re-measured on that base.

## Result

Whole-solve wall time with `--prove`, all-solutions / full search, median of 5:

| benchmark | before | after | speedup | proof-logging overhead |
|---|---|---|---|---|
| minicp n_queens 12 (not-equals + views), 208 MB proof | 2.65s | 0.97s | 2.7× | 7.2× → 2.7× |
| minicp magic_series 100 (linear), 149 MB proof | 2.41s | 0.81s | 3.0× | 17.2× → 5.8× |
| minicp magic_square 4 (vc-alldiff), 166 MB proof | 2.25s | 0.86s | 2.6× | 10.7× → 4.3× |

"Overhead" is `--prove` time over the same run without proof logging (0.36s / 0.14s / 0.20s, unchanged by this branch). Entirely CPU-side: proof text now renders at roughly 200 MB/s, where the baseline managed about 90, and disk (~2 GB/s here) was never the bottleneck. An I/O ladder (`/dev/null`, tmpfs, real fs) puts ~92% of the remaining proof cost in user time.

## What changed

- **Representation swaps, invisible outside the tracker:** rendered literal names cached in an id-indexed vector (`const string &`, both naming modes); condition→literal lookup becomes dense per-variable atom tables keyed by value, with the ordered structure kept only where the order-encoding chain walk needs it (`need_gevar`'s next/prev walk over the ge-atom value set).
- **Text assembly:** lines are appended into reused buffers with `to_chars` number rendering instead of a fresh `stringstream` + locale-aware `ostream <<` per line; 1 MB `pubsetbuf` on the proof stream; `simplify_literal` fast path for plain non-range conditions; `PolBuilder` builds one text buffer with spliced line references instead of a `vector<variant<string, ProofLine>>`.
- **Structural:** the `need_all_proof_names_in` pre-pass is fused into rendering (`EnsureNames::Yes`; definitions emitted mid-render land ahead of the buffered line, nested emissions take their own buffer from a lease stack; the model-writing path keeps its explicit pre-pass). Reified lines render directly via `emit_reified_inequality_to` — no materialised reified sum, reason conditions negated at the XLiteral level — with the reification arithmetic single-sourced in `reification_shape()` shared with `reify()`. `backtrack()` uses the same renderer; `infer()` reuses a unit-sum scratch.
- **Model writing:** the OPB streams to disk instead of accumulating in a `stringstream`. `solve()` sets the objective and preserve list between `create_state_for_new_search` and `create_propagators` and calls the new `write_preamble()`, after which every constraint goes straight to the file (peak RSS drops by about the OPB size). Driving a `ProofModel` directly without `write_preamble()` still works — `finalise()` writes the preamble late, buffered as before.
- **Inference side:** reasons are snapshotted once and handed to the logger rather than re-materialised through an `ExplicitReason` round-trip; `equals` states its not-equals reason outright instead of building a `ConcatReason` per inference (the only hot user of that pattern — `vc_all_different` already states reasons explicitly, `regex`'s two uses are cold).

## Proof output compatibility

Re-established commit by commit against `775e5853`, regenerating `.opb`/`.pbp`/`.varmap`/`.scp` for three problem families at each commit:

- **14 of the 16 commits are byte-identical** to the commit before them, on all four output files.
- `6353554f` (fusing the name pre-pass into rendering) emits the same number of proof lines, but a `red` definition line now lands immediately ahead of the line that first uses it rather than in a batch, and the relative `pol` references shift to match (`pol -2 -6` → `pol -1 -6`). The `.varmap` holds the same entries in a different order. VeriPB verifies the result, including a full-size 208 MB enumeration proof.
- `efd013c6` (OPB streaming) changes the `.opb` by exactly one line: the `* #variable= ... #constraint= ...` counts header, which VeriPB 3 does not require and nothing here consumes. The `.pbp` is untouched.

If that definition ordering matters to anything downstream, `6353554f` can be dropped and the rest reworked onto the two-pass shape.

## Validation

- 491/491 ctest, configured with `GCS_ENABLE_MINIZINC=ON` so the frontend tests actually run
- per-commit byte-comparison as described above, all 16 commits building and running standalone
- VeriPB 3.0.2 verification with complete-enumeration conclusions: the three small proofs, plus the full-size 208 MB n_queens 12 enumeration
- clang-format 21 clean

## Rebase notes

Four proof fixes landed on main after this branch was written, in the files it rewrites. Each was reconciled deliberately rather than textually:

- **#554** stops aliasing the `>= 1` gevar to the encoding bit for `{0,1}` DirectOnly variables and deletes `track_gevar`. This branch had rewritten `track_gevar` to feed the dense tables; that rewrite is dropped rather than revived, since the atom is now built lazily by `need_gevar`.
- **#559** fixes the inverted value-0 eq pol-item polarity. The fix lives in `set_up_direct_only_variable_encoding` and carries through the dense-table rewrite unchanged.
- **#557** adds `find_xliteral_for`, the non-throwing twin of `xliteral_for`, which `recover_am1` uses to spot a complementary `{~b0, b0}` pair over a bit-aliased `{0,1}` variable. It landed as a direct probe of `variable_conditions_to_x` — which is exactly the map the dense-atom-tables commit stops putting eq/ge/in atoms in, so the merge was clean but would have made the lookup silently miss every pair it exists to find. It now goes through the same `find_condition` as `xliteral_for`. The `among` and both global-cardinality tests were run 12 times over fresh seeds (the failure #557 fixes is branch-order dependent, ~6/20 seeds) with no failures.
- **#553** added `track_bound_rows` and the `introduce_bits_of` path; the new per-variable table it introduces is hashed like its neighbours (point lookups only, never iterated in order).

`SnapshottedReason` also had to keep a handle on the still-declarative reason: the conflict observers behind dom/wdeg read the contradicting reason, and they run with proofs off, where the snapshot deliberately materialises nothing. `eqvar_polarity_test`, `direct_only_gevar_bound_test`, `introduce_bits_test` and `conflict_observer_test` all pass, and each covers exactly the machinery involved.

## Not done, recommended next (measured residuals)

Remaining tail on these benchmarks is tracker probes ~15%, `memmove` ~10%, and solver-proper work; the constraint-side sweep is done, in the sense that what is left is genuine proof content (linear's wide reasons *are* the emitted lines), cheaper only by changing what gets proved.

https://claude.ai/code/session_01KG64fM66vNWW4TG1MLXW45

### DIFF
--- a/gcs/innards/proofs/proof_logger.cc
+++ b/gcs/innards/proofs/proof_logger.cc
@@ -155,6 +155,11 @@ struct ProofLogger::Imp
     // before assembly starts may itself emit (and so clobber this).
     string line_buffer;
 
+    // Scratch for the reified form of an inference under its reason, reused
+    // across emissions for the same reason (a reified linear sum can run to
+    // hundreds of terms, and a fresh allocation per line is measurable).
+    WPBSumLE reify_buffer{{}, 0_i};
+
     Imp(NamesAndIDsTracker & t) : tracker(t)
     {
     }
@@ -502,7 +507,8 @@ auto ProofLogger::emit_under_reason(const ProofRule & rule, const SumLessThanEqu
         .visit(rule);
 
     if (! reason.empty()) {
-        emit_inequality_to(names_and_ids_tracker(), reify(ineq, reason), rule_line);
+        names_and_ids_tracker().reify_into(ineq, reason, _imp->reify_buffer);
+        emit_inequality_to(names_and_ids_tracker(), _imp->reify_buffer, rule_line);
     }
     else {
         emit_inequality_to(names_and_ids_tracker(), ineq, rule_line);
--- a/gcs/innards/proofs/proof_logger.cc
+++ b/gcs/innards/proofs/proof_logger.cc
@@ -195,6 +195,19 @@ struct ProofLogger::Imp
     deque<string> line_buffers;
     std::size_t line_buffer_depth = 0;
 
+    // Scratch for the ubiquitous `1 * lit >= 1` inference shape, in its
+    // rendered LE form (one -1 term, rhs -1), so infer() does not allocate a
+    // one-term sum per logged inference. Safe to reuse because nothing
+    // reachable from an emission re-enters infer().
+    WPBSumLE unit_buffer{{}, -1_i};
+
+    [[nodiscard]] auto unit_holds(const Literal & lit) -> const WPBSumLE &
+    {
+        unit_buffer.lhs.terms.clear();
+        unit_buffer.lhs.terms.emplace_back(-1_i, ProofLiteral{lit});
+        return unit_buffer;
+    }
+
     Imp(NamesAndIDsTracker & t) : tracker(t)
     {
     }
@@ -305,12 +318,17 @@ auto ProofLogger::solution(const vector<pair<IntegerVariableID, Integer>> & all_
 auto ProofLogger::backtrack(const vector<Literal> & guesses) -> void
 {
     _imp->proof << "% backtracking\n";
-    WPBSum backtrack;
-    backtrack.terms.reserve(guesses.size());
+    // The backtrack clause is `at least one guess is false': exactly a
+    // reason-only reified line over the guesses, so route it through the
+    // reified renderer, which negates each guess at the XLiteral level
+    // rather than as a condition object.
+    ReasonLiterals guesses_as_reason;
+    guesses_as_reason.reserve(guesses.size());
     for (const auto & guess : guesses)
-        backtrack += 1_i * ! guess;
+        guesses_as_reason.emplace_back(ProofLiteral{guess});
     auto assert_or_rup = (_imp->assertion_level >= AssertionLevel::Inferences) ? ProofRule(AssertProofRule{}) : ProofRule(RUPProofRule{});
-    emit(assert_or_rup, move(backtrack) >= 1_i, ProofLevel::Current, AssertionAnnotation{.hint_name = hints::Backtrack::hint_name});
+    emit_under_reason(
+        assert_or_rup, WPBSum{} >= 1_i, ProofLevel::Current, guesses_as_reason, AssertionAnnotation{.hint_name = hints::Backtrack::hint_name});
 }
 
 auto ProofLogger::emit_learned_nogood(const vector<Literal> & decisions) -> ProofLine
@@ -414,7 +432,7 @@ auto ProofLogger::infer(
         // infer_explicitly(); this variant only carries the plain ones, so the
         // annotation is just the one passed in.
         if (! is_literally_true(lit) && ! std::holds_alternative<NoJustificationNeeded>(why)) {
-            emit_under_reason(AssertProofRule{}, WPBSum{} + 1_i * lit >= 1_i, ProofLevel::Current, reason, annotation);
+            emit_under_reason(AssertProofRule{}, _imp->unit_holds(lit), ProofLevel::Current, reason, annotation);
         }
         return;
     }
@@ -422,12 +440,12 @@ auto ProofLogger::infer(
     overloaded{
         [&]([[maybe_unused]] const JustifyUsingRUP<NoHint> & j) {
             if (! is_literally_true(lit)) {
-                emit_rup_proof_line_under_reason(reason, WPBSum{} + 1_i * lit >= 1_i, ProofLevel::Current);
+                emit_rup_proof_line_under_reason(reason, _imp->unit_holds(lit), ProofLevel::Current);
             }
         }, //
         [&]([[maybe_unused]] const AssertRatherThanJustifying & j) {
             if (! is_literally_true(lit)) {
-                emit_under_reason(AssertProofRule{}, WPBSum{} + 1_i * lit >= 1_i, ProofLevel::Current, reason);
+                emit_under_reason(AssertProofRule{}, _imp->unit_holds(lit), ProofLevel::Current, reason);
             }
         },                                    //
         [&](const NoJustificationNeeded &) {} //
--- a/gcs/innards/proofs/proof_logger.cc
+++ b/gcs/innards/proofs/proof_logger.cc
@@ -195,11 +195,6 @@ struct ProofLogger::Imp
     deque<string> line_buffers;
     std::size_t line_buffer_depth = 0;
 
-    // Scratch for the reified form of an inference under its reason, reused
-    // across emissions for the same reason (a reified linear sum can run to
-    // hundreds of terms, and a fresh allocation per line is measurable).
-    WPBSumLE reify_buffer{{}, 0_i};
-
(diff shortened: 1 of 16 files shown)


ciaranm:
Added a 10th commit (43e6ffd2): the OPB now streams to disk instead of being buffered whole in a stringstream.

Checking the premise first: the counts header was **not** the only reason for the buffering — `finalise()` also wrote the `min:` objective line and the (unconditional) `preserved:` projection line ahead of the constraints, both rendered from state that only exists later (`minimise()`/`preserve()` were called after all constraints, and a view objective's rendering depended on which views constraints had registered). Both turned out to be movable: the objective and preserve list are Problem state, and every variable they mention has its bits by the end of `create_state_for_new_search`. So `solve()` now sets them and calls a new `write_preamble()` between variable set-up and propagator creation; after that, constraints go straight to the file. Driving a `ProofModel` directly without the new call still works — `finalise()` writes the preamble late with everything buffered, as before.

One deliberate behaviour change: a **view objective** is registered eagerly, so `min:` is always rendered over `BinEnc(V)` — previously that happened only if some constraint happened to register the view, with a deview-through-the-underlying fallback otherwise (which dropped the view's constant offset, and left an unchecked `*find_view` dereference reachable in the soli solution path). Verified on knapsack (maximise ⇒ view objective): the OPB gains BinEnc(V)'s four rows plus the rewritten `min:`, and VeriPB reaches the identical `BOUNDS -9 <= obj <= -9` conclusion on both builds.

Validation: satisfaction and plain-objective OPBs byte-identical minus the dropped header; `.pbp` files byte-identical except view-objective numbering shifts; 404/404 ctest; wall time unchanged on all benchmarks; peak RSS drops by about the OPB size (magic_series 100: 130 MB → 113 MB, scaling with model size for table-heavy instances).

ciaranm:
Five more commits, continuing down the recommended-next list:

- **e13898b2 PolBuilder internals** (API untouched): tokens were a `vector<variant<string, ProofLine>>`, one small string per operand plus a temporary per token when rendering; now one text buffer with (offset, line) splice points, references still resolved relatively at emit time. magic_series −9%, magic_square −4%.
- **10a5e933 dense per-variable atom tables**: conditions resolve through per-variable eq/ge/in tables keyed by value, storing only the positive polarity (the negative op is the flipped literal — allocation sites already registered pairs). magic_series −16%.
- **a9fcad50 atom defining lines join those tables**: `need_pol_item_defining_literal` no longer walks value-ordered maps; the only ordered structure left is the per-variable set of ge values the order-encoding chain walk needs. (The byte-compare harness caught a subtle miss here that 404 green tests did not: `track_gevar`-registered atoms have to appear in the chain-walk value set too, or 0/1 variables silently lose their chain links.)
- **4228de19 snapshotted reasons** (inference_tracker.hh, no propagator-facing signature changes): eager reasons used to be materialised, wrapped in an `ExplicitReason`, then "materialised" again — a copy per inference, and via `infer_all` a second snapshot per literal. A `SnapshottedReason` now carries the pre-inference literals directly (lazy kinds still materialise per literal against the current state). Measured neutral on the benchmark set (reasons there sit inline in the small_vector) — kept because it removes strictly-redundant work and makes the timing-pinning explicit; drop it if you judge otherwise.
- **3fced645 fix from CI**: the frontend lanes caught the one real bug of the OPB-streaming commit — eagerly registering a *negated unbounded-int* objective view (FlatZinc gives such variables half the Integer range, and negating that needs one more bit than exists) overflowed `power2`. Registration is now gated on the new `bits_encoding_fits`, with the old deview fallback restored for that case, and the soli witness path no longer dereferences a failed `find_view`. minizinc-arrayintminmax passes and verifies; local testing now includes a `GCS_ENABLE_MINIZINC=ON` build.

Cumulative wall times (median of 5): n_queens 12 all-solutions 2.49s → **1.01s**, magic_series 100 2.20s → **0.72s** (overhead 15.7× → 5.1×), magic_square 4 2.18s → **0.87s**. Proof bytes identical to the commit-6 baseline throughout; 404/404 plain ctest and 476/476 with the MiniZinc frontend.

ciaranm:
One constraint-side commit (96824260), prompted by the question of whether propagator changes buy anything:

**equals: state inference reasons outright** — the not-equals propagation built `concat(singleton_reason(cond), ExactSingleValue{...})` per inference: two Reason variants, a heap-allocated vector inside ConcatReason, and a redundant `optional_single_value` re-read at materialise time for a value the propagator already held. It now passes `ExplicitReason{ReasonLiterals{cond, v1 == *value1}}` — both literals sit inline — and the three shared_ptr scope vectors that existed only to back the deferred wrappers are deleted. n_queens 12 −10%, magic_square 4 −5%, proofs byte-identical, 404+476 tests pass.

This was the only hot instance of the idiom: vc_all_different already states its reasons explicitly (with the want_reasons guard), and `concat`'s only other users are two cold paths in regex. So this is not the first of a long series — the constraint-side sweep is essentially complete, and what remains there is genuine proof content (linear's wide reasons and pol operands are the lines being emitted), which could only get cheaper by changing what gets proved.

Final numbers with everything: n_queens 12 all-solutions 2.49s → **0.91s** (proof overhead 8.0× → 2.9×), magic_series 100 2.20s → **0.70s** (15.7× → 5.4×), magic_square 4 2.18s → **0.83s** (12.1× → 4.9×).

gcs/innards/proofs/proof_logger.cc
```diff
@@ -86,6 +86,60 @@ namespace
...
+
+        [[nodiscard]] auto buffer() -> string &
+        {
+            return _buffer;
+        }
+    };
+
+    // The string-append twin of AssertionAnnotation's ostream operator; keep
+    // the two renderings in sync.
+    auto append_annotation_to(string & out, const AssertionAnnotation & annotation) -> void
```
ciaranm, line 139: This "keep the two renderings in sync" obligation is against a **dead operator**. `operator<<(std::ostream&, const AssertionAnnotation&)` (`gcs/innards/assertion_hints.hh:172`) had exactly two callers on `main` — `proof_logger.cc:423` and `:491` — and both are the ones this commit converts to `append_annotation_to`. So post-PR the operator has **zero callers**, yet this comment asks a future maintainer to keep it in step with the twin.

Suggest either:
- delete `operator<<` at `assertion_hints.hh:172` and drop the "keep in sync" note; or
- make `append_annotation_to` the single primitive and have `operator<<` delegate to it (`s << ...` builds via `append_annotation_to`), so there's one spelling and no hand-sync.

(Same shape as `append_label_to` in `proof_model.cc` — but that one twins a *live* `operator<<(ProofLineLabel)` at `proof_logger.cc:454`, so it's legitimate duplication; single-sourcing would still be tidier.)

---
<sub>Auditor-generated (grumpy code-quality pass), audited commit `96824260`.</sub>

Issue #292, opened 2026-06-10
examples/ doubles as the proof-logging suite; relocate RNG stress-harnesses to test/


ciaranm:
**Classification:** part mechanical cleanup, part testing-strategy discussion. Came out of an audit of `examples/` vs `test/`.

## The structural finding

Every program in `examples/` is registered as a ctest test via `run_test_and_verify.bash` (run with `--prove`, then VeriPB-checked). The constraint unit tests in `gcs/CMakeLists.txt`, meanwhile, are registered almost entirely with `run_test_only.bash` (no proof check): **1 `run_test_and_verify` vs 48 `run_test_only`**. None of MultBC, Regular, SmartTable, or Circuit have their proof logging verified in the unit suite.

So `examples/` is effectively the project's proof-logging regression suite. Worth making explicit, because it means the "examples" can't be treated as purely pedagogical — several are load-bearing proof tests.

## Immediate, low-risk step: relocate the RNG stress-harnesses

Five "examples" have no reader value — they generate random instances to fuzz one propagator and check its proof (each genuinely instantiates `mt19937` / `random_device`, not just includes it):

- `circuit_random`, `multiply_random`, `random_polynomial`, `regular_random`, `smart_table_random`

These are tests, not examples. **Move them to `test/`**, carrying their `add_test … run_test_and_verify` registration across. An overlap check confirms they are *not* redundant: the corresponding unit tests (`mult_bc_test`, `regular_test`, `smart_table_test`, `circuit_*_test`) are all `run_test_only`, so these harnesses are currently the only end-to-end proof check for those constraints — removing them would open a proof-coverage gap. (`mult_bc_test` does sweep random instances, but without proof verification.)

## Longer-term: promote proof verification into the unit suite

The data-driven constraint test framework can already run VeriPB; it is simply registered `run_test_only`. If the constraint unit tests gained proof-verified modes, the random harnesses would become genuinely redundant and could be retired entirely — and proof coverage would be systematic rather than incidental to whichever example happens to exercise a constraint. That is a maintainer call and a larger job, but it is the real underlying issue.

## Minor leftover spotted en route
`examples/n_fractions/n_fractions.cc` keeps `#include <random>` (line 13) and `using std::mt19937;` (line 32) that it never uses — dead. (`n_fractions` itself is a genuine example; keep it.)

[...]

Issue #17, opened 2023-05-03
Consistency cleanup for examples


ciaranm:
They should probably all have a consistent command line, etc.

Might also be good to group them into MiniCP / features / other.

ciaranm:
**Unblocked.** The constraint-API churn that kept touching example call sites (#299 — fluent `.with_*()` selection) is now code-complete and merged, so the examples are a stable target again. The `--propagator` style CLIs partly converged as part of that work (e.g. the circuit examples now select `circuit::SCC` / `circuit::Prevent` via `with_algorithm()`), but the broader ask here — a consistent command-line across all examples, and grouping them into MiniCP / features / other — is still outstanding.

Keeping open as now-actionable.

PR #563, opened 2026-07-25
Consistent proof-file disposal policy, via GCS_PRESERVE_PROOF_FILES

### LINKED ISSUE #562: regex and regex-regex tests share a proof basename, leaving order-dependent artifacts
The two example tests registered in `examples/regex/CMakeLists.txt`:

```cmake
add_test(NAME regex COMMAND ${GCS_BASH} ${CMAKE_SOURCE_DIR}/run_test_and_verify.bash $<TARGET_FILE:regex>)
add_test(NAME regex-regex COMMAND ${GCS_BASH} ${CMAKE_SOURCE_DIR}/run_test_and_verify.bash $<TARGET_FILE:regex> --regex)
```

both write `regex.{opb,pbp}` in the same working directory (the proof basename defaults to the binary name). The existing `RESOURCE_LOCK regex` property does its job — verified empirically that parallel `ctest` leaves clean last-writer-wins bytes, not interleaved corruption — but the shared basename still has three costs:

1. **The surviving `regex.{opb,pbp}` after a suite run is whichever test ran last**, which ctest does not contractually order. Any tooling that harvests per-test proof artifacts after a run (byte-comparison regression checks, post-mortem debugging of a verification failure) gets an ambiguous, order-dependent file.
2. **Harvesting proofs *during* a suite run can read a half-written file**, since the basename is reused mid-suite. This bit us in practice: an automated proof byte-comparison harness captured bytes matching neither clean variant.
3. The lock **needlessly serialises** two otherwise-independent tests.

Distinct basenames per test would fix all three and let the `RESOURCE_LOCK` be dropped. One wrinkle: `run_test_and_verify.bash` derives the proof basename from the binary name, so it would need either an explicit basename argument, or per-test working directories.

Found during automated proof byte-comparison work on the order-encoding-deletion branch (Claude-assisted).

ciaranm:
First of a four-PR stack tidying up how tests generate, name and dispose of proof files. Fixes #562, #17 and #292 between them; this PR is the shared groundwork the other three build on.

## The problem

Proof files were disposed of inconsistently. The data-driven constraint harness deleted them on success and kept them on failure — the behaviour we want — but nine other test binaries verified their proofs by calling `run_veripb()` directly and simply left the files behind, and `stacktrace_logging_test` wrote a proof, read it back, and never cleaned up at all. A full `ctest` run left **136 stale proof artifacts** under `build/gcs/`.

## The policy

Three layers, so every caller has the right shape available:

- `dispose_of_proof_files()` applies the policy to a proof that has served its purpose;
- `verify_proof_and_dispose()` is the non-throwing form, for tests that report the outcome themselves (a Catch2 `CHECK`, or a bool returned to a caller that prints its own diagnostic);
- `verify_proof_and_clean_up()` keeps its throwing behaviour, now expressed in terms of the other two.

**The default is unchanged**: delete once verified, always keep on failure.

That default needs no uniquifying counter, and it is worth saying why. A failed verification throws, and no test catches `UnexpectedException`, so the run stops at the first bad proof — the files left under the bare basename are exactly the failing instance's, identified by the test's own log line just above them. Verified by shimming a `veripb` that always fails: the test aborts with `SIGABRT` and leaves precisely those four files.

`GCS_PRESERVE_PROOF_FILES` overrides it:

| Value | Behaviour |
|-------|-----------|
| unset, empty, `0` | delete once verified (the default) |
| any other value, e.g. `1` | keep, letting the next instance overwrite in place |
| `all` | keep every instance, renamed with a zero-padded per-basename counter |

`all` is for harvesting proofs across instances — byte-comparing before and after a change — rather than debugging one failure. It is deliberately opt-in: `equals_test` alone writes 284 proofs under one basename, so combining it with `-DGCS_TEST_CAP_DEFAULTS=OFF` can fill a disk. Uniquification happens by renaming *after* verification, which keeps it entirely inside the helper: tests go on choosing whatever basenames they like.

## Call sites routed through it

`solve_test`, `smart_table_test`, `seq_precede_chain_test`, `power_test` and the five circuit scenario tests. `stacktrace_logging_test` disposes of its proof once it has finished reading it, keeping it when the check failed.

## Result

Leftover artifacts under `build/gcs/` after a full run: **136 → 36**. The remainder belong to the tests verified by `run_test_and_verify.bash`, whose disposal is the wrapper's job — that is PR 3 in this stack.

Suite green: 525/525.

### DIFF
--- a/dev_docs/constraints.md
+++ b/dev_docs/constraints.md
@@ -719,6 +719,42 @@ GCS_TEST_MAX_SOLUTIONS= GCS_TEST_MAX_RECURSIONS= ./build/foo_test
 The capped defaults are for keeping the *routine* parallel suite fast;
 correctness work wants the full enumeration check.
 
+### What happens to the proof files
+
+`verify_proof_and_clean_up` runs VeriPB and then, **by default, deletes**
+the `.opb`, `.pbp`, `.scp` and `.varmap` it just checked. This is not
+merely tidiness: a `.pbp` for an enumeration test can reach hundreds of
+megabytes, and a full parallel `ctest` that kept them all would hold
+gigabytes at once — enough to exhaust the disk mid-run and make an
+*unrelated* lane's proof write fail. Deleting each proof as it verifies
+bounds the footprint to the lanes running concurrently.
+
+Files are **always kept when verification fails**, whatever the policy.
+Since a failed verification throws, and no test catches
+`UnexpectedException`, the run stops at the first bad proof — so the
+files left under the bare proof name are exactly the failing instance's,
+identified by the test's own log line just above them. That is why the
+default needs no counter.
+
+`GCS_PRESERVE_PROOF_FILES` overrides the policy:
+
+| Value | Behaviour |
+|-------|-----------|
+| unset, empty, `0` | delete once verified (the default) |
+| any other value, e.g. `1` | keep, letting the next instance overwrite in place |
+| `all` | keep every instance, renamed with a zero-padded per-basename counter |
+
+Use `1` to inspect the proof a test just produced. Use `all` only when
+harvesting proofs across instances — comparing proof bytes before and
+after a change, say. A single test can write hundreds of proofs under one
+basename (`equals_test` writes 284), so `all` combined with
+`-DGCS_TEST_CAP_DEFAULTS=OFF` can fill a disk.
+
+```shell
+GCS_PRESERVE_PROOF_FILES=1 ./build/equals_test     # keep the last of each
+GCS_PRESERVE_PROOF_FILES=all ./build/equals_test   # equals_test.0001.pbp, ...
+```
+
 ## Adding a new constraint: checklist
 
 1. Header file with class declaration, Doxygen comments, the standard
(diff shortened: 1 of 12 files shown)


ciaranm:
Addressed #572 in 34122538 (branch restacked; #564/#565/#566 rebased on top).

**1. `stacktrace_logging_test` leaked its proof wherever `<stacktrace>` is missing.** Confirmed: the `solve()` is outside the `#ifdef`, the disposal was inside it, and the `#else` branch is macOS libc++ — so this leaked on two lanes of every CI run, and the "136 → 36" figure necessarily came from a Linux run. Now disposes in the `#else` branch too, unconditionally, since that branch asserts nothing. Compile-checked by forcing the `#ifdef` off.

**2. `solve_test`'s ten cases shared one basename.** Confirmed and fixed the way the issue recommends rather than by weakening the doc: each case now has its own basename (`solve_test_unsat`, `solve_test_unsat_restarts_binary`, …), declared once per case as a local so the two uses cannot drift. All 11 cases pass, the 10 proofs verify under their new names, no leftover artifacts.

`verify_proof_and_dispose`'s doc no longer claims a failing proof is *always* on disk. It now says that not stopping is what makes preservation the caller's problem, and names both sides: the callers that get it for free because they `return EXIT_FAILURE` (circuit, smart_table — thanks for checking those, the note records it) and the one that has to arrange it.

PR #564, opened 2026-07-25
Give every solver binary a --proof-files-basename (issue #17)

ciaranm:
Second of a four-PR stack. **Stacked on #563** — review that first; this PR's diff is against it.

Towards #17.

## The problem

Four incompatible ways of asking for a proof had grown up across the binaries:

| Family | Spelling | Can you name the proof? |
|---|---|---|
| `examples/*` (33) | `--prove` + `--proof-files-basename NAME` | yes, defaults to the program name |
| `minicp_benchmarks/*` (5) | `--prove` | no — hardcoded in C++ |
| `xcsp` | `--prove` | no — hardcoded as `"xcsp"` |
| `fzn-glasgow` | `--prove NAME` | yes, but as the flag's value |

Only the first can express "put this run's proof here". That is why the proof-basename collisions in `examples/` and `xcsp/` are currently papered over with `RESOURCE_LOCK` rather than fixed: for three of the four families there was no flag to pass. **This PR is what unblocks #562**, handled in PR 3 of the stack.

## The change

Standardise on the `examples/` spelling — `--prove` as a bare flag, the name supplied by `--proof-files-basename`, defaulting to the program name. 35 binaries were already correct; 7 needed changing.

`fzn-glasgow` is the awkward one, since `--prove` took the basename as its value. It keeps working as `--prove=NAME`, which now warns. The space-separated `--prove NAME` cannot coexist with a bare `--prove` in cxxopts — an option with an implicit value does not consume the following token — so the three `.msc` solver configs now declare `--prove` as a bool extraFlag alongside a string `--proof-files-basename`, and the two bash runners pass the new spelling.

## Also

Fixes a manual-run recipe in `dev_docs/view-proof-logging.md` that told you to pass `--prove eq_w11` to `equals_test` / `lex_test`. Those are data-driven test binaries with **no `--prove` flag at all**: they prove and run VeriPB themselves whenever `veripb` is on the `PATH`. The recipe now uses the view flags plus `GCS_PRESERVE_PROOF_FILES=1`, and names the proof files the harness actually writes.

## Scope

Deliberately limited to the *proof* CLI. The broader #17 asks — converging `--stats` / `--branch` / `--restarts`, and grouping examples into MiniCP / features / other — are untouched, and #17 should stay open. Directory reshuffling on top of a wide mechanical edit is what would make this risky rather than routine.

No behaviour change for any existing invocation except `fzn-glasgow`'s `--prove NAME`. Suite green: 525/525, including all 73 MiniZinc tests (72 of which exercise the proof lane through the updated `.msc`).

### DIFF
--- a/dev_docs/view-proof-logging.md
+++ b/dev_docs/view-proof-logging.md
@@ -237,14 +237,23 @@ cmake -S . -B build -DGCS_ENABLE_VIEW_WRAP_SWEEP=ON
 The harness and the `--view-wrap=N` / `--view-position=...` flags are always
 built; only the per-wrap ctest registrations are gated.
 
-Single manual run:
+Single manual run. These are data-driven test binaries, not example
+programs: they have no `--prove` flag, and instead prove and run VeriPB
+themselves whenever `veripb` is on the `PATH`, naming each proof after the
+test and its view configuration. So a manual run needs only the view
+flags, and `GCS_PRESERVE_PROOF_FILES=1` to stop the harness deleting each
+proof once it has verified (see `dev_docs/constraints.md`):
 
 ```shell
-./build/equals_test --view-wrap=11 --view-position=all --prove eq_w11
-veripb eq_w11.opb eq_w11.pbp
-./build/lex_test --view-position=mixed --prove lex_mixed   # heterogeneous views
-veripb lex_mixed.opb lex_mixed.pbp
+GCS_PRESERVE_PROOF_FILES=1 ./build/equals_test --view-wrap=11 --view-position=all
+ls equals_test_*.opb            # one per view configuration exercised
+
+GCS_PRESERVE_PROOF_FILES=1 ./build/lex_test --view-position=mixed  # heterogeneous views
+ls lex_test_*.opb
 ```
+
+Use `GCS_PRESERVE_PROOF_FILES=all` instead to keep *every* instance's
+proof rather than the last one per configuration.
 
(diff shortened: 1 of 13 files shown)


ciaranm:
Addressed #573 in 781651ee (branch restacked onto the updated #563).

The finding is right, and I reproduced all four rows of the table against the built binary rather than against the parser's documentation:

| invocation | before |
|---|---|
| `--prove --proof-files-basename myproof model.fzn` | correct |
| `--prove=myproof model.fzn` | correct, warns |
| `--prove myproof model.fzn` | tries to solve `myproof`; fails loudly |
| `model.fzn --prove myproof` | **solves, proof to `fzn-glasgow.{opb,pbp}`, exit 0, silent** |

Fixed by warning about **every** unconsumed positional, not only the `--prove` ones: nothing reads `unmatched()`, so `fzn-glasgow a.fzn b.fzn` silently discards `b.fzn` today too, and one check covers all of it. The `--prove`-specific hint naming the new spelling is added on top when `--prove` is present with no value. Warning rather than rejecting, to match how `--prove=NAME` is handled.

All four arrangements now produce a diagnostic, and so does the two-model case. The 73 MiniZinc tests pass, so nothing MiniZinc itself passes through the `.msc` trips it.

Also corrected the comment at the option declaration, which said the space-separated form "is gone" — it now says it cannot coexist with a bare `--prove` and is diagnosed after parsing instead.

PR #565, opened 2026-07-25
Distinct proof basenames per test; drop the RESOURCE_LOCKs

### LINKED ISSUE #562: regex and regex-regex tests share a proof basename, leaving order-dependent artifacts

ciaranm:
Third of a four-PR stack. **Stacked on #564** — review #563 and #564 first; this PR's diff is against #564.

Closes #562.

## The problem

Seven groups of tests shared a proof basename and were serialised with `RESOURCE_LOCK` to stop them racing: six in `examples/` (`regex`, `crystal_maze`, `nonogram`, `regular_random`, `skyscrapers`, `sudoku` — 18 tests) and `xcsp`, where all 36 tests wrote `xcsp.{opb,pbp}`.

The lock stopped the corruption but not the underlying problem. Whichever test ran last left the surviving artifacts, so anything harvesting per-test proofs after a run got an ambiguous, order-dependent file, and anything reading them during a run could see a half-written one.

One correction to the issue as filed: the parallelism argument is the weakest of its three. Measured from `CTestCostData.txt`, the locks cost about 7.8s (5.6s examples, 2.1s xcsp) in a ~two-minute suite. Determinism is the real driver, not speed.

## The change

Now that every binary accepts `--proof-files-basename` (#564), give each test its own. `run_test_and_verify.bash` takes an optional `--basename NAME` and passes it through; `run_xcsp_test.bash` names the proof after the test case. All seven `RESOURCE_LOCK`s are gone — the remaining `lifetime_annotation_build` lock is unrelated, serialising compiler invocations rather than proof files.

Proof disposal is also made uniform across the wrappers, matching the policy #563 applies inside the harness: delete once verified, always keep on failure, honour `GCS_PRESERVE_PROOF_FILES`. `run_test_and_verify.bash` had its cleanup commented out entirely and the two MiniZinc runners never had any, which is why a full `ctest` left **566 stale proof artifacts**. It now leaves **none**.

## Two registrations that had been silently broken

`multiply_random` and `random_polynomial` passed their arguments after a bare `--`, which terminates cxxopts option parsing. `--display-problem`, `-n` and `--stats` were all discarded and both tests only ever solved the default n=5 instance. Two further faults were hiding underneath: the flag is spelled `--display-problem`, not `--display_problem`, and `--n 50` is not valid cxxopts syntax for a one-character option name.

- `multiply_random` moves to the n=50 the registration always intended — 0.1s.
- `random_polynomial` **deliberately stays at the default**. Its instances are unseeded and the runtime cliff is steep and close: over 12 sample runs of the full check, n=5 stayed under 2.5s, while n=10 ranged from 0.1s to over 120s with three runs above 14s, and n=15 took 90s. Raising it needs a seeded instance, not a bigger number. The measurements are recorded in the CMakeLists comment so the next person does not rediscover this the hard way.

## Verification

525/525, and the 56 formerly-locked tests pass **five consecutive parallel runs** with no lock. Zero leftover artifacts.


ciaranm:
Addressed #574 in 9cf3184f (branch restacked onto the updated #564).

Promoted `run_scp_chain.bash`'s `dispose_proof` into `proof_file_disposal.bash` at the repo root; all five wrappers now source it relative to their own `$0`. ctest invokes each by absolute path, so that works whatever working directory the test runs in — and the one documented manual invocation (`./run_test_and_verify.bash …` from the repo root) resolves too.

The extension list lives in the helper and callers pass only the basename; `run_scp_chain.bash`'s two extra artifacts (`verifiedopb`, `corepb`) are extra arguments. The `case` now reads the same way round as the four it replaces — delete in the matched branch — rather than the inverted early-return.

Left the C++ side alone, as the issue suggests: it is the only caller needing the `KeepAll` counter. `dev_docs/constraints.md` now also names the one file to edit when what the wrappers delete changes, next to the existing explanation of the two-versus-three-state divergence.

Exercised all six `GCS_PRESERVE_PROOF_FILES` states directly against the helper, with and without extras, confirming in particular that `all` still means the same as `1` for a wrapper. Suite: 368/368 non-`scp_chain` tests pass, and the `cake_pb_cp`-absent fallback path was run end to end and disposes correctly.

⚠️ The 157 `scp_chain` tests fail on my machine both before and after this change — the `cake_pb_cp` on my `PATH` predates the current `.scp` format and rejects the input at step [2], before any disposal runs. I verified that by running the pre-change script from `git show`, which fails identically. So the full workflow-2 path's `dispose_proof "$base" verifiedopb corepb` line is covered by the direct helper test above but not by an end-to-end run here; worth a green CI lane before merge.

ciaranm:
**Correction to my note above, and a rebase.**

I said the 157 `scp_chain` failures I saw were a stale `cake_pb_cp` on this machine. That was wrong, and so was my second guess (that the solver's `.scp` writer was out of date). #567 fixed the format properly on `main` — the actual cause is that this stack was branched from a `main` **53 commits old**, predating #567 along with #534, #555, #556, #560, #561 and the whole `min_distance` batch.

The check I offered as evidence was worthless: I re-ran the *pre-change script* from `git show` and saw the same failure, and concluded "pre-existing". But both runs used the same stale worktree's `glasgow_scp_solver`. I varied the script and held the solver fixed, when the solver was the variable that mattered.

**Rebased the whole stack onto current `main`.** One conflict, in `examples/CMakeLists.txt`: `main` added `p_dispersion` while #566 moved `positive_table_random` out to `benchmarks/`. Resolved by keeping both changes; verified the directories and registrations line up afterwards.

Re-verified on the rebased tree: **531/531 pass**, including all 157 `scp_chain` cases running the **full workflow-2 chain** — cake re-derives the OPB, veripb elaborates, cake re-checks the core, opbdiff runs as the step-5 oracle where registered. So the `dispose_proof "$base" verifiedopb corepb` call site is now covered end to end, not just by my direct helper test. The earlier caveat on this PR is withdrawn.

Also in this push: `shellcheck` is now installed here, so the sourced-helper pattern is checked properly. It was already clean; I added `source-path=SCRIPTDIR` + `source=` directives so `shellcheck -x` follows the file (a bare `source=` resolves against shellcheck's cwd, not the script's, so it only worked from the repo root). Clean from the repo root and from inside `minizinc/`.

The commit messages that carried the wrong claim have been corrected too.

PR #566, opened 2026-07-25
Move measurement-only programs to benchmarks/

ciaranm:
Last of a four-PR stack. **Stacked on #565** — review #563, #564 and #565 first; this PR's diff is against #565.

Towards #292, but see the caveat below — **that issue needs editing before it is acted on further.**

## The change

`examples/` had grown into three things at once: user-facing examples, the end-to-end proof-logging regression suite (every program there is a ctest run with `--prove` and checked by VeriPB), and a collection of benchmarks that are none of the above. This splits the third out into `benchmarks/`:

```
bin_packing_bench  knapsack_bench  linear_prop_cost  linear_slack_bench
negative_table_random  positive_table_random  slack_watch  wake_cost
```

None was registered as a ctest, so the suite is unchanged at 525 tests. They take a size or repeat count and print timings — there is no fixed instance to check an answer against, which is exactly why they were never tests and why they do not belong beside material people are meant to read.

Mostly `git mv`, so it should read as a pure move plus CMake rewiring.

- `benchmark_cli.hh` stays in `examples/`, since `ortho_latin` and `langford` use it too and `minicp_benchmarks/` already includes it across directories.
- `skeleton_puzzle` stays: despite having no test it is a genuine example, not a benchmark. It has no registration because it does not finish in five minutes under `--prove`, so it cannot join the proof suite as it stands. Worth knowing before someone "fixes" the omission.
- Drops the dead `#include <random>` and `using std::mt19937` from `n_fractions.cc`, as #292 spotted.

## #292's premise has gone stale

The issue proposes relocating the five RNG stress-harnesses on the grounds that they were the only end-to-end proof check for their constraints. **That is no longer true**, and the header comment in `examples/CMakeLists.txt` now records why:

- **54 of the 59 constraint test files verify their own proofs internally** via `can_run_veripb()` — Multiply, Regular, SmartTable and Circuit among them. They are registered `run_test_only` precisely *because* the harness does the verification itself.
- The issue's headline figure of "1 `run_test_and_verify` vs 48 `run_test_only`" is long out of date: it is **55 vs 59** now.

So the issue's "longer-term" ask — promoting proof verification into the unit suite — is essentially already done. Retiring the RNG harnesses is now a judgement call about whether their random instance streams still add value over the curated unit tests, not a coverage regression. They stay for now.

I have not edited #292 myself. It would be worth correcting those figures before anyone acts on it, since as written it will send the next reader after a problem that no longer exists.

Suite green: 525/525, verified from a clean build.


ciaranm:
Addressed #575 in 8eff8495 (branch restacked onto the updated #565).

**1. The `examples/CMakeLists.txt` header.** Confirmed — I checked every directory and `skeleton_puzzle` is the only one without an `add_test`, so the claim was wrong by exactly one. The reasoning from this PR's body now lives in the comment, including the instruction not to "fix" the missing registration.

**2. The `GCS_ENABLE_EXAMPLES` gate.** Went with keeping one switch, but renamed to **`GCS_ENABLE_EXECUTABLES`** and documented what it is for: letting someone consuming this as a CMake dependency build the library without also building sixty-odd binaries. That argument does not distinguish between kinds of program, so grouping `examples/`, `benchmarks/`, `verified_encodings/`, `scp/` and `minicp_benchmarks/` is deliberate rather than an oversight — which is what the comment now says.

The old spelling still supplies the default and emits a CMake deprecation warning, so an existing `-DGCS_ENABLE_EXAMPLES=OFF` does not silently start building everything again; an explicit `-DGCS_ENABLE_EXECUTABLES` wins over it. All three paths checked (old spelling `OFF`, old-then-explicit-new, and the default). `python/pyproject.toml` and `.github/workflows/windows.yml` updated to the new name.

**3. `benchmarks/bin_packing/`.** `git mv`'d to `bin_packing_bench`, matching the other seven. `dev_docs/bin-packing.md` now names the binary and where it is built from, rather than swapping one wrong path for another.

Suite: 368/368 (`scp_chain` excluded — broken on this machine before and after, see the note on #565).

ciaranm:
**Rebased onto current `main`** along with the rest of the stack — it was 53 commits behind, which is what was really behind the `scp_chain` failures I mentioned; see the correction on #565.

One conflict here, in `examples/CMakeLists.txt`: `main` added `add_subdirectory(p_dispersion)` while this commit removes `add_subdirectory(positive_table_random)`. Kept both changes.

Worth noting for item 1 of #575: I re-checked the "`skeleton_puzzle` is the only example without an `add_test`" claim against the new base, since `main` has added `p_dispersion` since. It still holds — `p_dispersion` is registered.

**531/531 pass** on the rebased tree, `scp_chain` included this time.

Issue #432, opened 2026-06-30
Verified-encoding (cake_pb_cp chain) coverage: status + remaining gaps


ciaranm:
Audit of which constraints chain-verify end-to-end through workflow 2 (`glasgow_scp_solver --prove` → `cake_pb_cp` re-derives the OPB → VeriPB elaborates → `cake_pb_cp` re-checks). Determined **empirically** (running the chain + probing), not from docs. Tracking issue to coordinate the remainder. Umbrella: #353.

## Decisions (2026-06-30)

- **#358 is closed wontfix.** We will **not** try to make the solver's and cake's *variable* encodings line up identically. Byte-match (strict `opbdiff`) is not a goal — **chain-verify is the bar.** Most chained constraints are mode `none` (chain verifies, doesn't byte-match) and that is fine.
- **Conform only if it's easy and efficient.** For a remaining constraint, add the reader + a chain test when it's a quick win. A deep-dive re-probe found that **every remaining "structural" item is actually a naming conform** (cake uses the same encoding, named differently) — but they vary widely in effort: `knapsack` was one line (#440), `value_precede`/`seq_precede_chain` ~80 lines (#439), `sort` was a larger conform that also shrank its encoding (#441); only `arg_sort` remains (its internal `y` is structurally different — see below). Cake-blocked items (`notin`, `mult_bc`, nonlinear arithmetic, views) remain out.
- **Boundary units — do next, in a fresh context.** The one general, worth-doing fix that fell out of the #358 close (see the dedicated box below). It is *not* an encoding-conform; it's about making chains that route through min/max eq-literals work at all.

Views are excluded throughout — a view operand renders as an s-expr list `(-X + c)` which neither cake's `sexp_varc` nor `scp_reader` parse; that's upstream cake affine-operand support, not here.

## ✅ Boundary units (DONE — PR #438, stacked on #437)

[...]

PR #577, opened 2026-07-27
Move the cake_pb_cp measurement probe out of the test-utils header

### LINKED ISSUE #576: cake_probe_chain: ~100 lines of measurement scaffolding in the header every constraint test includes
`gcs/constraints/innards/constraints_test_utils.hh:73-172` (at `4a337e68`)

This one is on `main` today, not in any open PR. It came up while reviewing #566, whose argument
applies to it more sharply than to anything #566 actually moves.

`constraints_test_utils.hh` is included by every constraint test in the tree. Roughly a hundred
lines of it are a measurement probe that no test uses:

```cpp
// ---- PROBE (measurement only): workflow-2 cake_pb_cp chain -----------------
// Runs the full verified-encoding chain on a just-written .scp/.pbp, best
// effort, and logs a greppable "CAKECHAIN <name> <OUTCOME>" line. Never
// throws: this is for measuring the chain-verify rate across the random
// data-driven instances, not for gating the suite. Enabled by GCS_TEST_CAKE;
// cake_pb_cp path overridable via CAKE_PB_CP.
```

It is called from one place, `verify_proof_and_dispose` (line 271, after #563; `verify_proof_and_clean_up`
before that), and is a no-op unless `GCS_TEST_CAKE` is set.

## Why it is worth moving

The banner is honest about what it is — "PROBE (measurement only)", "not for gating the suite" —
and the code is careful within its own terms. The objection is location, not quality.

#566 makes the case that measurement-only programs should not sit beside material people are meant
to read, and moves eight of them into `benchmarks/`. This is the same class of thing in a
considerably more central place: not a standalone binary in a directory people can skip, but a
hundred lines in the middle of the header that every constraint test author opens. Someone reading
`constraints_test_utils.hh` to work out how to write a test has to scroll past
`cake_capture(popen(...))`, a hand-rolled last-non-empty-line parser, `std::system` calls building
shell strings, and a `mkdir -p ... && cp` invocation, none of which has anything to do with writing
a constraint test.

It also carries the most shell-shaped code in the C++ tree — `popen`, `std::system` with
interpolated paths, `2>/dev/null` redirections — in a header, which is a surprising place to find
it if you are not looking for it.

There is a related concern I am **not** claiming: whether the probe still earns its keep at all.
#432 reports that all non-cake-blocked constraints now chain-verify, which is the question this
probe was built to answer. Whether that makes it finished, or makes it a regression guard worth
keeping, is a judgement I do not have the context to make — hence no `claude` label on this issue.

## Options

1. **Move it to its own header** — `gcs/constraints/innards/cake_probe.hh`, included by
   `constraints_test_utils.hh` for the single call site. Smallest change; gets it out of the way of
   readers without deciding anything about its future.
2. **Move it next to the rest of the chain tooling** in `verified_encodings/`, which is where the
   `cake_pb_cp` workflow otherwise lives, and where a reader would look for it.
3. **Retire it**, if #432's coverage work has made it redundant. Cheapest to maintain; needs
   someone who knows whether the chain-verify rate is still a number anyone wants.

I would suggest (1) as an immediate tidy regardless, since it is nearly free and does not foreclose
(2) or (3). But the real question is which of those three, and that is yours rather than mine.

**Local. Minutes for (1); (3) needs a decision first.**

---
*Filed by Grumpy Auditor Claude — automated, independent code-quality audit. Findings as of commit `4a337e68` on `main`; please re-verify line references before acting.*

ciaranm:
Closes #576.

`gcs/constraints/innards/constraints_test_utils.hh` is included by every constraint test in the tree, and roughly a hundred lines of it were a measurement probe that no test uses: a best-effort run of the workflow-2 `cake_pb_cp` chain over a proof that has just verified, logging a greppable `CAKECHAIN` line, no-op unless `GCS_TEST_CAKE` is set.

The objection is location, not quality — the banner is honest about what it is and the code is careful within its own terms. But #566 has just made the case that measurement-only programs should not sit beside material people are meant to read, and this is the same class of thing somewhere considerably more central: someone opening this header to work out how to write a constraint test has to scroll past `popen()`, `std::system()` with interpolated paths, a hand-rolled last-non-empty-line parser and a `mkdir -p && cp`, none of which has anything to do with writing a constraint test.

Moved verbatim to `gcs/constraints/innards/cake_probe.hh`, included by `constraints_test_utils.hh` for its single call site, with a file comment saying what it is and why it is not in there.

**No behaviour change.** With `GCS_TEST_CAKE=1`, `equals_test` still logs all 284 `CAKECHAIN` lines, and none without it.

**Also excluded from install**, alongside `constraints_test_utils.hh` — the install rule now says why both are excluded. Checked by installing into a prefix and looking: 222 headers installed, neither of these among them.

This deliberately does not decide whether the probe still earns its keep now that #432 reports all non-cake-blocked constraints chain-verifying. That is a separate question, and moving it does not foreclose either answer.

**Based on #566**, which is where the argument for it comes from; the review comment in #576 suggested `main`, but #563 rewrites the surrounding part of the same header, so a `main`-based branch would only conflict on rebase.

Suite: 368/368. The 157 `scp_chain` tests are excluded — they fail on this machine before and after any of this work, on a `cake_pb_cp` that predates the current `.scp` format and rejects the input at step [2].

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -306,7 +306,9 @@ install(EXPORT GlasgowConstraintSolverTargets
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/gcs/
         DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/gcs
         FILES_MATCHING PATTERN "*.hh"
-        REGEX "constraints_test_utils\\.hh$" EXCLUDE)
+        # Test-harness headers, not part of the installed API.
+        REGEX "constraints_test_utils\\.hh$" EXCLUDE
+        REGEX "cake_probe\\.hh$" EXCLUDE)
 
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/util/
         DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/util
--- a/gcs/constraints/innards/cake_probe.hh
+++ b/gcs/constraints/innards/cake_probe.hh
@@ -1,0 +1,135 @@
+#ifndef GLASGOW_CONSTRAINT_SOLVER_GUARD_GCS_CONSTRAINTS_INNARDS_CAKE_PROBE_HH
+#define GLASGOW_CONSTRAINT_SOLVER_GUARD_GCS_CONSTRAINTS_INNARDS_CAKE_PROBE_HH
+
+#include <cstdio>
+#include <cstdlib>
+#include <string>
+
+/**
+ * \file
+ * \brief Measurement-only probe: run the workflow-2 cake_pb_cp chain over a
+ * proof a test has just written.
+ *
+ * This is instrumentation, not part of the test harness. It is called from
+ * verify_proof_and_dispose() in constraints_test_utils.hh, and does nothing at
+ * all unless GCS_TEST_CAKE is set in the environment; when it is, it drives
+ * cake_pb_cp and veripb through a shell and logs a greppable line per proof,
+ * so the chain-verify rate across the random data-driven instances can be
+ * measured. It never throws and never fails a test.
+ *
+ * It lives in its own header because none of that has anything to do with
+ * writing a constraint test, and constraints_test_utils.hh -- which every
+ * constraint test includes, and which a test author reads to find out how the
+ * harness works -- is the wrong place to meet popen(), std::system() with
+ * interpolated paths, and a hand-rolled output parser.
+ *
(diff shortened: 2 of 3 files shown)


ciaranm:
**Rebased onto the updated #566**, which is now on current `main` (the stack was 53 commits behind). **531/531 pass**, `scp_chain` included; my earlier note about excluding those tests is withdrawn — see the correction on #565 for why they were failing.

Issue #571, opened 2026-07-27
Global difference logic propagator (x - y <= d), plus a presolver that detects and lifts suitable constraints


ciaranm:
## Summary

gcs has no global propagator for difference constraints `x - y <= d`. Each one is
posted as its own constraint — a two-term `LinearLessThanEqual`, or a `Comparison`
over an offset `ViewOfIntegerVariableID` — and gets its own propagator. Kletzander,
Dekker, Schutt and Stuckey, *Global Difference Constraint Propagation for Constraint
Programming* (arXiv:2607.20022, 22 Jul 2026; journal expansion of their PPDP 2008
paper) show that this is not merely weaker but can be asymptotically worse, and give
the algorithms and the LCG explanations for doing it globally.

Two parts, separable:

1. a global difference-logic propagator, and
2. a **presolver** that scans a posted `Problem` for difference-shaped constraints and
   lifts them into it automatically, so models do not have to be rewritten to benefit.

*(Filed off the back of the MinDistance stack (#569) only because both papers say
"distance". They are unrelated: difference logic's graph has variables as vertices and
computes shortest paths; `MinDistance`'s has sites as vertices and looks distances up
in a fixed matrix. Nothing here affects that stack.)*

## Why

The paper's Example 8 constructs a system where one-propagator-per-constraint reaches
a fixpoint in `O(n^3)` while the global propagator does it in `O(n^2)` — and notes the
whole thing can recur down a branch of the search tree. Their Figure 4 measures it:
350 s versus under 70 ms at n = 1000. Example 1 is the small version — `x - y <= 0`
and `y - x <= -2` are trivially unsatisfiable, but a propagation engine peels one
value off at a time and takes `O(n)` steps to notice.

Nothing about that argument is specific to their solver; it is about propagator
scheduling, and gcs schedules the same way. A grep confirms there is no Bellman-Ford,
potential function or negative-cycle machinery anywhere in `gcs/`.

The counterweight, from their own numbers, is that this is **not** a broad win. On the
MiniZinc challenge benchmark set (626 instances) the global propagator scores 312.94
against the 310.00 baseline — noise — rising to 320.95 with their root simplification
stage. The wins are concentrated on scheduling: `ProdCons` 540.00 → 637.50 and
`RCPSPmax` 550.00 → 586.55, and on `RCPSPmax` the simplification stage alone detects
unsatisfiability on most instances in under 0.01 s. So this should be judged as a
[...]

PR #595, opened 2026-07-28
Difference logic: the MiniZinc and XCSP3 frontends (#571)

ciaranm:
Part 6 of #571: the MiniZinc and XCSP3 frontends. **Rebased onto `main` at
`36999eb5`**, which now contains the rest of the stack, so this is no longer
based on #593 and the diff shown here is against `main`.

Two things in the previous description were wrong, and both are corrected below
rather than quietly dropped: the six sites #596 broke (which is why this could
not be merged), and the claim that the XCSP3 peephole is what makes the
presolver reachable from XCSP3 at all (which #596 stopped being true).

## The six sites, fixed

#596 deleted `DifferenceLogicStats::skipped_unlabelled_comparison` — once
`Comparison` labels every row, nothing can be skipped for want of one — and
this branch read it in three places and described it in three more. A textual
merge was clean and the result did not compile.

`skipped_uncitable_row` is the semantic descendant of the deleted counter, not
`comparison_edges_lifted`: it counts exactly "a donor of the right shape whose
primary OPB row cannot be cited", generalised over both donor families. So the
assertion keeps its meaning and its value.

| site | now |
|---|---|
| `minizinc/fzn_glasgow.cc` | prints `differenceLogicSkippedUncitableRow`, plus `differenceLogicComparisonEdgesLifted` — one `%%%mzn-stat:` line per `DifferenceLogicStats` field |
| `xcsp/xcsp_glasgow_constraint_solver.cc` | the same set, as `d DIFFERENCE LOGIC …` lines |
| `xcsp/CMakeLists.txt` | asserts `SKIPPED UNCITABLE ROW 0`, and `COMPARISON EDGES LIFTED 0` in its place |
| `dev_docs/frontend-support-matrix.md` (`[^cmp]`), `dev_docs/xcsp.md`, `dev_docs/difference-logic.md` | rewritten, see below |

`EDGES LIFTED 4` / `NODES 4` / `SKIPPED NOT TWO TERMS 1` did not have to move,
as #596 predicted they would not.

## What the peephole actually buys, re-measured

The old description said the XCSP3 affine peephole "is what makes the presolver
reachable from XCSP3 at all", with a table showing 0 edges lifted without it.
That was true against #593. It is not true against `main`: since #596 labelled
`Comparison`'s rows, the auxiliary-variable form is liftable too. Measured on
`xcsp/tests/difference_logic.xml` with `--difference-logic`, by compiling a
scratch binary in which the peephole returns `false` unconditionally:

| | edges lifted | of which comparison donors | graph nodes |
|---|---|---|---|
| without the peephole | 4 | **4** | **7** |
| with it | 4 | 0 | **4** |

Same four edges either way. What changes is what the graph is *over*: without
the peephole each edge spans an invented auxiliary rather than the instance's
own variable, so the graph carries a node per constraint for no extra
propagation.

The size argument, which is the real one, is unchanged in shape and re-measured
on the rebased tree (no presolver in play, `--prove`, identical solution sets,
every proof verified by VeriPB 3.0.2):

| instance | int vars | OPB lines/bytes | proof lines/bytes |
|---|---|---|---|
| `tests/difference_logic.xml` (4 edges + 1 three-term) | 9 → **4** | 44 / 5,209 → **19 / 1,500** | 928 / 45,476 → **451 / 15,492** |
| `tests/intension_ordering.xml` (6 constraints) | 10 → **4** | 92 / 7,942 → **62 / 4,832** | 1,484 / 105,911 → **935 / 72,849** |
| `tests/cop_maximize_sum.xml` (one `le(add(x,y),5)`) | 4 → **3** | 18 / 1,397 → **13 / 860** | 270 / 13,012 → **180 / 8,127** |
| 12 variables, 31 difference constraints | 43 → **12** | 242 / 40,137 → **87 / 9,309** | 3,082 / 245,462 → **846 / 45,475** |

On the 31-constraint network that is −64 % OPB lines, −77 % OPB bytes, −73 %
proof lines, −81 % proof bytes.

**No regression on the existing suite**, re-swept the same way: across all 38
committed XCSP3 instances, exactly one of the 36 that predate this PR changes
at all — `cop_maximize_sum`, whose `le(add(x,y),5)` is the only affine
top-level ordering in the whole suite. The other 35 produce OPB and proof files
identical in line and byte count with the peephole on and off, every solution
set and optimum is unchanged, and all 76 proofs verify. That matches the
peephole's reach: the only other top-level ordering anywhere in the suite is
`intension_arithmetic`'s `ge(mul(x,y),6)`, which is not affine and falls back.

## MiniZinc: a flag, and deliberately no predicate

`fzn-glasgow` gains `--difference-logic` (off by default, matching
`gcs::DifferenceLogic`) and `--difference-logic-simplify on|off` (on by
default, matching `simplifying_at_root`), declared in all three `.msc` files'
`extraFlags`. **That is the whole binding.** MiniZinc 2.10's flattener already
emits `int_lin_le([1,-1],[x,y],d)` for `x - y <= d`, for `x + k <= y`, and even
for a bare `x <= y` — checked against 2.10.0, all three come out as
`int_lin_le` — which is exactly the two-term `LinearLessThanEqual` the
presolver lifts. The shape is in the model whether or not anybody asks for it.

A `glasgow_difference_logic` predicate was considered and **rejected**. There
is no standard FlatZinc builtin for a difference-constraint *system*, so it
would be a Glasgow-only extension; it would reach nothing the presolver does
not already reach automatically; and it would require every model that wants
the global propagator to be rewritten against it, which is the thing the
presolver exists to avoid. The general rule, now written down in
`dev_docs/minizinc.md`: add a predicate when the frontend cannot otherwise
express what your propagator needs; add a flag when it can already, and the
only question is whether to switch something on.

Verified end-to-end from the counters, not from timing:
`minizinc/tests/differencelogic.mzn`, a four-event time-lag network, reports
`differenceLogicEdgesLifted=4` over `differenceLogicNodes=4`, with the two
deliberately non-difference constraints landing in `SkippedNotTwoTerms=1` and
`SkippedCoefficients=1` — and `differenceLogicComparisonEdgesLifted=0`, which
is what pins the paragraph above: all four arrived as `int_lin_le`.

## XCSP3: an affine peephole

`recognizeSpecialIntensionCases` stays off. Instead `post_intension_top_level`
tries an affine peephole first for `le`, `lt`, `ge` and `gt`: if both operands
fold into coefficient/constant form (variables, integers, `add`, `sub`, `neg`),
the whole constraint is posted as one `LinearLessThanEqual` over the instance's
own variables. Otherwise nothing has been created and the ordinary walk runs.

Two rules it must not break, both covered by `tests/intension_ordering.xml`
(all four orderings, both operand orders, a negated operand, a non-affine
operand that must fall back), with ACE-generated expected output:

* **Only the four orderings.** `eq` is excluded on purpose: `Equals` over two
  variables is domain-consistent and a linear equality is only
  bounds-consistent, so rewriting it would silently weaken propagation.
  `Comparison` and a two-term linear inequality are both bounds-consistent, so
  for `le` / `lt` / `ge` / `gt` there is nothing to lose.
* **Signs, in both operand orders.** `ge` / `gt` call the helper with the
  arguments swapped and strictness becomes a slack of `-1`. A negated operand
  (`le(neg(x),sub(y,3))`, i.e. `-x - y <= -3`) is still posted — it is a
  perfectly good linear inequality, just not a difference constraint, and the
  presolver counts it under `skipped_coefficients`.

Why `recognizeSpecialIntensionCases` was not used instead: it reroutes
`eq(x,k)` and `ne(x,k)` to `buildConstraintExtension`, a one-variable `Table`
where we post `Equals` / `NotEquals`; its pattern list does not include
`le(add(x,y),k)`, so it catches *less* than this does; and it splits intension
handling across two entry points, which is the property that file was
deliberately built without.

## A pre-existing crash, fixed first (commit 1)

Every auxiliary variable the XCSP3 binding invents was named with a *literal*,
and `Problem` rejects a duplicate name, so two `<intension>`s each wanting an
`addresult` — or two `<abs>` anywhere in the instance, or a second bin packing
— threw `NamingError` and killed the process before search began:

```
$ xcsp_glasgow_constraint_solver --all twoabs.xml
terminate called after throwing an instance of 'gcs::NamingError'
  what():  Duplicate variable name 'absresult'
```

That is most non-trivial instances; it just happened that no committed test had
a pair. All 33 sites now go through one serial-numbering helper. Instance
variables keep their own names, so the caches are unaffected.

## Half-reified edges are out of reach from either frontend

Which matters, because they are where the paper's scheduling wins come from.
`minizinc/mznlib/` declares no `*_imp` predicates (checked: zero matches across
its 57 files), so MiniZinc's flattener never half-reifies and emits
`int_lin_le_reif`, counted under `skipped_reified`; and the XCSP3 binding turns
a top-level `imp(b, le(...))` into an `Or` over a fully-reified control.
Declaring `int_lin_le_imp` in `mznlib/` is the obvious next step for MiniZinc,
but it changes flattening for *every* model, so it wants measuring as its own
piece of work rather than being smuggled in here. Recorded in the support
matrix and in `dev_docs/difference-logic.md`.

## Asserting the presolver actually fired

Per the rule from #588: a presolver that lifts nothing preserves the solution
set, adds no OPB content and leaves every proof verifying — so it passes the
cross-solver diff, the ACE cache diff and VeriPB alike. Both harnesses
therefore grew two optional arguments: extra flags for the Glasgow/gcs run only
(the default solver and ACE never see them, so the cross-check stays honest),
and any number of extended regexes that must each match a line of its output.
Both new tests use them to assert on the counters. Verified the tripwire fails
when the flag is dropped or a count is wrong, and that omitting the new
arguments leaves every existing registration behaving exactly as before.

## Support matrix

Two new rows in the outside-core table — binary comparison as a family, and
difference logic as a system — plus three footnotes: `[^cmp]` on why both
frontends reach a *linear* inequality rather than `Comparison` (now a question
of size rather than of reach, since #596), `[^intaff]` on the peephole, and
`[^dl]` on why neither frontend gained a predicate and where half-reification
stands. The `intension` row picks up a note too.

## Validation

`ctest --preset release -j 6`: **597/597**, which is `main`'s 594 plus this
PR's three, all with `--prove` plus veripb.

Sanitizers, which the previous description had a long apology for: **#597 and
#599 are both fixed and merged**, so that section is gone. `cmake --preset
sanitize` now produces a build with sanitizers actually in it (41 `__asan`
symbols in `xcsp_glasgow_constraint_solver`) and with both frontends enabled.
All **113 frontend tests pass clean under genuine ASan + UBSan** — 38 XCSP3 and
75 MiniZinc, including the three new ones — and so does the whole suite:
`ctest --preset sanitize -j 6` is **597/597**, with no suppressions beyond the
committed `lsan_suppressions.txt` for the FetchContent-ed XCSP3-CPP-Parser.

Refs #571.

https://claude.ai/code/session_018Lx8KBTVNSg44EEcVYcBoN


ciaranm:
The rest of the difference-logic stack landed today (#643 → #583 → #584 → #585 → #588 → #587 → #590 → #592 → #593 → #596 → #600 → #602 → #603), `main` now at `36999eb5`. This PR is retargeted to `main` and left open, because it cannot be merged as it stands.

#596's description called this out, and it verifies — but the failure mode is worse than "conflict", so it is worth recording exactly what happens.

## `mergeable` is CLEAN and the result does not compile

Trial merge of `difflogic-10-frontends` into current `main`, in a scratch worktree:

```
$ git merge --no-commit --no-ff origin/difflogic-10-frontends
Auto-merging dev_docs/difference-logic.md
Automatic merge went well; stopped before committing as requested

$ git diff --name-only --diff-filter=U
(empty)
```

No textual conflict anywhere, and GitHub reports `mergeable=CLEAN` with all 10 checks green. Those checks ran against `difflogic-08-incremental`, which predates #596, so they say nothing about this. A deleted C++ member is not a textual conflict — git merges it happily and the compiler finds out later.

## The six sites

#596 deleted `DifferenceLogicStats::skipped_unlabelled_comparison` (once `Comparison` labels every row, nothing can be skipped for want of one). In the merged tree the field is gone from `gcs/`, but this branch still references it:

| site | breaks as |
|---|---|
| `minizinc/fzn_glasgow.cc:1227` (`differenceLogicSkippedUnlabelledComparison`) | **compile error** |
| `xcsp/xcsp_glasgow_constraint_solver.cc:2079` (`d DIFFERENCE LOGIC SKIPPED UNLABELLED COMPARISON`) | **compile error** |
| `xcsp/CMakeLists.txt:67` (`^d DIFFERENCE LOGIC SKIPPED UNLABELLED COMPARISON 0$`) | test failure |
| `dev_docs/frontend-support-matrix.md` (`[^cmp]`) | stale: says a `Comparison`'s unconditional row is emitted unlabelled |
| `dev_docs/xcsp.md:170` | stale: same counter |
| `dev_docs/difference-logic.md:1695` | stale: "four `skipped_unlabelled_comparison`" |

All six were correct when this branch was cut.

## One thing that should make the fix cheaper

`comparison_edges_lifted` is the natural replacement, as #596 suggests, and the asserted **values** should not have to move. Both test models are deliberately built so every liftable constraint reaches the presolver as a two-term linear rather than a `Comparison` — `xcsp/tests/difference_logic.xml` goes through the intension peephole this PR adds, and `minizinc/tests/differencelogic.mzn` relies on the flattener emitting `int_lin_le` for `t[i] + 2 <= t[i+1]`. So `skipped_unlabelled_comparison` was 0 and `comparison_edges_lifted` should also be 0, leaving `EDGES LIFTED 4` / `NODES 4` / `SKIPPED NOT TWO TERMS 1` untouched. Worth confirming with a run rather than taking from me.

Also merged cleanly and needing nothing: #600's flat → per-directory presolver move. `gcs/presolvers/difference_logic/` is picked up correctly and the `<gcs/presolvers/difference_logic.hh>` umbrella include in both frontends is unchanged.

Not fixing it here — a merge-time patch is code CI has never built, which is the thing that caused this in the first place. A rebase onto `36999eb5` with the six sites updated and a green run is the way in.

ciaranm:
Done, the way the comment above asked for: rebased onto `36999eb5`, the six sites updated, and a green run rather than a merge-time patch. Force-pushed as `e5669e2b`; the description is rewritten around what is now true. All ten checks pass, and this time they ran against `main`.

## The counter

`skipped_uncitable_row`, not `comparison_edges_lifted`. The comment above suggested the latter and #596 does too, but the former is the actual semantic descendant — "a donor of the right shape whose primary OPB row cannot be cited", generalised from the comparison-only case. So the assertion keeps both its meaning and its value, and `comparison_edges_lifted` goes in as a *new* assertion rather than a replacement, where it says something the old one did not.

Both frontends now print one line per `DifferenceLogicStats` field, so the seven-bucket accounting the presolver documents is checkable from the frontend's output.

Confirmed by a run, as asked: `EDGES LIFTED 4` / `NODES 4` / `SKIPPED NOT TWO TERMS 1` did not move, and `COMPARISON EDGES LIFTED` and `SKIPPED UNCITABLE ROW` are both 0.

## One thing that did not survive contact with `main`

The old description's claim that the XCSP3 peephole "is what makes the presolver reachable from XCSP3 at all" is **false against `main`**, and it was load-bearing in three docs, one code comment and one commit subject. #596 is why: once `Comparison` rows carry labels, the auxiliary-variable form is liftable too.

Measured by compiling a scratch binary in which the peephole returns `false` unconditionally, on `xcsp/tests/difference_logic.xml` with `--difference-logic`:

| | edges lifted | of which comparison donors | graph nodes |
|---|---|---|---|
| without the peephole | 4 | **4** | **7** |
| with it | 4 | 0 | **4** |

Same four edges either way. The peephole's remaining value is the graph they span — the instance's own four variables rather than seven including three invented ones — plus the size win, which was always the bigger half and is unchanged in shape. That is now what the docs, the code comment, the commit subject (`xcsp: --difference-logic, over the instance's own variables`) and the test comment all say, and `COMPARISON EDGES LIFTED 0` is what holds them to it.

## Everything else re-measured

Every number in the description was measured against #593 and none of them survived #596 unchanged. All re-taken on the rebased tree, including `cop_maximize_sum` (`18 / 1,397 → 13 / 860` OPB, not the `22 / 2,221 → 17 / 1,691` that was there), and the 38-instance sweep re-run: exactly one of the 36 pre-existing instances changes, all 76 proofs verify, every solution set and optimum unchanged.

## The sanitizer section is gone

#597 and #599 are both fixed and merged. `cmake --preset sanitize` now builds with sanitizers actually in it and with both frontends enabled, and `ctest --preset sanitize` is 597/597 locally. The `Sanitize (ubuntu-26.04)` lane on this PR means something again.

Not merging: waiting on a go-ahead.

PR #603, opened 2026-07-29
Look a constraint's OPB row up, instead of rebuilding its label

### LINKED ISSUE #604: ValuePrecede emits two OPB rows under the same @label, and they disagree with cake's
`ValuePrecede::define_proof_model` builds its per-position role from the position index but not from the chain *value*, inside a loop over values, so every value after the first collides with the first value's labels.

`gcs/constraints/value_precede/value_precede.cc`:

```cpp
// Upper bound: (vars[i] = v) -> pos[v] <= i.
for (size_t i = 0; i < n; ++i)
    model.add_labelled_constraint(_constraint_id, std::to_string(i) + "ub", ...);   // line 113

// Existence: [pos[v] >= i+1] OR (exists k <= i. vars[k] = v).
for (size_t i = 0; i < n; ++i)
    model.add_labelled_constraint(_constraint_id, std::to_string(i) + "ex", ...);   // line 122
```

Both sit inside `for (auto & v : _chain)`. Line 102 in the same function gets it right — `"ubn_" + v.to_string()` — which is what makes this look like a slip rather than a decision.

### Reproduced

```
$ glasgow_scp_solver --prove --proof-files-basename vp verified_encodings/scp_cases/value_precede_sat.scp
$ grep -o "^@c\[[^ ]*\]" vp.opb | sort | uniq -d
@c[_1][0ex]
@c[_1][0ub]
@c[_1][1ex]
@c[_1][1ub]
@c[_1][2ex]
@c[_1][2ub]
```

Six colliding names on a three-variable, two-value instance. The colliding rows are genuinely different constraints, over different values' `pos` variables:

```
22:@c[_1][0ub] -1 v[_1][0_0][pos] -2 v[_1][0_1][pos] 3 ~i[A][eq0] >= 0;
51:@c[_1][0ub] -1 v[_1][1_0][pos] -2 v[_1][1_1][pos] 3 ~i[A][eq1] >= 0;
```

`seq_precede_chain` delegates to the same code, so it has it too.

### cake_pb_cp already gets this right

Same instance, `cake_pb_cp value_precede_sat.scp`:

```
@c[_1][0ex_0]   @c[_1][0ub_0]   @c[_1][ubn_0]
@c[_1][0ex_1]   @c[_1][0ub_1]   @c[_1][ubn_1]
```

Role = position, `_`, value. So this is not a case where our label scheme and cake's need reconciling — **gcs has drifted from cake, and the fix moves it back**: spell the roles `std::to_string(i) + "ub_" + v.to_string()` and likewise for `ex`, matching what line 102 already does for `ubn_<v>`.

### Why nothing caught it

`verified_encodings/scp_cases/CMakeLists.txt` registers both constraints with opbdiff mode `none`:

```cmake
add_scp_chain_test(value_precede_sat            none)
add_scp_chain_test(seq_precede_chain_sat        none)
```

Step 5 of `run_scp_chain.bash` is the `opbdiff --match-labels` oracle, and `none` skips it. So the label-by-label comparison against cake is disabled for exactly the two constraints that have colliding labels. Everything else in the chain still passes, because nothing dereferences these labels: they are definitional rows that no proof step cites.

### Impact

Nothing is broken today — no proof step references `@c[id][<i>ub]`, so no reference resolves ambiguously. It is latent rather than harmless, though:

* the `.opb` contains repeated names, which is malformed as a naming scheme even where it is never dereferenced;
* any future proof step that wants to cite one of these rows cannot, unambiguously;
* `NamesAndIDsTracker::constraint_row` (#603) has to poison such entries and decline to answer for them, which is how this was found.

### Suggested fix

1. Include the value in both roles, matching cake: `<i>ub_<v>`, `<i>ex_<v>`.
2. Check whether these two cases can then come off opbdiff mode `none` onto a real mode, since the label sets would agree. That would be the thing that stops it recurring.

Found while implementing #603, which added the constraint-row registry that made the collision visible.

ciaranm:
**Based on #602** (`difflogic-12-enumeration-check`), which is this PR's base
branch, so the diff shown here is only this PR's four commits.

Follow-up to the review point on #588/#596: the presolver was reconstructing
another constraint's `@label` as a string rather than being handed it.

> [!NOTE]
> **This PR has been redesigned, not rebased.** The previous version added a
> `ProofModel` registry — `track_constraint_row` / `constraint_row`, an
> `unordered_map<pair<ConstraintID, string>, ProofLine>` with its own hasher, and
> duplicate-poisoning logic for when two rows claimed one label. Main has since
> landed #613, which makes two rows under one label a hard `ProofError` at
> emission time, so the poisoning path is now unreachable; and main
> independently fixed #604, which the old version had only filed. What is left
> is smaller and says more, so the registry is gone rather than rebased. None of
> its commits are cherry-picked.

## The problem

`difference_logic.cc` cited a donor's row as:

```cpp
graph.edge_lines.push_back(ProofLineLabel{"c[" + as_string(id) + "]"});
```

That hard-codes the assumption that the row the presolver wants is the donor's
*empty-role* one. It happens to be true for the two reification kinds the
presolver lifts. It is silently false for two it does not: a comparison's
`MustNotHold` and `NotIf` rows go out under the empty role too, and they state
the **negated** inequality with the operands the other way round. The loop is
correct today only because it filters on the reification condition several lines
earlier. Nothing would have caught it if that filter were ever widened, and
nothing at all would have caught a donor renaming a role.

That format is also owned by `ProofModel::emit_constraint_label`, which is
private, and this was the only place in the tree outside `proof_model.cc` that
spelled it out.

## The design

Citing another constraint's row is an inherent dependency — you cannot build a
`pol` on somebody else's row without depending on how they defined their model.
The goal is not to remove the coupling but to make it **typed, narrow and
visible from both ends**.

Two facts make that cheap. Citation is already by label (`ProofLine` is
`variant<ProofLineNumber, ProofLineLabel>`), so nothing needs new plumbing. And
a label is a pure function of `(ConstraintID, role)` — clone-independent,
thread-independent — where a line *number* is a position in a file.

**The constraint publishes the role.** `ConstraintProofModelData<C>` is declared
and deliberately left **undefined**, so asking a constraint that publishes
nothing is a compile error naming the type rather than a `nullopt` to handle. A
specialisation lives in the constraint's own header, beside its posted-argument
accessors — where it is visible to whoever is most likely to break it. What is
published is the role *name*, not a line number.

**The tracker says whether a row went out under it.**
`NamesAndIDsTracker::constraint_row_label(id, role)` answers "may I cite this?",
not "what does it say". `emitted_labels` moves out of `ProofModel::Imp` into the
tracker for one reason: it needs a second reader, and that reader is a
presolver, which holds a `ProofLogger` and no `ProofModel`. Both are constructed
with the same tracker. `claim_labels` keeps its signature and both call sites.

**The enumeration hands over the pair.**
`Problem::each_constraint_of_type_with_proof_data` yields
`(constraint, optional<ProofLineLabel>)`, so the dependency is visible at the
point of use rather than buried in a lookup.

Two properties fall out. A `yes` always names exactly one row — #613 makes two
rows under one label a hard error at emission time, so the ambiguity a lookup
would have to worry about cannot survive to be asked about. And nothing is
stored per solve: under the intended one-OPB, N-thread model,
`define_proof_model` runs once before the fork, `emitted_labels` is write-only
during it and read-only afterwards, and every thread derives the same label from
`(id, role)` without sharing anything.

## What the published roles say

`primary_row_role` returns the empty role for `MustHold` and `If` — the row
stating `sum <= value` under the reification condition if there is one — and
`nullopt` for the other three. The `nullopt`s are the part worth having: they are
exactly the rows the constructed label would have got wrong.

A linear `NotIf` publishes nothing for a different reason, and this PR does not
fix it. Its `ltn` row is emitted as `cond -> sum <= value`, which is what the
constraint says must **not** hold. Nothing constructs a `ReifiedLinearInequality`
with `NotIf` or `MustNotHold` — the six derived classes cover `MustHold`, `If`
and `Iff`, and `scp_reader` builds only those — so the branch is unreachable and
this has never been observable. It is recorded in the header and in `dev_docs`
rather than fixed, because fixing an unreachable proof path belongs with whatever
first makes it reachable.

## `skipped_uncitable_row`

`lift_row` declines when there is nothing to cite, and the new bucket counts it,
keeping the accounting invariant that every donor examined lands in exactly one
of `edges_lifted` or a skip. It is only ever non-zero with a logger: with proofs
off there is nothing to cite and so nothing to be uncitable, and the presolver
must not lift differently under `--prove`.

## Checking

`gcs/constraint_row_test.cc` treats the `.opb` as the oracle rather than trusting
the roles by inspection. It posts one constraint of every reification kind of
both published families, resolves each label through the real API **from inside a
presolver** — the point in the solve where a real one asks — and requires that a
resolved label names a row the file contains, and that a `nullopt` means it does
not. It also pins that the `Iff` and `NotIf` donors really did emit rows, so
"publishes nothing" is not confused with "emitted nothing".

Mutation-tested, both directions:

| mutation | caught by |
|---|---|
| comparison publishes `""` for `NotIf` (the constructed label's mistake) | `constraint_row_test` |
| `constraint_row_label` returns its constructed label without checking the emitted set | `constraint_row_test` |

End-to-end, the check the registry design was validated against, unchanged in
outcome: revert #596 so `Comparison` emits through the void-returning
`add_constraint` again, and on
`difference_chain --mode refute -n 40 -k 8 --variant presolved --donor comparison --prove`
the presolver goes from **901 comparison edges lifted to 0**, all 901 counted in
`skipped_uncitable_row`, and **VeriPB still verifies the proof**. Declining and
staying verifiable is the outcome to want: what is being defended against is an
*invalid* proof, not a weaker one. With proofs off the same model lifts all 901
either way.

## No output change

`.opb` and `.scp` byte-identical across all 91 example snapshots
(`tools/opb_snapshot.bash`) against #602. Full suite **590/590**.

Refs #571.

https://claude.ai/code/session_01Lb9fgE3SAiASdwREZVrU3p

### DIFF
--- a/dev_docs/difference-logic.md
+++ b/dev_docs/difference-logic.md
@@ -630,12 +630,21 @@ The other three reification kinds are *also* difference constraints:
   stand.
 - `reif::MustNotHold` and `reif::NotIf` state the integer negation,
   `y - x <= -d-1`. For a comparison, and for a linear `MustNotHold`, that row
-  goes out under the same empty role and so *is* citable; a linear `NotIf` uses
-  the role `ltn`. These are therefore a **deliberate gap, not an
-  impossibility** — nothing in gcs constructs a comparison with either (there is
-  no user-facing class, and `s_expr()` throws on both, so they have no `.scp`
-  spelling), and closing the gap is worth doing for both donor families at once
-  rather than for one of them.
+  goes out under the same empty role and so *is* citable. These are therefore a
+  **deliberate gap, not an impossibility** — nothing in gcs constructs either
+  form of either family (there is no user-facing class, and for a comparison
+  `s_expr()` throws on both, so they have no `.scp` spelling), and closing the
+  gap is worth doing for both donor families at once rather than for one of them.
+
+  A linear `NotIf` is a special case and not a citable one. Its row goes out
+  under the role `ltn`, but what that row *says* is `cond -> sum <= value` —
+  which is the thing the constraint states must **not** hold. Nothing constructs
+  a `ReifiedLinearInequality` with `NotIf` (the six derived classes cover
+  `MustHold`, `If` and `Iff`, and `scp_reader` builds only those), so the branch
+  is unreachable and the discrepancy has never been observable; it is recorded
+  here rather than fixed because fixing an unreachable proof path belongs with
+  whatever first makes it reachable. `ConstraintProofModelData` publishes nothing
+  for it, so the presolver cannot cite it by accident.
 
 All three are counted in `skipped_reified` rather than guessed at. Two exclusions
 remain soundness requirements rather than mere incompleteness:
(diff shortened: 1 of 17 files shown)


ciaranm:
Pushed 2f1a7ffa addressing both review points.

**`unordered_map`.** No reason for the `std::map` beyond `std::pair<string, string>` having no `std::hash` — which argues for writing a hasher, not for a tree. This file's other lookup tables are already `unordered_map` with a custom hash, for the reason stated above `hash_combine` (`names_and_ids_tracker.cc:80`): nothing iterates them. Nothing iterates `constraint_rows` either. Converted, with a `HashConstraintIDAndRole` next to the existing `HashSimpleOrProofOnlyVariable`.

**The ValuePrecede collision is #604**, and probing `cake_pb_cp` while writing it up made the fix unambiguous: cake spells those roles `<i>ub_<v>` / `<i>ex_<v>`, so gcs has *drifted from* cake rather than disagreeing with it, and the fix moves gcs back — no coordination needed. Same instance:

| | gcs | cake |
|---|---|---|
| upper bound | `@c[_1][0ub]` (twice) | `@c[_1][0ub_0]`, `@c[_1][0ub_1]` |
| existence | `@c[_1][0ex]` (twice) | `@c[_1][0ex_0]`, `@c[_1][0ex_1]` |
| `pos[v] <= n` | `@c[_1][ubn_0]`, `@c[_1][ubn_1]` | same |

The third row is the tell: `value_precede.cc:102` already includes the value and matches cake, so lines 113 and 122 are a slip rather than a scheme.

And the reason nothing caught it: `value_precede` and `seq_precede_chain` are both registered with opbdiff mode `none` (`scp_cases/CMakeLists.txt:500-503`), so step 5's `--match-labels` oracle is skipped for precisely the two constraints whose labels collide. #604 suggests checking whether they can come off `none` once fixed, since that is what would stop it recurring.

Full suite 575/575.

Issue #449, opened 2026-07-02
SeqPrecedeChain: proofs are wrong when two chains are posted in one problem (unnamed ValuePrecede children collide)


ciaranm:
Posting two `SeqPrecedeChain` constraints in one problem with proofs on produces an unverifiable proof — on main, independent of the #445 branch:

```cpp
Problem p;
auto xs = p.create_integer_variable_vector(3, 0_i, 2_i, "x");
auto ys = p.create_integer_variable_vector(3, 0_i, 2_i, "y");
p.post(SeqPrecedeChain{std::vector<IntegerVariableID>{xs.begin(), xs.end()}});
p.post(SeqPrecedeChain{std::vector<IntegerVariableID>{ys.begin(), ys.end()}});
// solve with --prove:
//   veripb: "Propagation engine error: The given solution is conflicting
//            with constraint 94 in the database."
```

The solver's solutions (196) are right; the OPB is not.

## Cause

`SeqPrecedeChain::install` installs a `ValuePrecede` child directly, without giving it an identity, so every chain's child renders as `unnamed`. `ValuePrecede` keys both its labelled constraints (`add_labelled_constraint(id, ...)`) and — worse — its position flags (`create_proof_flag_values_fully_reifying(_constraint_id, {v, k}, ...)`) on that identity. With two unnamed children:

- the OPB contains duplicated labels (26 × `@c[unnamed][...]` in the reproducer), and
- the second child's flag lookups dedupe onto the *first* child's flags, so its OPB constraints tie chain 1's position flags to chain 2's variables. The encoding is semantically wrong, and genuine solutions violate it.

This is the same child-identity failure class that #445 hit with its decomposition children (there it broke RUP; here it breaks the model). `GACGlobalCardinality`'s `In` children look unaffected (`In` emits no id-keyed labels).

## Fix options

- One-liner: give the child a derived identity via `innards::child_constraint_id` (added in #445) — `SeqPrecedeChain` sets it before installing `ValuePrecede`.
- Or fold `SeqPrecedeChain` into the #448 rework (exposed define/propagate rather than a child object), which removes the child entirely.

Either way, a regression test posting two chains in one proofs-on problem belongs in `seq_precede_chain_test`.

PR #613, opened 2026-07-30
value_precede: key the ub/ex roles by value, and reject duplicate @labels at the source (#604)

### LINKED ISSUE #604: ValuePrecede emits two OPB rows under the same @label, and they disagree with cake's

ciaranm:
Fixes #604.

## The fix

`ValuePrecede::define_proof_model` built its upper-bound and existence roles from the position index alone, inside a loop over chain values, so every value after the first reused the first value's labels — two genuinely different rows, over different values' `pos` variables, under one name:

```
@c[_1][0ub] -1 v[_1][0_0][pos] -2 v[_1][0_1][pos] 3 ~i[A][eq0] >= 0;
@c[_1][0ub] -1 v[_1][1_0][pos] -2 v[_1][1_1][pos] 3 ~i[A][eq1] >= 0;
```

The roles are now `<i>ub_<v>` and `<i>ex_<v>`, matching `ubn_<v>` on line 102 and, more to the point, matching cake: `vp_ub_pins` / `vp_ex_pins` in `cp_to_ilp_lexicographicalScript.sml` spell them exactly that way. So this moves back onto cake's scheme rather than inventing one. `seq_precede_chain` delegates here and comes along.

With the roles value-keyed, the label sets of the three `value_precede` cases are **identical to cake's**, and no case in the whole 160-case scp corpus has a repeated label.

## The second question: can these come off opbdiff mode `none`?

No — and for a reason that has nothing to do with this constraint. Measured on `value_precede_sat`:

```
Summary (label-matched, ordered fallback): 60 matches, 3 differing, 0 only in A, 0 only in B.
```

Every one of the constraint's own rows compares equal. All three differing rows are the variables' boundary reifications `@i[X][ge0][r]`, where cake pins the always-true `ge0` with a zero coefficient and the solver reifies it:

```
A: @i[A][ge0][r] 1 i[A][b0] 2 i[A][b1] 1 ~i[A][ge0] >= 0;
B: @i[A][ge0][r] 1 i[A][b0] 2 i[A][b1] 0 ~i[A][ge0] >= 0 ;
```

That is the lazy-vs-eager variable-encoding divergence of #358, which is why *every* case in that section of `scp_cases/CMakeLists.txt` is `none`. `strict`/`aux` here is one #358 away, not one constraint fix away. `seq_precede_chain` has a second, smaller gap on top: cake emits a per-position clamp row `@c[id][<i>cl]` unconditionally, the solver only when the declared domains actually exceed the chain length (they do not, in these cases), so cake's OPB carries 3–4 rows the solver's does not. Harmless — they are redundant there and the chain verifies — but a label-set difference. Both findings are now recorded in the CMakeLists comment, so the next person doesn't have to re-derive them.

## So the guard goes at the source instead

The issue's "the thing that stops it recurring" was the opbdiff oracle, which is unavailable. `ProofModel::add_labelled_constraint` is the funnel every labelled row passes through — including the two-way-reified and equality-half overloads — so it is the one place that sees the whole `.opb`'s naming. It now **throws on a repeated label**, with a message that names it and says what to fix. A duplicate is always a bug in the emitting `define_proof_model`, so it's a hard error rather than a first-wins pick.

This is not per-case opt-in: it covers every constraint, whether or not it has an scp case.

## Testing

Full suite **537/537**. That is also where the confidence in "no constraint legitimately reuses a label" comes from — those runs cover 160 scp chain cases, 36 XCSP models and 74 MiniZinc models, all with `--prove`, plus every constraint test.

Both directions checked by mutation, rather than assumed:

| mutation | result |
|---|---|
| revert the label fix, keep the guard | **10 ctest tests red** — `value_precede_constraint`, `seq_precede_chain_constraint`, their view-mixed twins, all six `scp_chain_*` cases |
| stub the guard out, keep the fix | **`duplicate_label_test` red**: "a second OPB row was accepted under the label 'first'" |

New `duplicate_label_test` pins the guard in two halves, since a check that rejected everything would pass the first alone: distinct labels are still accepted, and the `.opb` left behind stays well-formed with both surviving rows citable from a `pol` step under VeriPB — the rejected row must not leave a half-written line in the file.

Cost is one `set` insert per OPB row at model-definition time, not per proof step. Bounded by row count rather than by a wall-clock A/B: the largest models to hand emit ~8.5k–13.5k labelled rows (`ortho_latin 7` and `8`).

### DIFF
--- a/dev_docs/constraints.md
+++ b/dev_docs/constraints.md
@@ -113,6 +113,28 @@ that does install a child directly (`SeqPrecedeChain`'s `ValuePrecede`)
 must give it an identity, or id-keyed proof flags collide across
 instances (issue #449).
 
+The `role` half of `@c[id][role]` must name **everything the surrounding
+loops vary over**, not just the innermost thing. A role built from a
+position index inside a loop over values gives every value after the
+first the same labels as the first — several genuinely different rows
+under one name, none of them citable unambiguously (issue #604, where
+`ValuePrecede` spelled its upper-bound role `<i>ub` rather than
+`<i>ub_<v>`). This is a hard error, not a first-wins pick:
+`ProofModel::claim_labels` throws on a repeated label, so a role that is
+missing a key fails loudly at model-definition time rather than becoming
+an `.opb` with repeated names that nothing dereferences. It also means
+you cannot work around a collision by ordering the emissions; fix the
+role. The `role_le`/`role_ge` pair of the equality overload is covered
+too, and claimed together, so passing the same role for both halves is
+caught as well.
+
+The check is confined to the `c[id][role]` namespace — it runs in the
+`ConstraintID`-taking `add_labelled_constraint` overloads, which is what
+every constraint uses. The variable-encoding namespaces (`@i[name][...]`
+for a real variable, `@po[index][...]` for a proof-only one) are out of
+scope deliberately: those rows may be deleted and re-emitted to keep the
+proof database small, so a repeat there is by design.
+
 ## The header
 
 ```cpp
--- a/gcs/CMakeLists.txt
+++ b/gcs/CMakeLists.txt
@@ -499,6 +499,10 @@ if(GCS_BUILD_TESTS)
     target_link_libraries(proof_model_tautology_test PRIVATE glasgow_constraint_solver)
     add_test(NAME proof_model_tautology_test COMMAND ${GCS_BASH} ${CMAKE_CURRENT_SOURCE_DIR}/../run_test_and_verify.bash $<TARGET_FILE:proof_model_tautology_test>)
 
+    add_executable(duplicate_label_test innards/proofs/duplicate_label_test.cc)
+    target_link_libraries(duplicate_label_test PRIVATE glasgow_constraint_solver)
+    add_test(NAME duplicate_label_test COMMAND ${GCS_BASH} ${CMAKE_CURRENT_SOURCE_DIR}/../run_test_and_verify.bash $<TARGET_FILE:duplicate_label_test>)
+
     add_executable(eqvar_polarity_test innards/proofs/eqvar_polarity_test.cc)
     target_link_libraries(eqvar_polarity_test PRIVATE glasgow_constraint_solver)
     add_test(NAME eqvar_polarity_test COMMAND ${GCS_BASH} ${CMAKE_CURRENT_SOURCE_DIR}/../run_test_and_verify.bash $<TARGET_FILE:eqvar_polarity_test>)
--- a/gcs/constraints/value_precede/value_precede.cc
+++ b/gcs/constraints/value_precede/value_precede.cc
@@ -108,18 +108,22 @@ auto ValuePrecede::define_proof_model(ProofModel & model) -> void
                 "pge", WPBSum{} + 1_i * pos_v >= Integer{static_cast<long long>(k)}));
         pge.emplace(v, move(ges));
 
-        // Upper bound: (vars[i] = v) -> pos[v] <= i.
+        // Upper bound: (vars[i] = v) -> pos[v] <= i. The role is keyed by both the
+        // position i and the value v: these rows live inside the per-value loop, so a
+        // position-only role would collide across values (#604), giving the .opb several
+        // different rows under one name. cake keys them the same way ("<i>ub_<v>").
         for (size_t i = 0; i < n; ++i)
-            model.add_labelled_constraint(_constraint_id, std::to_string(i) + "ub", WPBSum{} + 1_i * pos_v <= Integer{static_cast<long long>(i)},
-                HalfReifyOnConjunctionOf{{_vars[i] == v}});
+            model.add_labelled_constraint(_constraint_id, std::to_string(i) + "ub_" + v.to_string(),
+                WPBSum{} + 1_i * pos_v <= Integer{static_cast<long long>(i)}, HalfReifyOnConjunctionOf{{_vars[i] == v}});
 
-        // Existence: [pos[v] >= i+1] OR (exists k <= i. vars[k] = v).
+        // Existence: [pos[v] >= i+1] OR (exists k <= i. vars[k] = v). Value-keyed for
+        // the same reason as the upper bound above.
         for (size_t i = 0; i < n; ++i) {
             WPBSum existence;
             existence += 1_i * pge.at(v)[i]; // [pos[v] >= i+1]
             for (size_t k = 0; k <= i; ++k)
                 existence += 1_i * (_vars[k] == v);
-            model.add_labelled_constraint(_constraint_id, std::to_string(i) + "ex", move(existence) >= 1_i);
+            model.add_labelled_constraint(_constraint_id, std::to_string(i) + "ex_" + v.to_string(), move(existence) >= 1_i);
         }
     }
 
--- a/gcs/innards/proofs/duplicate_label_test.cc
+++ b/gcs/innards/proofs/duplicate_label_test.cc
@@ -1,0 +1,107 @@
+#include <gcs/constraint_id.hh>
+#include <gcs/innards/proofs/names_and_ids_tracker.hh>
+#include <gcs/innards/proofs/proof_error.hh>
+#include <gcs/innards/proofs/proof_logger.hh>
+#include <gcs/innards/proofs/proof_model.hh>
+
+#include <string>
+#include <version>
+
+#if defined(__cpp_lib_print) && defined(__cpp_lib_format)
+#include <print>
+#else
+#include <fmt/core.h>
+#endif
+
+using namespace gcs;
+using namespace gcs::innards;
+
+using std::string;
+
+#if defined(__cpp_lib_print) && defined(__cpp_lib_format)
+using std::print;
+#else
+using fmt::print;
+#endif
+
+// Issue #604: a label is how a proof step cites an OPB row, so no two rows in the
+// c[id][role] namespace may carry the same one -- a duplicate makes both uncitable,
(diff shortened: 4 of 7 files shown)


ciaranm:
Pre-merge check. CI is green on all ten lanes including the three Windows/MSVC ones, and the rest of the checklist is clean: no unchecked assertions, no stray files, `duplicate_label_test` is properly wired (`add_executable` + `target_link_libraries` + `add_test` inside `GCS_BUILD_TESTS`, so it runs under both release and sanitize) and it goes through `run_test_and_verify.bash`, so the `pol @first` / `pol @second` steps mean VeriPB really does fail if a surviving row is malformed. The relabelling itself is safe — nothing anywhere reconstructs the old `<i>ub` / `<i>ex` names to cite those rows, only the two emitting sites match — and lining the roles up with `ubn_<v>` next door and with cake's `vp_ub_pins` / `vp_ex_pins` is clearly the right call.

Holding on one thing, though.

## The guard does not cover the equality overload

The check goes in `add_labelled_constraint(const string & label, const WPBSumLE & ineq, ...)`. The description says that is "the funnel every labelled row passes through — including the two-way-reified and equality-half overloads". The two-way-reified part is right: `add_two_way_reified_constraint` builds `base + "[r]"` / `base + "[f]"` and calls the LE overload, so both halves get checked. The equality half is not.

`add_labelled_constraint(const string & label_le, const string & label_ge, const WPBSumEq & eq, ...)` (`gcs/innards/proofs/proof_model.cc:201-228`) does not delegate. It emits both rows itself:

```cpp
    ProofLineLabel le{label_le};
    append_label_to(_imp->opb, le);
    ...
    ProofLineLabel ge{label_ge};
    append_label_to(_imp->opb, ge);
```

`append_label_to` just appends `@` + the label, so neither row is checked against `_imp->emitted_labels`, and — the part that widens it — neither is *recorded* there. So the gap runs both ways: a later row on the checked LE path that happens to reuse an equality-half label also passes silently.

That path is not a corner. Twelve call sites across `gcs/constraints/` reach it, and they include the two in `circuit.cc`:

```cpp
    model->add_labelled_constraint(_constraint_id, "pos_suc_0_" + std::to_string(idx) + "_le", "pos_suc_0_" + std::to_string(idx) + "_ge",
```

Index-computed roles emitted from inside a loop — the same shape as the bug this PR is fixing. It happens to be well-named today (the index is in there), but it is exactly the class of role the guard was added to protect, and it is not protected.

The reason I would rather not merge this as-is is not the missing coverage on its own — a partial guard is still a clear improvement — it is that three places now state the coverage is total:

- `proof_model.hh:155`: "Every labelled row funnels through here, so this is where the .opb's labels are checked for uniqueness"
- `proof_model.cc:242`: "Every labelled row funnels through here, so this is the one place that can see the whole .opb's naming"
- the PR description, as quoted above

Someone adding a constraint later will read that and believe their `role_le` / `role_ge` pair is covered. That is the "a doc comment describing behaviour is not evidence the behaviour exists" failure mode, and it is worth not baking in — especially in a PR whose whole subject is a latent naming collision that no test could see.

Either resolution is fine by me: route the equality overload's two labels through the same insert (a small shared helper, since it needs to check both halves and also catch `label_le == label_ge`), or narrow the three comments to say which paths are actually checked. The first seems better value, given `circuit.cc`'s computed roles.

## Minor, separate

`dev_docs/constraints.md` (the `@c[id][role]` paragraph, around line 104) is where the label scheme and the #449 id-collision lesson are written down. The new rule — a role must name everything its loops vary over, and a duplicate is now a hard error rather than a first-wins pick — belongs in that paragraph, so it is findable from the docs rather than only from the header.

Everything else is ready to go; happy to merge as soon as the above is settled.

ciaranm:
Both fixed, and the equality gap was real — thanks, that was the load-bearing catch. Force-pushed onto the same two commits so the overclaiming comments don't survive into main's history at all.

**The equality overload now claims its own labels.** You were right that it doesn't delegate: it emits both halves itself via `append_label_to`, so the check in the LE overload never saw them, and neither half was *recorded* — the gap ran both ways, exactly as you said. Verified it was load-bearing rather than belt-and-braces: with the new claim stubbed out, all three of the test's equality cases pass silently.

```
an equality whose LE half reuses role 'eqle' was accepted
an equality whose GE half reuses role 'second' was accepted
an equality using role 'samesame' for both halves was accepted
```

Both halves are claimed up front, before either is emitted, so a colliding pair leaves no LE row behind — and claiming them together is what catches `role_le == role_ge`.

**Scope is now stated where it's true, and drawn by API rather than by comment.** `ProofModel::claim_labels` is called only from the two `ConstraintID`-taking overloads. That's the whole `c[id][role]` namespace and nothing else: no constraint passes a raw label string (checked across `gcs/constraints/`), and the CNF-clause overload has no caller outside a test. So "every path a constraint uses" is now a checkable claim about which API you called, not a claim about a call graph that can drift.

The variable-encoding namespaces are deliberately excluded — `@i[name][...]` and `@po[index][...]`. Those rows get deleted and re-emitted to keep the proof DB size down on another branch in flight, so a repeat there is by design rather than a naming bug, and scoping by API keeps this out of that namespace's way entirely rather than relying on prefix-sniffing. Flag-definition labels are likewise untouched; flag-name collisions across instances are #449's business, not this check's. That's a narrowing from what I first pushed, where the check covered every label — worth being explicit about, since it means the guard is smaller than the version you reviewed, not larger.

The three overclaiming comments are gone: the `proof_model.hh` docs on the string overloads now say plainly that they do *no* checking and why, the two `ConstraintID` overloads say what they claim, and `claim_labels` itself carries the namespace rationale.

**`dev_docs/constraints.md`** — added to the `@c[id][role]` paragraph, right after the #449 lesson: a role must name everything its loops vary over, it's a hard error rather than first-wins, you can't reorder your way around it, and the `role_le`/`role_ge` pair is covered too. Plus a short paragraph on what's out of scope, so the exclusion is findable rather than folklore.

While I was in there I also checked the two `@i[name][lb]`/`[ub]` bound rows, which write their labels directly and are likewise unchecked — leaving them alone, same namespace, same reason.

Full suite **537/537** again. Mutation checks: stubbing the equality claim reddens exactly its three cases; stubbing the whole guard reddens all four; reverting the `value_precede` fix still throws on `@c[_1][0ub]`.

Issue #612, opened 2026-07-30
Order-encoding deletion: step-2 Brancher refactor (tracking)


ciaranm:
Tracking issue for the step-2 Brancher-API refactor, which generalises the
consolidate-then-delete order-encoding deletion behind
`GCS_DELETE_ORDER_ENCODING=literals`.

**The decided design is `dev_docs/brancher-design.md`**, and
`dev_docs/order-encoding-deletion.md` is the feature's standing note. Those two are the
authority; this issue is just the index.

The design drivers, campaign data and validated `.pbp` scenarios live in an
`order-encoding-deletion-artifacts/` scratch directory that is **not version controlled
and does not travel with a clone** — check it is present before planning around it.
Both self-contained VeriPB drivers are green on current hardware: `eq-window/run.sh`
**8/8** (the B'' foundation) and `objective-delc/run.sh` **11/11** (the stage-D
mechanics); neither needs a GCS build. `order_jump_check.cc` and `order_hoist_check.cc`
are in there too, still uncommitted — stage E owns promoting them, and until it does
the two foundations this design rests on survive only by luck.

## Stages

| stage | what | status |
|---|---|---|
| A | `BranchDecision` / `BacktrackAdvance` types and plumbing, byte-identical | done — #606 |
| B | split-family bound advances + advance-RUP-driven deletion | done — #606 |
| B' | residency-cause bookkeeping + evict/hoist primitives (behaviour-neutral) | done — #607 |
| B'' | the eq-atom sliding window | done — #615 |
| C | objective / frontier deletion exemption | done — #617 |
| D | `delc` superseded improvement constraints + evict their thresholds | #610 |
| E | benchmark the finished stack, decide defaults, clean up | #611 |
| — | re-measure every figure on current hardware (docs only) | #614 |
| — | delete the solution-blocking constraints (core promotion) | #616 |

The base deletion mode itself is #605. Stage D is the only stage that was gated on work outside this task (the `delc` mechanics, now
validated) and can be prepared ahead, since it reuses the stage-B'' primitives.

**The whole stack stays unmerged until the plan is complete** (owner, 2026-07-30): it adds a
lot of complexity to a central proofs feature, so stage E's benchmarking is a **go/no-go
gate for all of it**, not just a choice of defaults. Confining the risk and the development
cost to one branch beats merging incrementally and having to undo a lot later. Do not
propose merging individual stages when their CI goes green.

## Where it stands

[...]

PR #617, opened 2026-07-30
Brancher stage C: exempt the objective from order-encoding deletion

### LINKED ISSUE #609: Order-encoding deletion stage C: objective / frontier deletion exemption
Part of #612.

Stage **C** of the step-2 Brancher refactor (payload 3). Design:
`dev_docs/brancher-design.md`, "Payload 3 — objective / frontier deletion exemption".

seat-moving 2018's residual ~20.8k delete/reintroduce churn at the default gate of 16 is
concentrated on its **always-bound-tightened objective (chain 98) and cost (76)**
variables. Any win-preserving flat gate leaves those deletable, and they churn
**verify-neutrally** (OFF 305.9 s ≈ L16 298.3 s) for *zero* shrinkage — deletion doing
enormous work for nothing.

The flat chain gate cannot fix this, and that is the point: it is length-based, so it
cannot tell a **win-regime long chain** (a weak-propagation split variable — delete it)
from a **churn-regime long chain** (the perpetually re-tightened objective — keep it
resident). Only the frontier owner knows which is which.

## What to build

- [ ] `note_deletion_exempt(id)` on `NamesAndIDsTracker` — a model/solve-time note, like
      `note_order_encoding_stays_resident`
- [ ] A `FrontierExempt` residency cause, and its slot in `need_gevar`'s ladder: the
      priority becomes **boundary > aux > view > exempt > gate**. It sits just above the
      gate, because an exempt variable is resident regardless of chain length but below the
      structural pins that force residency for their own reasons
- [ ] `solve_with` (or `model->minimise`) exempts the **objective**, since `solve.cc` owns
      the B&B bound-tightening step that makes it a perpetually-re-tightened,
      backtrack-relaxed frontier. This is the primary, measured case
- [ ] Optionally expose it to a brancher for a frontier variable it does not want churned —
      but **not by default**: exempting a bound-branched variable defeats the split win,
      since the win *is* deleting its stepped-over chain

Reuses B'''s residency ladder. `FrontierExempt` and `WindowedFrontier` are sibling slots
and exact opposites — "resident *despite* being frontier" versus "deletable *because*
frontier" — and never apply to the same variable.

## Oracle

VeriPB verifies and search is identical; on seat-moving 2018 the objective/cost churn drops
toward 0; and the exemption must **not** touch non-objective wins — the synthetic sweep's
numbers have to be unchanged.

ciaranm:
Closes #609. Part of #612, stacked on #615 (→ #614 → #607 → #606 → #605).

Branch-and-bound re-tightens the objective at every improving solution and every backtrack
relaxes it again, so under `Literals` its `ge` thresholds are deleted and re-introduced
forever. On seat-moving 2018 that is essentially all of the residual delete/reintroduce
churn at the default gate, and it is churn in the exact sense: **verify-neutral, for zero
shrinkage**. The chain-length gate cannot suppress it, and that is the point — the gate
measures *length*, and the objective's chain is long for a churn reason rather than a win
reason. Only whoever owns the frontier can tell those apart.

So the frontier owner says so. `note_deletion_exempt` on the tracker, a `FrontierExempt`
residency cause, and its slot in `need_gevar`'s ladder, which becomes
**boundary > aux > view > exempt > gate** — below the structural pins (they force residency
whatever anyone wants), above the gate (the exemption holds *however long* the chain grows).
`ProofModel::minimise` makes the note rather than `solve_with`, so every path that declares
an objective gets it; a view objective is exempted through its underlying.

146 lines, most of it comment.

## Two departures from the design

**The exemption also refuses to *window* a variable.** `FrontierExempt` and B''s
`WindowedFrontier` are opposite answers to the same question — resident *despite* being
frontier, versus deletable *because* frontier — and the design says they never apply to the
same variable. Nothing stopped them: a branched objective is both, and `tour` branches on
its objective, so this was not hypothetical. Making the exemption win in
`need_direct_encoding_for` keeps that claim true by construction, and it is the right way
round — exempting a variable says "do not churn this encoding", and windowing its eq atoms
is churning it.

**The brancher-facing half is not built.** The design offers it "optionally… but not by
default"; an API with no caller is speculation, and a brancher that wants it can call
`note_deletion_exempt` when a use appears. Exempting a bound-branched variable would defeat
the split win anyway — that win *is* deleting its stepped-over chain.

## Measured — the oracle is two-sided and both sides hold

**It must not touch non-objective wins.** The synthetic split sweep
(`order_deletion_bench --unsat`, d250, gate 0) is **byte-identical** to the stack's previous
tip: no objective, so the note is never made and nothing changes.

**It must suppress the objective's churn.** Same synthetic with `--optimise`:

| | deletes | reintroductions |
|---|---|---|
| before, gate 0 | 812 | 180 |
| after, gate 0 | **726** | **172** |
| before, gate 16 | 656 | 141 |
| after, gate 16 | **591** | **137** |

with the proof slightly smaller (463 839 → 460 736 bytes), and **78 Top-resident `ge`
atoms — 97.5 % of all Top residency at gate 0 — now attributed to the exemption**. On the
real optimisation examples the churn goes to zero outright (tour 4 → 0 deletes, colour 2/1
→ 0/0 at gate 0), though those are small enough that the absolute numbers say little.

## Byte-identity

Mode `None` is untouched. Under `Literals` this stage deliberately **does** move the proof —
but only for optimisation instances, and only their `.pbp`, never the `.opb`. Of six
instances compared against the stack's previous tip: colour and talent differ; crystal_maze,
langford, money and sudoku are byte-identical. That asymmetry is the feature — the exemption
should be invisible to everything without an objective.

## Policy, not correctness

Unlike every other stage in this stack, nothing is stranded without this one, so there is no
proof that could reject to tell you it stopped working. `deletion_exempt_test` therefore
asserts the residency fact in C++ **both ways round** — an exempt variable's thresholds
survive a forget, an ordinary variable's do not — because an exemption that quietly held
*everything* resident would suppress the churn and the deletion win together. Mutation-
checked: a no-op `note_deletion_exempt` fails it. VeriPB still runs over the proof, and the
test uses the exempt literals *after* the forget that would have deleted them, so a
bookkeeping-only lie rejects rather than passing.

## Gates

Caps-off suite **547/547** in each of {None, Literals gate 0, Literals gate 16}.

### DIFF
--- a/dev_docs/brancher-design.md
+++ b/dev_docs/brancher-design.md
@@ -712,7 +712,8 @@ gate keeps its job for the **non-frontier short-chain tail**
 (crystal_maze / talent eq atoms that are *not* branched on), which the window
 never touches.
 
-This composes with payload 3's `FrontierExempt` (below) as its exact opposite:
+This composes with payload 3's `FrontierExempt` (below, **built in stage C**) as its exact
+opposite:
 `FrontierExempt` says "resident *despite* being frontier" (the churn-regime
 objective); `WindowedFrontier` says "deletable *because* frontier" (the win-regime
 eq branch). They are the two opposite frontier policies and only the frontier
@@ -727,7 +728,9 @@ it is a reason to stay *deletable*, and a ge born deletable takes no cause at al
 `windowed_eq_variables` set, marked when the first windowed definition on that
 variable is minted — before the definition is emitted, because emitting it is what
 mints the two thresholds that must come out deletable. `FrontierExempt` (stage C)
-*is* a Top cause and will want the enumerator.
+*is* a Top cause and duly took the enumerator, plus its own line in the stats dump and its
+own per-variable class there — otherwise the gate would be credited with variables the
+exemption held.
 
 ### The one hidden pin — eq⨯interval (bidirectional guard)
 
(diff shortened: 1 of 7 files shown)


ciaranm:
Two notes on what this stage does and does not measure.

**seat-moving 2018 is not measured here.** Its ~20.8k churn at gate 16 is what motivates the exemption, so the obvious question is what that number became — and the answer is that it belongs to stage E, not to a per-stage gate. The instance's proof passes **8 GB** and the solve runs well past half an hour; I started it, let it run 35 minutes, and stopped it. Stage E re-measures every figure as the stack's go/no-go anyway. Recorded in the design note rather than left to be assumed.

What *is* established here is that the mechanism does what it claims, on an instance small enough to check **both sides** of the oracle in one sitting — the synthetic must not lose its non-objective win (byte-identical) and must lose the objective's churn (812 → 726 deletes at gate 0).

**A technique worth reusing.** When only the solver-side counters matter, the proof need not be written at all:

```
mkfifo run.pbp ; cat run.pbp > /dev/null &
fzn-glasgow --prove --proof-files-basename run model.fzn
```

The `.opb`/`.scp`/`.varmap` stay tiny and the multi-gigabyte `.pbp` never touches disk, so `GCS_ORDER_ENCODING_STATS`, `--stats` and search-shape comparisons all work at zero disk cost. Useful on this VM, where a stray `--prove` has filled the disk before now.

PR #632, opened 2026-07-31
Add dev_docs/proof-benchmarks.md: a benchmark set for proof work

ciaranm:
`benchmarking.md` curates instances for *solve* speed and says explicitly not to put `--prove` in its numbers. There has been no counterpart set for when the proof itself is what is being measured, so proof-logging work has been sized ad hoc.

The two sets turn out to be almost disjoint. Sizes that make a good solve benchmark are usually far too large to proof-log: `ortho_latin --size=6 --all` and `tsp`, both in the solve table, each write over 4 GB of `.pbp`.

## Why grouped by failure mode rather than size

The set separates instances that stress proof **checking** from those that stress proof **writing**. These are close to opposites, and trading proof bytes for checking time is the normal shape of a proof-side optimisation, so a change that helps one group routinely hurts the other:

| extreme | proof | verify | cost per byte |
|---|--:|--:|--:|
| `odb_split2000` | 17 MB | 106.1 s | 6.2 s/MB |
| `knapsack_bench --instance 2` | 138 MB | 2.4 s | 0.017 s/MB |

That is a **360×** spread, and it widens past 1200× across everything screened. Measuring only the first rewards writing more proof to check it faster; measuring only the second rewards the reverse.

A third corner, reachable only through the MiniZinc frontend, is the checker being bound by the *model* rather than the derivation: `aircraft-disassembly` writes a **2.5 GB `.opb`** against an 8 MB `.pbp`. It is listed separately because it needs about 19 GB of veripb resident memory.

## What was screened out

Most of the work was rejection, and the document records it so the screening does not get repeated: instances too large to proof-log at all, and two parameters that turn out not to be parameters —

- **`colour` is instance-dependent, not size-monotone.** 41 vertices gives 16 MB, 42 gives 120 MB, 46 gives 736 MB, and a different 42-vertex graph blows an 8 GB cap.
- **`regular_random` has no usable size.** n=6 checks in 0.6 s, n=7 takes over 1200 s, n=9 exceeds the cap. `Regular` is consequently unrepresented at scale, which is noted as a known gap.

Plus the argument-shape traps that produce an immediate parse error: the `minicp_benchmarks` binaries and `bin_packing_bench` reject `--stats`; `ortho_latin`/`magic_square` spell it `--all-different gac|vc|not-equals`; `skeleton_puzzle` needs `--seed` for any non-default shape; `frequency_square` takes its size positionally and divisible by `--lambda`.

## Corrections to benchmarking.md

- Cross-referenced in both directions, and from the `dev_docs` index.
- `nonogram --random 20 --seed 1 --all` is recommended there as the structured `Regular` instance. Measured, it is **17 recursions and 3 solutions** — inside the "tens of recursions means a no-search bench" band that same document defines two sections earlier. The generator builds clues from a random picture, so the puzzle is nearly determined. Now says to use the `--dzn` instances instead.
- Notes that concurrency inflates small proofs far more than large ones (2.99× for a 205 MB proof against 1.07× for a 188 MB one under the same load), so a single correction factor cannot be used to extrapolate solo times across this set's range.
- The "put timed proof I/O on tmpfs" advice is not carried over: it was formed where the alternative was network storage, and on the larger entries tmpfs competes with veripb for the same RAM.

## Since the pre-merge check

The three inconsistencies flagged in review are settled against the retained solo pass, not by picking a side:

1. **The duplicated `knapsack --upfront` row.** The 4.5× version was the pilot-load measurement, left behind when the solo pass replaced the figures. The solo numbers are 7.00 s against 2.00 s — 3.5×, which is what "the axis that matters most" already quoted. One row now.
2. **`hitori`'s recursion count.** The two numbers are two different instances. The group A row is the *generated* `--size 5 --seed 1` at 72 361 recursions; the port validation was against `fzn-glasgow` on the Challenge puzzle read through `--dzn`, at 113 671. Both re-run on current main, both reproduce. The document now says which is which and why the row is the generated one.
3. **`odb_split1000`.** Rewritten against `odb_split2000`, which is tabulated. The spread within the set is 360×, and the across-everything figure now follows from `regular_random`'s 21 s/MB, at over 1200×.

The `.pbp`-per-recursion range had the same defect and is fixed with it: `colour46` is 2.4 KB per recursion, not under 1 KB. The tmpfs table was in `du -h` mebibytes against decimal megabytes everywhere else, which is the `481`/`504` discrepancy; converted, and it now says those three times are their own sitting. Two further measured claims did not match the data — `qap --size=11` does not verify in over 900 s, it fails to finish inside 900 s; `multiply_random` at `-n 1000000000` writes 1.02 MB, not under 1 MB.

## The rcpsp rows are re-measured

Checking the above meant re-running every recursion count in the document against `main` at `d3c9e58b`. All of them reproduce exactly **except `rcpsp`'s**, which are 57 recursions shorter in both forms — 171 480 → 171 423 and 42 115 → 42 058. Nothing in `examples/rcpsp` changed; it is the `Cumulative` work merged since the first pass, strengthening the initial propagation. Sizes and verify times for those rows are re-measured accordingly (`rcpsp_dl21` is now 1 242 MB and 139.5 s), and the deadline row gains the `solve`/`+proof` columns it was missing.

That is the trap the section already warned about, arriving from the direction it did not name: the warning was about the generator drawing one more random number, but a propagator getting one inference stronger stales the same rows just as silently, and nothing errors either way. The document now says both.

### DIFF
--- a/dev_docs/README.md
+++ b/dev_docs/README.md
@@ -46,6 +46,12 @@ library. For an introduction to *using* the solver, start with the top-level
   measuring the wall-time impact of a performance-sensitive change, the
   rationale for each pick, the harness pattern for comparing two builds,
   and what to capture. Use when quantifying a refactor's perf impact.
+- [Proof benchmarks](proof-benchmarks.md) — the counterpart set for when
+  the *proof* is what is being measured: proof-writing cost, proof size
+  and VeriPB checking time. Groups instances by whether they stress
+  writing or checking, records which candidates are too large to
+  proof-log at all, and lists the argument-shape traps. Use when
+  changing proof logging, scaffolding, encodings or hinting.
 - [Frontend support matrix](frontend-support-matrix.md) — single source of
   truth for which gcs propagators each frontend (MiniZinc, XCSP3, CPMpy)
   exposes, plus where the solver-side gaps are tracked. Update when adding
--- a/dev_docs/benchmarking.md
+++ b/dev_docs/benchmarking.md
@@ -10,6 +10,15 @@ The set was settled on while doing Phase 2 of issue #134. The intent is that
 future performance work re-uses it, so that results are comparable across
 PRs over time.
 
+**If the change under test is on the proof side**, use
+[proof-benchmarks.md](proof-benchmarks.md) instead: it curates a separate set
+of instances for measuring proof-writing cost, proof size and VeriPB checking
+time. The two sets are almost disjoint, because sizes that make a good solve
+benchmark are usually far too large to proof-log — `ortho_latin --size=6 --all`
+and `tsp` from the table below both write more than 4 GB of `.pbp`. The
+"Benchmarking proof-shape changes" section further down remains the methodology
+for that work; proof-benchmarks.md is the instance set to apply it to.
+
 ## What to run
 
 Eight benchmarks, picked to cover a mix of search-heavy / propagation-heavy
@@ -166,6 +175,13 @@ constraint variant, a refactor of how a propagator emits scaffolding,
 moving from per-call to upfront derivation — is a different kind of
 performance change with different signals and different traps.
 
+The rest of this section is the methodology. For the instances to apply it
+to, see [proof-benchmarks.md](proof-benchmarks.md), which curates a set sized
+so that VeriPB takes minutes rather than milliseconds, and which separates
+instances that stress proof *writing* from those that stress proof
+*checking* — the two are close to opposites, and a change that helps one
+routinely hurts the other.
+
 ### What to capture
 
 - **`solve` time** — usually goes up modestly (more proof I/O), but
@@ -263,6 +279,14 @@ instances of increasing size; `--random N` scales continuously without
 needing a data file. Enumerating all solutions (`--all`) of an
 under-clued or random instance is the proof-shape regime here, exactly
 as for `regular_random --all`.
+
+**`--random N` does not deliver that regime, though.** The generator
+builds the clues from a random picture, so the puzzle it poses is very
+nearly determined and `--all` barely searches: `--random 20 --seed 1
+--all` measures **17 recursions and 3 solutions**, which is inside the
+"tens of recursions means a no-search bench" band described two sections
+above. `--seed 3` gives 3 recursions. Use the `--dzn` instances, which
+are genuinely under-clued, and treat `--random` as a smoke test only.
 
 ### Cross-variant invariants
 
--- a/dev_docs/proof-benchmarks.md
+++ b/dev_docs/proof-benchmarks.md
@@ -46,12 +46,21 @@ local SSD. `veripb` 3.0.2.
 | `freqsq6_gac` | `frequency_square 6 --all --stats --consistency gac` | enum, 53 220 | 105 825 | 44 KB | 205 MB | 0.90 s | ×2.0 | **83.6 s** |
 | `freqsq6_bc` | `frequency_square 6 --all --stats --consistency bc` | enum, 53 220 | 105 825 | 44 KB | 80 MB | 1.10 s | ×1.3 | **81.2 s** |
 | `pdisp10_tuple` | `p_dispersion --grid 10 -p 4 --variant tuple --stats` | optimal | 1 037 | 12.6 MB | 158 MB | 0.10 s | ×4.0 | **88.8 s** |
-| `mzn_hitori` | `fzn-glasgow -s -a hitori.fzn` (Challenge 2025, `h5-1.dzn`) | optimal | 113 671 | 1.2 MB | **1 000 MB** | 1.20 s | ×3.1 | **273.3 s** |
-| `mzn_seatmoving` | `fzn-glasgow -s -n 1 2018_seat-moving.fzn` | first solution | 15 610 | 3.1 MB | 629 MB | 0.40 s | ×3.8 | **122.3 s** |
+| `hitori` | `hitori --size 5 --seed 1 --quiet --stats` | optimal | 72 361 | — | 937 MB | — | — | **260.5 s** |
+| `table_layout` | `table_layout --size 15 --seed 1 --stats` | optimal | 17 768 | — | 146 MB | — | — | **102.6 s** |
+| `seat_moving` | `seat_moving --dzn sm-10-12-00.dzn --quiet --stats` | first solution | 15 610 | 2.4 MB | 645 MB | — | — | **119.5 s** |
 
 `solve` is wall time with proof logging off; `+proof` is the multiplier when it
 is on. Note how far that varies — from ×1.2 on `colour46`, whose search dominates,
-to ×9.0 on `odb_cumulative8`. One pass over group A costs about 47 minutes.
+to ×9.0 on `odb_cumulative8`. One pass over group A costs about 55 minutes.
+
+The last three rows are the native ports from issues #633–#636; they carry no
+`solve`/`+proof` column yet because they were measured after the pinned pass.
+`hitori` and `seat_moving` replaced `fzn-glasgow` entries and were validated
+against them — **recursion counts match exactly** (113 671 and 15 610), so they
+reproduce the search rather than merely the answer. `table_layout` gives `Table`
+a size where proof shape is measurable at all; `examples/tables` is 10
+recursions.
 
 `langford11` at 11.0 minutes sits just outside the nominal ceiling. It is kept
 deliberately as the largest entry; with `langford10` at 60.4 s the pair brackets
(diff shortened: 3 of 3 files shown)


ciaranm:
Pre-merge check on `dev_docs/proof-benchmarks.md`. Three internal inconsistencies — flagging rather than fixing, since each needs the underlying measurement to resolve. Holding the merge until they're settled.

**1. Group C lists the `knapsack_bench --upfront` pair twice, with different numbers.**

Lines 94 and 95 are the same comparison (`--instance 1` vs `--instance 1 --upfront`), and agree on the size ratio ("**5.9× smaller**", 20.7 MB against 121.3 MB) but not on the checking cost:

- L94 — **3.5× longer**: 7.0 s against 2.0 s
- L95 — **4.5× longer**: 9.9 s against 2.2 s

"The axis that matters most" (L189–191) uses the 3.5× figures. It reads like a re-measured row that was added without the original being taken out. Since this pair is the document's headline SIZE≠TIME demonstration, cited in three separate places, it's worth having one set of numbers.

**2. `hitori`'s recursion count disagrees with itself.**

The group A row (L50) gives **72 361** recursions. The paragraph immediately below (L60–62) says the native ports were validated against their `fzn-glasgow` predecessors and that "recursion counts match exactly (**113 671** and 15 610)". `seat_moving`'s 15 610 does match its row (L52); `hitori`'s doesn't match either figure in the other's place.

That paragraph's whole claim is that the ports reproduce the search rather than merely the answer, so this is the one number in it that has to be right.

**3. `odb_split1000` is cited but never tabulated.**

L157 uses it as the high end of the seconds-per-megabyte range — "writes 6 MB and spends 10.2 s checking it (1.61 s/MB)". No table in the document has an `odb_split1000`: group A has `odb_eq1000` (37 MB) and `odb_split2000` (17 MB), and neither is 6 MB. Consistent with that, the "about 190×" spread claimed in the same sentence isn't what its own two examples give (1.61 / 0.017 ≈ 95×), though it is roughly what the tabulated entries give — so it looks like the example was left behind by a change to the set rather than the ratio being wrong.

---

Minor, and only worth a look while the above are being fixed: group A and the tmpfs table appear to disagree on units. `qap10` is 504 MB in one and 481 in the other; `pdisp10_tuple` is 158 and 151. Both are exactly the MB/MiB factor, and proof sizes shouldn't otherwise vary between sittings.

For the record, the `order_deletion_bench` entries are *not* a problem — that binary isn't on `main` (it lives on the order-encoding-deletion branch), but ciaranm confirms that's ongoing work the document is deliberately independent of. Everything else the document names was checked against the tree: all example and benchmark binaries exist, and the documented options and argument-shape traps (including `--stats` being absent from the `minicp_benchmarks` binaries and `bin_packing_bench`, and `frequency_square`'s `--consistency bc|gac`) match the sources.

ciaranm:
All three are fixed, each from the retained measurement rather than by picking a side. Rebased onto `main` while I was in there.

**1. The duplicated `--upfront` row.** The 4.5× version is the pilot-load measurement: it was added by "Record the `--seed=-1` trap", and "Replace pilot figures with the measured solo pass" added the 3.5× row without taking it out — exactly the read you had. The solo pass has `knapsack1` at 121 255 063 B / 2.00 s and `knapsack1_upfront` at 20 722 828 B / 7.00 s, so 5.9× smaller and 3.5× longer are the right pair, and they are what "the axis that matters most" was already quoting. One row now, carrying the stronger of the two wordings.

**2. `hitori`.** The two numbers are two different instances, which the document never said. The group A row is the **generated** `--size 5 --seed 1` at 72 361 recursions; the validation was on the Challenge puzzle read through `--dzn`, at 113 671, which is the run that had to match `fzn-glasgow`. `seat_moving`'s row *is* its validated `--dzn` run, which is why its 15 610 matched and `hitori`'s did not. Both re-run on current main and both reproduce exactly. The document now says which row is which and why `hitori`'s is the generated one — it needs no external data.

While fixing that I dropped the claim two sections down that a port must reproduce "`.opb` size, `.pbp` size and recursion count". Only the recursion count matched, and it is the only one that should be expected to: `hitori --dzn` writes 1.0 MB / 815 MB against the flattened model's 1.2 MB / 1 000 MB, and `seat_moving --dzn` 2.4 MB / 645 MB against 3.1 MB / 629 MB. Measured figures are in the document now.

**3. `odb_split1000`.** Real measurement, never tabulated — the entry was dropped from group A in favour of `odb_split2000` and the example did not follow it. Rewritten against a row that is in a table: `odb_split2000` at 6.2 s/MB against `knapsack2`'s 0.017, a spread of **360×**, not 190×. The across-everything-screened figure now follows from `regular_random --bacchus`'s 21 s/MB that "Known gaps" already quotes, so it reads as over 1200× rather than 1700×.

The `.pbp`-per-recursion range had the same defect: `colour46` is 2.4 KB per recursion, not under 1 KB, and neither is `frequency_square` under GAC (1.9 KB). It is 1700× from `freqsq6_bc`'s 0.8 KB.

**The units.** Good catch — the tmpfs table came from `du -h`, so it was mebibytes against decimal megabytes everywhere else. Converted, and it now says those three verify times are their own sitting and are only to be read along the rows.

---

**One thing your check turned up that neither of us was looking for.** Re-running the recursion counts to settle (2) meant re-running all of them, against `main` at `d3c9e58b`. Every row reproduces exactly **except `rcpsp`**, which is 57 recursions shorter in both forms: 171 480 → **171 423** and 42 115 → **42 058**. Nothing in `examples/rcpsp` changed — it is the `Cumulative` work merged in between, strengthening the initial propagation.

So those rows are re-measured, solo and pinned as before: `rcpsp_dl21` is 1 242 MB and 139.5 s, the optimisation form 4 104 MB and 702.7 s, and the deadline row picks up the `solve`/`+proof` columns it was missing. The variant triple survives — `global` and `presolved` still make identical search to `decomposed`, now within 0.8 % on proof size.

This is the trap that section already warned about, arriving from the direction it did not name. The warning was about the generator drawing one more random number; a propagator getting one inference stronger stales the same rows just as silently, and nothing errors either way. It says both now.

Also, since you raised it: added a line saying `order_deletion_bench` is not on `main`, so the three `odb_*` rows need that branch.

ciaranm:
Follow-up, on `colour` g40 — it was inconsistent in three ways and I have taken it out.

**It appears exactly once in the document.** Only as the low end of the load-inflation range in "Reproducibility". It is in no group, in no screening table, and not in the "`colour` is instance-dependent" bullet that is precisely about which `colour` instances behave how. That is the same defect as the `odb_split1000` citation you found — a number carried by an instance the reader has no other sight of.

**It is not one of this document's graphs.** Everything else here names `colour` instances by vertices and density (`colour46`, "41-vertex", "42-vertex (50 % density)"). `g40` is a bare filename stem from the order-encoding-deletion campaign's artifact directory, a different corpus entirely, and the document never says so.

**Its solo baseline is not recoverable.** The loaded 77.7 s is in the pilot data; the 72.8 s solo that makes the 1.07× exists only in my working notes, which are not in the repository. Every other reproducibility figure in the document can be recomputed from rows that are in it.

Replaced with rows that are. Twenty of them were verified both solo and under pilot load on byte-identical proofs, so it is now a table rather than a sentence, and it says more than g40 did: the range is **1.05×–2.99×**, the **largest** proof in the comparison is the **least** inflated, and on the one pair holding everything else fixed — the two `frequency_square` rows, same model, same search — it is the **smaller** proof that degrades less. That is the direct opposite of the "small proofs degrade worse" rule the passage exists to refine, which g40 could only gesture at.

**And while checking that, the instance-dependence bullet had the same weakness.** It read "41 vertices → 16 MB, 42 → 120 MB, 46 → 736 MB, and a different 42-vertex one blows the cap" — nearly monotone in vertices, which undercuts the heading it sits under; the only non-monotone evidence was the last clause. The screened graphs make the point outright once the edge count is held roughly constant (383–445 edges): 44 vertices writes 869 MB, **45 writes 5.6 GB and does not verify inside 1200 s**, and 48 writes 657 MB from an eighth of the search. Three more vertices on the same edge budget, an eighth of the proof. The 45-vertex graph is now in the screening table too, where it belonged anyway.

ciaranm:
All five done.

**1. The superlative.** You are right and it is worse than a slip — the table right above it is sorted by inflation and bolds `odb_cumulative8`, so the sentence contradicts the thing it is describing. `colour46` is third, not first.

Replaced with what the data supports. There is a weak inverse tendency, rank correlation **−0.41**, and the useful observation is the one g40 was reaching for, now on rows that are in the document: `langford10` at 101 MB inflates **1.19×** and `freqsq6_bc` at 80 MB inflates **2.41×** — nearly the same size, opposite ends of the range. The `frequency_square` pair keeps its own sentence, unchanged. Commit message corrected too, by amend.

**2. Twenty stated, eight shown.** Removed the selection instead of documenting it: the table is now **every group A and B row except the four measured after the pilot**, all fourteen. That is a stated rule rather than an unexplained subset, and the range is visible at both ends of the printed table rather than asserted over rows that are not there.

**3. `magic_square`.** Falls out of the table with (2), since it is group C. Its group C entry now carries the proof size and verify time its neighbours already had — 169 MB / 29.1 s against 164 MB / 30.1 s — so it is no longer the one row without numbers.

**4. The duplicated 139.5 s.** Worth the glance. **It is real.** Regenerating the proof gives the same 1 242 344 164 bytes, and four fresh runs give 140.58, 139.69, 139.23, 139.82 — the recorded 139.50 sits inside that spread, so the collision with `odb_eq1000` is coincidence.

Taking the minimum over all seven runs, which is what the rule two sections down says to do, makes the row **139.2 s** — and incidentally ends the coincidence. I kept the spread rather than discarding it, because the document had no solo run-to-run figure anywhere: **139.2–140.6 over seven runs is about 1 %**, and group A opens by claiming a single-digit-percent change is resolvable. That claim now has a number under it.

**5. The `cap` threshold.** Stale definition, not a stale figure — the screening did not run one limit. Reading the recorded sizes back, the kills sit at **3, 4, 6 and 8 GiB** depending on the sitting: `tsp` and the g50/g60 graphs at 4, the `seat_moving` find-first sizes and `rcpsp18` and `hitori --density` at 6, three `rcpsp` seeds at 3, everything else at 8. The definition now says the limit varied and that a capped row means "out of reach" rather than a size, which is all it ever supported. The `tsp` row says which limit it hit.

---

Separately, and not part of this PR: #709 adds `concurrency` groups to the three workflows. Three force-pushes on this PR each left ten jobs building commits that no longer existed.

PR #646, opened 2026-08-02
Pin what a pol names, via the union over its operands (Brancher stage G)

ciaranm:
Closes the `pol`-reference gap that `dev_docs/order-encoding-deletion.md` recorded as
theoretical. It is not theoretical, and the instance is in the **shipped** configuration.

## The bug

```
GCS_DELETE_ORDER_ENCODING=literals magic_square --size=4 --all-different gac --prove
  -> Proofgoal 3515 could not be autoproven  (at pbp:4310)
```

`main` verifies this in 28.7 s. With the feature on it is **rejected in 0.10 s** — same
search (77 983 recursions), proof only **228 bytes** larger. Found by the PR #632 benchmark
sweep, on a **group C control row** that was expected to be an exact no-op.

A `pol` line's literals are the *result* of a cutting-planes derivation, which gcs never
evaluates. So a `pol` landing at Top over operands naming a windowed eq atom is a permanent
reference nothing can see — the hoist-out rule walks an emitted sum's terms, and a pol has
no terms to walk. The window evicted the atom anyway, and the "re-introduction" at
pbp:4309–4310 was not one: the atom had never been free.

Constraint 3515 is all-different's Hall-set at-most-one row (`all_different/justify.cc:39`),
a `PolBuilder` fold of fifteen pairwise clauses naming `i[_13][eq8]`. Inserting
`del id 3515` before the failing step makes it pass, confirming it as the sole blocker.

It is the eq-side member of the family stage F (#645) fixed on the ge side, reached by a
path stage F's rule cannot see — `evictions` is **0** in every config here, so
`evict_order_literal`'s ancestor guard never runs.

## The fix, and why it is automatic

**Cutting planes cannot invent an atom.** Addition, multiplication, division and saturation
all yield a constraint over a subset of the operands' variables, and the only other sources
(`add_for_literal`, `weaken`'s flag) are handed to `PolBuilder` directly. So the union over
the operands' recorded names over-approximates what the result names — soundly, with
nothing for a caller to declare or forget. `PolBuilder::emit` computes it, pins it when the
line lands at Top, and records it against the new line so a pol over pols composes.

That is why this is not a second level argument on the emitters, which was the other option
considered: a declaration is a second source of truth that drifts from the expression it
describes, and a missing one is silent until deletion is enabled *and* an instance hits it —
exactly how this survived every green suite until a benchmark sweep tripped over it. It
would also have missed this bug entirely, since the offending site is a `PolBuilder::emit`
and not an `emit_rup_proof_line`.

## What it costs

A `ProofLine -> named eq atoms` map — the point at which we start tracking the live contents
of the proof, and the part worth watching. Scoped to what is live, not to the proof: entries
only for lines that name an atom, nothing recorded unless the window is live, and
`forget_eq_atom_names_for` drops each level's entries as its lines are `del`'d.

Also brings `emit_red_proof_line` onto the hoist-out rule — it renders a sum at a
caller-chosen level exactly as the two reifying `red` emitters do, and was the one such
funnel not on the list.

## Validation

- `magic_square --size=4 --all-different gac` **verifies** at gate 16 and gate 0
- **505/505** ctest, flag off / `literals` / `literals` + `MIN_CHAIN=0`
- **21/21 byte-identical to main with the flag off** — carrying this still costs nothing
- **Must-fail control**: with the union rule disabled, `eq_window_test` fails on the named
  invariant (`p: a Top pol over operands naming the atom did not take it out of the
  window`), not merely on the downstream VeriPB rejection
- A verdict-only pass over the whole PR #632 row set at gate 16 and gate 0 is still running;
  I will post the result

The new `eq_window_test` scenario `p` asserts the **pin**, not the proof: with the reference
missed the atom stays windowed and is evicted, and VeriPB only objects two steps later.

Per the #612 stack policy this is **not for merge without an explicit go-ahead**.

### DIFF
--- a/dev_docs/brancher-design.md
+++ b/dev_docs/brancher-design.md
@@ -1499,7 +1499,12 @@ recursions / propagations / solutions unchanged mode-off vs mode-on.
   `magic_square --size=4 --all-different gac` rejected in the **shipped** configuration
   because of it. Fixed automatically rather than by declaration — cutting planes cannot
   invent an atom, so the union over the operands over-approximates the result. Costs a
-  `ProofLine -> named eq atoms` map, scoped to what is live. Full write-up in
+  `ProofLine -> named atoms` map, scoped to what is live. Covers **ge** thresholds too, under
+  the new `LineHoist` residency cause -- but `line_hoist` is **0 on all 24** PR #632 rows, so
+  the ge half rests on the argument and a discriminating unit test, not on an instance.
+  Pinning ge from the ordinary reference walk was tried and **reverted**: a ge threshold is a
+  chain link, so hoisting it restructures the chain and breaks RUP steps that depend on its
+  positive-level shape. Full write-up in
   [order-encoding-deletion.md](order-encoding-deletion.md), "The `pol` gap".
 
 Stages A, B, B', B'', C, D, E, F and G have all landed in sequence. Stage D was the only one
--- a/dev_docs/brancher-design.md
+++ b/dev_docs/brancher-design.md
@@ -1510,6 +1510,26 @@ recursions / propagations / solutions unchanged mode-off vs mode-on.
 Stages A, B, B', B'', C, D, E, F and G have all landed in sequence. Stage D was the only one
 gated on work outside this task (the delc mechanics), and that gate is closed.
 
+### Status: PAUSED (2026-08-03)
+
+Work on this stack is **paused**, with every stage pushed to a PR and nothing uncommitted.
+The reason is that the full PR #632 benchmark sweep did not make the case:
+
+- **Two correctness bugs found by benchmarking**, both fixed (stage F, stage G) — the stage G
+  one, `magic_square --size=4 --all-different gac`, was a rejected proof in the **shipped**
+  configuration, found on a control row expected to be an exact no-op.
+- **On real instances the feature is a no-op or worse.** 10 of 26 rows never engage at all;
+  only `qap10` (1.20×/1.28×) and the two `rcpsp` forms (~1.17×/1.20×) win; groups B and C are
+  1.00× throughout. The 3.7×–15× wins are confined to the `odb_*` synthetics built to exhibit
+  them.
+- **One severe regression**, `table_layout15` at 0.09×/0.08× — diagnosed as a VeriPB
+  unhinted-propagation issue (see order-encoding-deletion.md) rather than a cost of deletion,
+  and raised upstream.
+
+The flag-off byte-identity claim holds exactly (21/21), so carrying the stack costs nothing
+while it sits. VeriPB's forthcoming **label-groups** mechanism may offer a different route to
+the same goal, which is the other reason to stop here rather than push on.
+
 ## Implementation gates (not owner calls)
 
 Three unknowns; all are validated empirically against VeriPB, not decided by the owner.
--- a/dev_docs/brancher-design.md
+++ b/dev_docs/brancher-design.md
@@ -1486,7 +1486,23 @@ recursions / propagations / solutions unchanged mode-off vs mode-on.
   unchanged: **this is a feature for weak-propagation, large-domain search, and real
   MiniZinc-shaped models are not that.**
 
-Stages A, B, B', B'', C, D and E have all landed in sequence. Stage D was the only one
+- **Stage F — do not evict an ancestor's order literal. DONE (#645).** A definition minted
+  at a level the search has since left can still be named by a live row from that level, so
+  `evict_order_literal` refuses when `level < logger->proof_level()`. Found by rebasing onto
+  main, which made `table_layout` reachable: it rejected at gate 0 and only at gate 0.
+  Conservative by design — the information to do better is not kept — and it costs nothing
+  measurable.
+- **Stage G — pin what a `pol` names, via the union over its operands. DONE.** The
+  eq-side member of stage F's family, reached by a path stage F's rule cannot see: a `pol`
+  at Top over operands naming a windowed eq atom is a permanent reference that no walk over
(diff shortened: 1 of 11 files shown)


ciaranm:
## ge thresholds now covered too — with two results that contradict the obvious expectation

Pushed `0036abe8`. The recorded atom set now carries a kind, the reference walk records both
eq and ge, and a Top `pol` pins the ge thresholds its operands name under a new
`OrderEncodingResidencyCause::LineHoist`.

**1. It fires nowhere measured.** `line_hoist` is **0 on all 24 PR #632 rows** at gate 0 — no
Top pol names a live deletable threshold on any of them. So the ge half rests on the argument
and a discriminating unit test (`order_evict_test` scenario `v`, which fails when the union is
disabled), *not* on an instance. It is defence in depth for a soundness property, not a
measured win, and shouldn't be read as one. `line_hoist: 0` is the expected reading in the
stats dump; a nonzero one means an instance has finally reached it.

**2. Pinning ge from the ordinary reference walk does not work.** The obvious symmetry —
extend the hoist-out rule so any Top line naming a threshold pins it, as it does for eq —
**breaks proofs**:

```
comparison_test le_iff --seed=4979515          -> not implied by reverse unit propagation
linear_constraint_le_iff_incremental (gate 0)  -> same
```

An eq atom stands alone, so hoisting one is local. A ge threshold is **a chain link**, and
hoisting it to Top stitches over its neighbours — pulling the chain out from under RUP steps
that depend on its positive-level shape. The walk therefore *records* ge atoms (which is what
the pol union reads) but does not pin them; only a pol pins, at pol-emit time, which is
outside any other line's emission.

That leaves the ge coverage exactly as wide as the pol gap and no wider — the same conclusion
stage F reached when the Top-pin idea was first tried against `table_layout`.

## Validation (whole change, eq + ge)

- **505/505** ctest, flag off / `literals` / `literals` + `MIN_CHAIN=0`
- **21/21 byte-identical to main with the flag off**
- `magic_square --size=4 --all-different gac` **verifies** at gate 16 and gate 0
- Must-fail controls for **both** halves: disabling the eq union fails `eq_window_test` on
  its named invariant; disabling the ge union fails `order_evict_test`

A verdict-only pass over the whole row set at gate 16 and gate 0 is running against this
build; I'll post the result.

ciaranm:
## Full benchmark sweep, stage G: 26 rows × 5 configs, groups A–D

Re-ran the whole PR #632 set against `0036abe8`, plus group D which had never been run.
`main` = `81232b90`, veripb 3.0.2, one core with ASLR off, machine idle, **REPS=1**
(4 h 56 m for A–C, 1 h 12 m for D).

**`all rows: identical search in every config, and every proof VERIFIED`** — clean for the
first time.

### The row this PR exists for

`magicsq4_gac` was `VERIFY_FAIL(rc=1)` at both gates before; now:

| config | deletes | verdict | verify | `.pbp` |
|---|---|---|---|---|
| `main` | — | VERIFIED | 29.42 s | 169 064 735 |
| `gate16` | 4 | **VERIFIED** | 29.62 s | 169 068 153 |
| `gate0` | 49 | **VERIFIED** | 29.02 s | 169 093 606 |

The fix costs **+3 418 bytes** (0.002 %) and 0.7 % of verify — inside single-sample noise.
Deletions drop 10 → 4 at gate 16, which is the mechanism working rather than a regression:
the pinned atoms are no longer evicted.

### Group D, run for the first time

| row | main | gate16 | gate0 | gate16nowin | gate0nowin |
|---|---|---|---|---|---|
| `rcpsp20_opt` | 663.5 | 565.6 **1.17×** | 560.7 **1.18×** | 662.4 1.00× | 657.8 1.01× |
| `mzn_aircraft06` | 209.3 | 211.5 0.99× | 210.9 0.99× | 215.1 0.97× | 212.6 0.98× |

`rcpsp20_opt` reproduces the doc exactly (656.8 s, 4 106 MB) and gives the strongest
attribution result in the set: **the eq window is the entire effect, chain deletion
contributes nothing** — identical to `rcpsp_dl21`, i.e. the infeasibility and optimisation
forms of the same instance agree.

`mzn_aircraft06` is flat, unsurprisingly: it solves in **1 recursion**, so there is no search
for deletion to act on. It reproduced its documented profile (2.52 GB `.opb`, 8.4 MB `.pbp`,
19.1 GB veripb resident).

### The overall picture is unchanged, and it is not flattering

- **Two wins**: `qap10` 1.20×/1.28×, the `rcpsp` pair ~1.17×/1.20×
- **One severe regression**: `table_layout15` at **0.09×/0.08×** — 11.4× slower — now
  confirmed a third time, and untouched by this PR (it is chain deletion, not the
  reference-pinning changed here)
- **10 of 26 rows never engage at all** (`deletes = 0` in every config)
- Groups B and C are 1.00× throughout; the large wins stay confined to the `odb_*` synthetics

`line_hoist` remains **0 on all 26 rows**, so the ge half of this PR is still carried on the
argument and its unit test rather than on any instance.

### Methodology note

REPS=1 rather than min-of-3. That holds up because `run_all.sh` is row-major, so a row's
five numbers are adjacent in time and the VM's ~18 % drift cancels in the division — every
ratio reproduced the REPS=3 run to within 2–4 % (`qap10` 1.22×→1.20×, `rcpsp_dl21`
1.14×→1.16×, `table_layout15` 0.08×→0.09×, `seat_moving` 0.98×→0.99×), while absolutes
drifted 5–10 %. What it costs: `freqsq6_gac` reads 1.07×/1.09× against 1.01× at REPS=3, so
anything under ~1.08× in this run should be treated as unresolved.

Full write-up, including the two kit defects fixed along the way and the VeriPB upstream
reproducer for `table_layout15`, is in `RESULTS.md` in the sweep kit.

Issue #549, opened 2026-07-21
Presolver: general lifted Cumulatives and certified RCPSP lower bounds (Sidorov stage 2)


ciaranm:
Part of the Cumulative proof-logging plan — tracking issue #541 has the blanket rules (no new OPB content; the shared validation vocabulary) and the shared reference list. Plan numbering used below: issue 01 = #542, 02 = #543, 03 = #544, 04 = #545, 05 = #546, 06 = #547, 07 = #548, 08 = #549, 09 = #550.

## Summary

Extend the issue 07 presolver from Disjunctive cliques to Sidorov's full programme:
lift cover inequalities to general valid inequalities π^T x ≤ π₀ over the occupancy
flags and post them as derived Cumulatives with non-unit coefficients — plus the
headline deliverable: **VeriPB-certified makespan lower bounds** for public RCPSP
benchmarks, several of which are current best-knowns that no CP solver has reproduced
(Sidorov Table 4). The research meat is certifying the lifting coefficients; the
prize is a bounds table where every entry has a machine-checkable proof.

## References

- Sidorov CP 2026 §3–4 (lifting workflow, Alg. 2), Appendix A (novel bounds), C
  (budgets). Open access; artifacts doi:10.5281/zenodo.20044686 (software) and
  doi:10.5281/zenodo.18663551 (logs, including which instances close and how).
- Padberg 1975 and Zemel 1989 for sequential lifting / DP computation of lifting
  coefficients (standard IP literature).
- Ask Ciaran for any paper you cannot obtain.

## Dependencies

Issues 03 (subset-sum certificates), 04 (derived mode), 05, 07 (cover machinery,
clique certificates, cross-check corpus). Issue 01's window-energy lemma is consumed
by the bounds derivation. Constants-only, presence-free, as in issue 07.

## Proof discipline

As throughout the plan: derived constraints and derived bounds live in the proof,
never the OPB (byte-diff enforced). Additionally: a lifted inequality's validity is
coNP-hard to check in general, so we must carry a *certificate schema* out of the
lifting phase — if the presolver cannot emit a certificate for a candidate
constraint, the constraint is **dropped and logged**, never posted uncertified and
never smuggled into the model.

### Lifting with certificates

Run Sidorov Alg. 2 (sequential lifting over a cover, coefficients from subproblem
optima v*), but accept a lifted constraint only when it fits a certificate schema we
can emit per time point:

[...]

Issue #548, opened 2026-07-21
Presolver: inferred Disjunctive constraints across Cumulatives (Sidorov stage 1)


ciaranm:
Part of the Cumulative proof-logging plan — tracking issue #541 has the blanket rules (no new OPB content; the shared validation vocabulary) and the shared reference list. Plan numbering used below: issue 01 = #542, 02 = #543, 03 = #544, 04 = #545, 05 = #546, 06 = #547, 07 = #548, 08 = #549, 09 = #550.

## Summary

First stage of Sidorov's "On Inferring Cumulative Constraints" (CP 2026) as a GCS
presolver, restricted to what his own data says captures most of the value: inferring
**Disjunctive** constraints (capacity 1) from cliques in the cross-resource conflict
graph. His Table 4 has 14 of 15 search-less record bounds at capacity 1, and the
biggest search gains came from Disjunctive discoveries. Crucially, this restriction
keeps every certificate short and fully polynomial: per-pair at-most-ones from a
donor capacity line, merged into a clique inequality, replayed per time point. The
general lifted case (non-unit coefficients, ILP lifting subproblems) is issue 08.

## Background

Sidorov's key observation, mapped to our encoding: his occupancy-vector polyhedron at
time t *is* our vector of `active_{i,t}` flags, and his resource inequalities are our
per-time `C_t` lines. Any cutting-planes derivation of a valid inequality from the
resource lines is time-independent, so one derivation recipe replays verbatim at
every time point against that time point's lines. An inferred constraint's proof cost
is (per-time certificate size) × horizon — which is why stage 1 sticks to cliques.

Two tasks u, v conflict if some resource r has a_{ru} + a_{rv} > b_r. A set of
pairwise-conflicting tasks — with different pairs possibly conflicting on *different
resources* — admits Σ active ≤ 1 at every time point. That cross-resource case is
exactly the inference no single posted Cumulative can make, and it is what the
validation plan targets.

## References

- Sidorov CP 2026 (open access; arXiv:2602.15635; artifacts
  doi:10.5281/zenodo.20044686 and doi:10.5281/zenodo.18663551). See index issue for
  URLs; ask Ciaran for anything you cannot obtain.
- Sidorov, Marijnissen, Demirović, "Unite and Lead", CP 2025 (open access) — the
  disjunctive-clique predecessor, useful for intuition only.

## Dependencies

Issues 04 (derived mode, multi-donor flag bridging) and 05 (enumeration). Presence-
[...]

Issue #541, opened 2026-07-21
Cumulative proof-logging plan: energy rules, optional tasks, and inferred constraints (tracking)


ciaranm:
Nine issues covering the plan developed from two CP 2026 papers (Cloutier &
Quimper's knapsack-augmented overload check; Sidorov's inferred Cumulative
constraints), plus optional tasks.

| # | Title | Depends on | Priority |
|---|-------|-----------|----------|
| 01 | Certified window-energy reasoning and an overload check for Cumulative | — | high |
| 02 | Optional tasks for Cumulative via per-task presence Booleans | — | high |
| 03 | Certified subset-sum strengthening utility | — | high |
| 04 | Derived-constraint mode for Cumulative | — | high |
| 05 | Typed constraint enumeration for presolvers | — | high (small) |
| 06 | Presolver: single-Cumulative strengthening (Schulz rules) | 03, 04, 05 | medium |
| 07 | Presolver: inferred Disjunctive constraints across Cumulatives | 04, 05 | medium |
| 08 | Presolver: general lifted Cumulatives and certified RCPSP lower bounds | 03, 04, 05, 07; uses 01 | medium |
| 09 | TTHE-OC and knapsack-augmented (KAOC) overload checking | 01, 03; interacts with 02 | low (parked) |

Issues 01–05 are independent of each other and can proceed in parallel, with two
coordination points: 01/09 must guard energy reasoning by task presence if 02 lands
first, and 04's flag-sharing decision shapes 06–08.

## Two blanket rules (restated in every issue)

**No new OPB content.** The OPB model is the statement being verified. With the single
exception of issue 02 (which adds presence variables to the *posted constraint's own*
encoding), nothing in this plan may add or alter OPB output: every inferred fact enters
the proof as a derivation (`pol` / RUP / redundance in the `.pbp`), never as a model
axiom. VeriPB will happily verify a correct proof of the wrong model, so "define it in
the OPB" is a soundness hole, not a shortcut. Each issue carries an OPB byte-diff test
enforcing this.

**Validation must show the rule fired, and fired on something hard.** A proof that
verifies is necessary, not sufficient: the rule may never have triggered, or the
arithmetic may be slack in a way that small/close-together integers let pass by
coincidence. Every issue therefore uses the same validation vocabulary:

- *Marker comments*: each new inference rule emits a distinctive
  `emit_proof_comment(...)` in its justification; tests count occurrences in the
[...]

PR #663, opened 2026-08-04
Inferred Disjunctive constraints from cross-resource conflict cliques (#548)

ciaranm:
The cutting-planes core of #548 (Sidorov stage 1), as a standalone utility with
no Cumulative anywhere near it. Stacked on #661; both retarget to `main` as the
#541 stack lands.

Given `~a_p + ~a_q >= 1` for every pair of a clique, derive `sum a_p <= 1`.

## The derivation, and why not the obvious one

Summing all `k(k-1)/2` at-most-ones gives `(k-1) * sum ~a_p >= k(k-1)/2`;
dividing by `k-1` lands on `ceil(k/2)`. That is `k-1` for two and three members
and short of it from four on. So the merge is the induction: with the inequality
for the first `m` members in hand,

```
    m*~a_m + sum_{i<m} ~a_i             >= m          the m pairs
            (m-1) * sum_{i<m} ~a_i      >= (m-1)^2    m-1 copies of the current
    ---------------------------------------------------------------------------
    m * ( ~a_m + sum_{i<m} ~a_i )       >= m + (m-1)^2
```

and `m + (m-1)^2 = m(m-1) + 1`, so dividing by `m` rounds the degree up to
exactly `m`. **That spare `+1` is the entire margin** — one less and the division
rounds down to `m-1` and the induction does not advance.

The one-shot version is kept as a runnable mutation rather than described in a
comment, because it is *accepted* for k=2,3 and *rejected* from k=4 — so the
argument for the induction is pinned to a boundary a test watches, and if that
boundary ever moves the comment explaining it is wrong.

## The `ia` pin is what makes any mutation a test

Every step of the merge is sound whatever was fed into it. Drop an input and the
`pol` lands on a **weaker but still true** line, which VeriPB is right to accept
— so without pinning the result there is no corruption of this derivation that
can be detected. Same lesson as #660's shortened chain and #661's bogus divisor.

It turns out to be stricter than "weaker fails", which is worth knowing
deliberately: leaving the division off the last merge gives
`m*(sum ~a) >= m(m-1)+1`, which is mathematically **stronger** than the clique
inequality — and it is rejected too, because a syntactic implication check cannot
see an equivalence that needs dividing to spot. The final division is therefore
load-bearing twice, and any future rearrangement of this arithmetic breaks the
pin even while staying sound. That has a mutation of its own so it cannot rot
quietly.

The pin also *normalises*: `ia` enters the target as a new line and that is what
comes back, so a caller always receives the literal-exact clique inequality.

## Where the at-most-ones come from, checked here because only running it settles it

`#548` specifies a `c_j <= C` side condition on the pair derivation. **It is not
needed.** Weaken the other terms out of a resource row, saturate, divide by the
margin `D = c_u + c_v - C`: saturation caps both coefficients at `D` and the
division rounds them back up to one, whatever they were. Verified with
`c_v = 12` against `C = 10`. The condition belongs to clique *discovery*, where a
task that can never run would otherwise pad every clique it touches.

A pair summing to **exactly** the capacity is correctly refused — that is where
an off-by-one in the conflict test lands, and it is #548's camouflage-task case.

## A shortcut the caller should know about before it is built

Where one row covers three or more clique members, the pairwise at-most-ones are
not the cheapest premises: weakening the row down to that set and dividing by
`d = min(c_max, Delta)`, with `Delta = sum_K c_i - C`, reaches the **whole
sub-clique inequality in one step** whenever `Delta > d(|K| - 2)`. Verified:

| demands | C | result |
|---|---|---|
| `{6,6,6,6}` | 10 | `sum a <= 1` in one pol (Delta 14 > 12) |
| `{6,6,9}` | 10 | `sum a <= 1` in one pol (Delta 11 > 9) |
| `{9,6,6,6}` | 10 | fails at `<= 1` (Delta 17 not > 18), succeeds at `<= 2` |
| `{4,4,4}` | 10 | `sum a <= 2` — and **no pair conflicts at all** |

Since most conflicting pairs in a real instance share a resource, that replaces
up to `(k_r choose 2)` pair derivations and their weaken chains with one step.
The last row is a cardinality cut the pairwise machinery is structurally blind
to. Neither belongs in *this* routine, which works from at-most-ones by
construction — they are recorded as checked facts for #548 and #549 to build on.

## For the caller, in the header

- Every intermediate is emitted at the caller's level. Only the returned line
  needs to outlive the call, so replaying this across a horizon at `Top` leaves
  `k^2/2` permanently-live constraints per time point, taxing every later
  unhinted RUP (the `table_layout15` pathology). Deriving intermediates at
  `Temporary` and forgetting them, or inlining into a single `pol`, is the fix,
  and it is #548's to make.
- Two things the pin cannot check, both the caller's to get right: that
  `members` are pairwise distinct (a repeat becomes a coefficient-two term and
  means something else), and that each line really is the at-most-one for its
  pair.
- `PolBuilder::weaken` renders the flag verbatim, so a negative-polarity
  `ProofFlag` would emit `~name w`, which VeriPB's weakening rule rejects. Not
  reachable from here; a trap for #548's row-weakening.

## Validation

`clique_from_amos_test`, on a **satisfiable** micro model (per #656 — against an
unsatisfiable one every RUP is vacuously valid):

- the induction verifies for k = 2..12;
- the result contradicts an axiom forcing two members active, `ia`-confirmed as
  a genuine contradiction;
- `DropAnAtMostOne` (from the *first* merge, so the weakness propagates through
  every later pass) — rejected;
- `ClaimOneMore` — rejected;
- `SkipFinalDivision` — rejected, though strictly stronger;
- `NaiveOneShot` — accepted at k=2,3 and rejected at k=4,5,9;
- the pair recipe over/under/at capacity, and the row shortcut table above.

Full ctest green (625/625, `GCS_TEST_CAP_DEFAULTS=OFF`); GCC and clang
warning-free; clean under ASan/UBSan.

`add_term_to` moves to `pseudo_boolean.hh`, having acquired a second user.

### DIFF
--- a/dev_docs/README.md
+++ b/dev_docs/README.md
@@ -150,6 +150,14 @@ library. For an introduction to *using* the solver, start with the top-level
   sound derivation of the wrong line), the deviations from the paper's statement
   of the rules, and why the deep-gap fixture everyone quotes cannot be a
   `Cumulative` instance.
+- [Inferring `Disjunctive` constraints across resources](inferred-disjunctive.md) —
+  the `InferredDisjunctive` presolver: conflict cliques spanning several posted
+  Cumulatives, posted in derived mode. Covers why the cross-resource case is an
+  inference no single Cumulative can make, the three-piece certificate (pairwise
+  at-most-one out of a witnessing row, flag bridge, clique merge), why two-task
+  cliques are never worth posting, the camouflage fixture the mutations need, and
+  the measured proof size — which is the thing to fix before pointing this at a
+  real instance.
 - [Restarts, nogoods, and dom/wdeg weighting](restarts-nogoods-weighting.md) —
   the search-side machinery from issue #315: the restart loop and its
   `SearchResult` unwind signal, `RestartSchedule`, the `ConflictObserver`
--- a/dev_docs/cumulative-proof-logging.md
+++ b/dev_docs/cumulative-proof-logging.md
@@ -558,24 +558,52 @@ The real presolver built on that recipe is `CumulativeStrengthening` — see
 [cumulative-strengthening.md](cumulative-strengthening.md), which turns the
 invisibility above into an automated soundness tripwire rather than a caveat.
 
-### Multi-donor, for later
+### Multi-donor
 
 Issues #548 and #549 infer a Cumulative over tasks drawn from *several* donors,
-each with its own flag copies for the same `(task, time)` semantics. Two ways to
-make one derivation speak about all of them:
+each with its own flag copies for the same `(task, time)` semantics. Two things
+make that work, and both are now here.
 
-1. **Bridge lemmas.** Pick one donor's flags as canonical and derive
-   `active^{(r)}_{i,t} ↔ active^{(1)}_{i,t}` per `(i, t)`, by `pol` over the two
-   reification halves — the start variable's bits cancel, exactly as in the
-   window-energy bridges above. O(tasks × times) extra Top lines per extra
-   donor, and it needs no new API: both donors' rows are citable by name.
-2. **Rewrite each donor's row.** One `pol` pass per row, over the flag-defining
-   rows directly, landing on the canonical flags without ever stating the
-   bridge.
+**The spec is per task.** `DerivedCumulativeTask` names the donor and the
+position within it for each task separately, so a constraint whose members come
+from different donors needs nothing special — `derived_cumulative_tasks_from`
+builds the all-of-one-donor case, which is still the common one. `row_donors`
+is separate from the tasks' donors, because a pairwise conflict is witnessed by
+whichever resource cannot hold both tasks, and that need not be where either
+task's flags are taken from. The recipe is handed the rows those donors wrote
+for that time point, by donor, and may return nullopt to decline a time point it
+cannot derive — which declines the whole constraint, since the propagator cites
+a row at every time it covers.
 
-Neither is implemented here. The first is the safer starting point (each lemma
-is checkable on its own); the second is smaller if it works. Whichever lands
-should measure the proof size, since that is the whole difference between them.
+**The bridge is one `pol`.** `derive_flag_bridge`
+([`flag_bridge.hh`](../gcs/innards/proofs/flag_bridge.hh)) turns one donor's
+flag into another's. A fully reified flag emits `g → ineq` under `[r]` and
+`ineq → g` under `[f]`; adding one flag's `[r]` to another's `[f]` puts the two
+inequalities in with opposite signs, so their terms cancel — every bit of every
+variable they mention — and saturation leaves the two-literal clause. The
+sketch this replaces assumed order literals and bit reasoning would be needed;
+they are not.
+
+`active ⇔ before ∧ after` needs one more step, since two `active` flags are
+reified over *different* flags and so do not cancel against each other.
+`derive_conjunction_flag_bridge` bridges the conjuncts first, after which each
+appears with both signs and drops out. Three `pol`s per `(task, time, donor
+pair)`.
+
+What the derivation needs is not that the conditions match but that the
+constants leave something behind: for `e ≤ p` and `e ≤ q` the sum has degree
+`q − p + 1`, so it goes through exactly when the implication is true. Identical
+conditions are the case this is for; a weaker target also works, a stronger one
+correctly does not.
+
+Two traps, both guarded. The halves are labelled from the flag's **full PB
+rendering**, not `name_of`, whose plain-flag form is the bare stem — the
+window-energy bridges get away with `name_of` only because `Cumulative`'s flags
+are values-named. And a negated flag has no halves of its own, so bridging one
+is refused rather than silently naming the positive flag's rows.
+
+Proof size is the thing to watch: the bridges are O(tasks × times) Top lines per
+extra donor, and at `Top` none of them ever dies.
 ## Optional tasks (issue #543)
 
 The optional-task constructor gives each task a `{0, 1}` presence
--- a/dev_docs/inferred-disjunctive.md
+++ b/dev_docs/inferred-disjunctive.md
@@ -1,0 +1,124 @@
+# Inferring `Disjunctive` constraints across resources
+
+Two tasks conflict when *some* resource cannot hold both at once. A set of tasks
+conflicting pairwise can have at most one of them running at any time — and when
+different pairs of that set conflict on *different* resources, that is an
+inference no single posted `Cumulative` can make.
+
+`InferredDisjunctive`
+([`gcs/presolvers/inferred_disjunctive.hh`](../gcs/presolvers/inferred_disjunctive.hh))
+finds those sets and posts each as a capacity-one derived `Cumulative`. It is the
+first stage of Sidorov (CP 2026), restricted to capacity one and unit
+coefficients — which is what keeps every certificate polynomial, and where his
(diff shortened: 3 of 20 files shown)


ciaranm:
Added the second primitive: **the cross-donor flag bridge**, which #657 sketched
two approaches for and implemented neither.

It costs **one `pol`**, which the sketch did not anticipate. A fully reified flag
emits `g -> ineq` under `[r]` and `ineq -> g` under `[f]`; adding one flag's
`[r]` to another's `[f]` puts the two inequalities in with opposite signs, so
their terms cancel — every bit of every variable they mention — and saturation
leaves the two-literal clause. The sketch assumed order literals and bit
reasoning would be needed; they are not.

`Cumulative`'s `active <-> before /\ after` needs one extra step, because two
`active` flags are reified over *different* flags and so do not cancel against
each other. Bridging the conjuncts first fixes that: each then appears with both
signs and drops out. Three `pol`s per (task, time, donor pair) in total.

## What the precondition actually is

Not "the conditions match" — that was my first guess and the tests disproved it.
For `from <-> (e <= p)` and `to <-> (e <= q)` over the same expression, the sum
has degree `q - p + 1`, so the derivation goes through exactly when `q >= p`,
i.e. exactly when the implication is true. Identical conditions are the case
this exists for and give degree one; a **weaker** target bridges too (`start <= 4`
really does imply `start <= 5`), and a **stronger** one correctly does not. The
tests pin all three, because the stronger-target direction is the one a caller
can get wrong and nothing else would say so.

## Two traps, both now guarded

- The halves must be labelled from the flag's **full PB rendering**, not
  `name_of` — whose plain-flag form is the bare stem, so the first version
  emitted `@from[r]` against a row actually called `@f[0][from][r]`.
  `window_energy` gets away with `name_of` only because Cumulative's flags are
  values-named. `ProofModel` says as much where it writes the labels; worth
  reading before using either.
- A **negated flag** has no reification halves of its own, so bridging one is
  refused rather than silently naming the positive flag's rows.

Full ctest green (626/626, `GCS_TEST_CAP_DEFAULTS=OFF`).

ciaranm:
#548's actual inference now works end to end, on top of the two primitives
above.

## Two more commits

**`install_derived_cumulative` generalised to per-task donors.** Keying every
flag lookup off one `spec.donor` plus the task's position in it cannot express a
clique whose members conflict pairwise on *different* resources — there is no
single donor covering all of it, and a canonical one containing every member
with nonzero demand need not exist. `DerivedCumulativeTask` now names the donor
and position per task; `derived_cumulative_tasks_from` builds the
all-of-one-donor case, which both existing callers use unchanged in substance.

`row_donors` is deliberately separate from the tasks' donors: a pairwise
conflict is witnessed by whichever resource cannot hold both tasks, which need
not be where either task's flags live. The recipe now receives the rows those
donors wrote for that time point, by donor, and may return nullopt to decline a
time point — which declines the whole constraint, since the propagator cites a
row at every time it covers.

**`InferredDisjunctive`.** Conflict graph over every posted Cumulative, tasks
identified by start variable so the same task on two resources is one node;
greedy maximal cliques taking longest-duration first; ranked by total duration;
subsumed cliques and anything under three members dropped. A two-task clique is
never worth posting — the resource witnessing the pair already keeps them apart.

Per time point the certificate is: pairwise at-most-one out of the witnessing
row (weaken, saturate, divide by the margin) → `derive_conjunction_flag_bridge`
where the witness is not where the task's flags live, cached per
(task, resource) rather than per pair → `derive_clique_from_amos`, whose pinned
output *is* the unit-height capacity-one row.

## The fixture family, and what it establishes

k tasks, k resources, pair `(i, j)` conflicting only on resource `(i + j) mod k`.
Every resource sees exactly one conflicting pair, so the clique exists only in
the union — a root refutation here cannot be one of the donors doing the work.

Three tasks of length two into five time points: six units of work, five slots.

- **with the presolver**: refuted at the root, proof verified, **15 flag bridges**
  derived (so the certificate genuinely spans resources);
- **without**: searches;
- **sharp twin** (one more unit of horizon): satisfiable, all six solutions,
  matching brute force — the clique is still posted and still harmless.

Also asserted: solution preservation at three shapes; **time-table neutrality**
as node-for-node equality (the inferred constraint ships with time-tabling off
for exactly that reason, so the test turns it back on); both budgets biting *and
being counted*, on a two-disjoint-edges fixture where the drops are real; an
optional-task donor declined; the OPB byte-identical; and the per-time
derivation markers present.

627/627 ctest with `GCS_TEST_CAP_DEFAULTS=OFF`, seeds 1–4; GCC and clang
warning-free; clean under ASan/UBSan.

## Still not done from #548

- **Sidorov's Pack/Pack_d cross-check** (his §5.1 artifacts) — needs the instance
  lists from the zenodo logs.
- **Mutation tests for the presolver's own certificate.** The primitives have
  theirs; the assembled per-time derivation does not yet have a knob that
  corrupts it (bridging the wrong task, claiming a clique containing a
  non-conflicting pair).
- **Budget-robustness sweeps** on larger instances.
- **Proof size.** Every intermediate lands at `Top` and never dies, and the
  bridges are O(members × times × donors). This is the concern flagged on the
  clique merge, now with a real consumer to measure.

ciaranm:
Worked through the rest of the list.

## Mutations of the assembled certificate

The primitives had theirs; the assembly did not. Three knobs, all corrupting the
**conclusion** rather than the route — which is what a conflict-shaped derivation
needs, since the route is where it forgives everything:

| mutation | what it corrupts | verdict |
|---|---|---|
| `ClaimRhsZero` | claims no member may run at all | rejected |
| `BridgeWrongTask` | bridges a task onto the *other* task's flags, so the at-most-one is about a task nothing cornered | rejected |
| `IncludeNonConflicting` | grows the clique with a compatible task, inventing the conflict record | rejected |

The last one needed a fixture the family did not have: a **camouflage task**, on
a resource of capacity two where every pairwise demand sums to exactly two.
Compatible, but only by one unit — which is precisely where an off-by-one in the
conflict test lands, and what #548 asks for. Honestly, it stays out of the clique
and the proof verifies; forced in, VeriPB rejects.

## Proof size, measured rather than argued

Three-task family over five time points: **251 lines, 11 KB, 139 `pol` steps** —
about 28 `pol` per time point for a three-clique.

Ten of those went by carrying the at-most-one onto this constraint's own flags in
the *same* `pol` rather than starting another: the bridges just go on the stack,
each cancelling its task's term, and one saturation clears up after all of it.

The shape is `O(k²)` pair derivations plus `O(k)` bridges plus `O(k)` merge steps
**per time point**, multiplied by the horizon. At eight members and a horizon of
a thousand that is order 10⁵ lines per clique — all at `Top`, where none of it
ever dies and all of it taxes every later unhinted RUP. Only the pinned per-time
row needs to outlive the recipe, so `Temporary` + forget, or inlining a whole
time point into one `pol`, is the fix. It is written down in
`dev_docs/inferred-disjunctive.md` as the thing to do before this meets a real
instance; I have not done it, because it wants measuring against a real instance
rather than this fixture.

## Not attempted

**Sidorov's Pack/Pack_d cross-check.** It needs the instance lists out of his
zenodo solver logs (doi:10.5281/zenodo.18663551), which is a data-acquisition
job rather than a code one — worth doing deliberately, with the artifact in hand,
rather than guessing at instance names.

627/627 ctest with `GCS_TEST_CAP_DEFAULTS=OFF`.

Issue #666, opened 2026-08-04
InferredDisjunctive: per-time-point scaffolding is emitted at Top and never deleted


ciaranm:
`InferredDisjunctive` (#548, PR #663) proves each derived capacity-one row from
scratch at every time point, and emits every intermediate at `ProofLevel::Top`,
where nothing is ever deleted. Only one line per time point is load-bearing; the
rest is scaffolding that stays in the constraint database for the remainder of
the proof.

## Measured

A k-clique on one resource, tasks of duration two, via
`rcpsp --dzn ... --infer-disjunctive --prove`, counting only the presolve prefix
(up to the last `presolve disjunctive: inferred a clique` marker) so the search
proof is excluded:

| k | horizon | presolve `pol` | per time point | pairs `k(k-1)/2` |
|---|---|---|---|---|
| 3 | 6 | 24 | 4 | 3 |
| 4 | 8 | 64 | 8 | 6 |
| 5 | 10 | 130 | 13 | 10 |
| 6 | 12 | 228 | 19 | 15 |

So the derivation is **≈1.3 × k(k-1)/2 `pol` lines per time point**, i.e.
`O(k^2)` per time point and `O(k^2 * horizon)` per clique — matching the shape
(a pairwise at-most-one per pair, a bridge per task that needs one, and the
merge).

Growth in the horizon is exactly linear, holding the clique fixed at three
tasks and padding with a precedence chain:

| horizon | pbp lines | `pol` | size |
|---|---|---|---|
| 20 | 1142 | 407 | 77 KB |
| 40 | 2302 | 827 | 165 KB |

## The actual cost is database bloat, not size

For k=6 over horizon 12, the presolve prefix emits **228 `pol` and 12 `ia`**, and
contains **zero deletions** — against 466 deletions in the rest of the proof,
so the machinery is working elsewhere and simply is not applied here.

Only the 12 pinned lines (one `ia` per time point) are ever cited again. The 228
intermediates are scaffolding, and every one of them stays live for the whole
proof, taxing every later unhinted RUP — the shape recorded in the
`table_layout15` propagation pathology, where a bloated live database cost 12x at
the root rather than at deletion time.

Extrapolating the measured rate to a realistic instance: k=8, horizon 1000 is
~28 pairs x 1000 x 1.3 ~ 36k `pol` per clique, and at the default `N_out = 5`
that is ~180k live constraints from one presolver on one model. Sidorov's own
budgets go to 20 cliques.

[...]

PR #668, opened 2026-08-04
Presolver: strengthen a Cumulative's heights, not just its capacity (#547, part two)

### LINKED ISSUE #547: Presolver: single-Cumulative strengthening (Schulz rules)
Part of the Cumulative proof-logging plan — tracking issue #541 has the blanket rules (no new OPB content; the shared validation vocabulary) and the shared reference list. Plan numbering used below: issue 01 = #542, 02 = #543, 03 = #544, 04 = #545, 05 = #546, 06 = #547, 07 = #548, 08 = #549, 09 = #550.

## Summary

A presolver applying three classical solution-preserving strengthenings to each
posted Cumulative, posting the strengthened version as a derived constraint (issue
04) with a per-time-point certificate. These rules are from Schulz's thesis and are
recapped compactly in Cloutier & Quimper CP 2026 §2.3; they are known to matter on
Pack-style instances (that paper's experiments) while being certificate-cheap — the
ideal first consumer for the derived-constraint infrastructure.

The rules (tasks i with consumptions cᵢ, capacity C):

- **R1, coefficient strengthening**: any cᵢ > C − min{cⱼ : j ≠ i, 0 < cⱼ < C} can be
  raised to C (task i can never run beside anything else).
- **R2, gcd rounding**: with d = gcd of the relevant consumptions, all cᵢ and C can
  be divided by d (equivalently: capacity reduced to d·⌊C/d⌋). Follow the paper's
  precise statement for which consumptions enter the gcd.
- **R3, knapsack capacity reduction**: capacity reduced to
  κ = max subset-sum of feasible consumption sets ≤ C, per the paper's per-time
  refinement (tasks not available at t are excluded first).
- **R4, precedence-based filling: out of scope** (needs precedence extraction from
  linear constraints; revisit once someone wants it).

## References

- Cloutier & Quimper CP 2026 §2.3 (open access; see index issue).
- Schulz, PhD thesis, TU Berlin 2013 (open access). Ask Ciaran for anything you
  cannot obtain.

## Dependencies

Issues 03 (subset-sum certificates, for R3), 04 (derived mode), 05 (enumeration).
Requires presence-free Cumulatives; bail out loudly on optional ones (issue 02).

## Proof discipline

The strengthened constraint is posted in derived mode: its capacity lines are
derived in the proof from the donor's lines. Nothing is written to the OPB — the
donor's encoding stays, unchanged, as the semantic anchor, and the byte-diff test
below enforces it. Do not "replace" the original constraint in the model.

## Design

- Presolver enumerates Cumulatives (issue 05), computes strengthened parameters
  (apply rules in the paper's order, once — document the order), posts derived
  Cumulatives via issue 04. Donors stay posted; measure propagation redundancy before
  deciding whether donor propagators want a disable knob (separate follow-up if so).
- Certificates per time point t:
  - **R2**: divide the donor's C_t by d, multiply back by d — two pol steps.
    (Equivalent divided-through formulation is acceptable if it simplifies the
    propagator; pick one and document.)
  - **R1**: from C_t, for each pair (i, j) with cᵢ + cⱼ > C, derive the at-most-one
    `activeᵢ + activeⱼ ≤ 1` by weakening all other tasks out of the line, saturating,
    and dividing (the side condition cⱼ ≤ C holds or task j is outright
    impossible — handle that degenerate case first with a unit derivation). Then
    combine i's AMOs, weighted by the cⱼ, with C_t to raise cᵢ to C. The exact pol
    program is design work; it is validated by veripb plus the mutation tests, not by
    eyeballing.
  - **R3**: pin tasks unavailable at t inactive (RUP under their bounds), then apply
    the issue 03 utility to the remaining line at `ProofLevel::Top`.
- Marker comments per rule (`presolve cumulative gcd`, `... amo`, `... kappa`).

## Expectation management (also a validation tool)

All three rules are provably *time-table-neutral*: profile values are subset-sums of
consumptions, so a profile exceeds C iff it exceeds d·⌊C/d⌋ (R2) and iff it exceeds κ
(R3), and R1 changes no time-table verdict either. Consequences:

1. With only time-table propagation enabled, search trees with and without this
   presolver must be **identical** — a strong automated tripwire: any node-count
   difference under TT-only search signals an unsound strengthening (or an unsound
   derived-mode wiring), before veripb even gets a say.
2. User-visible benefit arrives only with energy-family rules (issues 01/09). Do not
   expect speedups from this issue alone; say so in the PR.

## Validation plan

1. **Solution preservation**: random instance corpus, full enumeration with
   donor-only vs donor + strengthened: identical solution sets (caps off). This is
   the defining property of a presolve strengthening.
2. **TT-neutrality tripwire**: as above — identical node counts under TT-only search
   across the corpus, asserted automatically.
3. **Trigger evidence**: fixtures where each rule strengthens (assert the computed
   κ/d/raised coefficients against hand-computed values in test comments; assert the
   rule's marker in the `.pbp`); negative controls where no rule applies (already-
   coprime tight consumptions, min cⱼ small) — presolver posts nothing, and both the
   OPB *and* the proof are byte-identical to a no-presolver run.
4. **Sharpness fixtures** (hand-verified in comments):
   - R2: heights {6, 10, 4}, C = 13 → capacity 12 (rounding margin 1).
   - R3: heights {6, 10, 15}, C = 14 → κ = 10 (the deep-gap fixture from issue 03;
     overall gcd 1, so R2 cannot get it).
   - R1: heights {5, 4, 2}, C = 6 → c₁ raised to 6 (barely over: 5 = C − min + 1);
     control with c₁ = 4 = C − min exactly, which must **not** be raised.
5. **Mutation tests** (`run_test_and_expect_verify_failure.bash`):
   - claim κ − 1 (via the issue 03 knob);
   - R2 divide by a d not dividing every consumption;
   - R1 raise a coefficient not entitled to it (the c₁ = C − min control above, with
     the mutation forcing the derivation) — veripb must reject each.
6. **Value demonstration** (gate on issue 01 having landed): a Pack-style
   divisibility instance (all consumptions even, odd capacity — Cloutier & Quimper's
   Example 2 shape at larger scale): with R2 + overload check, root refutation of the
   optimality gap; without the presolver, search. Assert the differential via
   toggles.
7. **OPB byte-diff** presolver on/off (standard helper from issue 04); veripb on
   everything; distinct proof file names per registration.

## Out of scope

- R4 (precedences), multi-constraint reasoning (issues 07/08), optional tasks,
  donor-propagator disabling.

ciaranm:
Closes #547. Stacked on #667, which is where the at-most-one recovery this needs now lives.

Part one (#661) took the capacity down to `kappa`, the largest load the heights
can actually reach. This does the other half: the tasks that fill the resource
on their own are set aside from that sum, and their own heights come down to
`kappa` with it.

### One rule, not two

A task that cannot run beside any other task that consumes anything —
`h_i + h_j > C` for every overlapping `j` — occupies the resource whenever it
runs, whatever its height says. Such a task also **turns the capacity rule off**:
it reaches `C` by itself, so the largest reachable load *is* the capacity and
part one declines the donor outright.

Both of Schulz's height rules land on the same table. Coefficient raising (`c_i`
above `C − min_j c_j` goes to `C`) picks out exactly that set of tasks, and the
knapsack rule's "`c_i = C` can be reduced to `kappa`" is what happens to them
once the capacity moves.

| | capacity | full task | every other task |
|---|---|---|---|
| donor | `C` | `h_i` | `h_j` |
| derived | `kappa` | `kappa` | `h_j` |

Stated as the pairwise test rather than as the minimum: the two conditions are
the same, and the pairwise one is what the certificate needs anyway. Tasks whose
windows cannot overlap are left out, which is slightly more than the paper claims
and costs nothing. The set is deliberately not per time point — a `Cumulative`
has one height per task.

### The value

New fixture: five unit tasks of height three plus one of height eight, all able
to run in `[0, 3)`, capacity eight. `kappa` over every task is eight, so part one
declines. Setting the tall task aside takes `kappa` to six and its own height
with it, and the window then supplies 18 against the 21 those six tasks need —
root refutation. The donor needs 23 against a supply of 24, and searches.

### Raising a coefficient is a loop

The at-most-ones all come off *one* donor row, which is the case where #667's
`recover_am1_from_row` can bound a whole set in one step rather than pair by
pair. The rule's shape hides most of that win — a raise consumes the individual
pairwise lines, one per step, so they cannot simply be replaced — but a time
point where only full tasks can run is exactly `sum_F a_i <= 1` and could be one
`pol` instead of a RUP plus `|F|-1` sums. Left for the proof-size pass with
#666, since it changes bytes and wants measuring.

Cutting planes cannot raise a coefficient to the right hand side in one
division. Asking for the whole raise at once forces `(k−1)(T−R−1) < 1`, so each
step is `lambda` copies of the row so far plus the pairwise at-most-ones weighted
by their own coefficients, divided by `lambda + e`, with the step size bounded by
`k·(T − R) < T − c` — the margin the degree has to round through. A row whose
other tasks only just overshoot the capacity raises in one step; one that
overshoots by half pays a `pol` per unit of `kappa`. Hence `with_raise_budget()`,
predicted by the same function the derivation walks.

Two ends of that are not the loop: everything fitting alongside needs no division
at all (the at-most-ones summed *are* the row, relaxed by one `ia`), and nothing
else being able to run at all is a RUP.

### Testing

Time-table neutrality still holds, and now needs a second argument, since a
raised height really does change the profile: a raised task conflicts with
everything, so every verdict its height reaches is one the donor's capacity
reaches too. Asserted node-for-node on four fixtures, two of which raise.

Two new mutations, because the halves fail differently:

- **`RaiseTooFast`** takes one step past the bound. The step is sound and
  weaker, every later step compounds it, and only the row's `ia` pin objects.
- **`RaiseUnentitled`** raises a task that misses the pairwise test by one. The
  derivation then runs honestly and lands on a row the donor does not imply —
  the one mistake here that no care in the arithmetic could have caught.

The random corpus draws heights against each instance's own capacity now: a
fixed pool mostly produced donors declined for a height above the capacity, and
covered raising only by accident. It gets a second, smaller pass with proofs on
and veripb checking every one, which is where the raise arithmetic's corners get
exercised.

Fixture note: **`{6, 10, 4}` under `C = 13` is no longer a dynamic-programming
fixture.** The ten conflicts with both others, so it is now raised and `kappa`
over `{6, 4}` needs no subset sum. `{2, 6, 6}` under `C = 10` takes over as the
case where the gcd's multiples are not all reachable.

630/630 ctest, GCC and clang, ASan+UBSan clean on the touched tests.

### DIFF
--- a/dev_docs/README.md
+++ b/dev_docs/README.md
@@ -142,13 +142,16 @@ library. For an introduction to *using* the solver, start with the top-level
   unit propagation cannot do, and why a satisfiable test model and an `ia`
   content check are both needed to test it.
 - [Strengthening a `Cumulative` by integrality](cumulative-strengthening.md) —
-  Schulz's pre-solving capacity rules as the `CumulativeStrengthening` presolver,
-  posting the strengthened constraint in derived mode. Covers why the largest
-  reachable load is the real capacity, why the rules are *time-table neutral* and
-  how that turns into a node-for-node soundness tripwire, the `ia` step that
-  pins every row to the declared capacity (and is the only thing that catches a
-  sound derivation of the wrong line), the deviations from the paper's statement
-  of the rules, and why the deep-gap fixture everyone quotes cannot be a
+  Schulz's pre-solving rules as the `CumulativeStrengthening` presolver, posting
+  the strengthened constraint in derived mode. Covers why the largest reachable
+  load is the real capacity, why the tasks that fill the resource on their own
+  have to be set aside from that sum and given the capacity as their height
+  instead (which is both of Schulz's height rules arriving at the same place),
+  why raising a coefficient in cutting planes is a loop rather than one
+  division, why the rules are *time-table neutral* and how that turns into a
+  node-for-node soundness tripwire, the `ia` step that pins every row to the
+  declared capacity (and is the only thing that catches a sound derivation of
+  the wrong line), and why the deep-gap fixture everyone quotes cannot be a
   `Cumulative` instance.
 - [Inferring `Disjunctive` constraints across resources](inferred-disjunctive.md) —
   the `InferredDisjunctive` presolver: conflict cliques spanning several posted
--- a/dev_docs/cumulative-strengthening.md
+++ b/dev_docs/cumulative-strengthening.md
@@ -15,15 +15,58 @@ computes
                              that is at most C )
 ```
 
-and, when `kappa < C`, posts a
+— over the tasks that can run *beside something*, which is the subtlety the whole
+thing turns on, and posts a
 [derived `Cumulative`](cumulative-proof-logging.md#derived-cumulatives-an-implied-constraint-that-adds-nothing-to-the-model)
-over the same tasks with the same heights and a capacity of `kappa`. The donor
-stays posted and the OPB is untouched: each per-time capacity row of the derived
-constraint is *proved* from the donor's row for that time point, by
-[subset-sum strengthening](subset-sum-strengthening.md).
+at a capacity of `kappa`. The donor stays posted and the OPB is untouched: each
+per-time capacity row of the derived constraint is *proved* from the donor's row
+for that time point, by [subset-sum strengthening](subset-sum-strengthening.md).
 
 The rules are Schulz's pre-solving strengthenings, recapped by Cloutier and
 Quimper (CP 2026, §2.3).
+
+## The tasks that fill the resource, and why they are set aside
+
+Call a task **full** when it cannot run beside any other task that consumes
+anything: `h_i + h_j > C` for every `j ≠ i` with `h_j > 0` whose window overlaps
+its own. Such a task occupies the resource whenever it runs, whatever its height
+says.
+
+A full task ruins the subset sum. It reaches `C` on its own, so the largest
+reachable load *is* the capacity and `kappa = C` and nothing happens — a task of
+height `C` in the model turns the capacity rule off entirely. Excluding the full
+tasks from the sum is Schulz's refinement, and it is only sound if their own
+heights come down to `kappa` as well, which is what the derived constraint does:
+
+| | capacity | full task's height | every other height |
+|---|---|---|---|
+| donor | `C` | `h_i` | `h_j` |
+| derived | `kappa` | `kappa` | `h_j` |
+
+Both of Schulz's height rules arrive at that same table. His coefficient raising
+— any `c_i` above `C − min_j c_j` can be raised to `C` — is the same set of
+tasks, since `c_i > C − min{c_j : j ≠ i, c_j > 0}` says exactly that no other
+task fits alongside; and his knapsack rule's "`c_i = C` can be reduced to
+`kappa`" is what happens to those tasks once the capacity moves. So there is one
+rule here, not two, and `CumulativeStrengtheningStats::tasks_raised` counts it.
+
+Stating it as the pairwise test rather than as the minimum is deliberate. The
+two conditions are the same, but the pairwise one is what the certificate needs
+anyway — one at-most-one per pair, off the donor's own row — and it does not
+need the empty-minimum case argued separately. Tasks whose windows cannot
+overlap are left out of the test, which is a little more than the paper claims
+and costs nothing: if they can never be active together, no row ever mentions
+both.
+
+The set is **not** computed per time point, even though fewer tasks can run at
+one time point than over the whole horizon and the set would be larger for it. A
+`Cumulative` has one height per task, not one per time point, so a task that
+only fills the resource at some of them cannot be given a raised height at all.
+
+When *every* task is full, `kappa` is a subset sum over nothing and comes out at
+zero. That is not a strengthening but a disjunctive, and the donor is declined:
+inferring those from conflict cliques is
+[`InferredDisjunctive`](cumulative-proof-logging.md)'s job.
 
 ## Why `kappa` is the right number, and why the max is the max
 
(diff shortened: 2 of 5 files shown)


ciaranm:
Pushed a test fix: the verified sweep's "something was raised" assertion is seed-dependent, and fails on about one seed in fifty when all twenty-five drawn instances happen to have nothing to raise. That is what turned CI red here and on everything stacked above.

The sweep now keeps drawing until one instance has raised, up to a hundred. The assertion stays — dropping it would let a sweep that certified no raising at all pass quietly — and now means "the generator cannot produce a raisable instance in a hundred goes", which would be worth knowing. Eighty seeds pass, including seed 49, which reproduces the failure on the previous commit.

ciaranm:
An eight-hour randomised sweep over the Cumulative family (395,687 runs across GCC Release, clang Release and ASan+UBSan, caps off on the Release lanes) turned up **one defect, and it was in this file's other seed-dependent assertion**.

`cumulative_strengthening_presolver_test` draws sixty random instances and then asserts at least one had a height raised. Seed 268 draws sixty that do not; so do clang-lane seeds 10002223 and 10002570. It is the same shape as the verified-sweep assertion fixed earlier in this branch, and the same fix applies: draw up to 240, and say in the message how many were tried. Both assertions stay, because "it actually fired" is the only kind of check with any power over a presolver.

That is now twice in this file, which is worth knowing about the file rather than about either seed: any assertion of the form "the corpus found something" over a fixed draw is a CI failure waiting for the right seed.

Nothing else in the sweep failed. The strengthening presolver itself ran 10,321 times per lane against brute force with no lost solutions and no rejected proof.

ciaranm:
**Merge Check 14: wrong link target in `dev_docs/cumulative-strengthening.md`**

At this PR's head, L68–69:

> When *every* task is full, `kappa` is a subset sum over nothing and comes out at
> zero. That is not a strengthening but a disjunctive, and the donor is declined:
> inferring those from conflict cliques is
> [`InferredDisjunctive`](cumulative-proof-logging.md)'s job.

The link text names `InferredDisjunctive` but the target is `cumulative-proof-logging.md`. It should be `inferred-disjunctive.md`, which is where that presolver is written up and which does exist at this PR's head (added by #663).

It is the only one of its kind. Across the four new docs there are six links whose text names `InferredDisjunctive` or `InferredCumulative`, and the other five all point at the right file:

| link | target | |
|---|---|---|
| `cumulative-strengthening.md:69` | `cumulative-proof-logging.md` | ← wrong |
| `inferred-cumulative.md:22` | `inferred-disjunctive.md` | ok |
| `inferred-disjunctive.md:14` | `inferred-cumulative.md` | ok |
| `inferred-disjunctive.md:240` | `inferred-cumulative.md` | ok |
| `certified-makespan-bounds.md:7` | `inferred-cumulative.md` | ok |
| `certified-makespan-bounds.md:8` | `inferred-disjunctive.md` | ok |

Nothing would flag it: the target file exists, so even a link checker would pass it, and CI never renders the docs at all.

ciaranm:
Fixed at the head of the stack rather than here, in `4a7feeb2` on #699, because moving it to this branch restacks nineteen PRs for a one-word change. Happy to move it if you would rather the branch that introduced it also fixed it.

Checked the same thing across the rest of `dev_docs` while I was there: every markdown link whose text names `InferredDisjunctive`, `InferredCumulative` or `CumulativeStrengthening` now points at the right file, and this was the only mismatch. It is also the only broken relative target in the four new documents — the two `(when written)` placeholders in `regular.md` are deliberate and predate this stack.

PR #671, opened 2026-08-04
Presolver: infer Cumulatives by lifting cover inequalities (#549, single-resource half)

### LINKED ISSUE #674: Lifted cover cuts: partial weakening, and deleting the backward planner
Follow-up to #671, from a proof-consultant review of its certificate design. Everything below was tested against veripb 3.0.2; artefacts in `~/claude/tmp/fable-549/` (25 experiments, `run.sh`), key results reproduced independently in `~/claude/tmp/549-verify/`.

## The headline, which settles a question rather than opening one

**No checker-side rule can bridge from a capacity row to a lifted cover cut.** Tested and rejected: plain `rup`, hinted `rup`, `rup` against the full reified scaffolding, `ia` with the row as hint, bare `red`, bare `pbc`. The reason is structural — at `Top` nothing is assigned, PB unit propagation only fires when a coefficient exceeds a constraint's slack, and every constraint in play has slack to spare. Hints restrict which constraints propagate; they add no strength.

Every integrality-capable operation VeriPB has (`d`, `s`, and 3.0.2's `c`/`m`/`n`) takes solver-specified operands. So a solver-side model of the arithmetic is **intrinsic to the design space**, not a shortcut #671 took. That is worth recording, because it is the obvious thing to want and it does not exist.

## 1. Partial weakening, which #671's docs wrongly ruled out

`pol` can push a **literal axiom** (`x >= 0`), which cancels against that literal's complement and shaves *part* of a coefficient — where `w` can only remove all of it. `window_energy` already uses this; `lifted_cover_cut` does not.

With it, #671's motivating example is one `pol` rather than a cover plus a lifting step. Adding one copy of `a >= 0` to `5~a + 2~b + 2~c + 2~d >= 6` cancels one unit, giving `4~a + 2~b + 2~c + 2~d >= 5`; dividing by two lands exactly on `2~a + ~b + ~c + ~d >= 3`:

```
pol 1 aa + 2 d ;
e 2 ~aa 1 ~bb 1 ~cc 1 ~dd >= 3 : -1 ;
```

The `e` is an *equality* check, so this is exact, not merely implying. Claiming one better is rejected.

**The two families are incomparable**, so this does not retire the chain. Brute force over all four-member instances with demands at most five (44,226 valid tight cuts): 783 targets are chain-reachable and outside the whole one-`pol`-with-shaving family, 157 of those semantically new; 40 go the other way. Witnesses verified both directions. The right vocabulary is the union.

Modelling cost is small: `coefficient -= lambda; degree -= lambda` with `lambda <= c_i`. It does break `Normalised`'s no-cancellation invariant, which should become an explicit operation rather than an accident.

## 2. Deleting `plan_lifted_cover_cut`

The backward, target-directed planner exists only for window edges, where a row must be derived over a subset of the cut's members with the coefficients already fixed. That obligation can be made to **fall out of the full-support plan by construction**:

- one `red` per member, once at `Top`, introducing an always-false proof literal (`red 1 ~zz >= 1 : zz 0 ;`);
- inside each step's `pol`, extend the restricted row with `c_i * z_i` from those unit lines;
- replay the full-support plan **verbatim** — the derivation is literal-for-literal isomorphic, so it lands by construction;
- the final `ia` pin's weaken-cost arithmetic absorbs the leftover `z` terms without an explicit weaken.

Verified. Same per-time-point line count. This deletes the planner, its cover enumeration, its per-present-set cache, and its "declines the whole constraint" failure mode — i.e. it removes solver-side search rather than adding it, which is the direction #671's review asked for.

Care needed: the `z` literals are proof-only aux, so they must stay UP-determined for enumeration proofs. Unit lines do that; keep or `del` them, but do not make them free.

## 3. What VeriPB's `ia` check actually is (do not adopt uncritically)

Characterised from source (`veripb-formula/src/pb_constraint.rs:202-259`) and confirmed by probe: a one-pass weaken-cost check. Start `wc = deg_D - deg_C`; a source term absent from the target costs its full coefficient; a source coefficient above the target's costs the difference, *unless* the target coefficient is at or above the target degree; accept iff `wc <= 0`. No division, no gcd.

`round_onto`'s `== target` is therefore about 10x stricter than the pin actually requires (3,667 versus 37,624 targets on the grid). Widening it would find more cuts — **but it means reimplementing VeriPB's implication check in-tree, which is more solver-side PB semantics, not less**, and it is moot if item 2 is adopted. Recorded for completeness; not recommended.

One genuinely free win from the same analysis: the capacity row *occasionally* `ia`-implies a lifted cut directly, when every oversized coefficient sits at or above the target degree. A direct-`ia` attempt is a zero-cost shape 0 for the planner.

## Also checked and confirmed sound

No soundness bugs found in #671. The `Temporary`-at-`saved+1` / `forget(saved+2)` scaffolding dance is correct, the no-cancellation invariant genuinely holds for the current vocabulary, and inlining the derivation per conflict instead of citing a `Top` line was considered and is worse (the bound-push chain cites by line ID in hinted RUPs, which have no inline form).

## Suggested order

1. Item 2 — biggest structural simplification, removes search.
2. Item 1 — most real cuts collapse to one `pol`; keep the chain as the fallback the separation study says it must be.
3. Item 3's shape 0 only. Not the implies-acceptance.

### LINKED ISSUE #675: DECIDED: proof-log the lifting DP, rather than searching for short derivations
Supersedes most of #674, and would delete rather than extend the machinery in #671.

## The position we are in

`InferredCumulative` follows Sidorov's Algorithms 1 and 2 to decide what to infer, and then goes looking for a cutting-planes derivation of each constraint. Around **96% of inferred constraints certify**; the rest are dropped and counted. Reaching that 96% costs an in-tree model of VeriPB's normalised-form arithmetic, a search over `pol` programs, and an `ia` pin to catch the model drifting.

Every extension considered so far — literal-axiom shaving, variable-form division, MIR, merging pairwise at-most-ones — makes the search wider and the tail smaller, and none makes the tail empty. Over a corpus of 2,760 rows, 4.3% of inferred constraints resist every short shape tried; six of ten sampled are at-most-one constraints over a pairwise clique (which the capacity-one stage's certificate *would* reach), and the rest have no known route.

## The proposal

Stop deriving the *conclusion* of the lifting subproblem and derive the *computation* instead.

Algorithm 2 sets `pi_j = pi_0 - v*`, where `v*` is a knapsack optimum computed by dynamic programming. If the DP is itself proof-logged — a proof variable per state, a case split per transition — the lifting step becomes a mechanical replay of work we are already doing. That is **complete by construction**: every coefficient the procedure computes is one the proof establishes, so `cuts_uncertifiable` goes to zero and the reproduction becomes exact rather than 96% of exact.

Presolvers run once, so the cost lands somewhere much more forgiving than a propagator's per-call justification.

## We already have this machinery

`KnapsackUpfront` (#200, `gcs/constraints/knapsack/knapsack_upfront.cc`, documented in `dev_docs/knapsack.md`) builds exactly this object: a DP DAG over partial-sum vectors, with a **Top-level proof flag per state**, and paper-style proof logging over the transitions.

That also dissolves the obvious objection. A DP transition wants to condition on `sum c_i x_i <= w`, which is not a literal and so cannot be a `pol` operand; a per-state flag makes it one. `dev_docs/knapsack.md` also records a trap worth reading before starting: capping the state space by the initial total bound breaks the proof, because the per-call "eliminated" path needs a Top-level flag for the over-bound successor to reference.

## What it would delete

Nearly all of `gcs/innards/proofs/lifted_cover_cut.{hh,cc}`: the `Normalised` model of VeriPB's arithmetic, the divisor and copy-count search, the cover enumeration inside the planner, and the `ia` pin that exists only to catch the model being wrong. The window-edge restriction case goes too — the DP is simply rebuilt over the tasks present at that time point, with no separate planner and no cache.

That is the real argument for it. It is not only more complete, it removes the part of #671 that reproduces PB reasoning inside the solver, which was the original objection to that design.

## The cost, which is the thing to measure first

Size. Against one or two `pol` per time point today:

- a DP of `O(|S| * C)` states per lifting step, and up to `|S|` lifting steps per constraint, so `O(|S|^2 * C)` per time point;
- **times the horizon**, because the activity flags are per `(task, time)` and there is no way to derive the arithmetic once and instantiate it — a derivation over proof-only variables `z_i` cannot be transferred to `a_i[t]` for all `t` at once.

For `|S| = 10`, `C = 20`, horizon 1000 that is order `10^6` lines per constraint. Pseudopolynomial, and linear in the horizon rather than worse, but an ugly constant.

**Do not start by implementing it.** Start by emitting the line count analytically for the Pack/Pack_d instances in `~/claude/tmp/sidorov-548/`, and by checking what veripb does with a proof of that size — [[reference_veripb_perf_diagnosis]] has the tooling. If it is affordable there, this is clearly the right design; if it is not, the interesting question becomes whether the DP can be shared across time points at all.

## Relationship to the other issues

- #674's items 1 and 3 (literal-axiom shaving, widening the acceptance test) become moot if this lands. Its item 2 (dummy-literal replay for window edges) also becomes moot, for a better reason: there is no separate restriction planner to delete.
- #673 (multi-resource lifting) gets *harder* here, not easier: validity becomes multi-dimensional knapsack and the DP is `O(n * prod_j C_j)`, exponential in the number of resources rather than pseudopolynomial.

ciaranm:
Sixth of the [#541](https://github.com/ciaranm/glasgow-constraint-solver/issues/541) stack, on top of #668. Addresses the single-resource half of #549.

Closes #675, whose decision this implements — the certificate is a replayed knapsack programme rather than a search for short derivations. Closes #674, whose three items were all ways of widening that search, and are moot now there is no search to widen.

**#549 itself stays open.** Its title also covers certified RCPSP lower bounds, split out as #672 and still the headline deliverable, and multi-resource lifting, split out as #673.

A resource's capacity row says what fits. What it does not say is what integrality adds on top. Three tasks of demand two do not fit into a capacity of five, and a fourth task of demand five fills the resource on its own — so

```
2a + b + c + d <= 2
```

holds at every occupancy point the row allows, and at no rational relaxation of it. That is a *lifted cover inequality*, and `InferredCumulative` finds them and posts each as a derived `Cumulative` whose heights are the coefficients and whose capacity is the right-hand side. Stage one ([#663](https://github.com/ciaranm/glasgow-constraint-solver/pull/663)) was capacity one and unit coefficients; this is where the non-unit ones live.

## What a cut is worth, and what it is not worth

**Energy.** The tasks in a cut need `sum d_i pi_i` units of a resource supplying `pi_0` per step, so no schedule holding them can be shorter than the ratio — Sidorov's `L`, reported as `largest_capacity_bound` and the only output of this presolver that means anything without running a search. Where the ratio does *not* improve, the published procedure drops the cut: the donor already said it better.

**Nothing from time-tabling**, and the argument is one line this time. A cut is *valid*, so every 0/1 point the row allows satisfies it, and no verdict about a single time point can differ. Time-tabling ships off, as in #661 and #663, with the test asserting node-for-node equality with it back on.

## The inference is the paper's, unchanged

Algorithms 1 and 2 of Sidorov (CP 2026), verbatim, following the reference implementation where it and the paper disagree (Algorithm 2 lifts longest-duration-first, not shortest). The point of the exercise is a *certified reproduction*, so the constraints are whatever the published procedure infers and not whatever happens to be easy to prove — a reproduction whose inferences differed could not be compared against its published bounds, however good the inferences were.

The one deliberate divergence is that the lifting subproblem is over a single resource rather than the paper's Equation 4, which spans all of them. That is a restriction on what is *inferred* and not only on what is proved, and it is #673.

## The certificate: replay the knapsack programme, do not out-argue it

A lifted cover cut is true because a knapsack optimum says so — `sum pi_i a_i <= pi_0` holds at every 0/1 point the row allows exactly when `max { sum pi_i x_i : sum c_i x_i <= C }` is at most `pi_0` — and Algorithm 2 already computes that optimum. So rather than deriving the conclusion, this derives the computation, in the states-and-transitions shape of Demirović et al. (CP 2024) and in its **one-sided** form, the one their standalone knapsack solver uses: a state says "at least this much weight, at most this much profit" rather than "exactly this much of each".

`validate_lifted_cover_cut` builds the programme; `derive_lifted_cover_cut` emits it. Layer `i` holds the (weight, profit) pairs the first `i` members can reach; a successor either leaves the next member out or takes it and pays its demand; and **a successor that would overrun the capacity is not created at all**, which is the only place the donor's row is used and is exactly what makes the cut a consequence of it. Each state carries three extension variables and each transition an implication; each layer gets an at-least-one saying its states are between them complete; and at the end one flag reifies the cut, which every final state contradicts.

### This replaces a search, and closes its gap

The previous version of this PR went looking for a short cutting-planes derivation of each constraint's conclusion: three shapes of `pol`, an in-tree model of VeriPB's normalised-form arithmetic to predict where they would land, and an `ia` pin to catch that model drifting. It certified about 96% of what the procedure inferred. Each of the other 4% was a constraint the published method would have posted and we could not justify, so it was dropped and counted — and every way of widening the search traded a smaller tail for more of VeriPB's arithmetic reimplemented in-tree.

`cuts_uncertifiable` is now **zero**, by construction rather than by luck, and the verified sweep asserts it. Deleted along with the search: the `Normalised` model of VeriPB's arithmetic, the divisor and copy-count search, the cover enumeration inside the planner, and the backward planner the window edges needed.

The unit test's property changed to match. It used to be "never plan an invalid cut, and refusing a valid one is allowed"; it is now **exact agreement with brute-force enumeration in both directions**, over a few thousand random claims per seed. One of its cases, `3a + 2b + 2c + 2d <= 4` from `6a + 3b + 3c + 3d <= 8`, is there because no sequence of weakenings, saturations and divisions of that row reaches it.

### Dominated states, which is why this is affordable

A state taking no more of the resource while allowing no less on the cut says everything another one does, so the other can go. What survives runs strictly upwards in both coordinates, which means a layer holds at most one state per achievable profit — and a state whose profit exceeds `pi_0` would be a point breaking the cut, so **no layer is wider than `pi_0 + 1`**. Dropped states are never named: a transition landing on one is emitted straight into the state that covers it.

`pi_0` is a cover's size minus one, so it is two or three on the constraints the procedure actually infers.

### Proof size, measured, because an estimate is what argued against this

#675 budgeted `O(|S|^2 * C)` per time point — order `10^6` lines per constraint at `|S| = 10`, `C = 20`, horizon 1000 — assuming a programme per lifting step, indexed by residual capacity. Neither is needed: one programme certifies the finished cut, and the frontier is indexed by profit.

| members | capacity | `pi_0` | states | lines | bytes |
|--------:|---------:|-------:|-------:|------:|------:|
| 4 | 5 | 2 | 12 | 175 | 11 K |
| 6 | 10…80 | 3 | 23 | 331 | 23 K |
| 8 | 10…80 | 3 | 31 | 457 | 33 K |
| 12 | 10…80 | 3 | 47 | 697 | 59 K |

**The capacity does not appear.** Over a horizon, one such row per time point; for eight members and a capacity of twenty:

| rows | lines | bytes | veripb | peak RSS |
|-----:|------:|------:|-------:|---------:|
| 200 | 91 K | 6.9 MB | 0.19 s | 49 MB |
| 1000 | 457 K | 35 MB | 0.99 s | 164 MB |

Linear, and a second of checking for a horizon of a thousand. Scaffolding goes one proof level deeper than the caller's and is forgotten on the way out — extension variables included, since deleting a variable's two defining constraints deletes the variable — so only one line per time point survives. That is what keeps checking cheap as well as memory: live constraints tax every later unhinted RUP, and there are 456 per time point that do not outlive the row they establish.

### Restricting to a time point

A derived `Cumulative` has one height per task, so every time point's row carries the same coefficients. At the window edges only some tasks have flags, so the cut is restricted to them — still valid, because setting an absent task's flag to zero is a point it already covered — and the programme is simply built again over the members present, which is cheap and always succeeds. Answers are cached per distinct present set, so the middle of a window uses the one discovery built.

This used to be the one place the backward planner was needed, and the one place the search failed most often. It is now free.

## Testing

**A brute-force oracle that needs no proof checker.** A few thousand random claims per seed, most of them nonsense; every occupancy point enumerated; and the programme required to agree with that in *both* directions. Around 1,800 validated per seed, 1,200-odd with a non-unit coefficient and something to derive.

**"One better must fail"**, the signature test of a lifted constraint — a right-hand side one smaller, and each coefficient one larger, over every fixture. With small close-together integers a derivation that landed somewhere weaker still lands somewhere true, and only a rejected `+1` says the honest line is tight to what the constraint assumes of it.

**Mutations**, at the presolver, each of which veripb must reject. One less capacity and one more height are caught at the pin. The third used to skip a weakening of the donor's row; that mutation is gone, because it cannot be caught — saturation already forces the member out whatever else is left in the row, so the sweep is there to land the step on a two-literal clause rather than on something as wide as the donor. It is replaced by one that builds the programme against a capacity one below the donor's, so its states claim the row rules out a member it does not, which veripb refuses **inside the replay** rather than at the pin.

**The accounting assertion** — posted plus uncertifiable plus over-budget plus declined equals inferred — stays, and now sits beside an assertion that the uncertifiable count is zero.

632/632 ctest with caps off, clang clean, ASan+UBSan clean, five seeds swept with 100% certified on each.

## Not here

- **Multi-resource lifting** (#673). Harder under this design, not easier: validity becomes multi-dimensional knapsack, and the frontier is no longer bounded by the right-hand side.
- **Certified makespan bounds and an RCPSP artefact** (#672), which is #549's headline deliverable. `largest_capacity_bound` is still a number the presolver computed rather than a bound the proof establishes.
- A size trim worth roughly 40%: with hinted RUPs the per-coordinate transition clauses need not be materialised at all. Not done, because on the numbers above nothing is asking for it.


ciaranm:
Swept overnight: **395,687 runs over eight hours**, three lanes (GCC Release, clang Release, ASan+UBSan; caps off on the Release lanes), covering the Cumulative constraint and its optional and overload variants, the Disjunctive and Disjunctive2D globals, all three presolvers, the proof helpers, the six mutation entries that require veripb to *reject*, and fifteen `rcpsp` shapes end to end with veripb.

**No solver defect and no rejected proof.** The only real finding was a seed-dependent assertion in `cumulative_strengthening_presolver_test`, fixed in #668 below; the remaining failures were a job spec of mine and a binary I rebuilt underneath a running sweep.

What this PR's own machinery got:

- `inferred_cumulative_presolver`: 10,321 runs, each running a 25-instance verified sweep — about **258,000 randomised instances through the dynamic-programme certificate, every one checked by veripb**, with `cuts_uncertifiable` asserted zero throughout.
- `lifted_cover_cut`: 10,321 runs, each validating 4,000 random cuts against brute-force enumeration — about **41 million** checks that the programme is built exactly when the cut holds.
- The six mutation entries: 10,324 runs each, all passing, i.e. veripb **rejected** every deliberately corrupted proof, 61,944 times.

The 332 `toobig` and 240 `timeout` records are all `rcpsp` at sizes 12-14 hitting the sweep's own proof-size cap and five-minute timeout. They are instance hardness, recorded apart from failures on purpose — no proof was rejected.

ciaranm:
**Merge Check 14: "two orders of magnitude" does not follow from the figures given for it**

`dev_docs/inferred-cumulative.md` at this PR's head, L203:

> Measured, since the estimate that nearly sank this was out by two orders of
> magnitude.

Both numbers for that comparison are in the same section:

- **L215** — the #675 estimate, "around `10⁶` lines per constraint at `|S| = 10`, `C = 20`, horizon 1000";
- **L226** — the measurement, `| 1000 | 457 K | 35 MB | 0.99 s | 164 MB |`.

`10⁶ / 457 K ≈ 2.2`, which is well under one order of magnitude, not two. (Later in the stack the hinted replay re-measures that row to 375 K, giving 2.7 — same conclusion, so this is not something the rest of the stack fixes.)

The claim does hold of the **state** count, which may be what was meant. #675 budgeted `O(|S|² · C)` states per time point, so `10² × 20 = 2000`, against the 31 the table at L211 actually builds for eight members — about 65×, which is nearly two orders. But the sentence opens a section titled "Proof size" and is followed immediately by a lines-per-state table, and the document's own explicit estimate-versus-measurement comparison at L214–215 is stated in lines, so it reads as a claim about lines.

So: either the sentence wants to say states rather than lines, or the multiplier wants to be "a factor of two". Flagging rather than picking, since only you know which was meant.

Checked the rest of this version's arithmetic against itself while I was here and it holds up: the `21 / 2 = 10.5` against the donor row's `45 / 5 = 9`, the horizon-of-ten fixture sitting between them, and the lines-per-state ratios across all four rows of the L207–212 table.

ciaranm:
You are right, and picking one of the two readings would still have understated it. Fixed at the head of the stack rather than here, in `4a7feeb2` on #699, for the same restack reason as the link fix on #668 — say the word and it moves.

The estimate was wrong in *both* directions, and what "two orders of magnitude" was reaching for is one of them:

- **States.** #675 budgeted `O(|S|² · C)` per time point, which is 2000 at `|S| = 10`, `C = 20`. A row actually builds about 40 there (`members × (π₀ + 1)`, and the table measures 30 for eight members). **Fifty times out** — your 65× against the eight-member row, at the estimate’s own `|S|`.
- **Lines per state.** `10⁶ / (2000 × 1000)` is half a line per state. The measurement is twelve and a half. **Twenty-five times out the other way.**

So `10⁶` against 375 K is 50 / 25, and the line figure is two large errors nearly cancelling rather than a margin anything was decided on. The text now says that, and says the state count is the number to quote — which is also what the paragraph it opens goes on to argue, since "the capacity does not appear" is a claim about states.

Issue #672, opened 2026-08-04
Certified makespan lower bounds from lifted Cumulatives, and an RCPSP bounds artefact


ciaranm:
Follow-up to #549, split out of #671, which landed the inference but not the bound.

## What is missing

`InferredCumulativeStats::largest_capacity_bound` and `InferredDisjunctiveStats::largest_capacity_bound` are numbers the *presolver* computed. They are correct — a cut says its tasks cannot occupy more than `pi_0` of the resource at once, and between them they need `sum_i d_i pi_i`, so the schedule cannot be shorter than the ratio — but the proof does not contain that argument anywhere. What the proof contains is the derived constraint's per-time rows, and its propagator's energy reasoning at whatever nodes the search reached.

That is the gap between "we can report Sidorov's `L`" and "we can hand you a `.pbp` for it", and closing it is the headline deliverable of #549.

## The derivation

Issue 01's window-energy lemma (`gcs/constraints/innards/window_energy.hh`), applied to the *derived* capacity rows rather than a posted Cumulative's:

- under the objective order literal `[M <= mu]`, every task is confined to `[0, mu)`;
- `derive_window_energy` over that window gives each task `sum_{t in [0,mu)} active_{i,t} >= d_i`;
- summing those weighted by `pi_i`, against the derived rows summed over the window, contradicts when `sum_i d_i pi_i > pi_0 * mu`;
- the wrapping RUP concludes `M >= mu + 1`. Iterate `mu` up to the fixpoint.

Note `window_energy`'s restrictions, all of which bite here: constant length, and a `SimpleIntegerVariableID` start that is not `{0,1}`-domained. The clipped form its end-of-range literals produce is the one this needs.

## The artefact

A driver and scripts that model public instances, run the presolvers, emit the bound derivation, verify with veripb, and produce a table. `examples/rcpsp --dzn` and `examples/dzn.hh` already exist (#665), as does the Pack/Pack_d cross-check kit in `~/claude/tmp/sidorov-548/`.

Start with Pack/Pack_d — plain RCPSP. UBO/RCPSP-max needs difference constraints via the MiniZinc route and can come second.

## Validation

- **Every reported bound must be at most the best-known makespan** from Sidorov's tables. A bound exceeding a known feasible solution is a bug; automate the check.
[...]

PR #681, opened 2026-08-06
Hint every RUP the lifted cover cut replay emits

### LINKED ISSUE #676: Trim the lifted cover cut replay with hinted RUPs
Follow-up to #675, which landed in #671. Not urgent: the numbers below are comfortable, and this is recorded so it is not rediscovered rather than because anything is asking for it.

`derive_lifted_cover_cut` emits, per transition, two `pol` lines and three unhinted `rup`s: one clause for the weight half, one for the profit half, and then the state implication that follows from those two through the conjunction reification. The two half-clauses exist only so the state implication is one unit propagation away.

If the state implication's `rup` carried **hints** naming the reification lines it needs, the half-clauses need not be materialised at all. Measured on hand-written proofs of the same shape: about 10.5 lines per state hinted against about 15 as emitted, so roughly a 40% cut — six lines per source with a live take branch, three per source whose take the capacity kills.

Concretely, the transition
```
rup 1 ~S_prev 1 S_succ0 1 S_succ1 >= 1 : @sf_prev @wnt @ptk @sr_succ0 @pnt @wtk @sr_succ1 ;
```
closes because the weight half propagates the successor's `W` from the source's alone, the profit half propagates its `P`, and the conjunction reifications force the conflict. The two `pol`s stay; the two intermediate clauses go.

Costs: `hints.hh` machinery per emitted line, and a proof that is harder to read step by step — every line still says what it means, but the reader has to follow the hint list rather than the clause it replaced.

For scale, what is being trimmed (eight members, capacity twenty, one row per time point):

| rows | lines | bytes | veripb |
|-----:|------:|------:|-------:|
| 200 | 91 K | 6.9 MB | 0.19 s |
| 1000 | 457 K | 35 MB | 0.99 s |

So this is worth doing when a real RCPSP instance under #672 makes proof size or checking time an actual problem, and not before.

ciaranm:
Issue #676: the lifted cover cut replay's RUP steps carried no hints, so each of
them unit-propagated over the whole live constraint database. At the root of a
real model that database is the entire model, and the deletion that keeps the
*scaffolding* from accumulating can do nothing about it.

Every RUP the replay emits now names the lines it needs.

- A **transition** cites the source state's forward reification, which puts its
  halves in hand; the `pol` per half, each of which carries one bound from
  source to successor once the branch literal is fixed; and the successor's
  reverse reification, which turns the halves back into a state. **The clause
  naming each half is no longer emitted at all** — the state step was the only
  thing that ever wanted it, so it is a hint rather than a line. That is 94 of a
  single-resource row's 468 lines, about a fifth, and is the whole of what #676
  asked for.
- A **source's at-least-one** over its successors cites its transitions, plus —
  where a capacity rules the member out — the source's forward reification and
  the `pol` that does the ruling out. Unit propagation had often been reaching
  that through the rows in the database instead; a hinted step is only allowed
  what it names.
- A **layer's at-least-one** cites the layer before's and every one of that
  layer's successor steps, which is the resolution it is doing.
- The **conclusion** cites the last layer's at-least-one and the per-state
  clauses contradicting the cut flag.

Nothing about what is inferred changes. The one hint-free step left is the
degenerate window-edge case, where too few members have flags to reach the
right-hand side and the whole derivation is one line; hinting it works (the
negation is contradictory on its face, so citing nothing checks) but is
byte-for-byte invisible on the Pack instances, so it stays as it is.

## Sizes

Per single-resource row, by member count:

| members | `π₀` | states | lines before | lines after |
|--------:|-----:|-------:|-------------:|------------:|
| 4 | 2 | 12 | 179 | 147 |
| 6 | 3 | 22 | 338 | 272 |
| 8 | 3 | 30 | 468 | 374 |
| 12 | 3 | 46 | 728 | 578 |

Twelve and a half lines per state where it was fifteen and a half. Bytes fall
by less than lines, since a hint list is bytes.

## Checking time

Dropping the clauses and hinting the steps are separable, so both were measured.
Horizon of a thousand rows against a model that is one capacity row per time
point — minimum of five, one verify at a time, all in one sitting:

| | lines | bytes | veripb | peak RSS |
|---|------:|------:|-------:|---------:|
| clauses, hint-free (as it was) | 469 K | 42 MB | 0.84 s | 196 MB |
| clauses, hinted | 469 K | 44 MB | 0.63 s | 69 MB |
| no clauses, hint-free | 375 K | 35 MB | 0.71 s | 195 MB |
| no clauses, hinted (this PR) | 375 K | 36 MB | 0.53 s | 65 MB |

So against a bare model the hints buy a quarter of the time and two thirds of
the checker's memory, and dropping the clauses buys 15% more. Worth having, and
not obviously worth the trouble.

**Against a real model it is an order of magnitude.** The same comparison over
the root-refutation certificates the #672 sweep produces:

| instance | before | after | bytes before | bytes after |
|---|-------:|------:|-------------:|------------:|
| `pack008` | 3.84 s | 0.59 s | 48 MB | 42 MB |
| `pack012` | 3.68 s | 0.54 s | 43 MB | 38 MB |
| `pack018` | 12.10 s | 1.34 s | 112 MB | 100 MB |
| `pack025` | 16.27 s | 1.48 s | 124 MB | 111 MB |
| `pack039` | 6.05 s | 0.74 s | 57 MB | 50 MB |
| `pack043` | 17.69 s | 1.24 s | 98 MB | 84 MB |

Six to fourteen times faster to check, for a proof 12% smaller. The size is not
what did it.

The transferable half: **the value of hinting is set by how much is standing,
not by how much is being emitted.** A derivation measured in isolation and found
not to need hints may well need them once it is running inside a real proof.

## Testing

Nothing new tests the hints, because nothing can: a hint set that misses a line
its step needs is a step that does not check, so the existing cases fail.
Dropping the source's forward reification from the transition hints makes the
first fixture in `lifted_cover_cut_test` reject, which is what says the hints
are load-bearing rather than decorative.

632/632 ctest with caps off, clang clean, ASan+UBSan clean, and every mutation
in the family still refused.

Closes #676

### DIFF
--- a/dev_docs/benchmarking.md
+++ b/dev_docs/benchmarking.md
@@ -209,6 +209,28 @@ the no-`--all` bench at `n=22` produced shallow-search numbers that
 mislead about Bacchus's per-call value; the `--all` bench at `n=4..6`
 showed the per-call shape clearly (and reversed the proof-size verdict).
 
+### The bare-model trap, for anything about hints
+
+A hint-free `rup` propagates over the whole live constraint database, so
+what it costs depends on what else is standing — and a harness built to
+exercise one derivation deliberately has nothing else standing. Measure
+a hinting change that way and you will conclude it is not worth doing.
+
+Worked example (issue #676): hinting the lifted cover cut replay was
+worth 25 % of checking time against a model that was one capacity row
+per time point, and **six to fourteen times** against real RCPSP
+models, where the standing database is the whole model. Same code, same
+proof shape, two orders of magnitude apart in how the answer reads. The
+checker's peak memory went the other way — two thirds off on the bare
+model, nothing on the real one, where the model sets it.
+
+So for any change to hints, or to what lives at `ProofLevel::Top`,
+**bench against a real instance as well as a harness**, and report both.
+A root refutation is the cheap way to get one: it is a proof of the
+derivation and nothing else, with no search in it to confound the
+reading (`examples/rcpsp --deadline <bound-1> --prove`, and see
+`dev_docs/certified-makespan-bounds.md`).
+
 ### What size to pick
 
 For `--all` mode, scale `n` down until a single trial finishes in
--- a/dev_docs/inferred-cumulative.md
+++ b/dev_docs/inferred-cumulative.md
@@ -297,15 +297,15 @@ asserts a non-zero restriction count so it cannot quietly stop covering them.
 ## Proof size, which is what decided the design
 
 Measured, since the estimate that nearly sank this was out by two orders of
-magnitude. Over a single resource one derived row costs about **fifteen lines per
-state**, and the states are `members × (π₀ + 1)`:
+magnitude. Over a single resource one derived row costs about **twelve and a half
+lines per state**, and the states are `members × (π₀ + 1)`:
 
 | members | capacity | `π₀` | states | lines | bytes |
 |--------:|---------:|-----:|-------:|------:|------:|
-| 4 | 5 | 2 | 12 | 175 | 11 K |
-| 6 | 10…80 | 3 | 23 | 331 | 23 K |
-| 8 | 10…80 | 3 | 31 | 457 | 33 K |
-| 12 | 10…80 | 3 | 47 | 697 | 59 K |
+| 4 | 15 | 2 | 12 | 147 | 10 K |
+| 6 | 20 | 3 | 22 | 272 | 21 K |
+| 8 | 20 | 3 | 30 | 374 | 30 K |
+| 12 | 20 | 3 | 46 | 578 | 54 K |
 
 **The capacity does not appear.** Issue #675 budgeted `O(|S|² · C)` per time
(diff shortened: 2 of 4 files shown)


ciaranm:
Also adds the general form of this to `dev_docs/benchmarking.md`: a hinting change measured against a purpose-built harness has nothing standing for its hint-free steps to propagate over, so the harness will say it is not worth doing. Bench a root refutation as well.

ciaranm:
Second commit is unrelated to the hints and was found by the sweep run over them: `inferred_cumulative_test`'s random corpus asserts that at least one of sixty drawn instances posts a cut with a coefficient above one, and about one seed in ten thousand draws sixty that do not (905208 is one). That is an assertion about the draw, and it would eventually turn every PR in this stack red. Same fix as `cumulative_strengthening_test`'s, which had this twice: draw until the counters fire, capped at 240.

The sweep itself: 33 K runs over `lifted_cover_cut`, `inferred_cumulative_presolver` and `derived_cumulative` with the hinted code, one failure, that one.

PR #699, opened 2026-08-07
Say what the code does, where the review found it saying otherwise

ciaranm:
Part three of three answering the whole-stack review of #661–#694. Based on #698.

**Docs falsified by later work, which #690's sweep missed.**
`dev_docs/cumulative-strengthening.md`'s Restrictions section had its whole list
deleted where #682 meant to delete one bullet of it, leaving a colon with nothing
after it and a following paragraph pointing at "this list". Restored, and updated
for what the restrictions actually are now. `inferred_cumulative.hh` still said
"Restricted … to constant lengths, heights and capacities" and its `tasks`
counter still said "constant", both falsified by #687/#688 and both contradicted
later in the same header; `dev_docs/inferred-cumulative.md` still listed
"variable durations or demands" under "What is not here". And `with_makespan` in
both presolver headers promised that naming a wrong makespan variable makes
VeriPB refuse the derivation, where `makespan_links.hh` correctly documents the
opposite — no link means that task's energy is forgone, never a rejection. That
framing predates `find_makespan_links`.

**Two sentences a paper draft would crib, wrong as written.** The lifting
defaults are not "what his experiments used": 100 and 5 are the main
configuration's, which runs *unbudgeted* on calls, and 2·10⁴ is Appendix C's
termination variant, so our default is a combination he did not run. And our 29
closed instances and his "no less than twelve" are not the same measurement —
his counts unrounded bounds improving a prior bound, where his own closure test
gives 20 and the ceiling gives 36. Both now say so. The bounds artefact's numbers
also now carry an explicit "do not quote until rerun" at the point they are
stated rather than thirty lines later, and a list of what has changed under them
since — including this stack's own lifting-order and candidate-ordering fixes.

**The two Goal 1 divergences that are decisions rather than defects, written
down.** `lifting.py` abandons a cover whose optimistic elastic lower bound cannot
beat a full pool's worst; we have no counterpart, and the direction that costs
depends on the budget — unbudgeted we lift covers he abandons (a second
unattributed reason our L sometimes beats his), budgeted we burn budget on
hopeless ones, and our default is the budgeted one. Implementing it verbatim
would be wrong: his estimate is an upper bound only while unlifted coefficients
stay at most one. And `_min_clique_size = 3` never posts the two-task cliques his
binary covers are; the knob allows two, the default is a measurement question,
and the measurement is the artefact rerun. Both are now caveats with the
direction-of-effect argument, not silence.

**Alongside, the small consistency debts.** Three guards that read as live paths
and cannot fire, now saying so and why they are kept. A field written and never
read, and a comment describing a sweep that iterates something else. The recipe
contract — what it returns must be a line it derived, never a donor's own OPB
row, or the decline path deletes a model row. The two tightened-bound fallbacks
in `recover_constant_argument_row`, which have no fixture and whose soundness
rests on every root tightening being proof-logged. A `name_of` that is the right
accessor for these flags and the wrong one for others, which is worth not having
to discover. Loner tasks, which come out vacuously full and can cost a donor its
strengthening. The three different things the three presolvers do with a demand
above the capacity, only one of which was documented.

**And two that were more than comments.** A task the donor encoded but whose
length upper bound has since collapsed to zero landed in neither `usable` nor
`set_aside`, leaving its terms in every reduced row un-weakened — a row not
saying what the recipe using it claims. It is now asked about rather than assumed
away, at `lb(start)`, which is inside whatever window the donor used whatever has
happened to the length since. And the dominance drop #697 adds gets a fixture.

Verified: 635/635 ctest on GCC release, green on clang and Sanitize, and the Pack
cross-check still 20 of 20.


ciaranm:
**Merge Check 14: `dev_docs/cumulative-strengthening.md` keeps a superseded copy of its own Restrictions section**

Found running the pre-merge checklist over the stack. Not a mistake in either half — both read correctly on their own, which is what makes it easy to miss.

At this PR's head, `## Restrictions` (L310) runs:

- **L316–338** — the bullet list, six items: an irreducible capacity / a mandatory task above the capacity / a capacity too large to subset-sum over / every task full / nothing to gain / the two budgets.
- **L340–343** — "Optional tasks are deliberately **not** on this list. A presence is a conjunct…"
- L345–378 — variable lengths and heights, `CumulativeDonorView`, the conversion trade-off.
- **L380–390** — "Optional tasks are deliberately **not** on this list. The whole strengthening is…" — the same claim again, longer and differently worded.
- **L392–401** — the bullet list again, four items: a capacity that is a view / a height above the capacity / every task full / the two budgets.

The second list is a strict subset of the first. It drops "a capacity too large to subset-sum over" and "nothing to gain", and it reintroduces

> - **A capacity that is a view.** Its bits are not the ones the donor's rows mention…

as a fresh item, where L316 has already covered it as

> - **An irreducible capacity**, which today means a capacity that is a *view*.

Blame says the newer text is this PR's `23c0493d`, and the tail is older (`8a52e138` from #682, `db1050d2a` from #683, `b86c4235` from #668). The section appears to have been rewritten at the top with the old tail left standing below it — Check 14's "added alongside instead of replaced". The counting check: at #698's head "Optional tasks are deliberately" appears **once**; at this PR's head, **twice**.

Flagging rather than fixing, since which half is meant to survive is yours to say. From outside, the newer top list looks like the intended one, in which case L380–401 goes.

Worth noting nothing could have caught this: a docs-only change is unchecked by every CI lane, and both halves are individually true.

Two smaller things from the same pass, on other PRs in the stack, so they are not lost here: a wrong link target at `cumulative-strengthening.md:69` (#668), and an "out by two orders of magnitude" claim that does not follow from its own figures (#671).

ciaranm:
All three fixed, and thank you — the duplication is mine, and the reason it happened is worth recording.

**The Restrictions section.** The review item I was working from said #682 "deleted all five bullets instead of only Optional tasks; the section now opens a list with a colon and has no list". I read the twenty lines after the `## Restrictions` heading, saw a colon and prose, and believed it. The list was there — it had been *moved* to the bottom of the section, below all the "not on this list" paragraphs, which is why the colon dangles and why the review's reader and I both concluded it was gone. Taking a review finding at face value instead of reading the whole section is the mistake, and it is the same shape as the thing the finding was about.

Merged rather than deleted, since neither half was wholly right:

- the **top list is kept**, six items — it has the two the tail list is missing (a capacity too large to subset-sum over, from #697; nothing to gain) and it names the irreducible-capacity decline after the counter #690 renamed, where the tail still said "a capacity that is a view";
- the **tail's optional-tasks paragraph is kept**, and moved up to sit directly after the list, because it is the fuller of the two and because it is what makes the next paragraph's "not on this list *either*" refer to something;
- my shorter optional-tasks paragraph and the four-item tail list are gone.

Folded into `23c0493d` rather than added on top, so the PR reads as one thing.

**The other two** are in a commit of their own at the head of this branch, clearly labelled as not belonging here. Moving them to #668 and #671 would restack nineteen branches for two paragraphs of prose; say the word and they move.

- The link target is now `inferred-disjunctive.md`. I checked every markdown link across `dev_docs` whose text names one of the three presolvers — that was the only mismatch, and it is the only broken relative target in the four new documents.
- The "two orders of magnitude" claim: you are right that it does not follow, and right that it holds of the states. Rather than pick one, the text now says what actually happened, because it is more interesting than either reading. #675 budgeted `O(|S|² · C)` states per time point — 2000 at `|S| = 10`, `C = 20` — against the ~40 a row builds, so **fifty times out on the states**. It also costed a state at about half a line (`10⁶ / (2000 × 1000)`) where the measurement is twelve and a half, so it was **twenty-five times out the other way**. `10⁶` against 375 K is those two cancelling. The section now says the state count is the number that decided the design and the line figure is not a margin to quote.

Docs only, so no lane exercises any of it; the tree is otherwise byte-identical to what was pushed before.

PR #710, opened 2026-08-08
Eliminate all 326 GCC 15.2.0 build warnings

ciaranm:
A full GCC 15.2.0 build of the tree (Release preset, all 410 translation
units: library, examples, tests, XCSP and MiniZinc frontends) emitted
**326 warnings**. It now emits **0**.

```
$ cmake --preset release && make -C build -j32 -k -Otarget
```

| warning | before | after | whose |
|---|---:|---:|---|
| `-Wunused-result` | 219 | 0 | ours |
| `-Wunused-but-set-variable` | 73 | 0 | ours |
| `-Wmissing-field-initializers` | 12 | 0 | ours |
| `-Wswitch` | 6 | 0 | ours |
| `-Wunused-parameter` | 7 | 0 | 4 ours, 3 XCSP3 parser |
| `-Wsign-compare` | 5 | 0 | XCSP3 parser |
| `-Wdeprecated-copy` | 2 | 0 | XCSP3 parser |
| `-Wdeprecated-declarations` | 1 | 0 | XCSP3 parser (libxml2) |
| `-Wdangling-reference` | 1 | 0 | ours |
| **total** | **326** | **0** | 315 ours, 11 third-party |

The "after" number is from a from-scratch rebuild (`rm -rf build`,
reconfigure, 410/410 TUs compiled), not an incremental one.

326 warnings is only 30 distinct source locations — 288 of them are four
sites in one header that 72 translation units include.

## What was actually wrong

**`cake_probe.hh` — 288 warnings, four sites.** The measurement-only
cake chain probe is a header included by every constraint test.
`firstline` was a lambda nothing ever called, so it goes; GCC 15 reports
an uncalled local lambda under `-Wunused-but-set-variable`. The other
three are `std::system()` calls discarding their return value. glibc
declares `system()` with `warn_unused_result`, which — unlike
`[[nodiscard]]` — a cast to void does *not* silence, so the three go
through one `cake_run_ignoring_status()` helper whose name records that
the exit status is deliberately unexamined. It genuinely is: the probe
recovers from a failed command by inspecting what it did or did not
write, and must never fail a test.

**`GlobalCardinality` was carrying a dead argument — 4 warnings.**
`install_propagators` built `all_vars` (the variables concatenated with
the counts), captured a copy of it into each of the BC and GAC
propagator lambdas, and threaded it through to
`propagate_bounds_global_cardinality` / `propagate_gac_global_cardinality`,
neither of which ever read it. That is not just an unused parameter, it
is a live `vector<IntegerVariableID>` copy held for the lifetime of
every GlobalCardinality propagator for nothing, so the parameter is
removed rather than annotated.

**Two switches predate `InRange`/`NotInRange` — 6 warnings.**
`refined_watch_trigger_mask()` in `propagators.cc` was falling into its
conservative "wake on anything" default for both range operators. Not
wrong, but not free either, and now spelled out: `State::test_literal`
entails `x in [lo, hi]` exactly when both bounds have moved inside the
interval, so `InRange` only needs the bounds mask, while
`x not in [lo, hi]` becomes true when the domain stops intersecting the
interval — which a removal from the *interior* can do — so `NotInRange`
keeps the full `on_change` mask. The two oracles in `nogoods_test.cc`
cannot see a range condition at all (a `TestLit` holds a single value,
not an interval, and the random operator picker only produces the four
scalar ones), so those cases now fall explicitly into the never-taken
default the switches already had.

**`hints::Plus` / `hints::Minus` — 12 warnings.** The tabulated path
builds these as `Hint_{owner}`, correctly — it has no `pol` line to cite
— but `install_tabulation` is generic over the hint type and so cannot
name the remaining field. A default member initialiser on `pol_line`
fixes it at the definition instead.

**`LinearEquality` — 3 warnings.** Its tabulated path calls
`TabulationVariables::position_of` purely to register each term
variable, and `position_of` is `[[nodiscard]]`. The discard is now
explicit, with a note that the position is uninteresting precisely
because the sanitised terms are distinct, so position always equals term
index — which is what the acceptance test relies on when it indexes by
term index.

**Nothing here was a real defect.** No uninitialised read, no dangling
reference, no sign-compare that could go wrong: the `-Wsign-compare`
warnings are all in the fetched parser. The nearest thing to a bug found
is the dead `all_vars` plumbing above, which cost memory rather than
correctness.

## The suspected compiler bugs — there are two

Both are GCC being wrong about correct code, and both are handled the
way `element.cc:236` already handles a `-Wmaybe-uninitialized` false
positive: a narrow `#pragma GCC diagnostic push/ignored/pop` around the
single offending construct, guarded `#if defined(__GNUC__) && !
defined(__clang__)`. That guard is load-bearing — clang understands
`#pragma GCC diagnostic` but knows neither warning name, and answers
with `-Wunknown-warning-option` (verified). **One site each**, so a
pragma is far narrower than the tree-wide `-Wno-` the CMakeLists carries
for `-Wrestrict` and `-Wstringop-overread`; a tree-wide flag would also
disable the diagnostic for the ~150 other `visit(overloaded{...})` calls
in the tree, where it might one day be right.

### 1. `-Wdangling-reference`, `names_and_ids_tracker.cc:226`

`NamesAndIDsTracker::Imp::find_atoms` binds `const auto & table` to
`visit(overloaded{...}, id)`, where both lambdas return a reference to a
member of `*this`. GCC sees a reference-returning call with a temporary
(the `overloaded{...}` visitor) among its arguments and assumes the
result might point into that temporary. It cannot — it points at
`simple_variable_atoms` or `proof_only_variable_atoms`, which outlive
the visitor by the whole solve.

The tell that this is a heuristic and not an analysis: `atoms_for()`,
ten lines above, is the *same construct* — same `visit`, same
`overloaded`, same two members — and is not warned about. Only the
`const`-qualified copy returning a `const` reference trips it. A 30-line
reduction reproduces exactly that asymmetry:

```cpp
struct Holder {
    std::vector<int> xs, ys;
    auto find_nonconst(const V & id) -> int * {           // no warning
        auto & table = std::visit(overloaded{
            [&](const A &) -> std::vector<int> & { return xs; },
            [&](const B &) -> std::vector<int> & { return ys; }}, id);
        return table.empty() ? nullptr : &table[0];
    }
    auto find_const(const V & id) const -> const int * {  // warns
        const auto & table = std::visit(overloaded{
            [&](const A &) -> const std::vector<int> & { return xs; },
            [&](const B &) -> const std::vector<int> & { return ys; }}, id);
        return table.empty() ? nullptr : &table[0];
    }
};
```

Clang 21, whose lifetime analysis is stricter in general, is silent on
both. This is the long-running `-Wdangling-reference` false-positive
family; the closest documented shape is
<https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107532> — a reference into
the parent object, reported as dangling because a temporary happened to
be one of the arguments.

### 2. `-Wunused-but-set-variable`, `propagators.cc:586`

This one is the more interesting of the two, because the variable is
plainly read. `check_idempotent_claims` is read ~280 lines below, in the
`PropagatorState::EnableButIdempotent` arm. GCC still calls it set but
not used, at every optimisation level including `-O0`, so it is a
front-end decision, not an optimiser one.

Two things conspire. The read is inside `run`, which is a **generic**
lambda (`[&](auto & tracker)`). A `static` local is not captured, so
there is no capture to mark it read — and the only read that exists sits
in a lambda body that is still an uninstantiated template at the point
the enclosing function's scope pops, which is where the front end
decides this warning. Reduced to seven lines:

```cpp
#include <cstdlib>
extern void sink();
struct A { int a; };
auto f() -> bool {
    static const bool flag = nullptr != std::getenv("X");   // "set but not used"
    auto run = [&](auto & t) -> bool { if (flag) sink(); return t.a == 0; };
    A a{};
    return run(a);
}
```

Change `auto & t` to `A & t`, or drop the `static`, and the warning
disappears in either case — which pins both ingredients. Neither is a
change worth making to the propagation loop to appease a diagnostic
(`run` is generic because it is instantiated for both tracker types, and
the flag is `static` so the `getenv` happens once per process rather
than once per `propagate()` call). Closest bugzilla entry:
<https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96400>. Clang does not
warn.

## Third-party warnings (11)

The last 11 are in `build/_deps/xcsp3_cpp_parser-src` and are not ours
to fix — that tree is fetched at a pinned commit. They are only there
because `add_compile_options(-W -Wall)` at the top level applies to
every target in the build, so a third-party library gets compiled
against a warning set it was never written for. The last commit turns
warnings off for the parser's own two library targets, which is the
narrowest place to do it: nothing else changes, and the parser headers
we include are still compiled with our flags in our own translation
units. **That commit is separable — drop it if you would rather keep the
upstream noise visible.**

## Checks

- **GCC 15.2.0**, from-scratch Release build, 410/410 TUs: 326 → **0**
  warnings, 0 errors.
- **clang 21.1.8**, full Release build, 410/410 TUs: 153 warnings, 0
  errors — all pre-existing kinds (115 `-Winconsistent-missing-override`
  from `proof.hh` and `variable_weighting.hh`, 37
  `-Wunused-lambda-capture`, 1 `-Wdeprecated-copy`) at 4 distinct
  locations. **None of them is in any file this branch touches**, so
  this branch introduces no clang warnings and fixes none; clang's
  situation is untouched and out of scope here.
- **`ctest --parallel 32`**: 635/635 tests passed, 0 failed, on the GCC 15 Release build (`Total Test time (real) = 46.37 sec`).
- `clang-format` 21 clean on every touched file.

### DIFF
--- a/gcs/constraints/global_cardinality/bounds_global_cardinality.cc
+++ b/gcs/constraints/global_cardinality/bounds_global_cardinality.cc
@@ -51,8 +51,8 @@ using fmt::print;
 
 auto gcs::innards::propagate_bounds_global_cardinality(const vector<IntegerVariableID> & vars, const ConstraintID & owner,
     const vector<Integer> & values, const vector<IntegerVariableID> & counts,
-    const vector<pair<optional<ProofLine>, optional<ProofLine>>> & count_lines, const vector<IntegerVariableID> & all_vars, const State & state,
-    auto & inference, ProofLogger * const logger) -> PropagatorState
+    const vector<pair<optional<ProofLine>, optional<ProofLine>>> & count_lines, const State & state, auto & inference, ProofLogger * const logger)
+    -> PropagatorState
 {
     auto m = values.size();
 
@@ -395,10 +395,10 @@ auto gcs::innards::propagate_bounds_global_cardinality(const vector<IntegerVaria
 
 template auto gcs::innards::propagate_bounds_global_cardinality(const std::vector<IntegerVariableID> & vars, const ConstraintID & owner,
     const std::vector<Integer> & values, const std::vector<IntegerVariableID> & counts,
-    const std::vector<std::pair<std::optional<ProofLine>, std::optional<ProofLine>>> & count_lines, const std::vector<IntegerVariableID> & all_vars,
-    const State & state, SimpleInferenceTracker & inference, ProofLogger * const logger) -> PropagatorState;
+    const std::vector<std::pair<std::optional<ProofLine>, std::optional<ProofLine>>> & count_lines, const State & state,
+    SimpleInferenceTracker & inference, ProofLogger * const logger) -> PropagatorState;
 
 template auto gcs::innards::propagate_bounds_global_cardinality(const std::vector<IntegerVariableID> & vars, const ConstraintID & owner,
     const std::vector<Integer> & values, const std::vector<IntegerVariableID> & counts,
-    const std::vector<std::pair<std::optional<ProofLine>, std::optional<ProofLine>>> & count_lines, const std::vector<IntegerVariableID> & all_vars,
-    const State & state, EagerProofLoggingInferenceTracker & inference, ProofLogger * const logger) -> PropagatorState;
+    const std::vector<std::pair<std::optional<ProofLine>, std::optional<ProofLine>>> & count_lines, const State & state,
+    EagerProofLoggingInferenceTracker & inference, ProofLogger * const logger) -> PropagatorState;
--- a/gcs/constraints/global_cardinality/bounds_global_cardinality.hh
+++ b/gcs/constraints/global_cardinality/bounds_global_cardinality.hh
@@ -22,8 +22,8 @@ namespace gcs
          */
         [[nodiscard]] auto propagate_bounds_global_cardinality(const std::vector<IntegerVariableID> & vars, const ConstraintID & owner,
             const std::vector<Integer> & values, const std::vector<IntegerVariableID> & counts,
-            const std::vector<std::pair<std::optional<ProofLine>, std::optional<ProofLine>>> & count_lines,
-            const std::vector<IntegerVariableID> & all_vars, const State & state, auto & inference, ProofLogger * const logger) -> PropagatorState;
+            const std::vector<std::pair<std::optional<ProofLine>, std::optional<ProofLine>>> & count_lines, const State & state, auto & inference,
+            ProofLogger * const logger) -> PropagatorState;
     }
 
 }
--- a/gcs/constraints/global_cardinality/gac_global_cardinality.cc
+++ b/gcs/constraints/global_cardinality/gac_global_cardinality.cc
@@ -196,8 +196,8 @@ namespace gcs::innards
 
 auto gcs::innards::propagate_gac_global_cardinality(const vector<IntegerVariableID> & vars, const ConstraintID & owner,
     const vector<Integer> & values, const vector<IntegerVariableID> & counts, bool closed,
-    const vector<pair<optional<ProofLine>, optional<ProofLine>>> & count_lines, const vector<IntegerVariableID> & all_vars,
-    GacGlobalCardinalityScratch & scratch, const State & state, auto & inference, ProofLogger * const logger) -> PropagatorState
+    const vector<pair<optional<ProofLine>, optional<ProofLine>>> & count_lines, GacGlobalCardinalityScratch & scratch, const State & state,
+    auto & inference, ProofLogger * const logger) -> PropagatorState
 {
     auto m = values.size();
     auto n = vars.size();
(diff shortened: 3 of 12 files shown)


ciaranm:
Merged. One follow-up, doc-only, noticed during the pre-merge checklist and not worth holding the PR for.

`refined_watch_trigger_mask` gained two operators here (`propagators.cc:151-152`), but `dev_docs/refined-triggers.md` — which documents this exact function — still describes it as handling four:

- **L13**, in the definition of what refined triggers are: "arm watches on individual literals (``x = v``, ``x != v``, ``x >= k``, ``x < k``)".
- **L118-123**, the *Trigger masks* table, which has a row per operator and now covers four of the six. The prose under it at L125 says "`refined_watch_trigger_mask(literal)` derives this from the operator at arm time", so the table reads as the complete derivation.

For the record, since it saves re-deriving them: I checked both new arms against `State::test_literal` (`state.cc:761-775`) while reviewing, and they are right.

- `x in [lo, hi]` is `DefinitelyTrue` only when `lower_bound >= lo && upper_bound <= hi` — both bound conditions, so an interior removal cannot newly entail it, and `on_bounds` is correct.
- `x not in [lo, hi]` is `DefinitelyTrue` when the domain stops intersecting the interval, which a removal from the *interior* can do, so `on_change` is correct — and identical to the conservative default it used to fall into.

So this is purely the doc lagging the code, not a behaviour question.

Worth noting alongside it that nothing arms a watch on a range literal today: every `ctx.watch(...)` call site in the tree passes `>=`, `<=`, `<`, `==`, or a nogood literal, and a `Nogood` is a `vector<IntegerVariableCondition>` of decision literals. So the `InRange` arm is currently unreachable — which is fine and forward-looking, but does mean nothing exercises it, and the table is the only place a future caller would look to find out what it does.

PR #712, opened 2026-08-08
A diagnostic channel for solver decisions, instead of proof comments

### LINKED ISSUE #662: A diagnostic channel for solver decisions, instead of proof comments
## The problem

Several components currently tell the user what they decided by calling
`emit_proof_comment`. That is the wrong channel, and it is wrong in a way that
costs more than tidiness.

A presolver that declines a constraint has **silently weakened propagation**. The
only record of that decision goes into the `.pbp` — so it exists only when proofs
are on, which is not the configuration anyone runs when they care about
performance. A model that grows optional tasks, or a variable height, quietly
stops being strengthened and there is nothing to find out from.

`CumulativeStrengthening` (#661) is the immediate example, and its budget decline
is the sharpest case, because it carries a number the user would act on:

```
presolve cumulative: declining <id>, derivation would need 41000 dynamic
programming states against a budget of 20000
```

That is a "raise the budget" message, delivered by writing it into a proof file
that is deleted on success.

## This is not a complaint about proof comments in general

Most `emit_proof_comment` calls are genuine derivation annotation — `Step 1`,
`AM1 p[{}]`, `REACHABLE SET from {}`, `Regular: per-val backward chains` — and
those are landmarks inside a proof, read by someone reading the proof. They
belong exactly where they are and should not move.

The ones at issue are the messages *about the solver's decisions* rather than
about the derivation. A survey of the tree puts that set at roughly:

- `gcs/presolvers/cumulative_strengthening/` — three declines and a summary
  (#661);
- `gcs/constraints/cumulative/derived_cumulative.cc` — `derived cumulative: N
  capacity rows from <id>`;
- `gcs/presolvers/auto_table/` — `starting autotabulation`, `creating autotable
  with N entries`, `new table entry found`, `finished autotabulation`.

`AutoTable` is worth calling out: it has no stats block, so those comments are
the *only* observable record that it ran at all, and they disappear with proofs
off.

## There is already a precedent, half-built

`Stats::idempotence_downgrades` exists, in its own words, so that "a downgraded
propagator just keeps today's always-requeue behaviour; this counter makes the
downgrade observable". Same problem — a silent weakening the user should be able
to see — solved with a counter on `Stats` rather than a proof comment. `Stats`
already has an `operator<<`.

So this is less "invent a channel" than "finish generalising the one that is
already there".

## Proposed shape

A small vector of **typed** diagnostic notes on `Stats`, rendered in one place.

Typed rather than strings, for the same reason `Reason` is declarative, tuning
goes through `ArrayParam`, and consistency selection is a tag: a note should
carry the `ConstraintID` and the figures, and the one thing that knows how to
phrase it should be the renderer. A note built as a string at the call site is a
message nobody can filter, count, or assert on.

Sketch, to be argued with:

```cpp
namespace diagnostic {
    struct PresolverDeclined { ConstraintID constraint; DeclineReason why; };
    struct PresolverOverBudget { ConstraintID constraint; long long needed, budget; };
    struct ConstraintStrengthened { ConstraintID constraint; Integer from, to; };
}
using Diagnostic = std::variant<...>;
```

Two things to get right:

- **The per-component stats blocks stay, and become the source of truth.**
  `DifferenceLogicStats` and `CumulativeStrengtheningStats` are the precise
  machine-readable record, and the tests assert on them today. The renderer
  should read those rather than a second, parallel record — one source of truth,
  and no test needs to change.
- **The proof comments that are derivation annotation stay put.** Only the
  decision messages move. `#661`'s per-row `presolve cumulative gcd` / `kappa`
  markers are annotation and are counted in the `.pbp` by tests; they are not
  part of this.

## Alternatives considered

- **Named counters on `Stats` only.** Cheapest and most consistent with what is
  there, but "3 declined" throws away the numbers that make the message
  actionable, which is most of the point.
- **A `SolveCallbacks::diagnostic` callback.** More flexible, but push-based and
  opt-in, so the default experience stays silent — and the default experience is
  the thing being fixed.

## Testing

The obvious trap is that this is exactly the kind of feature that can be added
and then never fire. A test wants to assert that a *declining* configuration
produces the note — an optional-task Cumulative under `CumulativeStrengthening`
is a ready-made fixture — and, with proofs off, since that is the case the whole
issue is about.

## Out of scope

- Severity levels, filtering, and anything resembling a logging framework.
- `GCS_VERBOSE_LOGGING`, which is a proof-debugging tool and a different job.
- Changing any derivation annotation.

Prompted by review of #661.

ciaranm:
Closes #662.

## Read the comment, not the body

The issue's revision comment reverses several parts of the body, and this
implements the comment. A reviewer who remembers only the body should know that
the following were **deliberately not built**:

- **The central `std::variant`** (`using Diagnostic = std::variant<...>`) is
  gone. It is closed-world in an open-world tree: every new component would edit
  a Core header, and `stats.hh` would grow an include of every component's
  decline enum. Nothing here is a variant.
- **Severity levels are back in scope**, where the body put them out of it.
- **`SolveCallbacks::diagnostic` is back in scope**, where the body rejected it.
  What the body rejected was an *opt-in* callback defaulting to silence; this
  one defaults to `default_stats_report()`, which has the opposite property.
- **The "optional tasks or a variable height quietly stop being strengthened"
  premise is stale** and no fixture here relies on it. The capacity is the only
  argument that turns a whole donor away.
- **The two _proof-size_ budget declines are proofs-on only, but the `Important`
  note they raise is not.** `declined_over_budget` and
  `declined_over_raise_budget` do sit inside `if (logger)` at
  `cumulative_strengthening.cc:435`, because the figures are estimates of proof
  size. But they are two of *three* sites that increment
  `limit_declines_this_run`, and the third —
  `declined_capacity_too_large`, at `cumulative_strengthening.cc:228` — is
  deliberately outside that guard: the comment above it says it runs ahead of
  the proof budgets precisely because none of them runs with proofs off, and a
  donor with a capacity in the billions would otherwise spend hundreds of
  megabytes before anything decided there was nothing to strengthen.

  So **this PR does make the solver write to `cerr` in a default proofs-off
  run**: a `Cumulative` whose constant capacity exceeds
  `_max_subset_sum_capacity` (default 1000000) raises the aggregate `Important`
  note, and with no `stats_report` set that goes to `cerr` via
  `default_stats_report()`. That is the intended behaviour — a limit stopped the
  presolver doing what it was asked, which is what the rung is defined by, and
  it is the one limit decline reachable with proofs off — but it is a behaviour
  change for ordinary callers and is called out here rather than left to be
  discovered. See the discussion below on what that means for the three
  frontends.

  An earlier revision of this description said the budget declines were
  proofs-on only and stopped there, which was wrong in the direction that hid
  this; the coverage gap that let it stand is fixed too, in
  `cumulative_strengthening_test.cc`, which now asserts exactly one `Important`
  note on that path with proofs off.

## Typed notes, without a central variant

The open-world property comes from splitting what the body's variant conflated.

**Blocks stay the source of truth**, and derive from a narrow base so `Stats` can
render a block whose type it has never heard of:

```cpp
class ComponentStats
{
public:
    virtual ~ComponentStats() = 0;
    [[nodiscard]] virtual auto component_name() const -> std::string = 0;
    [[nodiscard]] virtual auto summary() const -> std::string = 0;
    [[nodiscard]] virtual auto entries() const -> std::vector<StatsEntry> = 0;
};
```

A component in its own directory joins the report by deriving from that and
calling `propagators.add_component_stats(_stats)`. No Core header learns that
the component exists. Every existing test on those blocks still passes
unchanged, which was the point of keeping them as the record.

**Decisions are pushed as events**, because a block is registered *empty* and
anything pulled from a finished block has lost the per-instance detail that made
the proof comments worth reading:

```cpp
struct StatsNote final
{
    StatsLevel level;
    std::string component;
    std::optional<ConstraintID> constraint;
    std::string text;
};
```

The `ConstraintID` is on the note rather than in the words, and `render()` is
the one thing that knows how to phrase it — same argument as `Reason` being
declarative and tuning going through `ArrayParam`. The typed part is the level,
the component and the constraint, which is what a caller filters on and what a
test asserts on; the figures are in `text` because the block already holds them
as numbers and a per-decline struct per component would be the closed world
again by another route.

`Propagators` is the bus, holding the `Stats &` `solve_with` constructs before
`create_propagators`, so a note fires while the presolver is running.
`SolveCallbacks::stats_report` is the handler, and is the one callback whose
unset behaviour is not silence: unset gets `default_stats_report()`, and silence
is asked for with `silent_stats_report()`. `solve_with` clears the handler before
returning, so a caller holding the result is not holding a live callback into a
finished search.

## The levels, and the name

`Debug < Detailed < General < Important`, filtered by comparison, with the
three-behaviours rule in the `StatsLevel` doc comment so it does not rot: a rung
is justified by a distinct default behaviour, not by a distinct shade of
importance.

I took the naming concern in the comment's *Deferred* section and picked
**`Important`** over `Severe` and over `Warning`. `Severe` asks for crashes.
`Warning` is the same mistake in a smaller font: in every logging framework it
means *provenance* — something went wrong — and everything at this rung is the
solver working correctly and getting the right answer. A user trained to read
`Warning: ...` as "a problem" will read "strengthening was skipped, answers are
still correct" as a problem, and a user trained the other way will start
ignoring the rung. `Important` says only "the reader needs this", which is
exactly what the rung is defined by, and it sorts naturally at the top of the
ladder.

## The proof comments

Deleted:

- the three `presolve cumulative: declining ...`;
- `starting autotabulation`, `creating autotable with N entries`, `finished
  autotabulation`;
- `derived cumulative: N capacity rows over M donors`.

Kept, being derivation annotation: `presolve cumulative gcd` / `kappa` / `amo`,
`presolve cumulative: strengthened <id> from capacity ...` — which is what keeps
the negative control at `cumulative_strengthening_test.cc` meaningful, since it
asserts the *absence* of the `presolve cumulative:` prefix for a donor that was
passed over — `new table entry found`, and `derived cumulative: declined at time
...`.

Checked by diffing real `.pbp` files against a build of `main`, since some test
lanes skip veripb:

- `auto_table --prove`: OPB byte-identical, and the `.pbp` diff is exactly the
  three deleted comment lines and nothing else.
- A two-donor Cumulative probe with `with_dynamic_programming_budget(0)`, so one
  donor is strengthened and one passed over: OPB byte-identical, `.pbp` diff is
  exactly the two deleted comment lines, `presolve cumulative: strengthened` and
  the three gcd/kappa markers still present, and VeriPB still verifies.

That probe's console output, which is the whole point of the issue:

```
solve time: 0.000802s
cumulative_strengthening: 1 of 2 posted Cumulatives strengthened, taking 2 off their capacities and raising 0 heights
cumulative_strengthening: passed over: the derivation would need 165 dynamic programming states against a budget of 0, see with_dynamic_programming_budget (_1)
Cumulative strengthening was skipped on 1 of 2 constraints because a size limit was reached; answers are still correct, but search may be slower
```

The last line also went to `cerr` live, when the presolver decided it.

## Components

- `CumulativeStrengtheningStats` becomes a `ComponentStats`, is always
  allocated, and carries a `DerivedCumulativeStats derived` sub-block.
- `AutoTableStats` is new: `ran`, `variables`, `tuples`, `search_nodes`. The
  worst case in the tree, since `tuples` was unrecoverable with proofs off.
  `search_nodes` is new information rather than a comment moved — the presolver
  solves a subproblem to exhaustion before search starts, and doing that
  expensively looked, from every other measurement, exactly like a slow model.
- `DerivedCumulativeStats` is an out-param on `DerivedCumulativeSpec`, folded
  into the caller's block, so a constraint installed once per donor contributes
  one aggregate rather than a component entry apiece. It has a fourth field the
  comment did not name, `constraints`, because `donors` alone conflates one
  constraint over three donors with three constraints over one.
- `declined_by_install` is **removed** rather than levelled, per the comment: it
  throws now. One guard the comment's reasoning does not cover — with assertions
  on the proof omits definitions, so a task with both a variable start and a
  variable length legitimately has no end-of-task line to pin through and the
  install declines for a reason the presolver cannot do anything about. That
  case `continue`s rather than throwing, with the reason in a comment.

## Testing

All three anti-never-fires tests from the comment, and each was **verified by
mutation** rather than assumed:

1. **A default-constructed presolver reaches `Stats::components()` with a
   non-empty summary.** As first written this did *not* catch a `clone()` that
   allocates a fresh block — what reaches `components()` is whatever the clone
   holds, so the summary looked right while the caller's handle read zero. It
   now also asserts, by pointer identity, that a caller-supplied block is the
   one registered. Mutating `clone()` to drop `_stats` fails both components.
2. **`entries().size()` equals the block's field count**, worked out from
   `sizeof` rather than written down, so adding a field moves one side of the
   comparison and not the other. Dropping one `entries()` line fails with "the
   flat view has 20 entries for 21 fields".
3. **The level is asserted, through a recording callback** rather than a parse of
   rendered output, which is exactly where a level gets thrown away. Dropping
   the view-capacity note from General to Detailed, and the aggregate from
   Important to Detailed, each fail on their own.

Three fixtures cover the level, one per reachable case, which is what the
corrected account of the decline sites above requires:

- **Proofs off, no `Important`**: a view capacity. Not a limit, so it stays at
  General — the negative control that stops `Important` becoming a rung
  everything reaches.
- **Proofs off, one `Important`**: a constant capacity beyond the subset-sum
  limit. This is the path an ordinary caller with no proof and no
  `stats_report` actually sees, so it asserts the *count* of `Important` notes
  rather than their presence: a second one is the solver talking over itself.
  Verified by mutation — dropping `++limit_declines_this_run` from the capacity
  site fails with "a capacity decline raised 0 Important notes with proofs off,
  not one".
- **Proofs on, one `Important`**: the dynamic-programming budget, asserting both
  the figures in the General note and the single Important note that names no
  constraint.

`ctest` is green on GCC (635/635) and on clang (639/639, Debug). Clang caught one
real defect GCC accepts silently: `GCS_LIFETIME_BOUND` on an unnamed parameter.

## Not done

- **`DifferenceLogicStats` is not migrated**, although the comment's level table
  sorts its fields and gives it an `Important` example (lifting zero edges from a
  non-zero number of donors). Deriving from `ComponentStats` makes it
  non-aggregate, which breaks seven designated-initialiser sites in
  `difference_logic_test.cc` — churn outside the comment's own survey of what
  moves. Mechanical follow-up, along with `InferredCumulativeStats`,
  `InferredDisjunctiveStats` and `DifferenceSimplificationStats`.
- **One judgement call to overrule if you disagree**: the comment's table names
  "either budget decline" for `Important`, and I also folded
  `declined_capacity_too_large` into the aggregate. It meets the stated test
  verbatim — a limit was reached, so the configuration being run is not the one
  that was asked for, and `with_subset_sum_capacity_limit` is a knob the caller
  can raise — and it is the only limit decline reachable with proofs off, so
  without it `Important` is a rung this presolver cannot reach in the
  configuration the issue is about. One `++limit_declines_this_run` to remove.
- **`python/gcspy.hh:61` is unchanged**, per the comment's *Deferred* section: it
  still returns `unordered_map<string, unsigned long long int>` and still works,
  since nothing existing moved.
- **The callback is not yet thread-safe**, but its contract now says it may
  eventually be called from any search thread, and `report()` being the only way
  in is what will make a `merge()` at the join easy.

### DIFF
--- a/gcs/CMakeLists.txt
+++ b/gcs/CMakeLists.txt
@@ -34,6 +34,7 @@ add_library(glasgow_constraint_solver
         constraints/count/count.cc
         constraints/cumulative/cumulative.cc
         constraints/cumulative/derived_cumulative.cc
+        constraints/cumulative/derived_cumulative_stats.cc
         constraints/cumulative/donor_view.cc
         constraints/difference/difference_constraints.cc
         constraints/difference/difference_graph.cc
--- a/gcs/constraints/cumulative/derived_cumulative.cc
+++ b/gcs/constraints/cumulative/derived_cumulative.cc
@@ -210,8 +210,6 @@ auto gcs::innards::install_derived_cumulative(
     // still the option of not installing a propagator whose inferences could
     // not be justified.
     if (logger) {
-        logger->emit_proof_comment(
-            "derived cumulative: " + to_string(rows_by_time.size()) + " capacity rows over " + to_string(spec.row_donors.size()) + " donors");
         for (const auto & [t, rows] : rows_by_time) {
             auto derived = spec.recipe(*logger, rows, t);
             if (! derived) {
@@ -286,7 +284,8 @@ auto gcs::innards::install_derived_cumulative(
 
         propagators.install_initialiser(
             [inputs, energy_tasks, energy_presences, makespan = *spec.makespan, capacity = spec.capacity, mutation = spec.makespan_mutation,
-                reached = spec.makespan_bound_reached](const State & state, auto & inference, ProofLogger * const logger) -> void {
+                reached = spec.makespan_bound_reached,
+                derived_stats = spec.stats](const State & state, auto & inference, ProofLogger * const logger) -> void {
                 // An optional task's length x height is guaranteed work only
                 // once it is known present, and at the root it usually is not.
                 // Counting one that is undecided would claim energy the
@@ -335,6 +334,9 @@ auto gcs::innards::install_derived_cumulative(
                 if (reached)
                     reached(bound->bound);
 
+                if (derived_stats)
+                    ++derived_stats->makespan_bounds_posted;
+
                 // Tests only: claiming one more than the argument reaches must
                 // be refused, since a derivation with slack in it verifies
                 // whatever it concludes.
@@ -378,6 +380,16 @@ auto gcs::innards::install_derived_cumulative(
             return propagate_cumulative(*inputs, state, inference, logger);
         },
         triggers);
+
+    // Only here, on the success path: every decline above returns without
+    // installing anything, and a block saying a constraint was derived when it
+    // was not is worse than one saying nothing. The caller's aggregate, not a
+    // component of this constraint's own --- see DerivedCumulativeSpec::stats.
+    if (spec.stats) {
+        ++spec.stats->constraints;
+        spec.stats->donors += spec.row_donors.size();
+        spec.stats->capacity_rows += inputs->capacity_lines.size();
+    }
 
     return true;
 }
--- a/gcs/constraints/cumulative/derived_cumulative.hh
+++ b/gcs/constraints/cumulative/derived_cumulative.hh
@@ -3,6 +3,7 @@
 
 #include <gcs/constraint_id.hh>
 #include <gcs/constraints/cumulative/cumulative.hh>
+#include <gcs/constraints/cumulative/derived_cumulative_stats.hh>
 #include <gcs/constraints/innards/makespan_energy.hh>
 #include <gcs/innards/proofs/proof_line.hh>
 #include <gcs/innards/proofs/proof_logger-fwd.hh>
@@ -14,6 +15,7 @@
 #include <cstddef>
 #include <functional>
 #include <map>
+#include <memory>
 #include <optional>
 #include <vector>
 
@@ -199,6 +201,19 @@ namespace gcs::innards
         /// all of them: it gets the same time-tabling and overload checking a
         /// posted Cumulative does, over the donors' flags.
         CumulativeRules rules;
+
+        /**
+         * \brief Where to add what this install came to, if the caller keeps a
+         * block.
+         *
+         * An out-param rather than a component of its own: the caller installs
+         * one of these per donor and reports one aggregate, so the natural
+         * owner is a DerivedCumulativeStats sitting inside the caller's own
+         * block. A shared_ptr rather than a raw one because
+         * makespan_bounds_posted is bumped from an initialiser, which runs
+         * after this call has returned.
+         */
+        std::shared_ptr<DerivedCumulativeStats> stats = nullptr;
     };
 
     /**
--- a/gcs/constraints/cumulative/derived_cumulative_stats.cc
+++ b/gcs/constraints/cumulative/derived_cumulative_stats.cc
@@ -1,0 +1,14 @@
+#include <gcs/constraints/cumulative/derived_cumulative_stats.hh>
+
+using namespace gcs;
+
+using std::string;
+using std::vector;
+
+auto DerivedCumulativeStats::add_entries_to(vector<StatsEntry> & entries, const string & prefix) const -> void
+{
+    entries.push_back(StatsEntry{prefix + "constraints", static_cast<long long>(constraints)});
+    entries.push_back(StatsEntry{prefix + "donors", static_cast<long long>(donors)});
+    entries.push_back(StatsEntry{prefix + "capacity_rows", static_cast<long long>(capacity_rows)});
+    entries.push_back(StatsEntry{prefix + "makespan_bounds_posted", static_cast<long long>(makespan_bounds_posted)});
+}
--- a/gcs/constraints/cumulative/derived_cumulative_stats.hh
+++ b/gcs/constraints/cumulative/derived_cumulative_stats.hh
@@ -1,0 +1,57 @@
+#ifndef GLASGOW_CONSTRAINT_SOLVER_GUARD_GCS_CONSTRAINTS_CUMULATIVE_DERIVED_CUMULATIVE_STATS_HH
+#define GLASGOW_CONSTRAINT_SOLVER_GUARD_GCS_CONSTRAINTS_CUMULATIVE_DERIVED_CUMULATIVE_STATS_HH
+
+#include <gcs/stats.hh>
+
+#include <cstddef>
+#include <string>
+#include <vector>
+
+namespace gcs
+{
+    /**
+     * \brief What install_derived_cumulative() did, summed over every derived
+     * Cumulative a caller installed.
+     *
+     * Not a ComponentStats of its own, and deliberately: a derived Cumulative
+     * is installed once per donor, so one component entry per derived
+     * constraint would be noise where one aggregate is what a reader wants. A
+     * caller keeps one of these inside its own block --- see
+     * CumulativeStrengtheningStats::derived --- and hands
+     * DerivedCumulativeSpec::stats a share of it, so every install adds to the
+     * same figures.
+     *
+     * \ingroup Innards
+     */
+    struct DerivedCumulativeStats final
+    {
+        /// Derived constraints installed: one per successful call.
+        std::size_t constraints = 0;
+
+        /// Donors those constraints took their capacity rows from, summed. A
+        /// derived constraint over several resources names more than one.
+        std::size_t donors = 0;
+
+        /// Per-time capacity rows derived, summed. Zero with proofs off, there
+        /// being no rows to derive: it is a measure of what the certificate
+        /// cost, not of what the constraint says.
+        std::size_t capacity_rows = 0;
+
+        /// Makespan lower bounds actually pushed. Only a spec that asked for a
+        /// makespan can contribute, and only where the energy argument reached
+        /// a bound at all, so this is the count that says the argument fired
+        /// rather than that it was asked for.
+        std::size_t makespan_bounds_posted = 0;
+
+        /**
+         * \brief Append these figures to a flat view, under `prefix`.
+         *
+         * For the enclosing block's ComponentStats::entries(), which is where
+         * these reach a caller: they are part of that block's flat view rather
+         * than a view of their own.
+         */
+        auto add_entries_to(std::vector<StatsEntry> & entries, const std::string & prefix) const -> void;
+    };
+}
+
+#endif
--- a/gcs/innards/conflict_observer_test.cc
+++ b/gcs/innards/conflict_observer_test.cc
@@ -64,7 +64,9 @@ TEST_CASE("Conflict observer attributes a wipeout to the failing constraint, not
     auto x = state.allocate_integer_variable_with_state(0_i, 1_i);
     auto y = state.allocate_integer_variable_with_state(0_i, 1_i);
 
-    Propagators propagators;
+    Stats stats;
+
+    Propagators propagators{stats};
     // Constraint _1: innocent, triggers on x only. Dense constraint index 0.
     propagators.install(NumberedConstraint{1}, a_propagator_that_does_nothing(), Triggers{.on_change = {x}});
     // Constraint _2: explicitly contradicts, with a non-empty reason. It is
@@ -103,7 +105,9 @@ TEST_CASE("Conflict observer fires for an inference-driven wipeout")
     State state;
     auto x = state.allocate_integer_variable_with_state(0_i, 1_i);
 
-    Propagators propagators;
+    Stats stats;
+
+    Propagators propagators{stats};
     // The contradiction here comes not from an explicit contradiction() call but
     // from an inference that empties x's domain (the Contradiction case inside
     // the tracker), exercising the other reason-stash path.
@@ -145,7 +149,9 @@ TEST_CASE("Conflict observer fires for a non-throwing (_or_stop) wipeout")
     State state;
     auto x = state.allocate_integer_variable_with_state(0_i, 1_i);
 
-    Propagators propagators;
+    Stats stats;
+
+    Propagators propagators{stats};
     propagators.install(
         NumberedConstraint{9},
         [x](const State &, auto & inference, ProofLogger * const logger) -> PropagatorState {
@@ -179,7 +185,9 @@ TEST_CASE("Propagation without a conflict observer still detects the wipeout")
     State state;
     auto x = state.allocate_integer_variable_with_state(0_i, 1_i);
 
-    Propagators propagators;
+    Stats stats;
+
+    Propagators propagators{stats};
     propagators.install(
         NumberedConstraint{1},
         [](const State &, auto & inference, ProofLogger * const logger) -> PropagatorState {
@@ -198,7 +206,9 @@ TEST_CASE("Every attached conflict observer is notified of a wipeout")
     State state;
     auto x = state.allocate_integer_variable_with_state(0_i, 1_i);
 
-    Propagators propagators;
+    Stats stats;
+
+    Propagators propagators{stats};
     propagators.install(
         NumberedConstraint{3},
         [x](const State &, auto & inference, ProofLogger * const logger) -> PropagatorState {
@@ -227,7 +237,9 @@ TEST_CASE("Propagator scope and variable-to-constraint adjacency")
     auto y = state.allocate_integer_variable_with_state(0_i, 10_i);
     auto z = state.allocate_integer_variable_with_state(0_i, 10_i);
 
-    Propagators propagators;
+    Stats stats;
+
+    Propagators propagators{stats};
     // Constraint index 0, propagator 0: over x and y (with a duplicate and a
     // view of y to check deduplication / view resolution).
     propagators.install(NumberedConstraint{1}, a_propagator_that_does_nothing(), Triggers{.on_change = {x, y}, .on_bounds = {y, -y}});
--- a/gcs/innards/idempotent_claim_checker_test.cc
+++ b/gcs/innards/idempotent_claim_checker_test.cc
@@ -34,7 +34,8 @@ TEST_CASE("A lying idempotence claim is caught by the checker")
     enable_checker();
 
     State state;
-    Propagators propagators;
+    Stats stats;
+    Propagators propagators{stats};
     auto x = state.allocate_integer_variable_with_state(0_i, 10_i);
 
     // Shaves one value off x per run, so an immediate re-run always infers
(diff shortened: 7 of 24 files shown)


ciaranm:
Merge-checklist review. CI here is green on all ten lanes, but that run is **43 commits behind current main** — it predates #710, #715, #716 and #717 — so I trial-merged onto `02b02543` and verified locally instead of relying on it: clean merge, GCC 15.2 release build with **0 errors and 0 warnings**, `ctest` **652/652** with veripb 3.0.2.

Three findings. None is a correctness defect in the channel itself, which reads well; the first is a factual error in the PR description that happens to conceal the one behaviour change here, and the second is the coverage gap that let it stand.

## 1. The description is wrong about when the Important note can fire

The body says, under "Read the comment, not the body":

> **The budget declines are proofs-on only.** Both sit inside `if (logger)` at `cumulative_strengthening.cc:448`

There are **three** sites that increment `limit_declines_this_run` (declared at `cumulative_strengthening.cc:192`), not two: L230, L454 and L463. The two at 454/463 are indeed inside `if (logger)`. The third is not:

```cpp
// cumulative_strengthening.cc:228
if (capacity > Integer{_max_subset_sum_capacity}) {
    bump(&CumulativeStrengtheningStats::declined_capacity_too_large);
    ++limit_declines_this_run;               // <-- no logger guard
```

and it is deliberately outside one — the comment immediately above it (L219-227) is explicit that this check exists *because* the proof budgets do not run with proofs off:

> It runs before any of the proof budgets, and with proofs off none of them ever runs at all --- so a donor posted in scaled units, with a capacity in the billions, spends hundreds of megabytes ... Magnitude is the wrong thing to find that out with, so decline on it first.

That counter is what raises the `StatsLevel::Important` note at L817. So with **proofs off and no `stats_report` set** — the default for every ordinary caller — a `Cumulative` whose capacity exceeds `_max_subset_sum_capacity` (default 1000000, L138) writes to `cerr`. Verified by running it rather than by reading:

```
$ ./harness            # proofs off, no callback, constant capacity 2000000
[harness] === anything between these markers came from the solver ===
Cumulative strengthening was skipped on 1 of 1 constraints because a size limit
was reached; answers are still correct, but search may be slower
[harness] === end ===
[harness] notes accumulated on Stats: 2
```

To be clear about what I think is wrong here: **the behaviour looks right.** A limit did stop the presolver doing what it was asked, that is a model-level consequence, and `StatsLevel::Important` is documented as being for exactly this reader. It is the *description* that is wrong, and it is wrong in the direction that would stop a reviewer noticing that this PR makes the solver write to `cerr` in a default proofs-off run. Worth correcting the body, and worth deciding deliberately that this note is wanted proofs-off rather than inheriting it.

## 2. Nothing exercises that path (Check 11)

The proofs-off stats test asserts the *opposite* case — a view capacity, which correctly does not count as a limit decline:

```cpp
// cumulative_strengthening_test.cc:1226
fail("a view capacity raised an Important note");
```

and every assertion about an Important note being *present* is inside `if (proofs)` (L1232, with the comment at L1230-1231 saying "Proofs on, necessarily"). So the single path on which this PR changes behaviour for ordinary proofs-off users is the one path with no coverage at all — which is the shape Check 11 exists for, and is presumably also why the description could say "proofs-on only" without anything contradicting it.

A fixture is cheap: post a `Cumulative` with a constant capacity above the subset-sum limit, solve with proofs off, and assert exactly one Important note. That is the same construction as the existing view-capacity case a few lines above.

## 3. All three frontends get `cerr` forced on them

`stats.hh:190` says of the callback:

> Set it to route notes into a frontend's or a client library's own idiom instead of having `cerr` forced on it.

but no frontend does. `minizinc/fzn_glasgow.cc:1139`, `xcsp/xcsp_glasgow_constraint_solver.cc:1999` and `python/gcspy.cc:163` all construct `SolveCallbacks` without `stats_report`, and this PR touches no frontend file. Given finding 1, that means `fzn-glasgow` can emit unsolicited non-FlatZinc text mid-solve, and `gcspy` can write to the host process's stderr from inside a library call.

This is a design decision rather than a defect — `StatsLevel`'s own documentation argues that the non-expert arriving through MiniZinc is precisely the intended audience, which is an argument for leaving it exactly as it is. Flagging it so it is a decision rather than a default.

## Verified and fine, but worth knowing

`silent_stats_report()` has **zero callers and zero tests** in the tree. That is more load-bearing than it looks, because `solve.cc:289` picks between the caller's callback and the noisy default on a truthiness test:

```cpp
stats.set_report_handler(callbacks.stats_report ? callbacks.stats_report : default_stats_report());
```

so a `silent_stats_report()` that returned an empty `std::function` would silently reinstate the default — asking for silence would get you noise, and nothing would fail. It is correct as written (`stats.cc:38` returns a non-empty lambda), and I confirmed by running it: silence really is silent, notes still accumulate on `Stats`, and a custom callback replaces the default rather than doubling it. No change needed; noted only because nothing in the tree would catch a regression.

## Everything else on the checklist passes

`Co-Authored-By` on all 7 commits; `derived_cumulative_stats.cc` wired into `gcs/CMakeLists.txt` in alphabetical order; `gcs/stats.hh` already in the umbrella, and `derived_cumulative_stats.hh` follows its sibling `derived_cumulative.hh` in staying out of it; no `AssertProofRule` / `AssertRatherThanJustifying`; no stray or generated files; no `overloaded{}` site touched; and `solve.hh`'s "Every callback is optional" was correctly amended rather than left to contradict the new default.

Two things I checked and decided were *not* findings, in case they come up: the `println(cerr, "diagnostics: ...")` lines in the tests are house convention rather than debug leftovers (64 of 111 test files on main use `println(cerr`), and the removed proof comments are unreferenced — the recipe markers the round-trip tests assert on (`presolve cumulative gcd` / `kappa` / `amo`) all survive, and `dev_docs/cumulative-strengthening.md:420`'s "no marker comment appears in the proof at all" is about the no-presolver negative control, which removing comments only makes more true.

ciaranm:
Findings 1 and 2 are both right, and both fixed in 5b179df6 plus a description
edit. Finding 3 I want to push back on for MiniZinc specifically, because the
FlatZinc spec turns out to have an opinion here — and it says the current
behaviour is correct.

## 1. The description, corrected

Confirmed by reading: three sites increment `limit_declines_this_run`, not two.
`declined_over_budget` (L454) and `declined_over_raise_budget` (L463) are inside
the `if (logger)` that opens at L435; `declined_capacity_too_large` (L230) is
not, and is deliberately not, for the reason the comment at L219-227 gives.

The body now says so, and says the consequence in the direction the old text
concealed: **this PR makes the solver write to `cerr` in a default proofs-off
run**, when a `Cumulative`'s constant capacity exceeds `_max_subset_sum_capacity`.
Agreed that the behaviour is right and the description was what was wrong.

## 2. Coverage for that path

The proofs-off fixture now comes in a pair, on either side of the rung: the
existing view capacity, which is not a limit and must stay at General, and a new
constant capacity of 2000000 against the default limit of 1000000, which is one
and must raise exactly one `Important` note. It asserts the *count*, not the
presence — what goes wrong on the one path a caller sees without asking for
anything is the solver writing more than it meant to.

Verified by mutation rather than assumed, since a note fixture that never fires
is exactly what let this through the first time. Dropping
`++limit_declines_this_run` from the capacity site fails with:

```
cumulative strengthening test failure: a capacity decline raised 0 Important notes with proofs off, not one
```

I also fixed the comment above the pair, which asserted that a proofs-off
`Important` fixture could not exist. That is true of the two proof-size budgets
and false of the limit that runs ahead of them, and it was the same error as the
body's, sitting directly above the gap.

`ctest` 635/635 on this branch, 0 warnings, 53 VeriPB verifications inside the
presolver test. (635 rather than your 652 because this is the branch as it
stands, not trial-merged onto current main.)

## 3. `cerr` is what the FlatZinc spec asks for

For `fzn-glasgow` I think the finding does not hold, and I would rather leave the
frontend alone than route notes somewhere the spec does not want them.

From the [FlatZinc specification](https://docs.minizinc.dev/en/stable/fzn-spec.html),
Output:

> All non-error output must be sent to the standard output stream.

> **Errors and warnings must be output to the standard error stream.**

> Print log messages (verbose solving) to the standard error stream. If solvers
> choose to print to standard output instead, all messages must be valid
> comments (i.e., start with a `%` character).

and on annotations it does not recognise, a solver "should print a warning on
the standard error stream if it does so."

`default_stats_report()` writes to `std::cerr`. So the FlatZinc protocol stream
— the solutions, `----------`, `==========`, the `%%%mzn-stat` block — is
untouched, and the notes land on precisely the stream the spec nominates for
warnings and log messages. This is not unsolicited non-FlatZinc text in the
output; it is not in the output at all.

`fzn_glasgow.cc` already does this for its own messages, including the spec's
ignored-annotation rule verbatim:

- `:1053` `Warning: treating unknown int_search variable heuristic {} as dom_w_deg instead`
- `:1077`, `:1082` — the same for value heuristics and search method
- `:303` `warning: ignoring unexpected argument '{}'`

An `Important` note appearing next to those is the same channel, used the same
way, for the same kind of thing. There is no `%%%mzn-` magic line for a warning
— the spec defines `%%%mzn-stat:`/`%%%mzn-stat-end` and `%%%mzn-core:`, and
nothing else — so plain text on stderr *is* the standard answer, and there is no
more idiomatic target to route to.

That leaves the other two frontends, where the argument is genuinely different
and the spec has nothing to say:

- **`gcspy`** is the one I would actually change. A Python library writing to the
  host process's stderr from inside a call is nobody's idiom, and the fix is one
  line: `silent_stats_report()`, with `Stats::notes()` available if the binding
  ever wants to surface them. Worth noting the binding at `python/gcspy.cc:193`
  returns a flat `unordered_map` and would need a shape for notes, which the
  deferred `gcspy.hh:61` work is already the right place for.
- **`xcsp`** already mixes both conventions — errors to `cerr` at `:1900`,
  `:1944`, `:1991`, but exception messages as `c ` comments on `cout` at `:1924`,
  `:1931`. stderr is outside the answer stream there too, so nothing is
  corrupted either way; I have not checked whether the competition rules say
  more than that, so I would leave it and not guess.

## The gap this did surface, in the other direction

`%%%mzn-stat:` is the standard channel for the *counters*, and `fzn-glasgow`
hand-writes them: the core ones at `:1219-1225`, then fourteen individually
written `differenceLogic*` lines at `:1230-1244`. `ComponentStats::entries()` is
exactly the generic flat name/value view that should feed that with a loop over
`stats.components()` — which is what it is for — and nothing wires it up. So
`AutoTableStats` and `CumulativeStrengtheningStats` are invisible to `-s` while
`DifferenceLogicStats` is enumerated by hand, and the deferred
`DifferenceLogicStats` migration would *remove* those fourteen lines from the
frontend with nothing replacing them.

That is a real follow-up, and it should land with or before that migration
rather than after it. One wrinkle: `entries()` names are snake_case
(`search_nodes`) and the existing stat names are camelCase (`peakDepth`,
`differenceLogicEdgesLifted`); solver-specific names are free-form so snake_case
is legal, but the loop probably wants to transform them to avoid a report in two
styles.

Happy to do that follow-up, and to take the `silent_stats_report()` line in
`gcspy` here if you would rather it not ship as is.

Issue #192, opened 2026-05-11
Plus, Minus: a GAC propagator over intervals (Minkowski sums), between bounds and tabulation


ciaranm:
*Rewritten on 2026-09-18 with Claude Code, against main at `cfb6da62`. The May 2026 text is in this issue's edit history. It was written when `Plus` did bounds propagation only, and most of it has since been overtaken.*

## Where Plus and Minus are today

- **`consistency::Auto` (the default)** tabulates the relation for GAC when the enumeration is small. `want_tabulation()` skips the largest-domained determined variable and compares the product of the remaining domain sizes against `GCS_TABULATION_THRESHOLD` (default 100). Otherwise it uses bounds propagation. (#452, `05c54e17`)
- **`consistency::BC`** runs the six bound rules, looping only when a written bound snaps across a hole. Since `a8ee28ca` it claims idempotence.
- **`consistency::Tabulated`** always tabulates.
- **`Minus` has its own propagator** (`2ec1297e`) rather than running `propagate_plus` over a `-b` view. With the view, the pol's bits failed to cancel: the model's row is over the user's `b`, but the reason literals landed on an unregistered view.

So the example this issue opened with, `dom(a) = [1, 3] ∪ [10, 12]` and `dom(b) = [0, 2]`, is already pruned to GAC by default, since its enumeration is 6 × 3. The gap only appears above the threshold. Take `a ∈ [1, 30] ∪ [100, 130]`, `b ∈ [0, 20]` and `result ∈ [0, 200]`: the enumeration is 61 × 21, so Plus gets bounds propagation only, and `result` keeps `[51, 99]`. A wide domain over `0..10^9` never gets past bounds propagation at all.

`examples/langford` is the in-tree workload that wants something stronger. It posts `position[i + k] + (i + 2) = position[i]` with `.with_consistency(consistency::Tabulated{})`, so that the example gets GAC whatever the width.

## Proposal: a genuine `consistency::GAC` arm

From one snapshot, prune

```
result  to  D(result) ∩ (D(a) ⊕ D(b))
a       to  D(a) ∩ (D(result) ⊖ D(b))
b       to  D(b) ∩ (D(result) ⊖ D(a))
```

Each set is built from intervals, and each pruning is one `infer_not_in_range` per gap run, found with `each_interval_minus()`. The primitives are the ones from #144 and #833.

[...]

PR #713, opened 2026-08-08
Delete logged solutions from proofs once something subsumes them

ciaranm:
We log every solution at `ProofLevel::Top` and keep it for the whole proof.
Nothing needs that: for optimisation the previous solution's constraint is
superseded the moment a better one is found, and for enumeration a solution's
blocking constraint is subsumed by the backtrack clause of the frame that
refutes the subtree it was found in. This deletes them. It also states the
solution itself in bits rather than in direct-encoding atoms.

**These proofs need a VeriPB from 2026-06-22 or later**, i.e. one containing
`veripb-dev` MR !193. An older checker rejects them at the root `rup >= 1`,
reported as if a constraint were missing, and `--unchecked-deletion` makes it go
away --- which is misleading, since there is nothing wrong with the deletions.
Every VeriPB build reports version 3.0.2 regardless of commit, so check the
commit rather than `--version`. What ships here verifies against public VeriPB
today; only the narrow promotion discussed below needs a newer checker.

The measurements are in `dev_docs/solution-constraint-deletion.md`; the
headline is that **enumeration checking gets an order of magnitude faster**,
and that optimisation is free but is not a speedup on anything we have.

| instance | solutions | size before | size after | time before (s) | time after (s) | speedup |
| --- | --- | --- | --- | --- | --- | --- |
| `frequency_square --all --size 6 --lambda 2` | 53220 | 204569673 | 224919034 | 117.19 114.45 125.08 116.78 | 12.52 12.34 12.69 12.38 | 9.4x |
| `regular_random --all --seed=1` | 32985 | 14231970 | 18870069 | 21.62 21.63 21.54 21.48 21.91 | 2.07 2.08 2.07 2.10 2.11 | 10.4x |
| `langford --all` | 52 | 964583 | 1024105 | 0.24 0.24 0.24 0.24 0.24 | 0.27 0.27 0.26 0.27 0.26 | 0.89x |
| `talent` (optimisation) | 23 | 6592428 | 6669941 | 1.13 1.18 1.14 1.12 1.12 | 1.15 1.14 1.13 1.11 1.12 | 1.00x |

Checker is `veripb-dev` at MR !217. Each pair is timed *alternately* so machine
drift lands on both sides; ratios use medians, because the `frequency_square`
before side has about ten per cent spread run to run.

Two things in that table are not good news and should not be glossed over.
`langford` is 12% slower and 6% bigger: 52 solutions is too few to pay back
the one-off promotion of the encoding to core --- and this is the one number the
narrow closure below would fix outright. And `talent` is the longest
improvement chain in the examples --- deleting 22 objective-improving
constraints moves the checking time by nothing, because such a constraint is
one short PB row over the objective terms and is not where the checker's time
goes. Take the optimisation half as tidiness, and as what stops a long-running
branch and bound accumulating rows without limit, not as a number. Peak RSS
moves the wrong way on the big enumeration case (`frequency_square`, 564 MB to
645 MB) while `regular_random` is flat.

## Solutions are stated in bits

`solx` / `soli` now get the assignment as bit literals at every assertion
level, where below `AssertionLevel::Definitions` they used to get eq atoms.
Both work --- the rule propagates whatever it is handed to a full assignment,
and unit propagation crosses between the two vocabularies through the same
reifications either way --- but the bits are the variables the OPB is written
over. The logged solution is then a solution to the *formula*, in the
formula's own terms, and the preserved set (so the projection the
`ENUMERATION_COMPLETE` conclusion is about) is the model's variables rather
than a set of atoms the proof introduced along the way. A trimmer has to keep
the model alive regardless; this way it does not also have to keep the
encoding definitions for whichever atoms the search happened to create.

It costs size, since a variable is one eq literal but several bit literals:
9% on `frequency_square`, 35% on `regular_random` (six variables, so a
solution line goes from six literals to eighteen against an otherwise small
per-solution proof), 5% on `langford`, and 2--4% of checking time. That is
separated from the deletion change in the dev doc.

## Optimisation

Two constraints per `soli`: the objective-improving constraint VeriPB creates
for the rule, which is **core**, so deleting it is a checked deletion; and our
own order-literal restatement, which is derived and free. The checked deletion
needs no scaffolding at all --- the constraint that discharges it is the
improving constraint the new `soli` has just put into core, over the same
terms. The only subtlety is ordering: the deletions have to follow the `e`
line, which addresses the new constraint as `-1`, and the new `soli`, which
supplies the discharging constraint.

## Enumeration, and the two things the issue did not anticipate

The issue expected the backtrack (and restart) constraints to need moving to
core. They do. Two more things came out of it.

**The lazily-introduced encoding definitions have to go to core too.**
`check_checked_deletion` sets `only_core`, so derived constraints are invisible
to it. Negating a blocking constraint fixes the preserved *bits*; the backtrack
clause is written in *atoms*; and the reifications joining the two are emitted
by `NamesAndIDsTracker` on first use, in the proof, so they are derived.
Without them nothing propagates and the goal fails. A witness or an explicit
subproof does not help. Promoting *every* Top line would be simpler and wrong,
because `DerivedCumulative` deletes abandoned Top lines (#666) and those
deletions would become checked and fail --- so the tracker marks its own, and
they are batched into `core id` steps issued lazily, the first time a deletion
needs them and never at all in a proof that deletes nothing.

What ships promotes all of the tracker's definitions, not just the ones a
deletion goal reads --- but that is a decision about the checker we can rely on
today, not a claim that the narrow rule is insufficient. It is not. Promoting
only the definitional closure of the atoms in the discharging clause (an eq
atom's two halves plus its two order cuts, and not the chain linking
neighbouring cuts) passes the whole suite, 636/636, and is a real reduction
rather than a no-op: `abs_test`'s first batch goes from 26 promoted constraints
to 10, and `langford --all` from 2604 over 124 `core id` steps to 76 over 76.
What it buys is better than tidiness. Measured blanket → narrow, same branch
both sides:

| instance | promoted ids | `.pbp` bytes | median time |
| --- | --- | --- | --- |
| `frequency_square` | 52786 → 52652 | 224919034 → 224918889 | 12.65 → 12.80 |
| `regular_random` | 7717 → 7679 | 18870069 → 18869929 | 2.08 → 2.03 |
| `langford` | 2604 → 76 | 1024105 → 1011205 | 0.26 → 0.24 |
| `talent` | 0 → 0 | 6669941 → 6669941 | 1.14 → 1.14 |

The saving lands where the enumeration is small, which is exactly where the
blanket rule costs us: `langford`'s 12% loss disappears completely, 0.24 s being
what the no-deletion baseline takes. On the big enumerations nearly every
definition is needed by some goal anyway, so there is almost nothing to save and
`frequency_square` is about 1% slower. `talent` promotes nothing either way ---
the optimisation half never needs a definition in core, so this is entirely an
enumeration question.

Blanket ships anyway because narrowing needs a checker containing MR !217 --- the
narrow rule's `skyscrapers 5` proof is one of the cases that fails without it,
and CI installs the checker by cloning public `VeriPB.git` at HEAD. Until that
contains !217, narrowing turns `skyscrapers-5` and `ortho_latin-5` red. Blanket
is the superset, so it is the safe end to sit at meanwhile, and nothing else in
the PR depends on which is used.

**Promotion cascades to the root.** Once a frame promotes its clause, the
parent's `forget_proof_level` deletes that clause, which is checked in turn,
discharged by the parent's own clause --- but only if the parent promoted as
well, which it has no local reason to. `ProofLogger` keeps
`core_lines_by_level` next to `proof_lines_by_level` so a frame can see that
the level it is about to forget holds core constraints.

Deletion goes through the level forget rather than next to the subsuming
clause: solution constraints are recorded at the proof *level* the subsuming
frame forgets, so `forget_proof_level` deletes them as part of the range it
already emits, which costs one fewer `core id` step and one fewer checked
deletion per solution.

## Deliberately not done

- The `solx` lines themselves stay --- `num_excluded_solutions` counts them,
  and that is what `ENUMERATION_COMPLETE` checks against.
- No *blocking* constraint is deleted under restarts. `RestartCutoffHit`
  forgets a level without emitting the frame's backtrack clause, so there is
  nothing to discharge the core deletions with. `solve_with` calls
  `ProofLogger::disable_solution_deletion` when restarts are on. This is
  specific to the enumeration half: the optimisation half deletes under
  restarts and at every assertion level, since what discharges it is the new
  `soli`'s improving constraint rather than a backtrack clause. Lifting it
  needs two things: the nld-nogood `solve.cc` already emits at Top for each
  refuted sibling promoted to core (it is a subset of the guesses, so it still
  subsumes the blocking constraints below it, and being at Top it survives the
  restart's forget), and `forget_proof_level` retaining rather than deleting
  the core lines of the abandoned spine, which nothing refutes
  (`IntervalSet::each_interval_minus`).
- Nothing is deleted above `AssertionLevel::Definitions`, which omits the atom
  definitions the deletion goals need.
- Nothing is deleted for a solution found at depth 0: there is no frame above
  it to refute it.

## Testing

`ctest` is 636/636 with GCC 15.2, including the lanes that actually invoke
veripb. `solution_deletion_test` is the guard against this quietly becoming a
no-op: it replays the proof's own relative addressing and asserts the `solx`
and `soli` constraints land in a later deletion, for both halves.
Mutation-checked --- defaulting `deleting_solution_constraints` to false gives
"0 of 12 solution constraints deleted", dropping the
`discard_superseded_objective_constraints` call gives "0 of 5".

## Related, not in this PR

VeriPB's "switching from stronger to weaker guarantee using unchecked
deletion" warning names no line, which makes a dropped guarantee very hard to
localise --- the symptom only shows up much later, as a rejected conclusion. A
minimal patch adding the proof line, the constraint ID and the proof goal is
sitting on the branch `claude/deletion-message-line-numbers` in the local
VeriPB checkout. It is worth upstreaming on its own merits.

https://claude.ai/code/session_01Wnubze1Z1SqBQxFyHg1PyY

### DIFF
--- a/dev_docs/README.md
+++ b/dev_docs/README.md
@@ -98,6 +98,13 @@ library. For an introduction to *using* the solver, start with the top-level
   distinction that governs which linking clauses are load-bearing — with the
   W1–W5 witness suite as the regression defence against re-simplification.
   Read when touching range/interval reasons, branching, or `infer_not_in_range`.
+- [Deleting logged solutions from a proof](solution-constraint-deletion.md) — why
+  a solution's constraint can be deleted once it is superseded, and what has to
+  move into VeriPB's core set first for the checked deletion to discharge:
+  trivial for optimisation, and for enumeration the backtrack clause, every
+  clause on the path back to the root, and all of the lazily-introduced
+  encoding definitions. Read when touching `ProofLogger::solution`,
+  `ProofLogger::backtrack`, or proof levels.
 - [View proof logging](view-proof-logging.md) — how the proof layer handles
 - [arithmetic-proofs.md](arithmetic-proofs.md) — how Multiply/Divide/Modulus/Power propagate and justify against cake's encoding: the slot-keyed emitters, the ConditionalBound justification layer, the sign-case driver, and the hard-won RUP/pol rules.
 - [Decision-diagram proof strategies](decision-diagram-proof-strategies.md) — for
--- a/dev_docs/solution-constraint-deletion.md
+++ b/dev_docs/solution-constraint-deletion.md
@@ -184,28 +184,32 @@ become checked and fail. So the tracker marks its own lines, with
 lazily --- the first time a deletion actually needs them, and never at all in
 a proof that deletes nothing.
 
-It is *all* of them, and not just the ones the deletion goal reads. That is
-worth knowing because the narrower rule is the obvious one to try and it looks
-right on small proofs. The goal needs the definitional closure of the atoms in
-the discharging clause: an order atom's two halves, an eq atom's two halves
-plus the two order cuts it is stated over, a range literal's likewise --- and
-not the order chain linking neighbouring cuts, since the bits settle each cut
-on its own. Promoting exactly that closure and nothing else takes `abs_test`'s
-first batch from 26 lines to 10, and it verifies. It does not survive contact
-with a real instance: `skyscrapers 5` and `ortho_latin 5` then fail, not at a
-deletion but at the root `rup >= 1`, with no unchecked-deletion warning
-anywhere before it. `--unchecked-deletion` makes the failure go away, which
-places it in the checked-deletion path; and the two lines that turn out to be
-missing (`grid2_2`'s `eq5` reification, `w_how_many_hidden[3][2]`'s `eq0`) are
-definitions of atoms that appear in *no* backtrack clause in the proof, so no
-rule phrased in terms of what the goals read can reach them. VeriPB's
-propagation engine keeps a persistent root-level trail and disables the derived
-propagation sets for the duration of an `only_core` episode
-(`propagation_engine.rs`, `init_trail`); the shape of the failure is
-root-level propagation through *derived* constraints not surviving that. A
-minimal hand-written reproduction has so far resisted: a single checked
-deletion over a derived clause, or a derived general PB constraint, both
-re-propagate correctly. Until that is understood, promote the lot.
+It is *all* of them, and not just the ones the deletion goal reads --- but that
+is parked rather than settled, and the honest position is worth writing down.
+The narrower rule is the obvious one to try: promote only the definitional
+closure of the atoms in the discharging clause, being an order atom's two
+halves, an eq atom's two halves plus the two order cuts it is stated over, and
+a range literal's likewise --- but not the order chain linking neighbouring
+cuts, since the bits settle each cut on its own. That takes `abs_test`'s first
+batch from 26 lines to 10, and verifies. Whether it is *sufficient* we cannot
+currently say, because the two experiments that would answer it disagree, and
+at least one of them is measuring the checker rather than us:
+
+- Against a current checker, the narrow rule fails `skyscrapers 5` and
+  `ortho_latin 5` at a shallow `rup`, with no unchecked-deletion warning
+  anywhere before it. Both of those failures are VeriPB regressions, bisected
+  to two unrelated commits and reported upstream as `veripb-dev` issue #210. A
+  checker predating either commit verifies both proofs.
+- Against that older checker, however, the narrow rule fails 24 *other* tests
+  that the blanket promotion passes, mostly `linear_constraint_*`. Those are
+  not diagnosed. They are either a real gap in the closure that the newer
+  checker's propagation happens to paper over, or further checker bugs.
+
+Sufficiency that moves with checker version is not something to build on in
+either direction, so the blanket promotion stays until #210 is fixed and the
+comparison can be re-run against a checker worth trusting. Nothing else here
+depends on which way that lands: the blanket rule is a superset of the narrow
+one, so it is the safe end to sit at while the question is open.
 
 **2. Promotion cascades to the root.** Once a frame has promoted its clause,
 its parent's `forget_proof_level` will delete that clause, and *that* is a
@@ -237,6 +241,13 @@ need a VeriPB from 2026-06-22 or later. An older checker rejects them at the
 root with an error that reads like a missing constraint rather than a checker
 bug, and `--unchecked-deletion` makes it go away, which is misleading --- there
 is nothing wrong with the deletions.
+
+That is not the only checker caveat, and "new enough" is not the same as
+"correct": two later propagation regressions are reported as `veripb-dev` issue
+#210. Neither affects what is shipped here, which verifies 636/636, but both
+affect the narrow-promotion variant discussed above, and both are reasons to
+treat a lone deletion-related verification failure as a question about the
+checker until a second checker version agrees with it.
 
 Deleting through the level forget is kept, but on its own merits rather than to
 dodge anything: it puts the blocking constraint at the level that actually
--- a/dev_docs/solution-constraint-deletion.md
+++ b/dev_docs/solution-constraint-deletion.md
@@ -185,7 +185,7 @@ dodge anything: it puts the blocking constraint at the level that actually
 refutes it, and it costs one fewer `core id` step and one fewer checked
 deletion per solution than deleting next to the subsuming clause does.
 
-**Restarts cannot have this.** A restart unwinds through
+**The enumeration half cannot have restarts.** A restart unwinds through
 `SearchResult::RestartCutoffHit`, which forgets each level *without* emitting
 the frame's backtrack clause. There is then nothing to discharge the core
 deletions in that level with. `solve_with` calls
@@ -198,6 +198,20 @@ this ever matters.
 Assertion levels above `Definitions` are excluded for a related reason: at
 `Links` and above the atom definitions are omitted from the proof entirely, so
 (1) has nothing to promote.
+
+**The optimisation half is unaffected by both.** This is worth stating
+separately, because "nothing is deleted under restarts" is the obvious thing to
+say and it is wrong. `discard_superseded_objective_constraints` is reached from
+`solution` whenever there is an objective (`proof_logger.cc:371`) and carries
+neither gate; `deleting_solution_constraints` and the `assertion_level` test are
+read only on the enumeration path (`proof_logger.cc:336`, `486`, `880`). Nothing
+here needs a backtrack clause: what discharges the deletion is the improving
+constraint the *new* `soli` puts in core, and `objective_value` is declared
+outside the restart loop (`solve.cc:381`), so each `soli` is still strictly
+better than the last across passes. `colour --restarts 1 --prove` duly deletes
+the previous solution's pair and concludes `s VERIFIED BOUNDS 2 <= obj <= 2`,
+which is exactly what gets rejected if the guarantee has been dropped.
+`solve_test.cc` covers both halves under restarts.
 
 ## Measurements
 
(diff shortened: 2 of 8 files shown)


ciaranm:
Pre-merge checklist findings. **Two, both documentation — no code defect in either.** Not merged: CI was still in flight when I looked.

For context on the rest: the branch is 24 commits behind `main`, so I trial-merged it onto current `main` (`69ddab8b`) rather than trusting `mergeable`. Clean merge, GCC 15 release build **0 errors / 0 warnings**, `ctest` **644/644** with veripb 3.0.2, including the new `solution_deletion_test`. `AssertProofRule` call sites are identical to main's, so nothing new is cheating a goal.

---

## 1. The headline `9.6x` does not follow from the stated method

`dev_docs/solution-constraint-deletion.md` L211 says "the ratio uses medians", and L220 says the first `frequency_square` "before" run is a cold-cache outlier that "the median ignores".

Taking the L215 row at its word:

| | values | median |
| --- | --- | --- |
| before | 117.83 113.07 113.23 113.24 | 113.235 |
| after | 11.91 11.87 11.85 11.79 | 11.860 |

113.235 / 11.860 = **9.55**, i.e. `9.5x`. Dropping the outlier before taking the median changes nothing (9.5472) — which is rather the point of using a median.

`9.6` is what the **mean including the outlier** gives (9.6451) — the one method the section says it is not using. The other three rows are exact by medians, which is what makes this look like a single stale cell rather than a different convention:

| row | by medians | printed |
| --- | --- | --- |
| `frequency_square` | 9.5476 | **9.6x** |
| `regular_random` | 10.8182 | 10.8x |
| `langford` | 0.8846 | 0.88x |
| `talent` | 1.0091 | 1.01x |

The conclusion is unaffected — 9.5x is still an order of magnitude — but this is the number the PR description leads with, so it is worth having right. Everything else in the section cross-checks: `talent`'s 576 bytes, `langford`'s 1.47% ≈ "1.5 per cent" and 0.26/0.23 ≈ "13 per cent slower", and the RSS figures against the PR body's MB.

## 2. "Nothing is deleted under restarts" is not what the code does

The claim appears three times — L250 ("Nothing is deleted under restarts, or above `AssertionLevel::Definitions`"), the **Restarts cannot have this** section at L188, and the PR description's "Deliberately not done".

`discard_superseded_objective_constraints()` is gated only on `optional_minimise_variable_and_value` (`proof_logger.cc:377`). It carries no restart gate and no assertion-level gate. `deleting_solution_constraints` is read in exactly three places — `proof_logger.cc:336`, `486` and `880` — and all three are the enumeration side.

Demonstrated rather than inferred, with `colour --restarts 1 --prove`:

```
430:soli i[vertex[0]][eq0] ... i[colours][eq2];
431:e -1 i[colours][b0] -2 i[colours][b1] -4 i[colours][b2] >= -1:-1;
432:rup 1 ~i[colours][ge2] >= 1;
433:del id -73;
434:del id -72;
```

— the previous solution's pair, deleted under restarts, in exactly the order the Optimisation section says is load-bearing (after the `e` line, after the new `soli`). And it verifies:

```
s VERIFIED BOUNDS 2 <= obj <= 2
```

**I think the code is right and the wording is what needs fixing.** The discharging constraint is the new `soli`'s improving constraint, which is in core and has nothing to do with backtrack clauses; and `objective_value` is declared outside the restart loop (`solve.cc:381`, whose comment says the incumbent bound persists across passes), so every `soli` is still strictly better than the last. That the `BOUNDS` conclusion is accepted above is the direct evidence — it is precisely what gets rejected if the guarantee has been dropped.

Note the header comment on `disable_solution_deletion` is already precise about this ("Stop deleting solution-**blocking** constraints"), so it is only the prose that generalises too far. The same L250 bullet also claims nothing is deleted above `AssertionLevel::Definitions`, which is true of the enumeration half (`proof_logger.cc:336` tests it) but not of the optimisation half.

Worth knowing for the rewording: `solve_test.cc` already has proof-verifying `"Optimise with restarts"` and `"Enumerate all solutions with restarts"` cases, so this interaction is covered rather than merely believed.

ciaranm:
Both fixed in `bbf85646`, documentation only — thanks, and the `9.6` was mine: it is the mean with the outlier left in, which is exactly the method the section disclaims.

- **Ratio.** Now `9.5x`, and the note about the cold-cache run says *why* a median makes dropping it by hand unnecessary, rather than implying the drop is what fixes it. The PR description above and the tip commit message are corrected too.
- **Restarts.** Reworded in all three places. I confirmed your reading against the code before changing anything: `discard_superseded_objective_constraints` is reached from `solution` whenever there is an objective (`proof_logger.cc:371`) with neither gate, and all three reads of `deleting_solution_constraints` (`336`, `486`, `880`) plus the `assertion_level` test are on the enumeration path. The doc now states the optimisation half's independence as its own claim rather than leaving it to be inferred, citing your `colour --restarts 1` evidence and the `solve.cc:381` reason each `soli` stays strictly better across passes.

No code changed, so the tree is the same one you trial-merged and got 644/644 from.

ciaranm:
Closing: superseded by #914, which implements the same feature — deleting `solx`
and `soli` constraints once they are spent — on a foundation this branch did not have.

ciaranm's call, and the reason is this branch's own history: its conclusions were
derived against a VeriPB that was not sound for this workload, so they are not
trusted and are not being carried forward.

  - `cef7a6aa` — veripb-dev issue #192, fixed by MR !193 ("never reset trailhead to
    higher position than current", merged 2026-06-22). The checker installed at the
    time **predated the fix while still reporting version 3.0.2**, which is what made
    the problem so hard to see.
  - `fbb6fee3` — veripb-dev MR !217: watches of derived constraints were not reset
    when the derived propagator was re-enabled after a core-only episode. Every
    checked deletion *is* such an episode, which is exactly why these proofs were the
    ones tripping it.
  - `2bff3933` — the narrow-promotion question parked rather than resolved, because
    resolving it needed !217. What shipped here is therefore the blanket promotion.

#914 reaches the same goal more simply, and the difference is not cosmetic. This
branch adds ten members to `ProofLogger` — `begin_recording_definitions()` /
`end_recording_definitions()`, a `DefinitionRecordingScope` RAII type, a batched
`move_to_core(const std::vector<ProofLine> &)`, `promote_definitions_to_core()` and a
`disable_solution_deletion()` hatch. #914 adds five, replacing that apparatus with
`ProofLevel::TopAndCore`, so core membership is stated at each emission site instead
of living in ambient scope state a caller cannot see. #914 evaluated this design
explicitly and measured the cost of rejecting it (1-3% of proof size, no measurable
checking time).

#914 also closes the hole that let a lenient checker matter in the first place: it
passes `--force-checked-deletion` from every script and helper that verifies a solver
proof, so a failed deletion check is an error rather than a silent downgrade to
unchecked deletion that still prints `s VERIFIED`. Adding that flag alone left the
suite green, which is what says it had been a no-op until now.

Nothing is stacked on this branch, and `solution-proof-deletion` is left in place at
`26a4d0b7` rather than deleted, so this is reopenable and the write-up remains
readable if anyone wants it. No content is being salvaged into #914.

ciaranm:
Correction to the comment above: the `solution-proof-deletion` branch has now been
**deleted** at ciaranm's request, so "left in place ... so this is reopenable" no
longer holds.

The 13 commits were never merged to `main`, so recording where they were is the
whole of the recovery path. Head was:

```
26a4d0b7beada9a97fc0120ef76b8fbe441f6bfc
```

GitHub keeps a closed PR's commits and diff viewable on the PR itself, so the
write-up and the code remain readable here. Reopening, however, needs the branch
back first — a PR whose base or head branch is gone cannot be reopened until the
ref exists again:

```shell
gh api repos/ciaranm/glasgow-constraint-solver/git/refs \
    -f ref=refs/heads/solution-proof-deletion \
    -f sha=26a4d0b7beada9a97fc0120ef76b8fbe441f6bfc
gh pr reopen 713
```

Superseded by #914, which is on `main` as of 78d2af7b. Nothing here was salvaged,
and per ciaranm its conclusions are not trusted — so treat the above as a way to
read the history, not as a starting point.

PR #715, opened 2026-08-10
Dual licence under Apache-2.0 or MIT

ciaranm:
Funding conditions now require that software developed under them is released under both the Apache License, Version 2.0 and the MIT License. The usual arrangement for this pair (the one the Rust ecosystem popularised) is to offer the work under *either*, at the recipient's option, so that is what this does: in SPDX terms, `MIT OR Apache-2.0`.

What changes:

- `LICENCE` becomes `LICENCE-MIT` (a rename, so the history follows), with the copyright year rolled forward to 2026.
- `LICENCE-APACHE` holds the Apache License, Version 2.0. The text is byte-identical to <https://www.apache.org/licenses/LICENSE-2.0.txt> (sha256 `cfc7749b…`), appendix included, as the licence asks.
- `COPYRIGHT` states the "either of these, at your option; you do not have to satisfy both" position in one place, with the SPDX expression and the inbound-contribution paragraph. This is the file to point people at, and the reason for keeping it separate is that neither licence file can say it without ceasing to be a verbatim copy.
- `README.md` gains a Licence section, and `CONTRIBUTING.md` a Licensing section stating that contributions are dual licensed the same way unless the contributor says otherwise, that there is no copyright assignment, and that third-party code in a contribution must be declared so it can be checked against both licences.
- `install()` now puts `COPYRIGHT` and the two licence texts in `${CMAKE_INSTALL_DOCDIR}`. Section 4(a) of the Apache licence asks that redistributions carry a copy, and until now an installed copy of the library carried none.
- `python/pyproject.toml` declares `license = "MIT OR Apache-2.0"`, so the `gcspy` wheel stops shipping with no licence metadata at all.

Attribution is unchanged: both files still say Ciaran McCreesh, and I have not added an "and contributors" that was not there before.

Checks: the generated `cmake_install.cmake` puts the three files in `share/doc/Glasgow-Constraint-Solver`; the `pyproject.toml` change parses to `License-Expression: MIT OR Apache-2.0` under Metadata-Version 2.4 with the current scikit-build-core (1.0.3), which is what the unpinned build requirement resolves to. No C++ changed, so clang-format has nothing to say.

Two things deliberately not done, both easy to add if wanted:

- No `SPDX-License-Identifier` headers in the source files. The tree has never had per-file headers, and adding them would touch ~950 files; the two dual-licensing conventions (per-file SPDX tags vs. root licence files) are both fine, and this follows the one the repository already implies.
- The wheel declares the licence but does not embed the licence *texts*, because `pyproject.toml` lives in `python/` and PEP 639's `license-files` globs may not escape the project directory. Fixing it properly means either moving `pyproject.toml` to the root or duplicating the licence files under `python/`, neither of which belongs in this PR. The sdist is unaffected, since it is built from the repository root.

---

A second commit adds the Advanced Research + Invention Agency to the README's Acknowledgements, alongside the Royal Academy of Engineering fellowship. It rides along because it is the same funding whose conditions prompt the relicence, but it is a separate commit and can be split out if you would rather keep this PR to licensing alone.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -380,6 +380,14 @@ install(FILES
         ${CMAKE_CURRENT_BINARY_DIR}/GlasgowConstraintSolverConfigVersion.cmake
         DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/GlasgowConstraintSolver)
 
+# We are dual licensed under Apache-2.0 or MIT, and section 4(a) of the former
+# asks that redistributions carry a copy of it, so install both texts.
+install(FILES
+        ${CMAKE_CURRENT_SOURCE_DIR}/COPYRIGHT
+        ${CMAKE_CURRENT_SOURCE_DIR}/LICENCE-APACHE
+        ${CMAKE_CURRENT_SOURCE_DIR}/LICENCE-MIT
+        DESTINATION ${CMAKE_INSTALL_DOCDIR})
+
 if(GCS_ENABLE_EXECUTABLES)
     add_subdirectory(examples)
     add_subdirectory(benchmarks)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,20 @@ All contributions should pass both the `release` and `sanitize` build and tests
 before submission, including the full test suite (with VeriPB installed) in
 both modes. See `README.md` for details.
 
+Licensing
+=========
+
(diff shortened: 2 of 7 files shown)


ciaranm:
@mmcilree I think technically this probably needs your explicit approval. It's to satisfy ARIA terms.

mmcilree (approved):
I hereby give my official approval for this

Issue #487, opened 2026-07-10
cake validation: In over a variable list uses a never-conformed flag encoding


ciaranm:
One of the three residual buckets from the cake validation sweeps (first characterised in #478's narrowing-down pass; split out here now that the small-domain work has landed). This is the largest remaining bucket: **34 FAIL_ELAB across the latest full sweep** (3 seed-repeats, with all of #478–#485 fixed), all from `in_test` with a variable value-list.

## Mechanism (established)

`In` over a variable list was never cake-conformed — the two encodings share no scaffolding:

**Ours** (`gcs/constraints/in/in.cc define_proof_model`): per list position, three fully-reified flags — `f[N][inlt<k>]` ⇔ var < V_k, `f[N][ingt<k>]` ⇔ var > V_k, and `f[N][in<k>]` ⇔ ¬lt ∧ ¬gt — plus an **unlabelled** at-least-one row over the selectors (mirroring Count's encoding).

**cake**: non-strict comparisons `@x[label][k][ge]/[le]` reified pairs, `@x[label][k][eq]` ⇔ ge ∧ le, and a **labelled** `@c[label][al1]` at-least-one over the eq flags:

```
@x[_1][0][ge][f] -1 i[_2][b0] ... 8 x[_1][0][ge] >= 1 ;
@x[_1][0][eq][f] 1 x[_1][0][eq] -1 x[_1][0][ge] -1 x[_1][0][le] >= -1 ;
@c[_1][al1] 1 x[_1][0][eq] 1 x[_1][1][eq] >= 1 ;
```

Our proofs cite the `f[N][...]` flags (e.g. `f[2][in0]`, `f[5][in1]` in the kept triples), which are free variables in cake's database, so elaboration dies at the first RUP step that leans on them.

**Conditional failure**: constant-only value lists pass (no flags needed), and variable lists pass when the proof happens never to lean on the flags — roughly 40/53 instances OK per seed.

## Minimal failing instance

```
(
    (
        (_1 5 6)
        (_2 3 5)
        (_3 2 3)
    )
    (
        (_1 in (_2 _3) _1)
    )
    (enumerate)
)
```

fails elaboration with "The constraint is not implied by reverse unit propagation (RUP)".

## Reproducing

[...]

PR #716, opened 2026-08-10
Read back every constraint the .scp writer can write, and keep it that way

ciaranm:
Falls out of an audit of what `cake_pb_cp` cannot encode. Everything here is
solver-side; nothing needs an upstream change.

## The gap

`Constraint::s_expr()` could write eight keywords `gcs::read_scp` had no case
for: `binpacking`, `cumulative_optional`, `difference`, `mdd`, `min_distance`,
`nogoods`, `at_most_one_smart_table` and `lex_smart_table`. Because the
workflow-2 chain harness re-solves the `.scp` as its first step, each of these
failed the chain *before* `cake_pb_cp` was reached — so the gap read as a solver
bug rather than the coverage hole it was, and no amount of upstream work would
have helped.

Three of the keywords were not even names anything accepts. `RegularLegacy` and
`RegularBacchus` wrote `regular_legacy` / `regular_bacchus`, and `LexSmartTable`
wrote a bare `lex`. The `.scp` describes the *constraint*, not which propagator
enforced it, so the two Regular variants now write `regular` and LexSmartTable
writes cake's spelling. RegularLegacy's OPB is byte-identical to Regular's, so
its proofs now chain-verify for free; RegularBacchus's upfront encoding labels
rows cake has no counterpart for, which is a separate, real conform.

Separately, *any* `.scp` from a model that did not name its variables was
unreadable: an anonymous variable is written `_N`, which is exactly the spelling
`Problem::check_name()` reserves and rejects. The reader now recreates such a
variable anonymously, mirroring `post_autonumbered`.

## A real writer bug on the way

`DifferenceConstraints::s_expr` rendered a half-reified edge's condition with
`s_expr_term_of(Literal)`, which renders a condition as the **bare variable
name**. `C = 1` and `D >= 2` both came out as just the variable, so the `.scp`
described a different problem than the one solved. Now written as the
`(var op value)` triple every other reified form uses.

## Keeping it that way

A hand audit is not enough — one done a fortnight before this branch missed
`difference` and `cumulative_optional` entirely, and the checks added here found
both on their first run. Two lanes:

- **Constraint tests, in process.** `verify_proof_and_dispose` reads the test's
  own `.scp` back and fails on `ScpUnsupportedConstraintError`, so a new
  constraint's own test reports the omission.
- **Examples, benchmarks, frontends.** `run_test_and_verify.bash` runs
  `glasgow_scp_solver --parse-only`, which exits 2 for an unknown keyword and 1
  for any other read failure; only 2 fails. The path comes from a
  CMake-generated `build/scp_solver_path`, so none of the ~200 `add_test`
  registrations needed changing. This lane earns its keep: `LexSmartTable` is
  only ever posted by an example, which is why its bad keyword survived.

Neither demands that every *instance* round-trip — a view operand renders as
`(-X + 17)`, a documented reader limitation rather than a coverage gap.
Constraints named `test_...` are exempt, for the ad-hoc fixtures in
`tabulation_test` and `product_justify_test`.

Both lanes were verified to fire by disabling a reader case and watching the
relevant test go red.

## Also

Five chain cases registered that already verified but had no test:
`gacglobalcardinalityclosed`, `disjunctive_strict`, `disjunctive2d_strict`,
`lin_equals_iff`, `lin_less_equal_iff`. The two strict cases put a zero-size
task inside a wider one, the only shape where strict and non-strict differ (7
solutions against 9; 67 against 81) — an all-positive-sizes case would exercise
the keyword while asserting nothing.

`dev_docs/workflow2_testing.md` gains the symmetry rule, a table of what
`cake_pb_cp` has no rule for at all (`binpacking`, `mdd`, `power`,
`min_distance`, `nogoods`) with which frontends reach each, and a note that `in`
over **two or more** variable members parses on both sides but fails at VeriPB
(#487; one variable chains, which is why `in_var_sat` passes). Part A's
three-way enum is marked superseded by the `cake_probe.hh` probe that landed
instead.

## Testing

`ctest` 648/648 green, GCC and clang 21. The chain harness is 165/165 with
`cake_pb_cp` on PATH.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -386,6 +386,16 @@ if(GCS_ENABLE_EXECUTABLES)
     add_subdirectory(verified_encodings)
     add_subdirectory(scp)
     add_subdirectory(minicp_benchmarks)
+
+    # run_test_and_verify.bash reads a written .scp back with glasgow_scp_solver
+    # --parse-only, so an example that posts a constraint gcs::read_scp cannot
+    # rebuild fails there rather than much later in the chain harness. The path
+    # goes in a generated file beside the binaries rather than an ENVIRONMENT
+    # property on each test: the script already knows the build directory (it is
+    # the directory of the program it was handed), so this needs no change to
+    # any of the ~200 add_test registrations.
+    file(GENERATE OUTPUT ${CMAKE_BINARY_DIR}/scp_solver_path
+        CONTENT "$<TARGET_FILE:glasgow_scp_solver>")
 endif()
 
 if(GCS_ENABLE_DOXYGEN)
--- a/dev_docs/constraints.md
+++ b/dev_docs/constraints.md
@@ -212,6 +212,16 @@ strategies). `s_expr` returns the constraint's `.scp` entry as a structured
 s-expression term `(name op args...)` (built with `innards::SExpr::list`/`atom`
 and `NamesAndIDsTracker::s_expr_term_of`); `innards::write_scp` serialises it.
 
+Whatever keyword `constraint_type` returns, **`gcs::read_scp` must have a case
+for it**: the writer and the reader are meant to be inverses, and the
+workflow-2 chain harness re-solves the `.scp` as its first step, so a keyword
+the reader does not know fails the chain before `cake_pb_cp` is reached. Your
+constraint's own test enforces this — `verify_proof_and_dispose` reads the
+`.scp` back and fails on an unknown keyword. See
+`dev_docs/workflow2_testing.md`. The keyword names the *constraint*, not the
+propagator: an alternative propagator for an existing constraint writes the
+existing keyword (all three `Regular` variants write `regular`).
+
 ## The three install phases
 
 `Constraint::install` is not virtual and is not yours to write. It is
@@ -901,7 +911,11 @@ that is a finding about the honest derivation.
 4. Add the `.cc` to the library and the test target to
    `gcs/CMakeLists.txt` (alphabetically), plus an `add_test` entry.
 5. Add the header to `gcs/gcs.hh` (alphabetically).
-6. Decide explicitly how the constraint behaves when the caller passes
+6. Add a `read_scp` case for the keyword `constraint_type` returns, in
+   `gcs/scp_reader.cc`, plus an `enumerate` and a write -> read -> write
+   test in `gcs/scp_reader_test.cc`. The writer and the reader are
+   inverses; the constraint tests fail if the keyword has no case.
+7. Decide explicitly how the constraint behaves when the caller passes
    the same `IntegerVariableID` (or a view that resolves to the same
    underlying variable) in two argument slots, or twice within an
    array argument. Pick one of three patterns and cover it with a test
--- a/dev_docs/workflow2_testing.md
+++ b/dev_docs/workflow2_testing.md
@@ -46,6 +46,84 @@ for the format lives outside this repo (cake's `SEXP_FORMAT.md`); the reader is
 the authority for what we accept, and `cake_pb_cp` for what the verified encoder
 accepts.
 
+## Writer/reader symmetry is a requirement, not a nicety
+
+`gcs::read_scp` must have a case for **every** keyword `Constraint::s_expr()`
+can write. The chain runner's first step re-solves the `.scp` with
+`glasgow_scp_solver`, so a keyword the reader has never heard of fails the chain
+before `cake_pb_cp` is even invoked — and the failure looks like a solver bug,
+not a coverage gap. This is not automatic: `s_expr()` is a per-constraint
+override, so a new constraint can introduce a keyword and nothing notices.
+
+Two checks enforce it, between them covering everything that writes a `.scp`:
+
+- **Constraint tests, in process.** `verify_proof_and_dispose` calls
+  `test_innards::check_scp_writer_reader_symmetry`, which reads the test's own
+  `.scp` back and fails if a keyword has no case
+  (`ScpUnsupportedConstraintError`). A new constraint's own test is what catches
+  the omission.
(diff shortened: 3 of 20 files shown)


ciaranm:
Merged. Four things from the merge checklist, none of them blocking and none a defect — recording them here rather than filing them away.

### 1. `cumulative_optional` has no round-trip test, against this PR's own new rule

`dev_docs/constraints.md` L914-917 now says a new keyword gets "a `read_scp` case … **plus an `enumerate` and a write -> read -> write test** in `gcs/scp_reader_test.cc`". Of the keywords this PR adds, `binpacking`, `difference`, `mdd`, `min_distance` and `nogoods` all got one; `cumulative_optional` (`scp_reader.cc:1177`) did not.

It is the one where that matters most, because it is four same-typed variable lists in a row — starts, lengths, heights, presences — so a re-ordering would still parse, still post, and still pass both of the new symmetry lanes, which only ask "does the keyword resolve".

I checked it by hand and **the ordering is correct**, so this is a missing regression test rather than a bug. The counts, if they are useful as a test:

```
( (version 1) (variables (S0 0 3) (S1 0 3) (L0 2 2) (L1 2 2) (H0 1 1) (H1 1 1) (P0 1 1) (P1 1 1) (C 1 1))
  (constraints (_1 cumulative_optional (S0 S1) (L0 L1) (H0 H1) (P0 P1) C)) (prob_type enumerate) )
```

- as written: **6** solutions (two length-2 tasks, height 1, capacity 1, so `|S0 - S1| >= 2`)
- heights 2 instead of 1, i.e. above capacity: **0** — this is what distinguishes the heights slot from the lengths slot
- `P0` fixed to 0 instead of 1: **16**, i.e. all of `S0 x S1`, which is what pins the fourth list as presences rather than anything else

### 2. `binpacking`'s error message names the wrong argument

`scp_reader.cc:706` says the tag is "binpacking's fourth argument". It is `terms[4]`, which is the *third* argument after the keyword — and the doc comment at L689 gets it right, calling it "the tagged fifth term". Looks like a 0-vs-1-indexed slip when the message was written; only the user-facing string is affected.

### 3. The `cake_pb_cp` gap table under-reports `Power`'s reach

`dev_docs/workflow2_testing.md:117` gives `Power` / `PowerTable` as reachable from "MiniZinc, XCSP". Python posts it too — `python/gcspy.cc:299`, `p.post(Power(var1, var2, result))`. The other two multi-frontend rows (`BinPacking`, `MDD`) really are MiniZinc/XCSP only, so it is just this row.

### 4. `check_scp_writer_reader_symmetry` is broader than its comment claims

`constraints_test_utils.hh:170` says "Deliberately narrow: only `ScpUnsupportedConstraintError` fails." The function catches `ScpUnsupportedConstraintError` (L193), `ScpReadError` (L198) and `SExprParseError` (L202) — so anything `read_scp` throws outside those two hierarchies propagates and fails a test that previously passed.

That is not hypothetical: the anonymous-variable case this PR fixes was exactly such an exception, since `_N` hit `Problem::check_name()`'s rejection rather than any `ScpReadError`. Every proving constraint test now runs this, so the blast radius of the next one is wide. Whether the fix is a broader catch or just an honest comment is your call — flagging it because the comment currently reads as a guarantee.

---

For context on what CI does and does not cover here: no lane runs the chain cases (they `SKIP_RETURN_CODE 77` without `cake_pb_cp`), so I parsed all five new fixtures and confirmed the counts in the new comments — `disjunctive_strict` 7 against 9, `disjunctive2d_strict` 67 against 81, both exact. I also confirmed the `--parse-only` contract the harness depends on: exit 2 for an unknown keyword, 1 for a known keyword in an unrebuildable shape, and 2 for the now-retired `regular_legacy` spelling. GCC 15.2 release build of the trial merge was 0 warnings.

ciaranm:
All four addressed in #717 (`1a5979dc`), which retargeted to `main` when this merged.

**1. `cumulative_optional` round-trip test** — added, using your three counts. I re-ran them against the solver first: 6 / 0 / 16, all three confirmed. You're right that it's the keyword where it matters most, and the reason is worth having in the test rather than only in this thread, so the three cases are commented as pinning each slot to its *meaning* rather than its type — heights-over-capacity separates heights from lengths, one-task-absent separates presences from the rest. A plain SAT case would have caught neither.

**2. `binpacking`'s "fourth argument"** — fixed to "third". Your reading is right: it's `terms[4]`, the third argument after the keyword, and the doc comment above already said "tagged fifth term" correctly.

**3. `Power`'s reach** — confirmed, `python/gcspy.cc:299` does post it. No change needed though: `fc61458b` in #717 deleted that table rather than correcting it, precisely because of what this item demonstrates. Four enumerated versions of "what cake can't encode" were live simultaneously and at least two were wrong; the section now says how to ask the tool (`cake_pb_cp` prints `unsupported constraint`, and the registered chain cases are by construction the set that works) instead of naming names. Your finding is a fifth data point for that, so thanks — it went into the reasoning rather than into a fix.

**4. The guard's comment vs its catch** — you're right that it read as a guarantee, and the blast radius argument is the important half. It now catches `std::exception` and explains why that's deliberate: re-posting into a throwaway `Problem` is an extra probe beside the solve the test already did, so anything it throws that isn't a missing keyword is a limitation of the probe, not a finding about the constraint under test — and anything genuinely wrong with posting would already have failed the real solve. As you note, the anonymous `_N` case was exactly such an exception, and a `NamingError` rather than any `ScpReadError`.

One thing I checked rather than assumed while making that change: an `UnexpectedException` thrown from inside the `ScpUnsupportedConstraintError` handler is *not* caught by the new sibling `std::exception` handler, so the guard still fires. Verified by disabling a reader case and watching `circuit_constraint` go red, not by reading the standard.

`ctest` 652/652 after all four.

PR #717, opened 2026-08-10
Conform In's selectors to cake_pb_cp's, so `in` chains over variables

### LINKED ISSUE #487: cake validation: In over a variable list uses a never-conformed flag encoding

ciaranm:
Stacked on #716. Closes #487.

## The gap

`in` over a candidate list with **two or more variables** parsed on both sides
but failed at VeriPB: the proof cited `f[N][in<i>]` flags cake's OPB has no
counterpart for. One variable happened to chain, which is why `in_var_sat`
passed and this went unnoticed.

## Two halves

**The encoding — cheaper than it looked.** cake's `cencode_in` *is*
`cencode_count_aux`, the helper `count` uses, so this is the conform count
already had in #354: flags become `x[id][i][ge|le|eq]` with `eq <=> ge & le` via
the position-indexed `create_proof_flag_fully_reifying` overload, and the
disjunction is labelled `@c[id][al1]`. The solver reified the strict
complements, and `ge = ~gt`, `le = ~lt` exactly — so the selector means what it
always did, which is all the propagator ever references. **The propagator is
untouched.**

**The partition — the part that needed a decision.** `In::prepare()` folded a
candidate whose *domain* was a singleton into the constant list, but `s_expr()`
runs on the stored constraint, before `prepare()`, so the `.scp` still named it
as a variable and cake still gave it a flag triple — shifting every later
candidate's index. `prepare()` now folds only *syntactically* constant
candidates, which the `.scp` renders as integers, so both sides agree.
`in_fixed_var_sat` is the regression test.

The general rule, now written down in `workflow2_testing.md`: `prepare()` may
not rewrite anything the `.scp` distinguishes.

## GlobalCardinality fallout

Its closed restriction is one child `In` per variable, so labelling the
disjunction made all *n* children collide on `@c[id][al1]` — which `ProofModel`
rightly rejects. Each child now takes the parent's identity and a `<i>_` role
prefix, landing on `@c[id][<i>_al1]`. That is cake's own label for the row
(`cencode_global_cardinality_aux`), so the closed cases' whole label set now
agrees with cake's; they stay `none` only for the #358 `@i[X][ge0][r]`
coefficient.

## Testing

`ctest` 652/652, GCC and clang 21. Nine `in` shapes chain-verified by hand (two
/ three variables, mixed, one variable, constants only, unsat, singleton-domain
variable, duplicate constants, duplicate variables). `in_const_sat` moves to
**strict**, joined by `in_two_vars_sat`/`_unsat`, `in_mixed_sat` and
`in_fixed_var_sat`. `in_var_sat` and `in_unsat` stay `none` for #358, which is
all that still differs.

Known cosmetic residue: duplicate *constant* candidates (`(in (2 2 B) A)`) chain
but do not byte-match — we dedupe the value list, cake does not. Degenerate
input, left alone.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,11 @@ Doxyfile
 *.pbp
 *.varmap
 *.scp
+# run_scp_chain.bash's intermediates: cake's re-derived OPB and veripb's
+# elaborated core. The script deletes both on success, so a tracked one is a
+# leftover from a hand run -- which is how six of them reached the root once.
+*.verifiedopb
+*.corepb
 # ... except the curated workflow-2 regression cases, which are source, not generated:
 !verified_encodings/scp_cases/*.scp
 # ... and the committed VeriPB smoke fixture used by the Windows CI lane:
--- a/dev_docs/workflow2_testing.md
+++ b/dev_docs/workflow2_testing.md
@@ -106,21 +106,21 @@ Two related rules that fall out of the same principle — the `.scp` describes t
 
 ## What `cake_pb_cp` does not encode
 
-Constraints whose keyword the verified encoder has no rule for, so they cannot
-chain at all no matter what the solver does. The solver still writes and reads
-them; the gap is upstream.
+Some constraints have no rule in the verified encoder at all, so they cannot
+chain no matter what the solver does. **This document does not list them**, on
+purpose: every enumerated version of that list in this repo went stale and was
+then believed. Ask the tool instead —
 
-| Constraint | keyword | reachable from |
-|---|---|---|
-| `BinPacking` | `binpacking` | MiniZinc, XCSP |
-| `MDD` | `mdd` | MiniZinc, XCSP |
-| `Power` / `PowerTable` | `power` | MiniZinc, XCSP |
-| `MinDistance` | `min_distance` | library only |
-| `Nogoods` (posted) | `nogoods` | library only |
+```
+cake_pb_cp <model>.scp        # `unsupported constraint: <keyword>` if it has no rule
+```
 
-Also upstream: `notin` (no solver writer) and views (affine operands).
+— or `ctest -R scp_chain`, whose registered cases are by construction the ones
+that do chain. A constraint with no case in `scp_cases/` is either an upstream
+gap or an unwritten test, and running the chain over a hand-written `.scp` tells
+you which in one command.
 
-`in` used to be on that list — over a member list with two or more variables it
+`in` used to have no rule — over a member list with two or more variables it
 parsed on both sides but failed at VeriPB, because the solver's
 `f[N][inlt/ingt/in]` flags were not cake's `x[id][i][ge/le/eq]` selectors
 (#487). It is conformed now, and byte-matches. Two things that made it cheap,
--- a/dev_docs/workflow2_testing.md
+++ b/dev_docs/workflow2_testing.md
@@ -118,11 +118,33 @@ them; the gap is upstream.
 | `MinDistance` | `min_distance` | library only |
 | `Nogoods` (posted) | `nogoods` | library only |
 
-Also upstream: `notin` (no solver writer), views (affine operands), and `in`
-over a member list with **two or more variables** — that last one parses on both
-sides but fails at VeriPB, because the solver's `f[N][inlt/ingt/in]` flags are
-not cake's `x[id][i][ge/le/eq]` selectors (#487; one variable chains fine, which
-is why `in_var_sat` passes).
+Also upstream: `notin` (no solver writer) and views (affine operands).
+
+`in` used to be on that list — over a member list with two or more variables it
+parsed on both sides but failed at VeriPB, because the solver's
+`f[N][inlt/ingt/in]` flags were not cake's `x[id][i][ge/le/eq]` selectors
+(#487). It is conformed now, and byte-matches. Two things that made it cheap,
+both worth reaching for next time:
+
+- **cake's `cencode_in` is literally `cencode_count_aux`**, the helper `count`
+  uses, so the conform was the one already done for count in #354: the
+  position-indexed `create_proof_flag_fully_reifying(ConstraintID, indices,
+  annotation, ineq)` overload, `eq <=> ge & le` in place of `eq <=> ~lt & ~gt`,
+  and the disjunction labelled `@c[id][al1]`.
+- **The propagator needed no changes at all.** It only ever references the
+  equality flag, whose meaning is unchanged; `ge = ~gt` and `le = ~lt` exactly.
+  Check that property first — it is what separates a conform that is a rename
+  from one that is a rewrite of every justification.
+
+The other half was not a naming question. `In::prepare()` folded a candidate
+whose *domain* was a singleton into the constant list, but `s_expr()` runs on
+the stored constraint, before `prepare()`, so the `.scp` still named it as a
+variable and cake still gave it a flag triple — shifting every later
+candidate's index. `prepare()` now folds only *syntactically* constant
+candidates, which the `.scp` renders as integers, so both sides agree.
+`in_fixed_var_sat` is the regression test. Any constraint that normalises its
+arguments in `prepare()` has this hazard; the general rule is that `prepare()`
+may not rewrite anything the `.scp` distinguishes.
 
 Two complementary mechanisms:
 
--- a/dev_docs/workflow2_testing.md
+++ b/dev_docs/workflow2_testing.md
@@ -184,11 +184,17 @@ Per-case `opbdiff` mode (3rd runner arg):
   to cake's and whose multiple identical-looking selectors defeat label matching
   (the deferred conform-the-solver item).
 
-**Domains are bits-encoded on purpose**, to dodge the direct-only-vs-bits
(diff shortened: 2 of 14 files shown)


ciaranm:
Merge checklist over `1a5979dc`. **One blocker**, one stale comment, one tidy.

First, the good news: the `1a5979dc` follow-ups to the #716 review all check out. I rebuilt and ran them — `scp_reader_test` is 73 cases / 1094 assertions, up from 71/1083. The `cumulative_optional` test goes further than the review asked, pinning the slot order literally as well as by solution count. And I compiled a minimal case for the load-bearing claim in that commit message, since a mistake there would silently disable the guard rather than fail loudly: the `throw UnexpectedException` inside the `ScpUnsupportedConstraintError` handler does escape the try block and is *not* caught by the sibling `catch (const std::exception &)`. Confirmed, the guard still fires.

---

### 1. Six generated chain artifacts committed at the repository root — blocking

Added by `253a7f48`:

```
global_cardinality_closed_sat.chain.corepb          12192 bytes
global_cardinality_closed_sat.chain.verifiedopb      2714
global_cardinality_closed_unsat.chain.corepb         2325
global_cardinality_closed_unsat.chain.verifiedopb    2714
global_cardinality_gac_closed_sat.chain.corepb      38949
global_cardinality_gac_closed_sat.chain.verifiedopb  3721
```

These are intermediates of `verified_encodings/run_scp_chain.bash` — `.verifiedopb` is cake's re-derived OPB (step 2), `.corepb` is veripb's elaborated core (step 3) — and the script deletes both itself on success (`dispose_proof "${base}" verifiedopb corepb`, L107). They look like leftovers from the hand chain-verification the PR describes.

They got past `.gitignore` because it covers `*.opb` (L11), `*.veripb` (L12), `*.pbp` (L13) and `*.scp` (L15), with deliberate `!` exceptions for `verified_encodings/scp_cases/*.scp` and `.github/veripb_smoke/*.opb|pbp` — but nothing matches `*.corepb` or `*.verifiedopb`. There's no precedent for tracking them: those two smoke fixtures are the only intentionally-committed proof artifacts in the tree.

They are ASCII and ~62 KB total, so this is not the permanent-bloat case that makes a stray *binary* unmergeable — but they are generated files at the top level that would sit there looking like fixtures indefinitely.

Suggested fix: drop the six, and add `*.corepb` / `*.verifiedopb` next to the existing patterns so it cannot recur.

### 2. `verified_encodings/scp_cases/CMakeLists.txt:315-321` describes the design `253a7f48` replaced

```
# The closed restriction is one child In per variable. Each now takes this
# constraint's identity and a `<i>_` role prefix, so its disjunction lands on
# @c[id][<i>_al1] -- cake's own label for that row, from
# cencode_global_cardinality_aux. (Before the In conform the row was unlabelled;
# labelling it without the prefix made all n children collide, which ProofModel
# rejects.)
```

That paragraph was written by `c347b153`, when the mechanism really was a child `In` per variable. `253a7f48` then removed child `In` entirely — `GlobalCardinality::define_proof_model` emits the rows directly, and the 18 lines that commit had added to `in.hh` to support the parent-identity/role-prefix route were removed again (`in.hh` is net-zero against main). `git grep 'In{' gcs/constraints/global_cardinality/` now returns nothing.

The `@c[id][<i>_al1]` label survives, so the paragraph's *conclusion* is still true — it is the derivation that is describing a design that no longer exists, including a collision hazard ("all n children collide") that can no longer arise. This is the shape Check 14 warns about: the rewrite landed and the prose it invalidated stayed, and each half reads authoritatively on its own.

The PR description's "GlobalCardinality fallout" section has the same staleness, if you are keeping it as the merge commit message.

### 3. Minor: `GlobalCardinality::prepare` is now vestigial

After `253a7f48` emptied it, `global_cardinality.cc:78` is an exact duplicate of the default at `gcs/constraint.hh:75-78`. The override and its declaration at `global_cardinality.hh:54` could both go.

---

For what it is worth, the rest is clean: all five commits carry the trailer, the branch sits on `762e7d68` (exactly what #716 landed), the trial merge onto current main builds with **0 warnings** under GCC 15.2, and the four new `in_*` fixtures are all present and registered with `in_const_sat` promoted to `strict`. I checked the new closed propagator against Check 11 too — it is genuinely exercised, since `bounds_global_cardinality_test.cc:85` and `gac_global_cardinality_test.cc:86` both post `.with_closed(closed)` with `closed` varied in the data table under `for (bool proofs : {false, true})`, so the new `infer_not_in_range` justification is proof-verified in CI rather than only by the chain cases that skip without `cake_pb_cp`. The doc claims verify as well: `9678233f` is exactly as cited, and `binary_less_equal_sat` / `binary_equals_unsat` are registered at `none` as stated.

ciaranm:
All three fixed in `38fd798e`. `ctest` 652/652.

**1. The six artifacts — my fault twice over.** Removed, and `.gitignore` now has `*.verifiedopb` / `*.corepb` next to the existing patterns. Your diagnosis is exactly right: they are `run_scp_chain.bash` intermediates from the hand chain-verification, swept up by a careless `git add -A`.

What is worth recording is that this is the *second* time in this stack. `bf919593` picked up the same class of file and I caught it by reading the diff before pushing; `253a7f48` did it again and I did not. A mistake I have already made once and can still repeat is a mistake the tooling should be catching, which is why the fix is the ignore patterns rather than just the deletion.

**2. The stale comment.** Rewritten. This one is the more interesting of the three, because of exactly what you point out: the paragraph's *conclusion* — the row lands on `@c[id][<i>_al1]`, cake's own label — is still true, so nothing about reading it feels wrong. It is the derivation underneath that stopped existing: there is no child `In`, no parent identity to pass down, and no n-way collision for a role prefix to avoid. A paragraph whose conclusion survives its premise is the version that gets believed, and I wrote it three commits before I removed the thing it describes.

**3. `prepare`.** Gone, override and declaration both.

---

On the checklist itself: thank you for compiling the `throw`-from-a-handler case rather than taking the commit message's word for it. I had verified it the same way — disabling a reader case and watching `circuit_constraint` go red — but the claim as written was a bare assertion about C++ semantics, and it is load-bearing in the direction where being wrong is silent. It should have carried the evidence rather than the conclusion.

Two things from your notes I am leaving alone deliberately, flagging in case you would rather I did not:

- **No CI lane runs the chain cases** — they `SKIP_RETURN_CODE 77` without `cake_pb_cp`, so all 169 are local-only. That is a real coverage hole, but it is a CI/toolchain question rather than something to fix in this stack.
- **`#487`** is `Closes #487` in this PR's first commit. Since the branch retargeted to `main` after #716 merged that may well fire correctly now, but it did not when this was stacked, so it is worth a glance after merge.

PR #720, opened 2026-08-10
Report every component block through %%%mzn-stat, generically

ciaranm:
Stacked on #712, which is where `ComponentStats` comes from. Follow-up to the
`%%%mzn-stat` gap raised in review there.

## The gap

`fzn-glasgow` listed the difference-logic counters by hand, one `println` per
field. That is how two of `DifferenceLogicStats`' twelve fields —
`propagator_installed` and `donor_propagators_disabled` — came to be filled in
and reported to nobody. Nothing was going to notice: they are counters on a
presolver that is invisible to solution equivalence and to proof verification
alike, which is the reason the block exists in the first place.

`ComponentStats::entries()` is the flat name/value view that removes the need
for a list. So the frontend now walks `Stats::components()` and knows about no
component in particular:

```cpp
for (const auto & component : stats.components())
    for (const auto & entry : component->entries())
        println(cout, "%%%mzn-stat: {}={}", mzn_stat_name(component->component_name(), entry.name), entry.value);
```

Fourteen hand-written lines out, three in. The report gains thirteen statistics
it was not carrying, and gains any future block by that block existing.

## Every existing name is preserved

This is the compatibility surface, and it is the whole reason `mzn_stat_name`
exists rather than the entry names being emitted raw. `component_name()` and the
entry name are joined and camel-cased, so `difference_logic` and `edges_lifted`
still make `differenceLogicEdgesLifted`, exactly as the hand-written line did.
All thirteen previously-emitted names come out unchanged.

The translation lives in the frontend rather than in the blocks. MiniZinc's own
statistics are camelCase (`peakDepth`, `solveTime`) and a solver's own are
free-form, so snake_case would be legal and still wrong to read next to them —
but a presolver in `gcs/presolvers/` should not have an opinion about a MiniZinc
spelling, and asking blocks to write their field names twice is how the two
copies drift.

One consequence worth stating plainly, since it is the one judgement call here:
**`DifferenceSimplificationStats::component_name()` is `difference_logic_simplify`,
not `difference_simplify` after its header.** The ComponentStats convention says
to name a block after the header it lives in. But `differenceLogicSimplifyRan` is
already pinned in `minizinc/CMakeLists.txt` and already public, and a stage that
reports under two different names across a version is worse than one whose
identifier does not match its filename. Overrule me if you disagree — it is one
string and one comment.

## Registration

`DifferenceLogicStats` registers at the top of `DifferenceLogic::run`, before
anything is decided, so a run that lifts nothing still reports having looked.
The counters are assigned into that same object at the end, so a caller's handle
and the registered block stay the same object — the identity trap #712's own
tests were extended for.

The simplification block registers inside `install_difference_propagator`,
because that is where both routes to a difference propagator meet: it reports
whether it was reached through the presolver or through posted
`DifferenceConstraints`.

Both blocks are now always allocated rather than only when a caller asks, as
`CumulativeStrengthening` already does it.

## Testing

Name preservation is what needed covering, and most of it already existed:
`minizinc-differencelogic` pins six stat names as exact line regexes, and it goes
through the real MiniZinc round trip. **Verified by mutation** — making
`mzn_stat_name` stop capitalising fails that test rather than silently renaming
every statistic the frontend emits.

Added on top of that:

- `differenceLogicPropagatorInstalled=1` joins the pinned patterns, so one of the
  two fields that previously reached nobody stays reachable.
- `run_report_tests` in `difference_logic_test.cc` asserts both blocks'
  `component_name()` and their entry names as an **ordered list**, so a rename
  fails as a rename rather than as a count.
- Each block's `sizeof` is pinned, which is the anti-rot half.
  `CumulativeStrengtheningStats` can count its fields by dividing `sizeof` by
  eight, because every field there is eight bytes wide; these two mix bools,
  sizes and a double, so the size is pinned directly instead, with the failure
  message saying to add the new field to `entries()` rather than just update the
  number. Skipped where `size_t` is not eight bytes.

`DifferenceLogicStats` stops being an aggregate, so the seven fixtures that
designated-initialised it use a test-local `Expected` mirroring the ten fields
the assertion actually reads. Mirroring only those also means a new field is not
silently defaulted into every fixture's expectations.

`ctest` 635/635 on GCC 15.2, 0 warnings.

## Not done

`xcsp_glasgow_constraint_solver.cc:2086` has the same hand-written list in the
XCSP idiom (`d DIFFERENCE LOGIC EDGES LIFTED ...`), and the same two fields are
missing from it. Left alone deliberately: that is a different output convention
with its own competition rules, and this PR is the MiniZinc follow-up. It should
get the same treatment, and the loop here is the shape to copy.

### DIFF
--- a/gcs/constraints/difference/difference_graph.cc
+++ b/gcs/constraints/difference/difference_graph.cc
@@ -1117,6 +1117,13 @@ auto gcs::innards::install_difference_propagator(Propagators & propagators, Stat
     if (graph.edges.empty() && graph.static_bounds.empty() && graph.disallowed_conditions.empty())
         return;
 
+    // Both routes to a difference propagator come through here --- the posted
+    // DifferenceConstraints and the presolver that builds a graph out of donors
+    // --- so registering the simplification block here is what makes it reach a
+    // report either way. A null block is a caller that asked for none, and
+    // add_component_stats ignores it.
+    propagators.add_component_stats(simplify.stats);
+
     incremental.audit = incremental.audit || difference_audit_from_environment();
 
     Triggers triggers;
--- a/gcs/constraints/difference/difference_simplify.cc
+++ b/gcs/constraints/difference/difference_simplify.cc
@@ -4,6 +4,7 @@
 #include <algorithm>
 #include <optional>
 #include <queue>
+#include <string>
 #include <utility>
 #include <vector>
 
@@ -18,6 +19,8 @@ using std::optional;
 using std::pair;
 using std::priority_queue;
 using std::size_t;
+using std::string;
+using std::to_string;
 using std::vector;
 using std::ranges::reverse;
 
@@ -265,4 +268,36 @@ auto gcs::innards::simplify_difference_graph(size_t n, const vector<DifferenceSi
     }
 
     return outcome;
+}
+
+auto DifferenceSimplificationStats::component_name() const -> string
+{
+    return "difference_logic_simplify";
+}
+
+auto DifferenceSimplificationStats::summary() const -> string
+{
+    if (! ran)
+        return "did not run";
+
+    if (base_negative_cycle)
+        return "stopped on a negative cycle in the base graph, leaving the refutation to the propagator";
+
+    return "removed " + to_string(redundant_edges_removed) + " redundant and " + to_string(dead_edges_removed) + " dead edges, fixed " +
+        to_string(conditions_fixed) + " conditions, over " + to_string(rounds) + " rounds of " + to_string(nodes) + " nodes and " + to_string(edges) +
+        " edges";
+}
+
+auto DifferenceSimplificationStats::entries() const -> vector<StatsEntry>
+{
+    return {StatsEntry{"ran", ran ? 1 : 0}, StatsEntry{"rounds", static_cast<long long>(rounds)}, StatsEntry{"nodes", static_cast<long long>(nodes)},
+        StatsEntry{"edges", static_cast<long long>(edges)}, StatsEntry{"conditional_edges", static_cast<long long>(conditional_edges)},
+        StatsEntry{"redundant_edges_removed", static_cast<long long>(redundant_edges_removed)},
+        StatsEntry{"redundant_conditional_edges_removed", static_cast<long long>(redundant_conditional_edges_removed)},
+        StatsEntry{"dead_edges_removed", static_cast<long long>(dead_edges_removed)},
+        StatsEntry{"conditions_fixed", static_cast<long long>(conditions_fixed)},
+        StatsEntry{"isolated_nodes_removed", static_cast<long long>(isolated_nodes_removed)},
+        StatsEntry{"zero_weight_cycles", static_cast<long long>(zero_weight_cycles)},
+        StatsEntry{"nodes_on_zero_weight_cycles", static_cast<long long>(nodes_on_zero_weight_cycles)},
+        StatsEntry{"base_negative_cycle", base_negative_cycle ? 1 : 0}, StatsEntry{"milliseconds", static_cast<long long>(seconds * 1000.0)}};
 }
--- a/gcs/constraints/difference/difference_simplify.hh
+++ b/gcs/constraints/difference/difference_simplify.hh
@@ -2,8 +2,10 @@
 #define GLASGOW_CONSTRAINT_SOLVER_GUARD_GCS_CONSTRAINTS_DIFFERENCE_DIFFERENCE_SIMPLIFY_HH
 
 #include <gcs/integer.hh>
+#include <gcs/stats.hh>
 
 #include <cstddef>
+#include <string>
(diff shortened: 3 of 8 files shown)


ciaranm:
Merge-checklist review. Verified against a trial merge of the stack tip onto current main (`da56a93d`, which now has #719's `-Werror` lane): clean merge, release build with **`-DGCS_WERROR=ON`, 0 errors and 0 warnings**, full `ctest` green.

The central claim — that every previously-emitted `%%%mzn-stat` name survives the move to a generic loop — is the one worth checking rather than reading, so I checked it by running both sides. Compiled `minizinc/tests/differencelogic.mzn` to FlatZinc once, ran `fzn-glasgow -s --difference-logic` under #712 and under this PR, and diffed the emitted names:

- **Names lost: none.** All 22 come through unchanged, including all 13 `differenceLogic*` ones.
- **Names gained: exactly 13**, matching "thirteen statistics it was not carrying":

```
differenceLogicDonorPropagatorsDisabled          differenceLogicSimplifyMilliseconds
differenceLogicPropagatorInstalled               differenceLogicSimplifyNodes
differenceLogicSimplifyBaseNegativeCycle         differenceLogicSimplifyNodesOnZeroWeightCycles
differenceLogicSimplifyConditionalEdges          differenceLogicSimplifyRedundantConditionalEdgesRemoved
differenceLogicSimplifyDeadEdgesRemoved          differenceLogicSimplifyRounds
differenceLogicSimplifyEdges                     differenceLogicSimplifyZeroWeightCycles
differenceLogicSimplifyIsolatedNodesRemoved
```

22 → 35, nothing renamed. So the compatibility surface is exactly as described.

## One finding: a comment got split in half, and both halves are now wrong

`run_report_tests()` was inserted into the *middle* of `run_comparison_label_tests()`'s comment block. The diff adds a single line at `difference_logic_test.cc:767`, and that line lands between the second and third lines of an existing four-line sentence. The result is two broken comments:

At L765-767, on top of the new function:

```cpp
// Every row ReifiedCompareLessThanOrMaybeEqual emits must carry an @label.
// The presolver's whole licence for lifting a comparison is that it can
// The flat view both blocks reach a report through.
```

The second line ends mid-sentence ("is that it can"), and the first two describe `@label`s on comparison rows, which has nothing to do with the flat view. Then at L838, the remainder is stranded on top of `run_comparison_label_tests()` with its subject gone:

```cpp
// cite the donor's row by name, and an unlabelled row cannot be cited at
// all -- a `pol` would name something the OPB never defines.
```

On #712 those four lines are one intact sentence. The fix is just to move L765-766 back down to rejoin L838, leaving `run_report_tests()`'s own comment to start at "The flat view both blocks reach a report through."

Purely cosmetic — no behavioural effect, and nothing in CI could see it — but it leaves two functions carrying each other's rationale, and the `@label` reasoning is load-bearing enough that losing its opening is a real loss.

## The naming judgement call: I would not overrule it

`DifferenceSimplificationStats::component_name()` being `difference_logic_simplify` rather than `difference_simplify` after its header is the right call. `differenceLogicSimplifyRan` is already public and already pinned in `minizinc/CMakeLists.txt`, and a stage that reports under two different names across a version is a worse problem than an identifier that does not match its filename. The comment at the site says so, which is what makes it a decision rather than an accident.

## Everything else passes

`Co-Authored-By` present; no `AssertProofRule` / `AssertRatherThanJustifying`; no `overloaded{}` site touched; no stray or generated files, and every changed path belongs to the change. On docs (Check 2) I checked the two places `dev_docs/difference-logic.md` describes these blocks — L915's account of what guards `DifferenceLogicStats`, and L1356's `reporting_simplification_to(...)` — and both are still accurate, since that entry point survives on both `DifferenceLogic` and `DifferenceConstraints` even though the blocks are now always allocated.

The `run_report_tests` anti-rot design is worth calling out as good: asserting entry names as an *ordered list* plus a `sizeof` pin, with a failure message that tells the reader to add the field to `entries()` rather than update the number, is precisely the shape that stops a field going quietly unreported — which is the bug this PR exists to fix, so the test defends against its own recurrence.

ciaranm:
Fixed in f8441d75. You're right on both counts — the insertion point and the fact that the `@label` reasoning is load-bearing enough that losing its opening clause matters.

Moved the two lines back to rejoin their sentence rather than rewriting either comment, so `run_comparison_label_tests()`'s block is byte-identical to #712's again and `run_report_tests()` starts at "The flat view both blocks reach a report through." `git diff stats-diagnostics-662` on that file now shows no removed lines other than the intentional `DifferenceLogicStats{` → `Expected{` retypes, which is the check that nothing pre-existing got reflowed on the way.

`ctest` 635/635, 0 warnings.

Thanks for running the name diff both ways rather than reading the loop — 22 → 35 with nothing renamed is the number I wanted to be able to state and could only assert from the six pinned patterns. And agreed on leaving `difference_logic_simplify` as it is.

Issue #708, opened 2026-08-07
Rerun the #672 RCPSP bounds artefact: every headline number in certified-makespan-bounds.md is stale


ciaranm:
Compute, not code. Filed so it does not go with #672 when that closes.

`dev_docs/certified-makespan-bounds.md` reports 92 of 110 instances bounded, 29
closed, and a 23-instance shortfall against Sidorov's `L`, plus per-instance
`.pbp` sizes and `veripb` times. **None of those numbers may be quoted**, and the
doc now says so at the point they are stated. They predate:

- **#678**, multi-resource lifting — expected to take 29 closed to roughly 33,
  and to close most of the shortfall, since the shortfall is what motivated
  lifting over every resource rather than one.
- **#681**, hinted replay — 6 to 14x on checking time *on exactly these
  certificates*, so every `verify_seconds` in `results.jsonl` is wrong by that
  much.
- **#697**, which changes the inference on three counts: the lifting order is by
  guaranteed duration rather than by variable identity, the candidate pairs the
  clique search grows are the best by bound rather than the first by scan order
  (forty of the 110 instances have enough conflicting pairs to notice), and a
  clique one posted capacity-one resource already contains is neither posted nor
  reported.

And the **Pack_d lifted stages were never run at all** — five of the largest
apparent shortfalls are that, not an inference gap.

`~/claude/tmp/rcpsp-bounds-672/run.bash` is the harness. Note
`NEVER rcpsp --dzn <a real Pack instance> --prove` without `ulimit -f`: it wrote
128 GB in ten minutes once. `(ulimit -f 100000; ...)` caps it at 100 MB, ample
for the presolve prefix.

The rerun is also what would answer #703 and #707, both of which are default
choices nothing has measured.

Issue #707, opened 2026-08-07
InferredDisjunctive: whether to post two-member cliques (min clique size 2) wants measuring


ciaranm:
From the whole-stack review of the #661–#694 cumulative stack, section B3. The
caveat is written into `dev_docs/inferred-disjunctive.md` by #699; this is the
decision it defers.

`_min_clique_size` defaults to 3, on the argument that a two-task "clique" is
just a conflicting pair and the resource witnessing it already rules that pair
out. That argument is right about *propagation* and incomplete about everything
else:

- Sidorov's binary covers are exactly two-task cliques, and his pipeline keeps
  them. Where a conflicting pair has no common neighbour, its
  `L = d_u + d_v` can top the ranking and be the constraint his reported `L`
  comes from. Ours would have nothing there, so this is a divergence from the
  procedure we claim to reproduce, not only a tuning choice.
- With `with_makespan`, a posted pair derives a makespan bound that the
  witnessing resource does not derive for itself (see #702). So even a
  propagation-redundant pair is not inference-redundant.

Against: `_max_posted` is 5, so a pair with a high `L` displaces a larger clique,
and pairs are numerous. Whether that trade is worth taking is a measurement, and
`with_minimum_clique_size(2)` already makes it measurable.

The measurement is the #672 artefact rerun, over both collections and all three
stages. It costs nothing on the twenty cross-check targets either way — their
cliques run to ten members and more.

ciaranm:
Measured by the #708 artefact rerun (2026-08-11, `main` at `83b31a5e`, VeriPB 3.0.2), and the answer is **the default of three stands**.

With `with_minimum_clique_size(2)`, over Pack in both disjunctive-bearing stages and Pack_d in the capacity-one stage — 165 runs — the posted set changes on **nine** of them: one clique becomes five members, and three- or four-member cliques become eleven or twelve. The certified bound changes on **none**.

Exactly one `L` improves, `pack_d/pack003`, from 7 to 210. Its critical path is 406, so even the improved value says nothing the model does not already imply, and it does not reach the bound.

So the concern in the issue body — that a conflicting pair with no common neighbour can carry the `L` his pipeline reports, and that `with_makespan` gives even a propagation-redundant pair some inference value — is real in principle and worth nothing on these collections. The `_max_posted = 5` displacement argument is what wins: pairs are numerous, and the larger cliques they displace are worth more.

[...]

PR #721, opened 2026-08-10
Reach with_minimum_clique_size from the rcpsp example

ciaranm:
Prerequisite for the #708 artefact rerun, so that it can also answer #707.

`InferredDisjunctive::with_minimum_clique_size` has existed since #663, but
nothing outside the unit tests could set it: the `rcpsp` example wires
`with_budgets` and not this, so the default of three is the only value the
presolver has ever been run at on a real instance. #707 asks whether two is
better, and says the measurement is the #672 artefact rerun over both
collections and all three stages. That sweep drives the solver through this
example, so the flag has to exist first.

Wiring only. The default is unchanged, so anything that does not pass
`--infer-disjunctive-min-clique-size` behaves exactly as before.

Rebased onto `0e7cf85a`, so it now sits on top of #712 rather than two merges back.

## That it is not a no-op

A flag that is parsed, accepted and then dropped passes every check this
example has, so it was checked by response. Over `data_pack/pack008.dzn`:

| `--infer-disjunctive-min-clique-size` | cliques_found | cliques_posted | capacity_bound |
|---|---|---|---|
| 2 | 10 | 5 | 44 |
| 3 (default) | 10 | 5 | 44 |
| 5 | 9 | 5 | 44 |
| 8 | 1 | 1 | 44 |
| 11 | 0 | 0 | 0 |
| 12 | 0 | 0 | 0 |

Monotone in the right direction, and the bound collapses once no clique
survives at all.

Two and three agreeing there is a result rather than a gap: #707 predicts
exactly that for the cross-check targets, whose cliques run to ten members
and more, so a two-member clique never displaces one under `_max_posted = 5`.
It is why the question needs the whole collection, which is what the rerun
will give it.

## The lane, and why it is shaped the way it is

`rcpsp_dzn_inferred_min_clique_size` covers the forward.

A plain verify lane would not have. Raising the minimum *removes* a derived
Cumulative, and a proof with one fewer honest derivation in it verifies
exactly as happily as one with it — so a `run_test_and_verify.bash` lane at
any value passes with the forward deleted. That is the shape the comment above
the two mutated lanes already warns about, one flag further along, and it is
why this lane is a sibling of `rcpsp_dzn_inferred_disjunctive_mutated` rather
than a variant of `rcpsp_dzn_inferred`: the two disagree on the same instance
under the same mutation, and disagree only because the flag arrives.

- at three, `sample.dzn`'s one clique is posted, its makespan bound is derived,
  `--mutate-makespan-bound` corrupts it, and veripb refuses — the existing lane;
- at five no clique survives, so there is no bound, the mutation has nothing to
  corrupt, and the honest proof verifies — this lane.

Five rather than two, per the review: `sample.dzn`'s only clique has three
members, so a lane at two is byte-identical to the default and would pass with
the flag deleted.

**Verified by mutation.** With `.with_minimum_clique_size(...)` deleted from the
wiring, `rcpsp_dzn_inferred_min_clique_size` fails and the other three
`rcpsp_dzn_inferred*` lanes still pass, so it isolates the forward rather than
the presolver. Restored, all 21 `rcpsp` lanes pass.

### DIFF
--- a/examples/rcpsp/CMakeLists.txt
+++ b/examples/rcpsp/CMakeLists.txt
@@ -70,6 +70,26 @@ add_test(NAME rcpsp_dzn_inferred_cumulative_mutated COMMAND ${GCS_BASH} ${CMAKE_
     --basename rcpsp_dzn_inferred_cumulative_mutated
     $<TARGET_FILE:rcpsp> --dzn ${CMAKE_CURRENT_SOURCE_DIR}/sample.dzn --infer-cumulative --mutate-makespan-bound)
 
+# The minimum clique size, which the mutation is what makes testable at all.
+# A plain verify lane cannot see this flag: raising the minimum removes a
+# derived Cumulative, and a proof with one fewer honest derivation in it
+# verifies exactly as happily as one with it, so such a lane would pass with the
+# forward deleted --- the same shape as the paragraph above, one flag further
+# along. Running it against --mutate-makespan-bound inverts that, and the lane
+# is the sibling of rcpsp_dzn_inferred_disjunctive_mutated rather than a variant
+# of rcpsp_dzn_inferred: at the default of three, sample.dzn's one clique is
+# posted, its makespan bound derived, the mutation corrupts it and veripb
+# refuses, which is what that lane asserts. At five no clique survives, so no
+# bound is derived, the mutation has nothing to corrupt, and the honest proof
+# verifies --- which is what this lane asserts. So the two disagree on the same
+# instance under the same mutation, and disagree only because the flag arrives.
+# Five rather than two: sample.dzn's only clique has three members, so a lane at
+# two is byte-identical to the default and would pass with the flag deleted.
+add_test(NAME rcpsp_dzn_inferred_min_clique_size COMMAND ${GCS_BASH} ${CMAKE_SOURCE_DIR}/run_test_and_verify.bash
+    --basename rcpsp_dzn_inferred_min_clique_size
+    $<TARGET_FILE:rcpsp> --dzn ${CMAKE_CURRENT_SOURCE_DIR}/sample.dzn --infer-disjunctive --mutate-makespan-bound
+    --infer-disjunctive-min-clique-size 5)
+
 # Enumeration rather than optimisation, so this one reaches the COMPLETE
 # ENUMERATION conclusion, which none of the others do.
 add_test(NAME rcpsp_all COMMAND ${GCS_BASH} ${CMAKE_SOURCE_DIR}/run_test_and_verify.bash --basename rcpsp_all
--- a/examples/rcpsp/rcpsp.cc
+++ b/examples/rcpsp/rcpsp.cc
@@ -322,6 +322,10 @@ auto main(int argc, char * argv[]) -> int
             ("infer-disjunctive-posted",                                                                 //
                 "Cap how many cliques are posted (Sidorov's N_out)",                                     //
                 cxxopts::value<std::size_t>()->default_value("5"))                                       //
+            ("infer-disjunctive-min-clique-size",                                                        //
+                "Smallest clique worth posting; two makes every conflicting pair a candidate, which "    //
+                "is what Sidorov's binary covers are",                                                   //
+                cxxopts::value<std::size_t>()->default_value("3"))                                       //
             ("infer-cumulative",                                                                         //
                 "Run the InferredCumulative presolver, which lifts cover inequalities over the "         //
                 "resources' capacity rows into implied Cumulatives with non-unit heights",               //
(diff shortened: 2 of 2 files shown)


ciaranm:
Merge-checklist review. Trial-merged onto current main (`da56a93d`; the branch is based on `83ffd4e7`, two merges back) and built with `-DGCS_WERROR=ON`: clean, **0 warnings**.

The wiring is right, and I checked it by response rather than by reading, on `examples/rcpsp/sample.dzn`:

| `--infer-disjunctive-min-clique-size` | cliques_found | cliques_posted |
|---|---|---|
| 2 | 1 | 1 |
| 3 (default) | 1 | 1 |
| 5 | 0 | 0 |
| 8 | 0 | 0 |
| 11 | 0 | 0 |

Same shape as the table in the description — monotone, collapsing once nothing survives — on a different instance, which is the useful confirmation. The sample's only clique has three members, so it dies between 3 and 5 rather than at 8.

**The default really is unchanged**, which is the claim worth being sure of: no flag, explicit `3`, and a binary built without this PR all give `cliques_found: 1, cliques_posted: 1` and `makespan: 6`.

`InferredDisjunctive`'s own default is `_min_clique_size(3)` (`inferred_disjunctive.cc:121`), matching the new `default_value("3")` — and the sibling flags line up the same way, `100` against `_max_candidates(100)` and `5` against `_max_posted(5)`.

## One finding: the reason given for adding no test is not the case

> Built and run against GCC release; no test changes, since the example has no
> test binary and the behaviour under test belongs to the presolver, which
> already has fixtures for the parameter.

The second half is true — `inferred_disjunctive_test.cc` carries `min_clique_size` in its `Setup` (L80) and passes it at L128. But the first half is not: `examples/rcpsp/CMakeLists.txt` registers about twenty ctest lanes, and several pass CLI flags, including `rcpsp_dzn_inferred` at L56 which runs `--infer-cumulative --infer-disjunctive` through `run_test_and_verify.bash`.

That matters more than a wording slip, because the comment sitting right below it (L59-65) makes the argument against the conclusion:

> What this covers that the presolvers' own mutation fixtures do not is the
> *forwarding*: the spec field, the stats callback and the flag that reaches them
> all live here rather than in a presolver, and a forward that bit-rotted would
> leave an honest proof behind and nothing to say so --- the shape a no-op
> presolver passing every check has, one layer out.

Which is exactly this PR's own premise. `with_minimum_clique_size` has existed since #663 and no lane could reach it, so it went four months as a parameter nothing outside a unit test could set — a forward that was never made rather than one that bit-rotted, but invisible for the same reason. Adding the flag without a lane leaves the new forward defended by nothing, and the presolver's fixtures cannot see it, by the same argument the file already makes about the makespan bound.

**If a lane is added, the value matters.** `2` would be inert here — it is byte-identical to the default on `sample.dzn`, since the only clique has three members, so a lane at `2` would pass just as happily with the flag deleted. `5` is the cheap one that bites:

```
--dzn sample.dzn --infer-disjunctive --infer-disjunctive-min-clique-size 5
```

takes `cliques_found` from 1 to 0 while `makespan` stays 6, so it still solves and still verifies, and a dropped forward turns it back into 1. Verified both, above.

This is a flag rather than a request to hold the PR: the wiring is correct, the default is genuinely untouched, and the measurement it exists to unblock does not depend on the lane. But it is worth either adding the lane or saying that the artefact rerun is the coverage and the lane is deliberately skipped — the thing to avoid is the reason as currently written, since it points at a test binary that is there.

## Everything else passes

`Co-Authored-By` present; single commit; one file changed and it belongs to the change; no `overloaded{}` touched; no proof-logging change, so no Check 7 obligation; no cheating-proof rules; no stray or generated files. On docs, `dev_docs/inferred-disjunctive.md:33` discusses `with_minimum_clique_size(2)` at the API level and no dev_doc enumerates the rcpsp CLI flags, so nothing goes stale — the two sibling budget flags are undocumented there too.

ciaranm:
Finding accepted — the reason I gave was wrong, and `examples/rcpsp/CMakeLists.txt` registers twenty-one lanes, several passing CLI flags. Lane added, and the argument you made from the L59-65 comment is the right one: this PR's own premise is that a parameter went four months unreachable, so adding the forward without defending it repeats the thing it fixes.

One correction to the shape you proposed, though, because I checked it before writing it and it does not bite.

**A plain verify lane at `5` passes with the forward deleted.** `run_test_and_verify.bash` runs with `--prove` and checks veripb; it has no output assertion. Raising the minimum *removes* the derived Cumulative, and a proof with one fewer honest derivation verifies exactly as happily as one with it. So `cliques_found` going 1 → 0 is invisible to the lane — both configurations produce accepted proofs, and the lane would pass just as happily with the flag deleted as one at `2` would. It is the same trap as `2` being inert, one level further in: the *inference* differs and the *check* cannot see it.

What makes it testable is the mutation, so the lane is a sibling of `rcpsp_dzn_inferred_disjunctive_mutated` rather than a variant of `rcpsp_dzn_inferred`:

```
--dzn sample.dzn --infer-disjunctive --mutate-makespan-bound --infer-disjunctive-min-clique-size 5
```

under `run_test_and_verify.bash`. At three the clique is posted, its bound derived, the mutation corrupts it and veripb refuses — which is what the existing lane asserts. At five no clique survives, so there is no bound, the mutation has nothing to corrupt, and the honest proof verifies — which is what this lane asserts. Same instance, same mutation, opposite expectations, and they differ only because the flag arrives. Measured directly: `veripb` rc=1 at three, rc=0 at five.

Your `5` is still the right value and for your reason — `sample.dzn`'s only clique has three members, so `2` is byte-identical to the default.

**Verified by mutation**, since a lane defending a forward is exactly the thing that should not be taken on faith: with `.with_minimum_clique_size(...)` deleted from the wiring, `rcpsp_dzn_inferred_min_clique_size` fails and the other three `rcpsp_dzn_inferred*` lanes still pass, so it isolates the forward rather than the presolver. Restored, all 21 `rcpsp` lanes pass.

Also rebased onto `0e7cf85a` — your trial-merge was against `da56a93d`, which predates #712 landing — and amended rather than added a commit, to keep the single-commit check. PR body updated; the sentence you quoted is gone.

Issue #742, opened 2026-08-14
Edge-finding's window x task scan is O(n^3), and is the whole of the rule's cost


ciaranm:
Follow-up to #733, which certified edge-finding for `Cumulative` in both directions. The certificate and the propagation are done; what is left is purely a propagation-performance matter, which is why it is split out rather than held in the same branch.

## The measurement

The rule's cost can be isolated exactly, because capping the number of firings at zero leaves the scan running in full while making the search identical to having no rule at all. Whatever the difference is, it is scan time.

| | nodes | wall |
|---|---|---|
| off | baseline | baseline |
| scan runs, nothing fires | identical to baseline | **1.5x** baseline |
| the rule as it ships | 72% of baseline | 1.10x baseline |

(35 generated RCPSP instances that close, timed serially. Before the two directions were merged into one pass it was 2.26x and 1.60x.)

Two things follow, and the first is the important one.

**The inferences cost nothing measurable.** Node throughput is the same whether or not firing is allowed — 32% of baseline either way on the 53 `data_pack` instances that time out at 60s. Every bit of the tax is the search for firings, not the firings.

**Whether the rule pays depends on the instance.** On `data_pack/pack021` it is a 1.94x win (19.2s -> 9.9s, recursions -82%). On the harder `data_pack` instances that do not close it is a loss: at 60s the incumbent is worse on 4 instances and equal on 49. So it is off by default, and should stay off until this is fixed.

## What it is

`Cumulative`'s (OC') overload check already sweeps O(n^2) windows `[a, b)`. Edge-finding adds, per window, a scan over the candidate tasks to find the ones it can push, which makes the whole thing O(n^3). Two guards keep most windows out of it — `tallest`, on the largest height, and `heaviest`, on the largest single-task energy — and inside the scan the candidates are walked tallest-first so that `rest <= 0` can `break` rather than `continue`. That took the scan from 6.1e9 task iterations to 2.0e9 on the same set, and is where the easy constant factors have run out.

## What would fix it

Vilim's Theta-Lambda tree computes the same prunings in O(n log n) per pass. The complication for us is that it computes the update *without materialising the responsible set*, and the certificate needs that set by name: the `pol` cites one window-energy row per member of Theta. So this is not a drop-in replacement of the search — it needs the tree to report the set it used, or a cheap reconstruction of it afterwards.

[...]

Issue #550, opened 2026-07-21
TTHE-OC and KAOC: both rungs certified and merged; the optional-task interaction and a variable height are what is left


ciaranm:
> **Status, 2026-09-04 — both rungs are in, and the "parked" note below is superseded.**
>
> **PR #747 landed TTHE-OC and KAOC, certified.** They are
> `CumulativeRules::elastic_overload` and `CumulativeRules::knapsack_overload`,
> both off by default, tested by `gcs/constraints/cumulative/cumulative_kaoc_test.cc`
> and documented in `dev_docs/cumulative-proof-logging.md` ("The overload ladder:
> one certificate, a tighter line per time point"). The design here held: the three
> rungs came out as **one comparison with a different cap on one side of it**, the
> pol shape never changes, and KAOC's integrality reasoning goes through #544's
> `derive_subset_sum_strengthening` — which until then only a presolver had used.
>
> **What it is worth**, from #747, on top of the edge-finding family rather than on
> top of `(TTOC)`, over the 39 instances both arms closed:
>
> | arm | recursions | vs TTEF | closed | wall |
> |---|---|---|---|---|
> | TTEF | 7,250,390 | 1.000× | 39 / 95 | 98.2 s |
> | + (KAOC) | 2,342,923 | **0.323×** | **49 / 95** | **76.0 s** |
>
> So the knapsack cap pays for its own `O(n²·horizon)` scan and then some, and it is
> **not** subsumed by the energetic family — it refutes windows where the *shape* of
> the heights is what does not fit, which no aggregate energy argument sees. The
> elastic cap alone does not pay: 0.693× of the search, but it closes *fewer*
> instances than leaving it off.
>
> **Why this is still open.** Three things from the plan above are not in:
>
> - **The optional-task interaction.** The *Optional tasks* section's explicit
>   deliverable — presence falsification by energy, and validation item 5's fixtures
>   for it — is not built. The rungs take the v1 policy only: the propagator "skips
>   a task not yet known present" (`gcs/constraints/cumulative/cumulative.cc:285`),
>   and `cumulative_kaoc_test.cc` carries no presence fixture. The time-table push
>   scan falsifies presences; the window-energy rungs do not.
> - **A variable height.** `cumulative.cc:1503` declines *both* rungs outright when
>   any active task has one, since a bit-linearised contribution is not a coefficient
>   the knapsack's item list can read. This issue put that out of scope, but
[...]

Issue #544, opened 2026-07-21
Certified subset-sum strengthening utility


ciaranm:
Part of the Cumulative proof-logging plan — tracking issue #541 has the blanket rules (no new OPB content; the shared validation vocabulary) and the shared reference list. Plan numbering used below: issue 01 = #542, 02 = #543, 03 = #544, 04 = #545, 05 = #546, 06 = #547, 07 = #548, 08 = #549, 09 = #550.

## Summary

A reusable proof component: given an already-derived PB line L: Σ cᵢxᵢ ≤ B (where the
xᵢ are literals or proof flags with known coefficients), derive the strengthened line
L′: Σ cᵢxᵢ ≤ B′ where B′ is the largest subset-sum of {cᵢ} that is ≤ B. This is the
integrality argument at the heart of three consumers: the knapsack-augmented overload
check's availability bound (issue 09), Schulz's knapsack capacity-reduction presolve
rule (issue 06), and single-resource lifting certificates (issue 08).

## Background

The claim "no subset of the cᵢ sums into the gap (B′, B]" is NP-hard in general but
pseudo-polynomial by dynamic programming, and the DP has a direct proof-logging
mirror: layered reified flags over reachable partial sums. The in-tree precedent is
the knapsack propagator (`gcs/constraints/knapsack/knapsack.cc:150-330`), which
proof-logs exactly this kind of layered DP with `create_proof_flag_reifying` over
partial-sum facts. This issue extracts the idea into a standalone utility with a
sharper contract.

There is also a one-line special case worth its own fast path: when d = gcd(cᵢ) > 1
and B′ = d·⌊B/d⌋, the strengthening is just VeriPB division by d followed by
multiplication by d (Chvátal–Gomory rounding).

## References

- Cloutier & Quimper CP 2026 §3–4 (h^K_max and the bitset DP; open access — ask
  Ciaran for any paper you cannot obtain).
- Schulz thesis (TU Berlin 2013, open access) for the κ presolve rule this feeds.
- `gcs/constraints/knapsack/knapsack.cc` for the DP proof-flag pattern.

## Proof discipline

The utility writes **only** proof derivations. Enforce this structurally: the API
takes a `ProofLogger &` and has no access to `ProofModel` at all, so OPB writes are
impossible by construction. (This is the pattern to prefer throughout the plan
wherever a component is derivation-only.)

## Design

- API sketch (final shape at implementer's discretion):
[...]

PR #747, opened 2026-08-14
cumulative: knapsack-augmented and horizontally elastic overload checking

ciaranm:
Closes the last commonly-run cumulative rule that was uncovered: the time-table
horizontally elastic overload check (TTHE-OC, Kameugne et al., C&OR 172 (2024))
and Cloutier & Quimper's knapsack-augmented one (KAOC, CP 2026). Issue #550.

Stacked on #745, which merges first.

## The three rungs are one certificate

`(TTOC)`, `(TTHE-OC)` and `(KAOC)` turn out to be one comparison with a
different cap on one side of it. Charge the window each contained task's energy
less whatever its compulsory part already accounts for, and supply it **one time
point at a time**:

```
required = e_Θ − Σ h_i·|comp_i ∩ window|
supplied = Σ_{t ∈ [a,b)} A_t

A_t = capacity − f(I,t)              → rearranges term for term into (TTOC)
A_t = min(that, optional heights)    → (TTHE-OC)
A_t = knapsack(heights, that)        → (KAOC)
```

So the pol shape never changes; only the line at some time points gets tighter.
Each availability line is the capacity row with every compulsory contribution
pinned off it (`pin_contributor`, which `(TTOC)` already uses) and every term
that is not a contained task's optional one `weaken`ed away — which leaves a
statement over exactly the heights the knapsack reasons about. `(KAOC)` then
puts it through `derive_subset_sum_strengthening`, the utility #544 added for
exactly this and which until now only a presolver had used.

Against those, each contained task's `derive_window_energy` line scaled by its
height, over the task's own `[est, lct)`, with its compulsory times weakened
back out.

**The knapsack DP is applied biggest-gain-first and stopped as soon as the
comparison tips**, so a conflict the elastic cap can carry pays nothing for it;
the marker records `strengthened=k/w`.

## Measured — `data_bl` + `data_pack` at 60 s, over the 36 instances every arm closed

| arm | recursions | vs baseline | propagations | closed |
|---|---|---|---|---|
| time-tabling + (OC') + (TTOC) | 7,227,100 | 1.000× | 72,057,194 | 38 / 95 |
| + (TTHE-OC) | 5,008,908 | 0.693× | 51,145,525 | 36 / 95 |
| + (KAOC) | 2,832,478 | **0.392×** | 29,227,548 | **48 / 95** |

Fewer recursions on 30 of 36 and more on none; no arm disagrees about an
optimum. **The knapsack cap is what pays** — it more than halves the search and
closes ten more instances at the same wall time. **The elastic cap alone does
not**: 0.693× of the search, but it closes *fewer* instances than leaving it
off, because its per-time-point scan costs more than the pruning returns. Same
verdict not-first / not-last got, and for the same reason.

That baseline is the identical figure to #744's "no edge-finding" row, so the
two families are directly comparable: edge-finding alone is 2,359,885.

## ... and on top of the edge-finding family, which is the question that matters

(TTOC) is not what anyone runs. Over the 39 instances both arms closed:

| arm | recursions | vs TTEF | propagations | closed | wall |
|---|---|---|---|---|---|
| TTEF | 7,250,390 | 1.000× | 74,388,179 | 39 / 95 | 98.2 s |
| + (KAOC) | 2,342,923 | **0.323×** | 29,856,904 | **49 / 95** | **76.0 s** |

Fewer recursions on 26 of 39 and more on none, ten more instances closed, and
faster in wall time as well — so the rule pays for its own `O(n²·horizon)` scan
and then some.

Which also settles that it is **not subsumed by the energetic family**.
Edge-finding and TTEF move bounds from a window's total energy; this refutes
windows where the *shape* of the heights is what does not fit, and no amount of
aggregate energy reasoning sees that.

## What the fixtures could not catch

Worth reading even if the rest is skimmed. **Every hand-built fixture verified
through two real certificate bugs**; 11 of 60 proofs on *generated* instances
did not. Both were the same mistake — a quantity the check computed that the
certificate never derived:

- **The elastic cap was computed and not derived.** Where the optional heights
  fall *below* what the profile leaves, the check charged
  `min(left, heights)` while the pol cited the capacity row, whose right hand
  side is `left`. The binding fact there is not the capacity row at all — it is
  each task's own literal axiom.
- **The energy lines ran over the window rather than over the task**, leaving a
  negative coefficient on every `(task, time)` the task cannot reach.

Both were hidden by the same property: a fixture built to *demonstrate* a rule
is symmetric and generous. Every task can run at every time point (so the
elastic cap never binds) and every task spans the whole window (so the energy
residue is zero). Cloutier & Quimper's own Example 2 has both.

This is the other half of #744's lesson — that one said a demonstration fixture
will not make a *mutation* bite; this says it will not make a *bug* show either.

Both are now cross-checked structurally: the justification adds up what the pol
actually charges and throws if it disagrees with what the rule fired on, with
the two figures computed from opposite sides of the propagator.

## One change outside the constraint

`derive_subset_sum_strengthening` now takes the reason its source line was
derived under. A reason-backed source carries the negated reason's literals
alongside its own terms, so a dead state's "this prefix cannot be completed"
holds only where the reason does, and the transition RUPs have to say so. Its
two existing callers are presolve-time, pass an empty one, and emit exactly what
they emitted before.

Also two stale doc claims corrected, both saying TTEF is not certified and
refuses a proof logger. Neither has been true since #744 certified it; only
`energetic_edge_finding` throws. They belong further down the stack, but moving
a two-line doc fix would mean rebasing.

## Testing

- `cumulative_kaoc_test`: three KAOC-only differentials with negative twins —
  `cloutier_ex2` (the paper's parity example, which takes the strengthening's
  divisibility fast path), `dp_path` (gcd 1, so the layered programme runs), and
  `compulsory` (added because neither published fixture has a task with a
  compulsory part, so the certificate's one asymmetric step was a no-op on both).
  Each asserts brute force finds no solution, that `(TTOC)` and `(TTHE-OC)` both
  decline, and that the conflict carries the `kaoc` marker.
- Nine mutation lanes: `claim_one_better`, `strengthen_one_fewer` and `capacity`
  on each fixture, all rejected. There is deliberately **no** lane for the
  compulsory-time weakening: corrupting it is accepted, because unit propagation
  assigns the terms it would have cancelled from the reason's own bound
  literals, the same way `(TTOC)`'s pins are droppable on 235 of 248 instances.
- Enumeration against brute force with the proof verified, caps on and off.
- 600 generated instances enumerated exhaustively at sizes 8 and 10, zero
  solution-count differences.
- 60 of 60 generated proofs verify on both arms.
- 53 of 53 cumulative and subset-sum ctest entries pass. OPB byte-identical.

## Restrictions

Constant heights only — a variable height puts bit-linearised contribution terms
in the capacity row, which neither the knapsack's item list nor the term-dropping
can read, so v1 declines rather than approximates and the plain rules still run.
A capacity above 4096 falls back to the elastic cap, since the bitset is
`capacity + 1` bits per time point (Cloutier & Quimper report `C ≤ 122` across
their benchmarks; the local instances run 5 to 22).

Propagation is `O(n²·horizon)` rather than their `O(Cn²)`: their doubly linked
Profile collapses the runs where the profile is constant, and this keeps a flat
array plus the incremental bitset (their Algorithm 3 shift-or) per time point.
Same trade as #742 for edge-finding's scan, and the same answer — what is
missing is propagation performance, not proof content.

### DIFF
--- a/dev_docs/cumulative-proof-logging.md
+++ b/dev_docs/cumulative-proof-logging.md
@@ -1429,6 +1429,21 @@ not**: 0.693x of the search, but it closes *fewer* instances than leaving it off
 because its per-time-point scan costs more than the pruning returns --- the same
 verdict not-first / not-last got, and for the same reason.
 
+And on top of the edge-finding family, which is the question that decides
+whether it is worth having at all --- over the 39 instances both arms closed:
+
+| arm | recursions | vs TTEF | propagations | closed | wall |
+|---|---|---|---|---|---|
+| TTEF | 7,250,390 | 1.000x | 74,388,179 | 39 / 95 | 98.2 s |
+| + (KAOC) | 2,342,923 | 0.323x | 29,856,904 | 49 / 95 | 76.0 s |
+
+Fewer recursions on 26 of 39 and more on none, ten more instances closed, and
+**faster in wall time as well** --- so the rule pays for its own O(n^2 * horizon)
+scan and then some. It is not subsumed by the energetic family: edge-finding and
+TTEF move bounds from a window's total energy, and this refutes windows where
+the *shape* of the heights is what does not fit, which no amount of aggregate
+energy reasoning sees.
+
 Soundness: 600 generated instances enumerated exhaustively at sizes 8 and 10,
 zero solution-count differences against the rules off.
 
--- a/dev_docs/cumulative-proof-logging.md
+++ b/dev_docs/cumulative-proof-logging.md
@@ -1343,11 +1343,155 @@ accepts, which fails the lane, so the verdict is veripb's either way.
 lane was corrupting an edge-finding firing on the same instance and reporting a
 rejection that said nothing about not-first / not-last.
 
+## The overload ladder: one certificate, a tighter line per time point (#550)
+
+`(TTOC)` charges a window `capacity x width` in bulk and subtracts the profile.
+Two published rules improve on the same comparison by capping what each
+*individual* time point supplies:
+
+* **(TTHE-OC)**, the time-table horizontally elastic overload check (Kameugne,
+  Fetgo Betmbe, Noulamo & Tayou Djamegni, C&OR 172 (2024); the formulation used
+  here is Cloutier & Quimper's equivalent one, CP 2026 SS2.2.5). A time point no
+  task can reach with more than its own tasks' heights does not supply the whole
+  capacity: resource nobody can use is not available.
+  `CumulativeRules::elastic_overload`.
+* **(KAOC)**, the knapsack-augmented overload check (Cloutier & Quimper, CP
+  2026). The tasks that could run at a time point have integer heights, so what
+  they can between them consume is the largest *subset sum* of those heights
+  fitting under what the profile leaves --- not that figure itself.
+  `CumulativeRules::knapsack_overload`, which implies the rule above and
+  dominates it.
+
+### The three rungs are one comparison
+
+Charge the window each contained task's energy less whatever its compulsory part
+already accounts for, and supply it one time point at a time:
+
+    required = e_Theta - sum_i h_i * |comp_i ^ window|
+    supplied = sum over t in [a, b) of A_t
+
+With `A_t = capacity - f(I, t)` this **is** `(TTOC)`: each contained task's
+compulsory load comes off the required side and goes back on as the supply the
+profile removes, and the two rearrange into `e + F > C(b-a)` term for term. Cap
+`A_t` by the optional heights at `t` and it is (TTHE-OC); cap it by the largest
+total they can reach and it is (KAOC). Nothing else changes --- so this is one
+certificate with a tighter line at some time points, not three certificates.
+
+### The certificate
+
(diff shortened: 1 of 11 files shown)


ciaranm:
Added a one-line test-hygiene commit: `cumulative_kaoc_test` was the only test in the Cumulative family whose `main` never called `establish_and_announce_seed`.

Nothing here draws random *data*, so it looked harmless — but the small enumeration branches through `random_branch_with_optional_seed`, which falls back to `random_device` when nothing has stored a seed. So `--seed=N` was accepted and ignored, no seed was announced, and `cumulative_kaoc_enumeration.pbp` came out different on two runs of the same binary (589 lines against 625). Not ASLR — `setarch -R` does not settle it. A failure in this lane could not have been reproduced, and the proof was unusable as a byte-diff reference against a later change.

With the call in place, the same seed gives a byte-identical proof, a different seed a different one, and 40 seeds pass. The three root-refutation fixtures were always stable, because they refute before anything branches.

Found while byte-diffing #748 against this branch.

PR #750, opened 2026-08-17
cmake: build the Sanitize configuration at -g1, not -g3

ciaranm:
The Sanitize lane on main is red, and not because a test failed: [run 31973615777](https://github.com/ciaranm/glasgow-constraint-solver/actions/runs/31973615777/job/95236658047)'s `Test (Sanitize)` step is still `in_progress` in the job record, its only annotation is `System.IO.IOException: No space left on device` thrown by the runner's *own* worker log writer, and `gh run view --job … --log` answers `log not found` — nothing was uploaded because there was nowhere to write it. Every other lane on that commit is green, and rebuilding `2ccabd97` locally with the same preset and the same GCC 15.2.0 passes **703/703**.

## What fills the disk

Debug info, multiplied by the link model. Every test binary statically links the whole solver, so `-g3` hands each of the 173 of them its own copy of the same DWARF:

| | `-g3` (before) | `-g1` (after) |
|---|---|---|
| `build-sanitize` | 93 GiB | **56.1 GiB** |
| 173 executables | 84.6 GiB | 52.1 GiB |
| mean binary | 501 MiB | 309 MiB |
| `among_test`, one of the larger ones | 819 MiB (434 MiB of it DWARF) | 503 MiB (118 MiB of it DWARF) |
| `libglasgow_constraint_solver.a` | 1.9 GiB | 901 MiB |

The build *fitted* on the runner; the tests did not, because a proof file needs somewhere to go. A full parallel `ctest` holds around 3 GiB of them at once (measured at -j24; largest single `.pbp` is 565 MiB, `linear_equality_test_ne_w0_pall_t0`), and that is what tipped it over. The margin had been shrinking for a while: the previous run linked 170 executables and passed, this one links 173 — `cumulative_{ttef,nfnl,kaoc}_test` from the #743→#749 stack are about 2.4 GiB between them.

## What -g1 gives up, checked rather than assumed

Nothing a CI log is ever read for. Two probes built with the lane's exact flags, once at `-g1` and once at `-g3`, produce frame-for-frame identical reports — same demangled function names, same `file:line`, and for UBSan the same column:

```
    #0 0x… in overflow_here(int*) /home/ciaranm/claude/tmp/asan_g1_probe.cc:5
    #1 0x… in main /home/ciaranm/claude/tmp/asan_g1_probe.cc:11
```

That is expected: the sanitizers carry their own location data for the diagnostic line, and symbolising a backtrace needs line tables, which `-g1` emits. What `-g1` drops is macros and local-variable DWARF, and those want a debugger — which is what the **Debug** configuration is for. It stays at `-g3`, and no CI lane builds it (the CMake workflow configures only `--preset release` and `--preset sanitize`; the Windows lanes are MSVC, where this whole block does not apply).

## Testing

`cmake --preset sanitize` + `ctest --preset sanitize -j24` on this branch: **703/703 passed** in 137 s, `sanitizers_enabled_test` and `stacktrace_logging_test` among them, so the sanitizers are still in the binaries and Release-grade backtrace info is still there.

This is the cheap half of the problem. The expensive half is that 173 binaries each carry a copy of one library; `BUILD_SHARED_LIBS=ON` for this lane would be the real fix, and is worth a separate look if 56 GiB turns out still to be too close to the line.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -5,11 +5,15 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 ## Build Commands
 
 Three build types are supported: `Release` (default), `Debug`, and `Sanitize`.
-All keep debug information, but scaled to purpose: `Release` uses `-g1` (`/Z7`
-on MSVC) — function names plus line tables, enough for backtraces and the
-`GCS_VERBOSE_LOGGING` stacktrace annotation — while `Debug` and `Sanitize` use
-`-g3` for full interactive debugging. Use named presets from
-`CMakePresets.json` for convenience:
+All keep debug information, but scaled to purpose: `Release` (with `/Z7` as the
+MSVC equivalent) and `Sanitize` use `-g1` — function names plus line tables,
+enough for backtraces, for the `GCS_VERBOSE_LOGGING` stacktrace annotation, and
+for an ASan or UBSan report to name a function and a `file:line` — while `Debug`
+uses `-g3` for full interactive debugging. `Sanitize` is `-g1` for size: every
+test binary links the solver statically and so carries its own copy of the same
+DWARF, which at `-g3` made a 93 GiB build tree and ran the CI runner out of
+disk; at `-g1` it is 56 GiB. Build `Debug` when you are going to attach a
+debugger. Use named presets from `CMakePresets.json` for convenience:
 
 ```shell
 cmake --preset release  && cmake --build --preset release   # optimised (default)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -230,15 +230,32 @@ if(PROJECT_IS_TOP_LEVEL)
         #               prints name + file:line. Keeping it in Release means crash
         #               reports and verbose proofs stay useful in optimised builds.
         #   Debug    -> -g3: full information, including macros, for interactive
-        #               debugging.
-        #   Sanitize -> -g3: fully-annotated ASan/UBSan stack traces.
+        #               debugging. This is the configuration to build when you are
+        #               going to attach a debugger; `cmake --preset debug` is not
+        #               used by any CI lane, so its size costs nobody anything.
+        #   Sanitize -> -g1 as well, which is a deliberate trade rather than an
+        #               oversight. Every test binary statically links the whole
+        #               solver, so each carries its own copy of the same DWARF, and
+        #               under -g3 the tree's 173 programs came to 84.6 GiB of a
+        #               93 GiB build tree -- around half of that being debug info,
+        #               going by among_test, one of the larger ones, at 434 MiB of
+        #               DWARF in an 819 MiB binary. That is more than a GitHub
+        #               runner has: the build fitted, and then the Sanitize lane
+        #               died mid-test with "No space left on device" when the tests
+        #               went to write their proof files (run 31973615777, 16 Aug
+        #               2026). At -g1 the same programs come to 52.1 GiB and the
+        #               tree to 56 GiB. An ASan/UBSan report still symbolises to
+        #               function + file:line there, which is all a CI log is ever
+        #               read for; what is given up is macros and local-variable
+        #               DWARF, and those want a debugger, i.e. the Debug
+        #               configuration above.
         #
-        # Anything other than Debug/Sanitize (Release, or an unset build type)
+        # Anything other than Debug (Release, Sanitize, or an unset build type)
         # gets -g1 so it always carries at least backtrace-quality debug info.
         # stacktrace_logging_test guards the Release end of this: it fails if a
         # build ever drops the debug info the verbose proof logging relies on.
         add_compile_options(-gdwarf-3)
-        add_compile_options($<IF:$<OR:$<CONFIG:Debug>,$<CONFIG:Sanitize>>,-g3,-g1>)
+        add_compile_options($<IF:$<CONFIG:Debug>,-g3,-g1>)
         add_compile_options(-pthread)
         add_link_options(-pthread)
 


ciaranm:
Pre-merge check. CI is green on all ten lanes for `8b3bd88d`, and I confirmed the diagnosis independently rather than from the description: run 31973615777's Sanitize job has `Build (Sanitize)` → `success`, `Test (Sanitize)` → `in_progress`, everything after pending, and its only annotation is the `IOException: No space left on device` from the runner's own worker log writer. The build fitted and the tests did not, exactly as described. The generator-expression change is right, `Sanitize` is GCC/Clang-only so the MSVC side correctly needs nothing, and `stacktrace_logging_test` is registered unconditionally so it runs in the Sanitize build and guards the `-g1` end too.

Three things in the prose, though, and one of them lands permanently in the tree. Nothing here affects behaviour.

### 1. `mean binary | 815 MiB` cannot be right

The table's own totals give `84.6 GiB / 173 = 501 MiB`. Conversely `815 MiB × 173 = 137.7 GiB`, which is more than the 93 GiB the same table gives for the whole build tree, so the figure is impossible rather than merely off.

The `-g1` column is exact — `52.1 GiB / 173 = 308.4 MiB`, printed as 309 — so it is this one cell rather than the totals. The `among_test` row is clearly a real measurement: its non-DWARF part comes out at **385 MiB in both columns** (819−434 and 503−118), which is what should happen when only the debug level changes. `815` looks like `819` having landed in the mean row. `among_test` is simply a larger-than-average binary: 1.63× the mean, matching its 316 MiB saving against the 192 MiB mean saving.

### 2. The same figure generalised, in `CMakeLists.txt:238-239`

> `-g3` gave each of the ~173 of them its own ~434 MB copy of the same DWARF -- 53% of an 819 MB binary

434 MiB is `among_test`'s own DWARF, and applying it to all 173 gives **73.3 GiB of DWARF, i.e. 87% of the 84.6 GiB of executables, not 53%**. It also contradicts this PR's own measurement of what `-g1` recovered: the tree fell by 36.9 GiB, roughly half of what 73.3 GiB of DWARF would imply. The mean per-binary DWARF works out at about 265 MiB.

The commit message has the same numbers but presents them as one example ("434 MiB of an 819 MiB binary, 84.6 GiB of a 93 GiB build tree"), which reads correctly. It is only the "each of the ~173 of them its own ~434 MB" phrasing that generalises — and that comment is the permanent justification for the trade, so it seemed worth catching before it lands. `CLAUDE.md:13` carries a softer version of the same.

### 3. `CLAUDE.md:8-9`: the `/Z7` parenthetical widened

> `Release` and `Sanitize` use `-g1` (`/Z7` on MSVC)

`/Z7` is only applied under `$<$<CONFIG:Release>:/Z7>`, and per `CMakeLists.txt:12` the Sanitize configuration is GCC/Clang-only, so it does not exist on MSVC at all. The previous wording attached the parenthetical to `Release` alone and was accurate.

### On whether 56 GiB is still too close to the line

The description leaves this open, and the job's step list bounds it. Because `Build (Sanitize)` *completed* at 93 GiB and the disk only filled once tests began writing, the runner's free space was somewhere around 93–96 GiB. At 56 GiB that leaves roughly 37–40 GiB against the ~3 GiB peak of concurrent proof files measured here — about 12× headroom rather than a marginal fix. That assumes the runner's tree sizes track the local GCC 15.2.0 figures; the ordering it establishes (build fits, tests do not) holds regardless. So `BUILD_SHARED_LIBS=ON` looks like a genuine nice-to-have rather than something needed soon.

Happy to merge once the numbers are settled — the code change itself I have no concerns about.

ciaranm:
All three are right, and I checked each rather than taking them on trust. Fixed in `c4029bb7`.

**1 and 2 are the same mistake.** The mean is `84.6 GiB / 173 = 501 MiB`; `815` came from eyeballing individual binaries (`cumulative_ttef_test` 816, `cumulative_edge_finding_test` 815, `among_test` 819) instead of dividing the total, so I quoted a sampled size in a row labelled "mean" — and, as you say, then generalised the same binary's *absolute* DWARF to all 173. Applying the *fraction* rather than the figure keeps it consistent: 53% of 84.6 GiB is ~45 GiB of DWARF, of which `-g1` removed the measured 32.5 GiB and kept ~12 GiB of line tables. That also lands on your ~265 MiB mean (53% of 501 MiB), which is the number the tree-wide arithmetic wants.

Rather than swap in an extrapolated mean, the comment at `CMakeLists.txt` now states only what was measured — 84.6 GiB of a 93 GiB tree, 52.1 GiB and 56 GiB after — with `among_test` named as "one of the larger ones" where its 434/819 appears. Same in the commit message and `CLAUDE.md`. The PR table above is corrected to 501 MiB.

**3**: correct, and it was a real widening rather than a typo — `/Z7` is `$<$<CONFIG:Release>:/Z7>` at `CMakeLists.txt:225`, and `Sanitize` lives inside `if(NOT MSVC)`, so it does not exist there at all. The parenthetical is back on `Release` alone.

**On the headroom**: agreed, and thanks for bounding it from the step list — that is a better argument than the one in the description, which left it open. ~37–40 GiB against a ~3 GiB proof peak is not marginal.

For what it is worth, the shared-library idea has since been measured, at Ciaran's suggestion and in the shape he asked for: private to the build tree, never installed or exported, so no ABI promise. `-g1` plus that is **5.7 GiB** (mean binary 12 MiB, one 533 MiB `.so`), 703/703 in 144 s against 137 s, with the default build still installing the static archive and the private tree refusing `cmake --install` outright. It is committed locally on top of this branch but not pushed, pending his call — so nothing here depends on it.

Issue #757, opened 2026-08-17
Disjunctive not-first/not-last certifies a strict weakening, and the published unary rule looks certifiable a different way


ciaranm:
Follow-up to #752, and the disjunctive counterpart of #746. **Nothing here is a soundness problem**: what #752 implements is sound, certified, mutation-tested and enumeration-checked. It is weaker than the rule as published, this issue records exactly how much weaker, and — unlike #746 — it has a concrete route to closing the gap.

## The conclusions match; the detection does not

Both published statements push where we push — not-first raises `lb(s_j)` to `min_{i∈Ω} ect_i`, not-last lowers `ub(s_j)` to `max_{i∈Ω} lst_i − p_j`.

The detection differs. #752 inherits `Cumulative`'s, which is the guarded window-energy lemma over the negated conclusion's start range:

```
p(Ω) + window_energy_bound(j, [a, b), negation range)  >  b − a
```

The unary rule as published (Baptiste, Le Pape and Nuijten; Vilím's Θ-tree presentation) argues about a *different window* instead. For not-first: if `s_j < min ect(Ω)` then for every `i ∈ Ω`, `s_j < ect_i`, so `j` cannot be after `i`, so `j` is before `i`, so all of `Ω` lies in `[ect_j, lct(Ω))` — giving

```
p(Ω) > lct(Ω) − ect_j            not-first
p(Ω) > ub(s_j) − est(Ω)          not-last, the mirror
```

## Measured: ours is a strict subset

Both conditions computed side by side per (window, task) pair, over 30,000 random unary instances at the time-tabling / detectable-precedence fixpoint:

| | firings |
|---|---|
| ours (window-energy) | 3,073 |
| published (unary) | 4,218 |
| both | 3,073 |
| **ours only** | **0** |
| **published only** | **1,145** |

So on this family we certify a strict weakening, and it is not a near-thing: the published condition fires 37% more often. That is a cleaner verdict than #746 got on the cumulative side, where each condition had firings the other did not.

The reason is visible in the arithmetic. The guarded row states one bound uniformly over the whole negation range from two guard literals, so it keeps the "ends by `t`" survivors the low guard decides and concedes a unit for every "starts after `t`" survivor the high guard does not — two worst cases that need not occur at the same start value. Not-first's negation range runs from outside the window to inside it, straddling the hump, so the concession is routine rather than exceptional.

## Why this one looks certifiable, where #746's did not

[...]

Issue #754, opened 2026-08-17
Set-based detectable precedences for Disjunctive: Vilim's ect(Omega), not the pairwise max eet


ciaranm:
#734 certified detectable precedences for `Disjunctive`, and deliberately took
the **pairwise** form only. This issue is the set-based form, which is the last
strength left on that rule and the only remaining scheduling gap that had no
issue of its own.

Recorded in `dev_docs/disjunctive-proof-logging.md`'s open follow-ups:

> **The set-based form of detectable precedences.** The rule above pushes to
> `max_k eet_k` over the detected predecessors; Vilím's pushes to `ect(Ω)`, the
> *set's* earliest completion time, which is larger when the predecessors cannot
> all fit before it. That is an energy statement about a set, so it is the same
> obstacle as the overload check below, not a variation on the pairwise proof.

## The rule

For a task `j`, let `Ω = { k : k ≪ j is detectable }` — the predecessors forced
before `j` by bounds alone, `lb(s_j) + lb(l_j) > ub(s_k)` in the merged rule's
spelling.

#734 pushes `lb(s_j)` to `max_{k∈Ω} eet_k`: the latest single predecessor's
earliest end. Vilím pushes to

```
ect(Ω) = max over Ω' ⊆ Ω of  ( est(Ω') + Σ_{i∈Ω'} p_i )
```

which is larger exactly when the predecessors **cannot all fit** before that
point — the set needs `Σ p_i` of room after `est(Ω')`, and if that runs past
every individual `eet_k` then `j` cannot start until it clears. At capacity one
the energy is just `Σ_{i∈Ω'} lb(l_i)`. Vilím's Θ-tree computes `ect(Ω)` and the
maximising subset in `O(n log n)`; the certificate does not have to follow the
algorithm, only the conclusion.

Take the exact form from Vilím rather than from this summary.

## Why it is not a variation on #734's proof

#734's result was that **the detection condition is precisely the condition
under which the refuting `pol`'s degree is positive** — the rule and its
certificate are the same statement, which is what made it "new strength for
free" and needed no new machinery. That property is specific to the pairwise
form: one `pol` over one pair's separation row and two bound-literal
definitions.

`ect(Ω)` is a statement about the total energy of a *set* over a window. It has
no pairwise witness, so there is no pair whose separation row carries it, and
the pairwise vocabulary in `disjunctive-proof-logging.md` cannot reach it. It
needs the same energy argument the overload check and edge-finding need.

[...]

Issue #751, opened 2026-08-17
Certified edge-finding for Disjunctive: the capacity-one case against the order-based encoding


ciaranm:
Successor to #733, which was auto-closed as COMPLETED at 21:27:15 on 16 August
2026 — one second after PR #743, *"Certified edge-finding for **Cumulative**,
in both directions"*, merged at 21:27:14. #733 was titled "for **Disjunctive**,
then Cumulative" and #729 ordered it that way deliberately. The cumulative half
landed; the disjunctive half was never started, and the tracker has shown no
open work for it since. Filed fresh rather than reopened so the closed issue
keeps its (accurate) cumulative history.

## Where the coverage table actually stands

#729's table, corrected to `origin/main @ d04983f9`:

| rule | Disjunctive | Cumulative |
|---|---|---|
| time-tabling | ✓ | ✓ |
| overload check | ✓ #737–#741 | ✓ |
| detectable precedences | ✓ #734 (pairwise, not Vilím's `ect(Ω)`) | — |
| **not-first / not-last** | **✗ #752** | ✓ #745 (a weakening — #746) |
| **edge-finding** | **✗ this issue** | ✓ #743 |
| TTEF | — | ✓ #744 |
| KAOC / TTHE-OC | — | ✓ #747 |
| optional tasks | ✓ #736 | ✓ |
| variable duration / demand in the energy set | n/a | ✓ #748, #749 |

Two empty cells, both on the disjunctive side, both from the same pair of
early-closed issues.

## The rule

For a set Ω and `j ∉ Ω`, if `Ω ∪ {j}` cannot fit in Ω's window then `j` cannot
end within it, so `j` must come after everything in Ω and `lb(s_j)` rises to
`ect(Ω)`. At capacity one the energy is just `Σ_{i∈Ω} lb(l_i)` against the
window's width. Vilím's Θ-Λ-tree computes all such pruning in `O(n log n)`; the
certificate does not have to follow the algorithm, only the conclusion.

Take the exact form from Vilím or from Baptiste, Le Pape and Nuijten rather
than from this summary.

## What has changed since #733 was written, and why it should now be cheap

#733 recorded the obstacle from `dev_docs/cumulative-proof-logging.md`:

> A *set* of tasks blocks an interval, not a single task at a single time. The
> pol arithmetic would need to sum across the set; the chain idea no longer
> fits directly.

and asked whether the energy statement and the bound chain compose. **On the
[...]

Issue #730, opened 2026-08-11
Disjunctive has no overload check, though Cumulative does and disjunctive is its capacity-one case


ciaranm:
Step 1 of #729. `Disjunctive` runs three inferences, all time-tabling strength — the mandatory-overflow contradiction, the lb-push and the ub-push (`dev_docs/disjunctive-proof-logging.md`, "The three inferences"). It has **no overload check**.

`Cumulative` has one: `prepare_cumulative_overload_check` plus the (OC')/(TTOC) conflict rule added by #542/#655, sitting between time-tabling's overflow scan and its bound pushes, and consuming the window-energy lemma in `constraints/innards/window_energy.{hh,cc}`.

A disjunctive resource is the capacity-one case of a cumulative one, so as things stand **a model gets a weaker certified inference by using the more specific global**, which is backwards. `InferredDisjunctive` already routes some of this the other way — it posts a discovered clique *as* a capacity-one `Cumulative`, which then does get the overload check.

## The rule

For a set Ω of tasks on the resource, if

```
Σ_{i ∈ Ω} lb(l_i)  >  max_{i ∈ Ω} ub(s_i) + lb(l_i)  −  min_{i ∈ Ω} lb(s_i)
```

then Ω cannot fit in the window it must occupy, and the state is infeasible. Conflict-only, like (TTOC): no bound pruning, which keeps it the cheapest possible first step.

## Why this should be mostly reuse

The window-energy lemma derives, per task and per window, `Σ_{t ∈ [a,b)} active_{i,t} ≥ activity` — energy a task must spend inside a window. Summing that over Ω against a supply of `capacity × width` is exactly what `Cumulative`'s overload check already does, and at capacity one the supply is the width. So the arithmetic is the existing one specialised, not a new one.

What is *not* automatic is the vocabulary. `Cumulative` argues over `active` flags reified per (task, time); `Disjunctive`'s encoding is pairwise — `before_{i,j}` separation clauses — and `dev_docs/disjunctive-proof-logging.md` documents that vocabulary as the thing its three inferences are written in. So the real work is deciding whether to

- **(a)** give `Disjunctive` the per-time `active` scaffolding the lemma needs, which duplicates `Cumulative`'s encoding and costs OPB rows on every disjunctive model whether or not the rule fires, or
- **(b)** derive the energy statement from the pairwise rows instead, which keeps the encoding as it is but needs a new derivation, or
[...]

PR #761, opened 2026-08-18
disjunctive: Vilim's set-based detectable precedence, measured and certified (#754)

ciaranm:
Stacked on #760. **Stage one of #754: the measurement, not the certificate** — and unlike #757, this one says *build it*.

## Why measurement first

#734 shipped the pairwise form, pushing `lb(s_j)` to `max_{k∈Ω} ect_k` over the detected predecessors. Vilím's pushes to the **set's** earliest completion time:

```
ect(Ω) = max_{Ω' ⊆ Ω} ( est(Ω') + p(Ω') )        lb(s_j) ← ect(Ω)
lst(Ω) = min_{Ω' ⊆ Ω} ( lct(Ω') − p(Ω') )        ub(s_j) ← lst(Ω) − p_j
```

larger exactly when the predecessors **cannot all fit** before that point. #754's stage zero counted *firings past a bound* and found `ect(Ω)` reaching past everything on 2.9% of nodes. #757 then showed that firing counts are not search — a 37% detection gap there bought 0.6% of the recursions — so this is the propagation measurement stage zero did not do.

`DisjunctiveRules::detectable_precedences_set`, by a **left-cut scan** rather than a Θ-tree: the maximum over subsets is attained at a cut, since taking every predecessor with `est ≥ a` never lowers `est(Ω')` below `a` and only adds duration. Uncertified, throwing under `--prove` — but only where its target is past what #734's own justification supports, a target the domain clip caps being one that justification still covers.

## Sound, checked against enumeration

A rule that removed a solution would measure as a large win, so the sweep cannot make this check. Both scans transcribed back out of the propagator, every push reaching past the pairwise target checked against a full enumeration of its instance's solutions:

| draw | pushes | past the pairwise target | removed a solution |
|---|---|---|---|
| 30,000 instances, 4 tasks | 125,060 | 33,307 | **0** |
| 20,000 instances, 5 tasks | 114,500 | 41,566 | **0** |
| 12,000 instances, 6 tasks | 87,214 | 37,710 | **0** |

## And it pays

Same 68 generated instances and 60 s timeout as #756's, #758's and #760's tables. `ef` and `ef + nfnl` reproduce their rows **to the digit**, which is the comparability check.

| arm | against | summed | median | geomean | better | closed |
|---|---|---|---|---|---|---|
| **set-based** | pairwise | **0.547x** | **0.386x** | **0.289x** | 34/36 | **44/68** |
| edge-finding | pairwise | 0.169x | 0.127x | 0.099x | 36/36 | 46/68 |
| **ef + set-based** | ef | **0.820x** | **0.941x** | **0.824x** | **23/36** | **47/68** |
| ef + nfnl | ef | 1.024x | 1.000x | 1.001x | 1/36 | 46/68 |
| ef + nfnl + set-based | ef + nfnl | 0.807x | 0.955x | 0.827x | 23/36 | 46/68 |

**On its own it is worth nearly half the search** and closes three more instances — a lot for a rule that adds a linear scan where edge-finding adds a cubic one.

**On top of edge-finding it is the first rung here that adds anything.** Better on **23 of 36**, worse on 6, unchanged on 7, and one more instance closed — including `--size 18 --seed 5`, which *no other arm closes at all*. Against not-first/not-last's 1 of 36 and the published detection's 4 of 36, both at a median of exactly 1.000x, that is a different verdict rather than a smaller one.

**Quote the median and the geomean, not the summed figure.** The largest instance carries **52%** of the summed total by itself, and dropping it takes 0.820x to 0.971x. The rule's real size is 0.941x at the median with a long tail — the best instance goes to 0.128x.

## What comes next

The certificate, which the dev doc now records in full. It is **#757's shape mirrored**: a window `[est(Ω'), T − 1)` whose *right* edge the negated conclusion derives, where #757 derives the left. Each `i ∈ Ω'` gives the two-literal clause `[s_j ≥ T] ∨ ¬[s_i ≥ T − p_i]` from #734's own refutation pol, and `T − p_i` is exactly the high guard of `i`'s guarded window-energy row over that window — so the low guard falls to the reason, the high guard to the clause, and what survives derives the conclusion rather than assuming it.

Simulate it first, as #730, #751 and #757 all did.

https://claude.ai/code/session_01PAZYGJdsP3dmWAepRGi5T5

### DIFF
--- a/dev_docs/disjunctive-proof-logging.md
+++ b/dev_docs/disjunctive-proof-logging.md
@@ -723,6 +723,97 @@ one measured here to be worth nothing. That is not a reason to close
 #754, whose gap is against a different rule's fixpoint rather than a
 weaker form of its own, but it is the number to beat before building it.
 
+## The set-based detectable precedence, measured before it is certified (#754)
+
+#734 pushes `lb(s_j)` to `max_{k∈Ω} ect_k` over the detected
+predecessors — the latest *single* predecessor's earliest end. Vilím's
+rule pushes to the **set's** earliest completion time,
+
+```
+ect(Ω) = max_{Ω' ⊆ Ω} ( est(Ω') + p(Ω') )        lb(s_j) ← ect(Ω)
+lst(Ω) = min_{Ω' ⊆ Ω} ( lct(Ω') − p(Ω') )        ub(s_j) ← lst(Ω) − p_j
+```
+
+which is larger exactly when the predecessors **cannot all fit** before
+that point. `DisjunctiveRules::detectable_precedences_set`, off by
+default, computed by a left-cut scan rather than a Θ-tree: the maximum
+over subsets is attained at a cut, since taking every predecessor with
+`est ≥ a` never lowers `est(Ω')` below `a` and only adds duration, so
+one pass over the ests sorted descending gives it.
+
+**It is deliberately uncertified and throws under `--prove`**, and only
+when the target it takes is past what #734's pairwise justification
+supports — a target the domain clip caps is one that justification still
+covers. This is #757's discipline: a stronger rule with no certificate
+is measured before it is built, because the certificate is the expensive
+part and a detection gap need not be a search gap.
+
+### It is sound, checked against enumeration
+
+A rule that removed a solution would measure as a large win, so the
+sweep cannot make this check. Both left-cut scans were transcribed back
+out of the propagator and every push reaching past the pairwise target
+checked against a **full enumeration** of its instance's solutions:
+
+| draw | pushes | past the pairwise target | removed a solution |
+|---|---|---|---|
+| 30,000 instances, 4 tasks | 125,060 | 33,307 | **0** |
+| 20,000 instances, 5 tasks | 114,500 | 41,566 | **0** |
+| 12,000 instances, 6 tasks | 87,214 | 37,710 | **0** |
+
+### And unlike #757 it is worth building
+
+The same 68 instances and 60 s timeout as the tables above:
+
+| arm | against | summed | median | geomean | better | closed |
+|---|---|---|---|---|---|---|
+| **set-based** | pairwise | **0.547x** | **0.386x** | **0.289x** | 34/36 | **44/68** |
+| ef | pairwise | 0.169x | 0.127x | 0.099x | 36/36 | 46/68 |
+| **ef + set-based** | ef | **0.820x** | **0.941x** | **0.824x** | **23/36** | **47/68** |
+| ef + nfnl | ef | 1.024x | 1.000x | 1.001x | 1/36 | 46/68 |
+| ef + nfnl + set-based | ef + nfnl | 0.807x | 0.955x | 0.827x | 23/36 | 46/68 |
+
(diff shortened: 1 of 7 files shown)


ciaranm:
## Stage three: the propagator, and it is certified

The certificate is in, and `--prove` no longer throws. It is **#757's mechanism, and both orientations of it**:

| push | derived window | edge the negated conclusion supplies | guard the clause discharges |
|---|---|---|---|
| lb, threshold `T = est(Ω′) + p(Ω′)` | `[est(Ω′), T − 1)` | the **right** one | the **high** guard |
| ub, threshold `L = lct(Ω′) − p(Ω′)` | `[L + 1, lct(Ω′))` | the **left** one | the **low** guard |

Either way the rows cited are the **standard contained-task guarded ones** — guards `(a, b − p_k + 1)`, the same call edge-finding makes — so the only new step is that one of the two guards is discharged by a derived two-literal clause instead of by the reason. That clause carries the conclusion literal at the guard's coefficient, so the conclusion accumulates across the cut and the summed pol *derives* it.

`precedence_clause` is four pols: #734's own refutation of the reverse ordering (whose degree is positive exactly when the precedence is detectable — the same statement), the separation clause forcing the surviving ordering, that ordering's arithmetic against the two thresholds at degree exactly one, and discharging the ordering. `emit_before_pol` gained an optional divisor for the first: a row that will be *added* to another wants to be a clause, and saturation alone leaves its coefficients at the degree.

## Testing

`disjunctive_set_precedences_test`: `sharp` for the lb push, `mirror` for the ub one, a control apiece saying the pairwise rules make neither, and a `loose` fixture a unit the other side of the margin. **Both pushes land exactly where enumeration puts the bound** — which not-first/not-last never manages — so the conclusion is checked against the truth and not only against being sound.

**Six mutation lanes, all rejected.** `drop_clause` and `rup_clause` are the two aimed at what is new; the second says unit propagation cannot reach the clause on its own, the same cross-variable limit `RupOverloadBridge` finds for the bridge. The mutation fixture is generated with `|Ω′| = 3`: on a two-task cut the fold is a single bridge row the closing RUP reconstructs and two of the six lanes survive — the same threshold the simulation found.

`--search` verified **60 generated instances**, the rule firing on 51, with no solution-count mismatch. All seven disjunctive test binaries pass.

## The trap, which no sweep would have caught

The first version read the bounds its arithmetic needed back out of `state` inside the justification, and **every generated proof above a handful of tasks was rejected**. By the time a justification runs, an earlier push in the same propagation has landed, so the state holds a bound *the reason does not support*, and a `pol` built on it is arithmetic about a fact nothing has established. The ub push already carried a comment recording exactly this for #734's own certificate.

Every bound is now captured at detection time and carried into the closure — that is what `SetTask` is for. Worth stating as a rule, and the dev doc now does: **a justification may read the reason and the model, and nothing else.**

Also: unlike #757 the final division is *not* load-bearing here, the surplus being exactly one, so the closing RUP reads the pol off as edge-finding's does.

## Still open

The rule is **off by default**, and for one measured-nothing reason: the derived windows are keyed on the detected set, so their flags, bridge rows, folds and guarded rows share nothing with the energetic sweep's. What that costs per firing is the one number this stack does not have.

PR #783, opened 2026-08-27
Add gcspy PyPI release workflow, docs, and bump to 0.1.10

mmcilree:
Adds an automated PyPI release path for the `gcspy` Python bindings, plus docs, and bumps the version to 0.1.10.

## What's here
- **`.github/workflows/release-gcspy.yml`** — builds the sdist (the portable fallback: it bundles the whole solver source, so anyone with a supported compiler can `pip install gcspy`) plus best-effort wheels for macOS (arm64 + x86_64) and manylinux_2_28 x86_64, CPython 3.10–3.13. Publishes via **Trusted Publishing (OIDC)** — no stored token — on a `gcspy-v*` tag push, with a manual TestPyPI dry-run option.
- Linux wheels install **gcc-toolset-13** inside manylinux_2_28 (GCC 13 = our compiler floor) and **static-link libstdc++/libgcc** so the wheel doesn't require a newer runtime than the manylinux policy guarantees. Wheels are built from the sdist tarball, so they double as a check that it builds.
- Wheels are **best-effort**: a failing wheel lane still ships the sdist plus the wheels that built.
- **`dev_docs/releasing-gcspy.md`** — procedure (bump → TestPyPI dry-run → tag), the compiler-floor rationale, and the one-time PyPI Trusted Publishing setup. Indexed in `dev_docs/README.md`, pointed to from `CLAUDE.md`.
- **Version bump to 0.1.10** in `python/pyproject.toml` (0.1.9 is already on PyPI).

## Before the first release
Trusted Publishing needs a one-time PyPI-side config (repo + workflow `release-gcspy.yml` + environment `pypi`/`testpypi`) — see the dev_docs note. Recommend a TestPyPI dry-run first; the manylinux wheel lane is the part most likely to need a first-run fixup.

### DIFF
--- a/.github/workflows/release-gcspy.yml
+++ b/.github/workflows/release-gcspy.yml
@@ -44,14 +44,30 @@ jobs:
       with:
         python-version: '3.12'
 
-    # scikit-build-core builds the sdist in an isolated environment (no compiler
-    # needed here). The package lives in python/, but its cmake.source-dir is the
-    # repo root, so the tarball bundles the whole solver source -- that is what
-    # lets a plain `pip install gcspy` compile on a machine with no wheel.
+    # The sdist must be rooted at the REPO ROOT, not python/: it has to carry the
+    # whole solver source (gcs/, the root CMakeLists.txt, ...) so a from-source
+    # `pip install gcspy` can compile. python/pyproject.toml uses cmake.source-dir
+    # = ".." for the in-tree developer install (`pip install ./python`), but an
+    # sdist cannot include files above its own pyproject, so building from python/
+    # packages python/ alone -- an 8 KB stub that cannot build. Synthesise a root
+    # pyproject from the dev one instead: drop cmake.source-dir (so it defaults to
+    # the repo root, now inside the tarball) and force GCS_BUILD_TESTS=OFF (at the
+    # root it defaults ON and would compile the whole test suite on install). The
+    # two edits are asserted below so a formatting change that breaks them fails
+    # loudly rather than shipping a broken sdist.
+    - name: Assemble the release pyproject at the repo root
+      run: |
+        sed -e '/cmake\.source-dir/d' \
+            -e 's/cmake\.args = \[/cmake.args = ["-DGCS_BUILD_TESTS=OFF", /' \
+            python/pyproject.toml > pyproject.toml
+        echo "--- generated root pyproject.toml ---"; cat pyproject.toml
+        grep -q -- '-DGCS_BUILD_TESTS=OFF' pyproject.toml || { echo "::error::GCS_BUILD_TESTS flag not injected"; exit 1; }
+        ! grep -q 'cmake\.source-dir' pyproject.toml     || { echo "::error::cmake.source-dir not stripped"; exit 1; }
+
     - name: Build sdist
       run: |
         python -m pip install --upgrade build
-        python -m build --sdist --outdir dist python
+        python -m build --sdist --outdir dist .
 
     # Confirm the sdist actually compiles and imports with a stock distro compiler
     # (ubuntu-24.04 ships GCC 13, this project's oldest supported compiler). This
@@ -78,8 +94,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { os: macos-14, archs: arm64 }    # Apple silicon
-          - { os: macos-13, archs: x86_64 }   # Intel Macs (older Apple Clang -- riskier lane)
+          - { os: macos-15, archs: arm64 }         # Apple silicon
+          - { os: macos-15-intel, archs: x86_64 }  # Intel Macs
           - { os: ubuntu-24.04, archs: x86_64 }
 
     steps:
--- a/.github/workflows/release-gcspy.yml
+++ b/.github/workflows/release-gcspy.yml
@@ -1,0 +1,176 @@
+name: Release gcspy
+
+# Builds and publishes the gcspy Python bindings to PyPI. Produces a source
+# distribution (the portable fallback -- anyone with a supported compiler can
+# build it) plus prebuilt wheels for macOS (arm64 + x86_64) and manylinux x86_64,
+# then uploads them.
+#
+# There are two ways to trigger a release:
+#   * Push a tag matching gcspy-v*  (e.g. gcspy-v0.1.10)  -> publishes to real PyPI.
+#   * Run the workflow manually (Actions -> Release gcspy -> Run workflow) and pick
+#     "testpypi" (the default, a safe dry run) or "pypi".
+#
+# The version that gets published is whatever is in python/pyproject.toml, NOT the
+# tag name -- keep them in step. See dev_docs/releasing-gcspy.md for the full
+# procedure, the one-time PyPI Trusted Publishing setup, and how to iterate on a
+# failing wheel build.
+
+on:
+  workflow_dispatch:
+    inputs:
+      target:
+        description: "Where to publish"
+        type: choice
+        options: [testpypi, pypi]
+        default: testpypi
+  push:
+    tags: [ "gcspy-v*" ]
+
+# Never let two release runs upload at once.
+concurrency:
+  group: release-gcspy
+  cancel-in-progress: false
+
+jobs:
+  build-sdist:
+    name: Build sdist
+    runs-on: ubuntu-24.04
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.12'
+
+    # scikit-build-core builds the sdist in an isolated environment (no compiler
+    # needed here). The package lives in python/, but its cmake.source-dir is the
+    # repo root, so the tarball bundles the whole solver source -- that is what
+    # lets a plain `pip install gcspy` compile on a machine with no wheel.
+    - name: Build sdist
+      run: |
+        python -m pip install --upgrade build
+        python -m build --sdist --outdir dist python
+
+    # Confirm the sdist actually compiles and imports with a stock distro compiler
+    # (ubuntu-24.04 ships GCC 13, this project's oldest supported compiler). This
+    # is the path every from-source user takes, so a red here is a real release
+    # blocker even if the wheels are fine.
+    - name: Smoke-build the sdist from source
+      run: |
+        python -m pip install --verbose dist/*.tar.gz
+        python -c "import gcspy; print('gcspy import OK')"
+
+    - name: Upload sdist artifact
+      uses: actions/upload-artifact@v4
+      with:
+        name: dist-sdist
+        path: dist/*.tar.gz
+
+  build-wheels:
+    name: Wheels (${{ matrix.os }} / ${{ matrix.archs }})
+    needs: build-sdist
+    runs-on: ${{ matrix.os }}
+    # Best-effort: one platform failing must not sink the whole release. The
+    # publish job runs whatever wheels succeeded plus the always-present sdist.
+    strategy:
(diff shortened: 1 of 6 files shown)


mmcilree:
Once this is merged we should be able to more easily publish a new gcspy release, and unblock #783 as well.

ciaranm:
Pre-merge check pass. CI is fully green (all ten checks, Windows included), the branch is on current `main`, the version bump is correct (PyPI has 0.1.9 from 2025-10-23 while the repo read 0.1.8, so 0.1.10 is right), and the docs are properly indexed. But there's a blocker: **the sdist this workflow builds contains no solver source**, so the release path can't succeed as written.

## The sdist is a 8 KB stub

The design rests on this claim, made in the workflow comment and again in `dev_docs/releasing-gcspy.md` under "What a release produces":

> The package lives in `python/`, but its `cmake.source-dir` is the repo root, so the tarball bundles the whole solver source

Running the workflow's own step on a clean clone of this branch:

```console
$ python -m build --sdist --outdir dist python
*** scikit-build-core 1.0.3 (sdist)
Successfully built gcspy-0.1.10.tar.gz          # 8,343 bytes

$ tar tzf dist/gcspy-0.1.10.tar.gz
gcspy-0.1.10/CMakeLists.txt
gcspy-0.1.10/gcspy.cc
gcspy-0.1.10/gcspy.hh
gcspy-0.1.10/pyproject.toml
gcspy-0.1.10/python_test.py
gcspy-0.1.10/PKG-INFO
```

Six files — only `python/`'s own contents, nothing under `gcs/`. `cmake.source-dir = ".."` points *outside* the packaged tree, so scikit-build-core packages `python/` alone, and the CMake source dir then resolves outside the extracted tarball. The "Smoke-build the sdist from source" step duly fails:

```console
$ pip install --verbose dist/gcspy-0.1.10.tar.gz
RUN: /usr/bin/cmake -S.. -Bbuild -DCMAKE_BUILD_TYPE:STRING=Release ...
CMake Error: The source directory "/tmp" does not appear to contain CMakeLists.txt.
*** CMake configuration failed
ERROR: Failed to build installable wheels for some pyproject.toml based projects (gcspy)
```

Not a scikit-build-core regression: same stub from 1.0.3 and from 0.11.6 (the version current when 0.1.9 was published), and the same from `main` as from this branch. It's a pre-existing property of `python/pyproject.toml` that this PR doesn't change but does depend on. The config is fine for the developer path — `pip install ./python` from a checkout, where `..` genuinely is the repo root — it just can't produce a distributable sdist.

The failure mode is at least safe rather than harmful: `build-sdist` goes red, `build-wheels` (which needs it) is skipped, and `publish` is gated on `needs.build-sdist.result == 'success'`, so a `gcspy-v0.1.10` tag push would publish nothing rather than publishing something broken.

**What the working 0.1.9 did differently**, which is the useful part. The published `gcspy-0.1.9.tar.gz` is 253 KB / 331 files with the full `gcs/` tree and a root `CMakeLists.txt`. Its bundled `pyproject.toml` sits at the *root* of the tarball, and:

```toml
cmake.args = ["-DGCS_ENABLE_PYTHON=ON", "-DGCS_ENABLE_XCSP=OFF", "-DGCS_ENABLE_MINIZINC=OFF", "-DGCS_ENABLE_EXAMPLES=OFF",
"-G Unix Makefiles"]
#cmake.source-dir = ".."
```

— `cmake.source-dir` commented out, so the sdist root *is* the repo root. That pyproject isn't in the repo, so the manual process being automated here differs from the in-repo config in exactly the load-bearing place. Either the pyproject needs to live at (or the sdist be rooted at) the repo root as 0.1.9's was, or `tool.scikit-build.sdist.include` has to pull the solver sources in with `cmake.source-dir` resolving inside the tarball. Worth settling with a local `python -m build --sdist` and a `pip install` of the result before wiring CI around it, since that's a few seconds and catches exactly this.

## Two dead runners in the wheel matrix

- **`macos-13`** (the x86_64 lane) is **retired** — there's no macOS 13 row left in `actions/runner-images`' README, and actions/runner-images#13402 records it. That lane can't get a runner at all.
- **`macos-14`** (the arm64 lane) is **deprecated**: actions/runner-images#13518 gives deprecation from 2026-07-06 and fully unsupported from **2026-11-02**, roughly two months out.

Our own CI already uses `macos-15` and `macos-26`; the current x64 labels are `macos-15-intel` / `macos-26-intel`. The published wheels have all been `macosx_12_0_arm64`, so `MACOSX_DEPLOYMENT_TARGET=12.0` stays right on `macos-15`.

## Smaller notes

- **Nothing in PR CI exercises this file** — it triggers only on `workflow_dispatch` and `gcspy-v*` tags, so the ten green checks say nothing about the 176 lines added. That's why both of the above got this far, and it's an argument for the TestPyPI dry run being mandatory rather than recommended.
- `CIBW_BUILD` is cp310–cp313. Python 3.14 has been out since October 2025, so 3.14 users fall back to the sdist — currently the broken path.
- Trusted Publishing prerequisites aren't in place yet: the repo has **zero** GitHub Environments, and the PyPI-side publisher config is manual. Merging arms the tag trigger, so worth doing the PyPI setup and the dry run before any `gcspy-v*` tag exists.

Everything else on the checklist came back clean: GCC 13 as the floor matches `README.md` and `CLAUDE.md`, the AI declaration is present both ways, no stray files, no doc-vs-code contradiction elsewhere, and the job graph and publish conditions are correct.

<sub>Posted from the pre-merge check pass, which is AI-assisted (Claude Opus 5); the sdist and runner findings above were reproduced locally rather than inferred.</sub>

ciaranm:
The sdist fix works — I verified it end to end rather than by reading. On a clean worktree of `295bc558` I ran the workflow's new synthesis step verbatim; both its assertions pass, and the resulting sdist is **1 915 571 B with 524 `gcs/` files and 1125 total**, carrying the root `CMakeLists.txt` and `pyproject.toml`, against the 8 KB / 6-file stub before. `pip install`ing that tarball into a clean venv exits 0 and `import gcspy` gives `['GCS']`. The `GCS_BUILD_TESTS=OFF` injection is genuinely load-bearing too: the root option defaults to `${PROJECT_IS_TOP_LEVEL}`, so without it every `pip install` would compile the whole test suite. `macos-15` / `macos-15-intel` are the right labels.

While confirming that, though, I hit something else, and it only bites on the wheels — so it has never mattered before and would be easy to ship without noticing.

## The wheels would be compiled for the CPU that happened to build them

The build compiles with **`-march=native`**:

```
g++ ... -O3 -DNDEBUG -std=c++23 -fPIC -W -Wall ... -march=native -c .../gcs/proof.cc
```

That comes from `CMakeLists.txt:264`:

```cmake
# Native CPU tuning is worthwhile for release performance but unhelpful for debugging.
if(COMPILER_SUPPORTS_MARCH_NATIVE)
    add_compile_options($<$<CONFIG:Release>:-march=native>)
endif()
```

and scikit-build-core configures with `-DCMAKE_BUILD_TYPE:STRING=Release`, so it applies to every gcspy build.

### What the flag actually does

`-march=native` tells GCC: *detect the CPU you are compiling on, and freely emit any instruction it supports.* It is not a hint or a tuning preference — the compiler will emit AVX-512, BMI2, FMA and so on directly into the object code, with no runtime check and no fallback path. A binary built that way runs only on a CPU that has every instruction the builder had. On anything older it dies with **SIGILL — illegal instruction**, typically at import or the first time the affected code runs.

For a **source** build that is exactly what you want, and it is presumably why the flag is there: the user compiles on their own machine, so "native" is their machine, and they get a solver tuned for it.

For a **wheel** it is the opposite. A wheel is a prebuilt binary compiled on one machine and installed on thousands of others. `manylinux_2_28_x86_64` is a promise about glibc and the x86-64 baseline; it says nothing about CPU features, so pip will happily install that wheel on any x86-64 machine. Compiled with `-march=native` on a GitHub runner, the wheel silently acquires whatever that runner's CPU supported as a hard requirement — and GitHub does not pin the runner CPU model, so **the minimum CPU a published wheel demands becomes an accident of which machine GitHub allocated that day, and can differ from release to release.**

### What that looks like concretely

The module I just built from this branch's sdist, on this machine (a Zen 5, which has AVX-512), disassembles to:

| instruction class | occurrences |
|---|--:|
| AVX-512 (`%zmm` registers) | 12 176 |
| AVX-512 masking (`%k` registers) | 188 |
| AVX / AVX2 (`%ymm` registers) | 23 958 |
| BMI2 (`shlx` / `sarx` / `bzhi` / `pdep` / `pext`) | 333 |

```
vmovdqa64 0x59bf1d(%rip),%zmm0
vmovdqu64 %zmm0,0x30(%rsp)
```

As a local build that is fine and fast. Uploaded to PyPI as a wheel it would SIGILL on any CPU without AVX-512 — which is most consumer x86 hardware, since Intel disabled AVX-512 on consumer parts from Alder Lake onward and AMD only added it in Zen 4. The user gets a crash with no useful message, and no way to tell that the wheel was the problem.

### Why nothing in the pipeline would catch it

This is the part worth dwelling on, because every check in the release path looks like it ought to find this and none of them can:

- **`auditwheel`**, which `cibuildwheel` runs on every manylinux wheel, inspects shared-library dependencies and glibc symbol versions. It does not decode instructions, so an AVX-512 wheel passes as cleanly as a baseline one.
- **`CIBW_TEST_COMMAND`** (`python -c "import gcspy"`) runs on the machine that just built the wheel, which by construction supports every instruction in it. It always passes.
- **The sdist smoke-build** likewise builds and imports on one machine.
- **PR CI** never runs this workflow at all — it only fires on `workflow_dispatch` and `gcspy-v*` tags.

So the first evidence would be a bug report from a user whose machine is older than the runner's, and it would be hard to attribute.

### Why it is new here

The wheels on PyPI today are all `macosx_12_0_arm64`, hand-built on one Mac. Adding the manylinux x86_64 and macOS x86_64 lanes is what turns this from a non-issue into a real one. Worth checking whether `CHECK_CXX_COMPILER_FLAG(-march=native)` even passes under Apple Clang on arm64 — that check runs per lane, so the arm64 lane may or may not be affected, and I could not test it here.

### Options, and it is your call

There is no knob for this today — `-march=native` is unconditional for Release on non-MSVC, and an environment `CXXFLAGS` will not override it, because `add_compile_options` puts the flag *after* `CMAKE_CXX_FLAGS` on the command line and the last `-march` wins. So:

1. **Add a CMake option** (say `GCS_NATIVE_ARCH`, defaulting ON so nothing changes for developers or sdist users) and have the wheel lanes pass it OFF via `CIBW_ENVIRONMENT`. This is the clean fix and keeps everyone's local builds as fast as they are now. It does mean touching the root `CMakeLists.txt`, so it is a separate change from this PR.
2. **Ship the sdist only for now** — drop the wheel lanes until such an option exists. Costs nothing, and sdist users arguably get the better artifact anyway, since they compile natively for their own machine.
3. Ship the wheels as they are and accept it. I would not, but you have far more context than I do on who installs gcspy and on what.

One last note for whichever way you go: `-march=native` in the *sdist* path has a mild version of the same trap, worth a line in `releasing-gcspy.md`. `pip install gcspy` inside a `docker build` compiles for the build host and bakes that into the image, so the image can SIGILL when the container is later run somewhere else. That is a user-side consequence rather than a defect here, but it is the sort of thing people lose an afternoon to.

<sub>Posted from the pre-merge check pass, which is AI-assisted (Claude Opus 5). The sdist verification, the build flags and the disassembly counts above were all produced locally on this branch rather than inferred.</sub>

mmcilree:
Ah yeah I think some version of this problem was why I only ever did sdist builds before. 

I'll be a good bot messenger though and see if my Claude can make more reliable wheels anyway 🤪

Issue #637, opened 2026-07-31
Graph constraints: `connected`/`dconnected` and the reachability–tree family have no propagator


ciaranm:
`Circuit` is the only graph constraint the solver has. Every other member of
MiniZinc's `globals.graph` group falls through to the stdlib decomposition:
`connected`, `dconnected`, `reachable`, `dreachable`, `subgraph`, `tree`,
`dtree`, `path`, `dpath`, `bounded_path`, `bounded_dpath`, `steiner`,
`dsteiner`, `dag`, `network_flow`, `weighted_spanning_tree`, `subcircuit`.

`minizinc/mznlib/` overrides exactly one graph predicate — `fzn_circuit.mzn` —
and `minizinc/fzn_glasgow.cc` names none of the others. XCSP3-core defines none
of this family, so it is a MiniZinc-side gap plus a Glasgow-extension
opportunity rather than an XCSP3 one.

## What the fall-through actually costs

The chain for `connected`, from the 2.9.7 stdlib:

```
connected(from, to, ns, es)
  -> fzn_connected:  let { var NODE: r } in reachable(from, to, r, ns, es)
  -> fzn_reachable:  duplicate every edge in both directions, then
                     fzn_dreachable(N, 2E, dfrom, dto, r, ns, des)
  -> fzn_dreachable: a parent + distance labelling, plus subgraph
```

So `connected` on a graph with `N` nodes and `E` undirected edges — `2E`
directed after the doubling — flattens to:

- `2N` auxiliary integer variables, `parent[n] ∈ 0..N` and `dist[n] ∈ 0..N-1`,
  plus the existentially quantified root `r ∈ 1..N`;
- `2N` half-reified **variable-indexed element** constraints: `ns[parent[n]]`
  over the node array, and `dist[parent[n]]` over the distance array;
- roughly `4N` further half-reified implications tying `ns`, `parent` and
  `dist` together, one of which (`ns[n] -> (n = r \/ parent[n] > 0)`) is a
  reified equality against the *variable* root;
- one half-reified disjunction per node over its in-edges, each term a reified
  equality `parent[n] = from[e]` — `2E` terms across the graph;
- `4E` `subgraph` implications over the doubled edge set.

The structural point is `dist`: the decomposition encodes reachability as a
**distance labelling**, so transitive closure becomes an arithmetic chain. Any
connectivity inference has to travel that chain — up to `N` element-plus-
[...]

PR #784, opened 2026-08-28
Add the Reachable / DReachable propagators (#637)

ciaranm:
Stage 1 of #637: `Reachable` and `DReachable`, the propagators behind MiniZinc's
`reachable`, `dreachable`, `connected` and `dconnected`.

## The issue's premise was wrong, and that decides the design

#637 says "the hard part is the proof logging, not the propagation", and offers
two routes: keep the stdlib's distance labelling and pay to re-derive the chain,
or build AM1/dominance scaffolding as `CircuitSCC` does. Neither is needed.

The Glasgow Subgraph Solver has logged connectivity as a bare `rup` for years
(`gss/innards/proof.cc`, `create_connected_constraints` /
`not_connected_in_underlying_graph`), against a pairwise repeated-squaring
encoding. That one is `O(n³ log n)`, which is fine at MCCS pattern sizes and no
good at 400 nodes — but it shows the property is available, and a cheaper
encoding keeps it.

The stdlib's labelling fails for a specific reason: unreachability is an
*induction*. A node in a cut-off region would need a parent in that region, and
its parent one too, forever, contradicted only because distances are bounded
below. Unit propagation does not do induction. Write the same labelling **unary**
— a breadth-first unfolding, one level of flags per possible step — and the
fixpoint becomes an explicit chain. Unit propagation over those levels *is* the
breadth-first search the propagator ran, so every inference is a plain `rup`
whose whole content is its reason.

The real cost is encoding size, `O(nodes × edges)`, and that is where this stops:
0.65 MB of `.opb` at 25 nodes, 16.4 MB at 121, `O(n⁴)` in a grid's side. A
different trade than the issue anticipated, and `dev_docs/connectivity-proofs.md`
is mostly about it.

## Measured, on hitori h5-1

`examples/hitori --connectivity decomposition|propagator|propagator-no-cuts|none`
posts the same requirement four ways, so this is one instance in one binary.

| `h5-1` | `decomposition` | `propagator` | `propagator-no-cuts` | `none` |
|---|--:|--:|--:|--:|
| recursions | 113 671 | **24** | 38 | 19 |
| `.opb` | 1 038 149 B | 649 792 B | 649 792 B | 68 327 B |
| `.pbp` | 815 217 012 B | 246 783 B | **129 545 B** | 97 774 B |
| veripb | 796.28 s | 0.14 s | **0.09 s** | 0.01 s |

3 300× on proof size and 5 700× on verify time against the decomposition. On
`h11-1` the decomposition does not solve inside 900 s *even without proofs*,
while the propagator takes 0.26 s; both propagator settings verify there (679 MB
in 1 485.3 s, and 93.7 MB in 626.7 s).

## It is GAC, and the tests check that

`reachable_test.cc` runs every case through `solve_for_tests_checking_gac`, which
asserts at every search node that every remaining value has a support. Both
spellings pass.

Getting there needed the cut-vertex and bridge forcing, which turns out not to be
a heuristic strengthening *towards* GAC — undirected, it is exactly the rest of
it. The two spellings need different algorithms, and the GAC check is what
validates the fast path against the slow one:

- undirected, one Tarjan pass gives articulation points and bridges together,
  with "would removing this separate the selected nodes" falling out as a count
  per DFS subtree. `O(nodes + edges)`.
- directed, the analogue is a dominator, but the root is existentially quantified
  and the witnessing selected node may differ per candidate root, so no single
  dominator tree answers it. This asks the question directly, per node and edge.

## Two things to look at

**The forcing is on by default and costs proof.** It wins on search (h11-1: 4 920
→ 1 932 recursions) and loses on proof (93.7 MB → 679 MB), because `connected`
leaves the root existentially quantified and hitori decides it last, so nearly
every forcing pays a proof line per candidate root. Default follows
best-with-proofs-off; `with_cut_forcing(false)` and
`--connectivity propagator-no-cuts` keep the control. `proof-benchmarks.md` lists
the pair as a **label-groups candidate**, since the cost is large-database RUP
penalty on a 141 506-row `.opb`.

**Directed forcing is `O(nodes × roots × (nodes + edges))` per call**, and there
is no directed benchmark in the repo to size it. Correct — the GAC check confirms
that — but its behaviour at scale is genuinely unexercised.

## Deliberate deviations from the acceptance list

- **No `fzn_connected.mzn` / `fzn_dconnected.mzn`.** They would be dead files:
  the stdlib defines both in terms of `reachable`/`dreachable`, so they already
  route through the four overrides here. The `--fzn-pattern glasgow_reachable`
  guards on `connectedtest`/`dconnectedtest` are what prove the propagator is
  actually reached.
- **No `Connected`/`DConnected` classes.** The root must be a `Problem` variable,
  because search only branches on those and nothing determines a root by
  propagation. Callers quantify it themselves, as `fzn_connected`'s `let` does.
- `Subgraph` and the later stages (`tree`, `path`, `steiner`, `dag`) stay open
  under #637; the support matrix says so.

## Also here

Mutation lanes (`reachable_mutation_*`) drop one part of a reason each and insist
veripb notices. Two lanes verified at first, and both were fixture bugs worth
recording in the test: a node pinned in the *model* is pinned in the OPB too, so
dropping it from a reason changes nothing; and enumeration's solution-exclusion
clauses re-supply a dropped literal on a path graph, so the fixture forks and
cycles.

`--connectivity none` is kept as the attribution floor, but note that the
attribution which *sized* #637 is an `h5-1` fact: on `h11-1` the same relaxation
costs 446 479 recursions and 9.65 GB, two orders of magnitude worse than solving
the real problem, because dropping connectivity makes the objective much harder.

760/760 tests green, sanitize clean, `clang-format` clean.

https://claude.ai/code/session_01SJ8fRUWKCktdwGSFcUEdi1


ciaranm:
Pre-merge check pass. All ten checks green, merge-base is exactly current `main` (`87b24555`), the trial merge is clean, and a full release build of it reaches 100% with **0 warnings** and 126 test binaries — which for GCC 15.2 is a real result rather than an empty grep, the baseline having been zero since #710. Nothing here blocks merging. Two things to report, plus a nit.

## 1. The two new `--connectivity` modes have no ctest lane

`examples/hitori/CMakeLists.txt` isn't in the changeset, so the hitori lanes are still `hitori` (decomposition), `hitori-no-connectivity`, `hitori-dzn` and `hitori-dzn-answer`. Nothing runs `--connectivity propagator` or `--connectivity propagator-no-cuts`; `Connectivity::Propagator` appears only in `hitori.cc` and the docs.

Those are the modes the headline table comes from, and the only place the propagator meets a 25-node instance with an existentially quantified root — the `force_in` per-candidate-root proof path at scale. Run by hand on the trial merge, on the built-in instance:

| mode | `.opb` | `.pbp` | veripb | verdict |
|---|--:|--:|--:|---|
| `decomposition` (the lane that exists) | 453 158 B | 1 382 121 B | 0.22 s | `s VERIFIED BOUNDS -13 <= obj <= -13` |
| `propagator` | 260 637 B | 65 990 B | 0.00 s | same |
| `propagator-no-cuts` | 260 637 B | 56 741 B | 0.00 s | same |

Both new modes solve to the same optimum and verify, and either lane would be **cheaper than the decomposition lane already registered** — so this isn't a cost trade, and it isn't in the PR's list of deliberate deviations. A lane wants `--basename` to avoid the #562 proof-file race, as the existing ones have.

Worth being clear about the scope: the unit tests *do* exercise the open-root forcing path with proofs, since most fixtures leave the root a range. So this is a scale and integration gap, not an uncovered path.

Smaller, same family: `reachable_test` parses `--mutate=none`, but no lane registers it, so nothing asserts that the mutation fixture's *honest* proof verifies — the six mutation lanes only assert that veripb refuses.

## 2. The summaries quote the `propagator-no-cuts` column under the name `propagator`

Against the source table (`connectivity-proofs.md:263`), where `propagator` is 24 recursions / 246 783 B / 0.14 s and `propagator-no-cuts` is 38 / 129 545 B / 0.09 s, three places take the second column while naming the first:

- **`dev_docs/proof-benchmarks.md:118`** — the row is "`--connectivity decomposition` vs `--connectivity propagator`", and the figures given are "the native `Reachable` propagator **38** for **0.13 MB / 0.09 s**". All three numbers are the no-cuts column. Line 119 of the same file has them right, so the file disagrees with itself.
- **`dev_docs/connectivity-proofs.md:305`** — the `h11-1` DB-tax table's row is labelled `propagator` and carries 93.7 MB / 626.7 s, which the prose eight lines above attributes to "without the forcing" (with it: 679 MB / 1 485.3 s). The `.opb` half of that row is fine, since the encoding doesn't change with the forcing.
- **`dev_docs/README.md:216`** — "130 KB / 0.09 s", the no-cuts column again.

Since the forcing is on by default, a reader who runs `--connectivity propagator` gets 0.25 MB / 0.14 s. The conclusion is untouched: `connectivity-proofs.md`'s own headline uses the right column, and the arithmetic checks out — 815 217 012 / 246 783 = 3 303 for "3 300×", 796.28 / 0.14 = 5 688 for "5 700×". So does the rest of it: 1.9×, 7.2×, 2.4× and 103× / 1.24× are all exact, and the per-megabyte rates (12.4 against 0.15) give 83× for "about eighty times dearer to verify". Only the setting the numbers are pinned to is wrong.

One more in the same document: **`connectivity-proofs.md:253`** says "`examples/hitori --connectivity decomposition|propagator|none` posts the same requirement **three ways**" — directly above a **four**-column table that includes `propagator-no-cuts`. The PR description's version of that sentence says four ways, so the doc looks like the stale half.

## 3. Nit

`reachable_test.cc`: the comment "Edges pinned, so the subgraph rules have something to say." sits above `path3_wide_root`, but describes `square_edges` two lines below; the "root declared wider than the node numbering" sentence is the one that belongs where it is.

## What the rest of the checklist said

- **Proof coverage is real, positively confirmed.** `reachable_test` makes 27 veripb calls — one `--help` probe plus 26 proofs — and every one answers `s VERIFIED COMPLETE ENUMERATION`. All six `reachable_mutation_*` lanes pass, i.e. veripb refuses each corrupted reason. (The 59 ms runtime is genuine: VeriPB 3.0.2 is the Rust build and the fixtures are tiny. Worth saying, because it looks too fast to have run.) 13/13 of the reachable, hitori and minizinc lanes pass locally.
- No `AssertProofRule` / `AssertRatherThanJustifying`; no `overloaded{` site touched; no debug leftovers, hard-coded paths or flipped defaults; no stray or binary files.
- Frontend coverage: MiniZinc wired end to end (four `mznlib` overrides, `fzn_glasgow.cc`, four test models, four ctest lanes with the `--fzn-pattern` guard that makes them prove the override fired), support matrix updated with a footnote, XCSP `n/a` and Python `?` matching the column's existing convention. `reachable.cc` is in the library list, `reachable_test` in both the executable list and the `foreach` line in the right alphabetical slot, and `gcs.hh` gets `reachable.hh` between `power.hh` and `regular.hh`.
- All five commits carry `Co-Authored-By:` and `Claude-Session:` in the current house form.
- Every API named in a new doc comment exists.

**Merge-order note.** `gcs/CMakeLists.txt` is the known hot spot and is also touched by #781, #779, #776 and #713; this PR adds `reachable_test` to that `foreach` line, so whichever of them lands second pays a rebase. `dev_docs/README.md` is shared with #776, #713 and #783, all appends at the end of the list.

<sub>Posted from the pre-merge check pass, which is AI-assisted (Claude Opus 5); the build, the proof runs and the hitori figures above were produced locally on the trial merge rather than inferred.</sub>

Issue #508, opened 2026-07-13
Compact-table-style extensional GAC for Table (Gecode parity)


ciaranm:
## Compact-table-style extensional GAC for Table (Gecode parity)

`propagate_extensional` (positive `Table`, and the tabulated arithmetic encodings
that share it) is a naive GAC: pass 1 rechecks every selectable tuple's
feasibility, pass 2 re-seeks a support for every (variable, value) by scanning all
selectable tuples. On identical search trees (same GAC, verified by matching node
counts) GCS is ~9x slower than Gecode 6.3.0 on a d=15 table (75s vs 8.4s).

Residual supports (PR: "Extensional GAC: cache residual supports") make pass 2
incremental, but that only shifts the bottleneck to **pass 1** — the per-wake
feasibility rescan — which is exactly what a compact-table scheme removes:
maintain a bitset of valid tuples, AND out `supportMask[var][removed value]` on a
removal (pass 1), and check `validTuples & supportMask[var][value]` per remaining
value (pass 2), both in O(tuples/word).

Design notes:
- **Do not assume a reversible sparse bitset.** On modern hardware a plain dense
  bitset that is recomputed (or a small delta trail) is often faster than the
  reversible-sparse structure the classic compact-table paper uses; benchmark
  before committing to a representation.
- Preserve proofs by keeping the existing `selector != tuple` (NoJustification)
  and `var != val` (RUP) inferences: the bitset drives the algorithm, the selector
  and its inferences stay in sync so the proof is byte-identical.
- The refined-watch mechanism (#502) fits the change-detection ("which values were
  removed since last wake").

Related: #503 (measured throughput audit), #364 (incrementality survey).

ciaranm:
Closing as done. The compact-table arm and the work around it are all merged:

- #786: domain iteration without coroutine generators
- #795: reason hoist and a concrete selector
- #796: private live-tuple set, proof-only selector
- #799: domain rasterising
- #804: the compact-table arm, chosen per instance
- #809: the tabulation arm
- #831: shared support masks
- #832: stop rather than throw on table failure

At the end of that work, on matched search trees against Gecode 6.3.0, Crossword was 1.1–1.3× and Dubois 2.0–2.1×. On every instance but one, making the table propagator infinitely fast would not close the remaining gap: what is left is the engine, which is #503.

---
Closed by an AI agent (Claude) during an issue-housekeeping pass, checked against `main` at `e0bc7039`.

Issue #503, opened 2026-07-13
Measured audit of per-propagator throughput (cross-solver, matched trees)


ciaranm:
## Measured audit of per-propagator throughput (vs Gecode / Choco)

Cross-solver benchmarking has, for the first time, let us measure **raw per-node
propagation throughput** on *matched* search trees. We should turn this into a
systematic, measured audit of every propagator — the one propagator we happened
to benchmark closely is almost certainly not the only slow one.

### The measurement that prompted this

Running the same random binary-CSP instances through GCS, Gecode 6.3.0 and Choco
4.10.15 with a **fixed, non-learning search** (`input_order` + `indomain_min`) and
**matched propagation strength**, verified by *identical node counts* across the
three solvers:

| propagator exercised | GCS nodes/s | Gecode nodes/s | GCS vs Gecode |
|---|---|---|---|
| pairwise `!=` (colouring) | ~305k | ~670k | ~2.2× slower |
| `table` (extensional GAC) | ~2.9k | ~40k | **~13× slower** |

Same tree, same GAC strength (identical node counts prove it), proofs off — so
this is raw propagator throughput, not extra search or proof overhead. `table` is
just the propagator we looked at; the point of this issue is that we now have a
repeatable way to find the others.

### The ask

A **measured** audit of every propagator's throughput, with **careful
per-propagator benchmark selection**, producing a ranked table of the slowest
ones to optimise. This is the empirical counterpart to the static incrementality
survey in #364: #364 lists *candidate fixes* by reading the code; this issue
*measures* which propagators are actually slow so we know where the fixes pay off.

### Methodology (illustrated by the `table` measurement above)

For each propagator:

1. **Pick instances where that propagator is the bottleneck** — it should dominate
   run time (heavy use, real search), not be drowned out by others.
2. **Match propagation strength across solvers** and **verify by identical node
   counts** — this is the load-bearing check. If node counts differ, the
   consistency levels differ and a raw-time comparison is meaningless.
3. **Fixed, non-learning search** so the trees are identical and deterministic.
4. **Proofs off**; measure per-node time / nodes-per-second (deterministic node
   counts make this environment-robust).
5. Compare to Gecode and/or Choco where an equivalent propagator + FlatZinc path
   exists.

### Caveats to design around

- Some propagators have **no cross-solver equivalent** (GCS-specific) or a
  **different default consistency** (can't match trees). For those, fall back to
[...]

PR #786, opened 2026-09-02
Extensional GAC: iterate domains without coroutine generators

ciaranm:
`propagate_extensional`'s three domain loops used `each_value_mutable()` /
`each_value_immutable()`, which allocate a `std::generator` frame per loop and
route each value through a `std::function`. The non-coroutine
`for_each_value_*()` alternatives do the same thing with a callback.

Purely mechanical: same values, same order, same inferences. `recursions`,
`propagations` and solution counts are identical on every shape measured, so
this is a pure per-call speed change with no effect on the search.

### Measurements

`positive_table_random`, fataepyc-10 pinned (`numactl --cpunodebind=0 --membind=0
taskset -c 4 setarch -R`), boost disabled, median of 3:

| shape | before | after | speedup |
|---|---|---|---|
| `--domain=12` | 0.241 s | 0.169 s | 1.43x |
| `--domain=15` | 5.649 s | 4.005 s | 1.41x |
| `--domain=18 --vars=25 --constraints=150` | 3.295 s | 2.346 s | 1.40x |
| `--domain=25 --vars=20 --constraints=100 --tightness=0.5` | 0.606 s | 0.433 s | 1.40x |

`recursions` matched exactly on all four (439, 11252, 20528, 659).

A leaf-attribution `perf` profile on `--domain=15` says why: before the change,
coroutine machinery plus the allocations it drives was **37%** of samples
(`each_value_generator` actor frames, `IntervalSet::each()` actor frames, the
`std::function` in the generator's view application, and the promise
allocations). After, it is **0.1%**.

### Scope

This touches the shared helper, so `Table`, the `AutoTable` presolver and every
runtime-tabulated constraint (`Multiply`, `Divide`, `Modulus`, `Power`, `Plus`,
`Minus`, GAC `Linear`) all benefit. It is deliberately separated from the table
propagator work in #508 so that a later compact-table change cannot take credit
for it.

Groundwork for #508; relevant to the throughput audit in #503.

### Testing

- `ctest` 749/749 with `GCS_ENABLE_XCSP=ON` and `GCS_ENABLE_MINIZINC=ON`
- `ctest` 631/631 with `GCS_TEST_CAP_DEFAULTS=OFF`, so enumeration completeness
  is actually checked rather than capped
- full workflow-2 chain re-run with a real `cake_pb_cp` on `table_sat`,
  `table_unsat`, `negative_table_sat` and `negative_table_unsat`: all
  `s VERIFIED`

### DIFF
--- a/dev_docs/constraints.md
+++ b/dev_docs/constraints.md
@@ -388,9 +388,20 @@ Inside the propagator body, the `State` parameter exposes:
 - `state.bounds(v)` → `pair<Integer, Integer>` (lower, upper).
 - `state.in_domain(v, val)` → bool.
 - `state.has_single_value(v)` → bool.
+- `state.for_each_value_immutable(v, cb)` / `state.for_each_value_mutable(v, cb)`
+  → call `cb(value)` for each value in the current domain, ascending. Use
+  `_immutable` if you only read; `_mutable` if you might infer a removal
+  mid-iteration. A `cb` returning `bool` stops iteration by returning `false`;
+  one returning `void` runs to completion. **Prefer these anywhere that runs
+  per wake** — see the note below.
 - `state.each_value_immutable(v)` / `state.each_value_mutable(v)` →
-  ranges over the current domain. Use `_immutable` if you only read;
-  `_mutable` if you might infer a removal mid-iteration.
+  the same two contracts as ranges, for a `for (auto val : ...)` loop. They
+  allocate a `std::generator` frame per loop and route each value through a
+  `std::function`, which is real per-call cost in a propagator body: converting
+  the three loops in `propagate_extensional` to the callback forms above was
+  worth 1.4x end-to-end on every shape measured (PR #786), with nothing else
+  changed. Reach for the range form when a propagator runs once, or when the
+  loop genuinely reads better that way.
 - `state.domains_intersect(v1, v2)` → bool. Does the variables' domains
   share any value? Walks both stored interval sets in merge order
   without copying for the common case. Use this instead of
--- a/dev_docs/state-and-variables.md
+++ b/dev_docs/state-and-variables.md
@@ -304,6 +304,16 @@ constants:
   not rely on that, since it is allowed to be optimised in future to
   avoid the copy. `CurrentState` exposes a single `each_value` (forward)
   and `each_value_reversed` (descending) for callback-time consumers.
+- `for_each_value_immutable(var, cb)`, `for_each_value_mutable(var, cb)` —
+  the same two contracts, delivered by calling `cb(value)` in ascending
+  order rather than returning a generator. A `cb` returning `bool` stops
+  iteration on `false`; one returning `void` runs to completion (that is
+  `IntervalSet::for_each`'s contract, selected with `if constexpr`).
+  These skip the coroutine frame, the `std::function` the view
+  application needs in the generator version, and the `IntervalSet`
+  copy, so they are the right default inside a propagator — see
+  [propagator-performance.md](propagator-performance.md), and note that
+  the `_immutable` one does **not** snapshot (below).
 - `copy_of_values(var)` — full snapshot as an `IntervalSet<Integer>`.
 - `domain_intersects_with(var, IntervalSet)` — does the variable's
   domain share any value with the given set? The common case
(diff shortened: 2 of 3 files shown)


ciaranm:
## Pre-merge review

Trial-merged onto `main` (`17344ec0`), built release, and measured. **No findings** — the "purely mechanical" claim holds, and I was able to check it rather than take it on trust.

### The equivalence argument, from the source

The three conversions preserve semantics exactly, and the reason is a little stronger than the PR body claims:

- `each_value_generator` takes its `IntervalSet` **by value** (`gcs/innards/state.cc:632`), and both `each_value_immutable` and `each_value_mutable` have literally the same body. So the old code was *already* snapshotting in both cases, and `for_each_value_mutable`'s explicit `auto snapshot = state_of_for_iteration(v)` reproduces that. `state_of` and `state_of_for_iteration` are the same expression, so nothing moves there either.
- `for_each_value_immutable` is the one that genuinely changes behaviour — it iterates the live set instead of a copy. That's fine for the site it replaces: the inner support search only reads `residues`, and the `inference.infer` that touches `table.vars[idx]` happens after that loop has finished.
- `break` → `return false` and `continue` → `return` match `IntervalSet::for_each`'s documented contract (`gcs/interval_set.hh:452`): a `bool` callback stops on `false`, a `void` one runs to completion. The outer lambda in the residues loop deduces `void`, the inner one is explicitly `-> bool`, so both take the right branch of the `if constexpr`.

Also worth noting for anyone reading this as a precedent: it isn't a new pattern. There are already 22 call sites of these helpers in `element`, `min_distance`, `gac_all_different`, `bounds_global_cardinality` and `gac_global_cardinality`, including both the `-> bool` early-exit form and `for_each_value_mutable`.

### Reproduced, on this box

Same tree both times, only `constraints/extensional_utils.cc` reverted for the "before" column, `taskset`-pinned, release:

| shape | recursions | propagations | before | after | speedup |
|---|--:|--:|--:|--:|--:|
| `--domain=12` | 439 | 65 250 | 0.092 s | 0.066 s | 1.39x |
| `--domain=15` | 11 252 | 1 408 987 | 2.166 s | 1.579 s | 1.37x |
| `--domain=18 --vars=25 --constraints=150` | 20 528 | 772 862 | 1.254 s | 0.919 s | 1.36x |
| `--domain=25 --vars=20 --constraints=100 --tightness=0.5` | 659 | 59 393 | 0.231 s | 0.168 s | 1.38x |

Solutions, recursions, failures and propagations are identical on every shape, and the four recursion counts are exactly the ones quoted. Ratios land a little under the reported 1.40-1.43x, which is what a different machine should do.

**The check I'd point at if anyone doubts "same inferences":** with a fixed `--seed=1`, `table_test`'s 16 `.opb` and 16 `.pbp` files are **byte-identical** before and after. Not just the same conclusions — the same proof lines, in the same order. (`table_test` is seed-randomised, so this only means anything at a pinned seed; comparing two unseeded runs shows twelve `.pbp` differences that are entirely the seed.)

`ctest` 764/764 on the trial merge, zero compiler warnings in the touched file.

### Checklist

Nothing to say on documentation, proof coverage, stray files, debug hacks, umbrella headers, unexercised paths or the AI trailer — one file, one function, no interface change, `Co-Authored-By:` present.

### Not a finding, but adjacent

`dev_docs/propagator-performance.md` already says to prefer `for_each_value_*` over `each_value_*`, but `dev_docs/constraints.md` (the doc a propagator author actually reads first, "Inside the propagator body, the `State` parameter exposes...") still lists only `each_value_immutable` / `each_value_mutable`, and `dev_docs/state-and-variables.md` doesn't mention the callback forms at all. That drift predates this PR, but this PR is the measurement that makes the case — a 1.4x end-to-end win on a shared helper. Might be worth a sentence in one or both while it's fresh.

*Reviewed with Claude Code.*

ciaranm:
Picked up the "not a finding, but adjacent" documentation drift in `4652d178`, since this PR is indeed the measurement that makes the case.

- **`constraints.md`** — the "Inside the propagator body" list now leads with `for_each_value_immutable` / `for_each_value_mutable` and their early-exit contract, and the generator bullet says what it costs and cites the 1.4x here, so an author reading this doc first reaches for the right one by default.
- **`state-and-variables.md`** — the callback forms are now in the accessor list next to the generators.

While writing it, one thing turned up that is worth more than the drift itself. `state-and-variables.md` says `_immutable` and `_mutable` are "behaviourally interchangeable today" because both generators snapshot the `IntervalSet` into the coroutine frame, and warns only that a future optimisation could remove that. **That future is already here for the callback forms, in the opposite direction**: `for_each_value_mutable` copies (`state.hh:457`), but `for_each_value_immutable` walks the stored set directly (`state.hh:436` — `state_of_for_iteration` returns `const IntervalSet<Integer> &`). So inferring a removal from inside a `for_each_value_immutable` callback mutates the container being iterated, which is undefined behaviour now rather than unsupported later. That is documented as a live hazard, not a future-proofing note.

Which is exactly the distinction your review drew for this diff, so for the record the three converted sites are on the right side of it: pass 1 and the outer loop of pass 2 both call `inference.infer` on the variable being iterated and are `_mutable`; the inner selector scan is `_immutable` and only touches `residues` and a local flag.

Documentation only — no source change, so the measurements and the byte-identical proof check above still stand.

Issue #167, opened 2026-05-08
XCSP3 `<circuit>` semantics: gcs `Circuit` is stricter than the XCSP3-core spec


ciaranm:
## Background

While generating cached expected outputs for the XCSP3 binding tests via ACE (the reference XCSP3 solver), I found a real semantic disagreement on a 3-element `<circuit>` instance:

- **gcs's `Circuit`** finds **2 solutions** — the two Hamiltonian circuits of K₃: `(1,2,0)` and `(2,0,1)`.
- **ACE** finds **5 solutions** — the same two Hamiltonian circuits, plus three solutions where one vertex is a self-loop and the other two form a 2-cycle:
  - `(0,2,1)` — vertex 0 isolated, `{1,2}` cycle
  - `(1,0,2)` — vertex 2 isolated, `{0,1}` cycle
  - `(2,1,0)` — vertex 1 isolated, `{0,2}` cycle

## What the spec says

XCSP3-core v3.2 (section 4.1.7.1, "Constraint circuit"):

> "If a vertex `i` has its successor `x[i]` equal to `i`, then the vertex is isolated. The constraint requires that there is exactly one circuit in the (functional) graph composed of vertices having distinct successors..."

So self-loops (isolated vertices) are explicitly allowed — XCSP3's `<circuit>` is the "at most one non-trivial cycle, the rest are self-loops" semantics. ACE matches the spec; gcs's `Circuit` is the stricter Hamilton-circuit-only semantics.

## Where this matters

The XCSP3 binding's `buildConstraintCircuit` posts gcs's `Circuit` directly — so it imposes a stricter constraint than XCSP3 specifies. For the cached test output, the test currently uses gcs's strict 2-solution cache and is documented as such; the test won't catch a regression in our XCSP3-conformance.

## Possible fixes

1. **Add a separate gcs constraint** (e.g. `XcspCircuit`) that matches the XCSP3 semantics (allows isolated vertices). Use it in the binding. Existing `Circuit` users keep the strict semantics.
2. **Decompose** in the XCSP3 binding: post `AllDifferent` on the successor array, plus a "at most one non-trivial cycle" constraint. The latter doesn't have an obvious propagator.
3. **Change `Circuit`** to match XCSP3 (and add a new `HamiltonCircuit` for the strict case). This is the most invasive — risks breaking other callers.

Option 1 is probably right.

## Test cache

[...]

PR #787, opened 2026-09-02
Add Subgraph and the tree / path family (#637)

ciaranm:
Stage 3 and stage 4 of #637: `Subgraph`, `Tree`, `DTree`, `Path` and `DPath`, as
C++ constraints rather than as mznlib overrides, so the family is reachable from
`.scp` and from a program as well as from MiniZinc.

## The thing that made this worth doing

`fzn_dtree` does **not** call `fzn_dreachable`. It carries its own copy of the
stdlib's parent-and-distance labelling, so `tree`, `dtree`, `path`, `dpath`,
`steiner`, `dsteiner`, `bounded_path`, `bounded_dpath` and both
`weighted_spanning_tree` spellings all still got the encoding #784 diagnosed —
with the cure merged and sitting next to them.

## No new encoding, no new proof idea

Each new constraint posts one reachability child, one `LinearEquality` child for
`sum(es) = sum(ns) - 1`, and a few counting rows of its own.

| | reachability child | its own rows |
|---|---|---|
| `Tree` | `Reachable` from the root | — |
| `DTree` | `DReachable` from the root | at most one selected edge into each node |
| `Path` | `Reachable` from the start | degree ≤ 2, ≤ 1 at each end, none at all when the ends coincide |
| `DPath` | `DReachable` from the start | in/out-degree ≤ 1, none in at the start, none out of the end, end is selected |

A connected subgraph with one fewer edge than nodes is a tree, and conversely.
Directed, reachability gives each non-root node an edge coming in, "at most one"
makes it exactly one, and the count leaves the root with none. A tree whose
degrees are at most two is a path, and its ends are the degree-one nodes.

**One reachability encoding does each of them**, where the stdlib doubles every
undirected edge and, for `dpath`, asks for two spanning trees. That matters more
here than for reachability: the encoding is `O(nodes × edges)`, so doubling the
edge set doubles it and a second tree doubles it again.

## Two things to look at

**The `start = end` corner, which is why these are classes and not six `.mzn`
files.** "At most one edge at each end" looks obviously right and admits the
start joined to one more node — a path of length one, not a path from a node to
itself. `Path` needs a row per node saying that when *both* ends are that node it
has no incident edge at all; `DPath` gets it for free. In a `.mzn` file that
error has no unit test. Here it has one.

**The counting rows are the constraint's own rather than more children, and not
by preference.** Two OPB rows may not share a `c[id][role]` label; a child builds
its label from its own id and role; every plain `LinearLessThanEqual` uses the
same empty role. So a second linear child under the parent's id is refused by
`ProofModel`, with a message that says exactly what the rule is — *a role must
name everything that varies*. One reachability child and one linear child coexist
because their roles differ. `constraints/innards/graph_rules.{hh,cc}` holds the
rows and a propagator that enforces them forwards and backwards; every inference
is a plain RUP against exactly the row its rule wrote.

## Deliberately not GAC

`Reachable` is GAC and the cardinality equality is GAC; their conjunction is not.
Nothing here notices that an edge would close a cycle until the count runs out of
room. The tests use `solve_for_tests` rather than `solve_for_tests_checking_gac`
and the class docs say so, because these sit next to a constraint that does claim
GAC. A union-find cycle check is the obvious later strengthening, with this as
the baseline to measure it against. `Subgraph` *is* GAC and its test checks that.

## Measured, on a grid Steiner tree

`steiner` is `tree` plus a weighted sum, so it reaches `Tree` through the
stdlib's own wrapper — one solver, one model, the only difference being whether
`mznlib/fzn_tree_*.mzn` is present. Same optimum both ways.

| `steiner`, 3 × 3 grid | decomposition | `Tree` |
|---|--:|--:|
| nodes | 4 135 | **91** |
| `.opb` | 326 118 B | 69 135 B |
| `.pbp` | 39 689 062 B | **238 799 B** |
| veripb | 4.87 s | **0.01 s** |

| `steiner`, 4 × 4 grid, proofs off | decomposition | `Tree` |
|---|--:|--:|
| nodes | 7 506 461 | **3 209** |
| solve | 164.49 s | **0.029 s** |

As with #784's hitori table, most of that gap is the distance labelling being a
bad thing to search over rather than these rules being clever; they are the
straightforward ones.

## Also here

Ten mznlib overrides (`fzn_subgraph`, `fzn_tree`, `fzn_dtree`, `fzn_path`,
`fzn_dpath`, each `_int` and `_enum`), and seven MiniZinc differential tests with
`--fzn-pattern` guards. Two of them are `steiner` and `bounded_dpath`, which
prove the free members really do ride the new overrides through the stdlib's own
wrappers rather than being assumed to.

`xcsp_glasgow_constraint_solver.cc` now qualifies `XCSP3Core::Tree`: that file
pulls in both namespaces, and gcs has a `Tree` of its own now.

`dag`, `network_flow` and `subcircuit` stay open under #637 and are not on this
ladder — `dag` is an acyclicity labelling, `network_flow` decomposes to plain
flow conservation, and `subcircuit` belongs with `Circuit` (#167). The support
matrix says so.

774/774 tests green, `clang-format` clean, no new clang warnings.


ciaranm:
## Pre-merge review

Trial-merged onto `main` (`17344ec0`) — clean merge — built release with `GCS_ENABLE_XCSP=ON` and `GCS_ENABLE_MINIZINC=ON`, and ran everything. **774/774 green**, matching the PR body. Read `graph_rules`, all three constraints and their tests in full.

Two things, one of which is real.

### 1. `hints::GraphRules` is dead

`gcs/constraints/innards/graph_rules.hh:26` still defines

```cpp
struct GraphRules
{
    ConstraintID originator;
    static constexpr std::string_view hint_name = "graphrules";
};
```

but dbd3d1d2 is exactly the commit that stopped anything using it — `propagate` now takes the hint from its caller, and `hints::Tree` / `hints::Path` live in the per-constraint headers. Grepping the whole tree for `GraphRules` finds one occurrence: that definition. So the commit message's "The shared helper now names nothing" is true of the behaviour but not of the header, which still carries the name it was retiring — and `dev_docs/constraints.md:478` is explicit that hint structs belong in `gcs/constraints/<foo>/hints.hh`, which this isn't. The `<string_view>` include goes with it; `<gcs/constraint_id.hh>` stays, since `define` still takes a `ConstraintID`.

### 2. Nit: the new `add_library` entries are out of alphabetical order

```
constraints/seq_precede_chain/seq_precede_chain.cc
constraints/subgraph/subgraph.cc
constraints/path/path.cc
constraints/tree/tree.cc
constraints/innards/graph_rules.cc
constraints/smart_table/smart_table.cc
```

Two of the three internal transitions descend, and the block sits between `seq_precede_chain` and `smart_table` rather than at the four points where the entries sort. That list *is* maintained alphabetically — 134 of its 139 entries are in order, the exceptions being `guaranteed_contribution`, `justify_not_in_range`, `regex` and `comparator_network`. In case it looks inconsistent to fix only this one: the `add_executable` and `add_test` lists are **not** alphabetical (42 of 125 and 11 of 45 descending, grouped by topic instead), so the new entries there are fine as they are.

### What I checked and found nothing wrong with

**The rules are sound and the reasons are complete.** All four inference directions in `graph_rules::propagate` RUP against exactly the row `define` wrote. Forward `AtMost`: reason is the true conditions plus the ones, negating `v == 0` gives `limit + 1` ones and falsifies the row. Backward `AtMost`: negating `!open.front()` reinstates the condition, and the ones already exceed the limit. Both `Selected` directions likewise. Nothing is inferred with more than one condition open, which is incompleteness rather than unsoundness and is documented as such.

**The self-loop corner is tested, with proofs.** `AtMost`'s "a variable appearing twice counts twice" means `define` emits a row with the same literal twice, and the propagator counts it twice into `ones`. `path_test`'s `loop_and_parallel` and `subgraph_test`'s `loop_and_parallel` both exercise that under veripb, so the duplicated term is verified rather than assumed.

**Proofs run and verify.** 30 fixtures in `path_test`, 22 in `tree_test`, 7 in `subgraph_test`, every one `s VERIFIED`. `path_test` finishing in 0.071 s is real, not skipped verification — veripb 3.0.2 is a Rust binary and the fixtures are tiny.

**The MiniZinc lanes are genuinely differential.** `run_minizinc_test.bash` diffs the full solution set against MiniZinc's default solver, and the `--fzn-pattern` guards mean `minizinc-steiner` and `minizinc-bounded-dpath` demonstrably reach `glasgow_tree` / `glasgow_dpath` through the stdlib's own wrappers rather than passing against the decomposition. All seven new lanes pass locally, as do the three new constraint lanes. The `.mzn` models index nodes from 2, so the `offset` shift in the `_enum` overrides is exercised rather than assumed.

**`prepare` returning false is right in both places.** `Tree` (undirected) always has empty `_rules`, so it installs no rows and no propagator of its own — everything is the `Reachable` and `LinearEquality` children, which `prepare` has already installed. Same for `Subgraph` with no edges, which `subgraph_test`'s `empty` fixture covers.

**The GAC claims match the tests.** `run_subgraph_test` uses `solve_for_tests_checking_gac`; tree and path use `solve_for_tests`, with the comment explaining which conjunction loses it. That is the doc and the test agreeing, which is what I check for.

**Everything else.** `.scp` round-trip test present and the arities line up with what `s_expr` emits (7 / 8 / 6). `gcs.hh` updated, alphabetically. New headers are picked up by the `FILES_MATCHING PATTERN "*.hh"` install glob, so nothing to wire. No duplicate `hint_name` anywhere in the tree. No stray files (44 files touched, 44 files ever touched on the branch — nothing added then removed), no debug output, no `AssertRatherThanJustifying`, no `TODO`, zero compiler warnings. `Co-Authored-By:` on both commits. Docs are thorough and I found no reference document that now contradicts itself — the support matrix rows, the `connectivity-proofs.md` section and the `dev_docs/README.md` summary agree with each other and with the code.

*Reviewed with Claude Code.*

ciaranm:
Both fixed in ffcbe2eb.

**1 — `hints::GraphRules` was dead, and the finding is sharper than "leftover".** It was not left behind on purpose: dbd3d1d2 was meant to delete it in the same edit that moved the callers, and the deletion silently did not happen — so I wrote "the shared helper now names nothing" in that commit message on the strength of the change I intended rather than the file I ended up with. The struct is gone now, with the `<string_view>` include that was only there for it; `<gcs/constraint_id.hh>` stays, since `define` still takes a `ConstraintID`, exactly as you said.

**2 — the `add_library` entries are sorted.** `constraints/innards/graph_rules.cc` after `cake_truthiness`, `path` after `parity`, `subgraph` after `sort/sort`, `tree` after `table/table`. Thanks for checking the other two lists rather than assuming they matched — the `add_executable` and `add_test` entries are left grouped by topic where they were.

774/774 still green, and the 22 / 30 / 7 tree, path and subgraph proof lanes still verify.

Issue #788, opened 2026-09-02
SubCircuit: `Circuit` has no opt-out sibling, and it is the one `globals.graph` member the Challenge actually uses


ciaranm:
`Circuit` has no opt-out sibling. MiniZinc's `subcircuit` falls through to the
stdlib decomposition, and XCSP3's `<circuit>` with a size argument is a hard
`report_unsupported` rather than a decomposition at all. This is the one member of
`globals.graph` with real usage in the MiniZinc Challenge that gcs still has
nothing native for.

Filed separately from #637 deliberately. #637 is the reachability ladder —
`subgraph`, `reachable`, `tree`, `path` and everything the stdlib defines in terms
of them — and its own body already puts `subcircuit` outside that ladder. Now that
#787 has finished the ladder, the ladder's shared breadth-first unfolding is the
wrong shape for this: `subcircuit` is not a connectivity property but a
permutation-plus-single-cycle one, so its natural home is beside `Circuit` and its
proof story is `Circuit`'s, not `Reachable`'s.

## The demand is real, unlike the rest of that issue

Across the whole `mzn-challenge` corpus, 2008–2026, five models call `subcircuit`,
covering 27 instances:

| model | instances | successor array |
|---|--:|---|
| `2012/tpp` | 7 | `1..numcities`, 9–35 |
| `2013/mario` | 5 | `1..nbHouses` |
| `2014/mario` | 5 | `1..nbHouses` |
| `2016/tpp` | 5 | `1..numcities`, 9–35 |
| `2017/mario` | 5 | `1..nbHouses`, 30 in the medium instances |

(`2023/multi-agent-graph-coverage` mentions it, but the call is commented out.)

For comparison, the whole reachability ladder that #637 covered has exactly one
challenge model behind it — `2025/hitori`, through `connected` — and `tree`,
`path`, `steiner`, `dag` and `network_flow` have none at all.

## What the fall-through looks like

`fzn_subcircuit` is the stdlib's position labelling, and structurally it is the
same thing gcs's own `Circuit` was written to replace:

```
array [S] of var 1..n: order;
alldifferent(x) /\ alldifferent(order) /\ ...
forall (i in S) ((ins[i] /\ x[i] != firstin) -> order[x[i]] = order[i] + 1)
forall (i in S) (ins[i] \/ order[lastin] < order[i])
```

so per node: a **variable-indexed element** into `order` (`order[x[i]]`), a
[...]

PR #790, opened 2026-09-02
A SubCircuit constraint, with check and prevent certified

ciaranm:
Step 1 of #788: `SubCircuit` over a successor array, MiniZinc semantics, with both of
Francis and Stuckey's cheaper propagation algorithms and a VeriPB certificate for each.

The MiniZinc binding, the XCSP3 size-carrying overloads (which subsume #167) and the
certified SCC arm are deliberately not here — see the [step-0
comment](https://github.com/ciaranm/glasgow-constraint-solver/issues/788#issuecomment-5505574380)
for the staging and the measurements behind it.

## What it is

A sibling class in `gcs/constraints/circuit/`, not a mode on `Circuit`. `Circuit`'s opening
inference is `succ[i] != i` for every `i`, which is exactly what this constraint has to
allow: a node off the tour points at itself. The three semantic corners are in the class
documentation, because they are the kind of thing one otherwise discovers later — the empty
subcircuit is a solution, the smallest non-empty one has two nodes, and the successors are a
permutation whether or not a node is on the tour.

`with_algorithm()` selects:

- `subcircuit::Check` — follow each chain of fixed successors and, when one closes into a
  cycle of two or more, force every other node to be a self loop. Where `Circuit` simply
  contradicts, here the cycle only pins down everyone else.
- `subcircuit::Prevent` (the default) — adds the lookahead, and with it Francis and
  Stuckey's *evidence node*: a chain must not close while some node outside it cannot be a
  self loop. With no such node there is nothing to infer at all, because a closing chain
  plus everyone else opting out is a perfectly good solution.

`with_gac_all_different()` as on `Circuit`.

## The encoding

Circuit's position encoding, conditioned on membership, and still linear — the tour's length
is the sum of the "not a self loop" literals, so it needs no auxiliary variable of its own
and the wrap row is

```
(succ[i] = j and first[j]) -> pos[j] - pos[i] + |tour| = 1
```

with `first[j]` a fully reified flag for "j is on the tour and every lower-numbered node is
off it". A node off the tour is given position **zero**: nothing here needs the positions to
be distinct, and an off-tour node takes part in no position row at all — it has no successor
but itself, and by all-different nothing can point at it — so the only job those rows have is
to leave `pos` determined by the successors under unit propagation, which is what solution
checking needs.

**Where this costs more than Circuit, and why that is in the encoding rather than the proof.**
Each edge needs *two* row families rather than one, because which edge wraps is not known
statically: it is the edge into whichever node is `first`, and Circuit anchors on node 0 so it
always knows. `define_proof_model()` carries a comment on this, because it is the one place
here where the encoding carries something for the proof's benefit rather than saying the
constraint as compactly as it could.

Dropping the wrap rows leaves an encoding that is *smaller than the stdlib decomposition's*
and still exact — a cycle avoiding the anchor still chains to `0 = k`. But the counting fact
those rows provide then has to be derived instead, and deriving it needs "the on-tour
positions are exactly `0..L-1`": a pigeonhole induction **conditional on the variable `L`**,
which is precisely the obstacle that stops `circuit_scc.cc`'s machinery being ported here. So
that is not a free reformulation into the proof; it is blocked on the same open problem. The
other way out is a caller-nominated anchor, which collapses this to exactly Circuit's shape,
since only the n edges into a nominated on-tour node can wrap.

On `2013/mario/mario_easy_2` (n = 15), the subcircuit encoding's share of the `.opb` is 1,868
rows against the decomposition's 1,531 — 22% more, and that 22% is the wrap family.

## The certificate, and the test that earns it

Circuit anchors on node 0, so it knows statically which edge wraps around. Here it is the
edge into whichever node is `first`, so each firing pays a case split: sum the step rows
round the cycle for "none of these is first", sum `|cycle|` more times for "this one is",
then add the first to the rest to resolve the flags away, leaving the bound that the tour is
no longer than the cycle.

**The plain enumeration test is what keeps the certificate honest, from n = 5 up**: swap
`derive_tour_at_most` for a plain `JustifyUsingRUP` and VeriPB rejects the proof at
`subcircuit_test.cc`'s n = 5 case. That is worth stating because it was not true of the first
version of this encoding: while off-tour positions were numbered *after* the on-tour ones,
unit propagation could reach the conflict unaided at every size the enumeration runs, and only
a hand-built scenario with the anchor left undetermined could tell the difference. Pinning
off-tour positions to zero moved that check into the cheap test, which is where it belongs.

`subcircuit_prevent_test.cc` still drives the evidence-node inference with the anchor
undetermined, which is a narrower case than the enumeration reaches and worth its own
regression test — but it is no longer the test that carries the argument, and its comment says
so.

## Consistency claim

Neither algorithm is arc consistent — both wait for a chain of successors to be fixed before
saying anything, so a value can easily survive in a domain with no solution behind it. The
enumeration tests therefore use `solve_for_tests`, not `solve_for_tests_checking_gac`, same
as `Circuit`.

## Testing

- `subcircuit_test.cc` — enumeration against a check written straight off MiniZinc's
  definition, `n` = 1..5 for both algorithms, plus the empty array (which MiniZinc
  special-cases to `true` and a C++ caller can still build), all under the view sweep.
  `n = 4` is the smallest size that says anything beyond all-different: with three nodes you
  cannot fit two disjoint cycles, so every permutation is a subcircuit.
- `subcircuit_prevent_test.cc` — the undetermined-anchor scenario above.
- `subcircuit_dup_test.cc` — aliased successor handles rejected at construction, as
  `Circuit` does, for both algorithms.
- `scp_reader_test.cc` — enumeration plus the write → read → write round trip.
- Full suite: 779/779 pass. Clean under `clang++` with `-DCMAKE_BUILD_TYPE=Sanitize`
  (the two pre-existing `-Werror` failures under clang, in `variable_weighting.hh` and
  `element.cc`, are untouched and unrelated). Both cap settings run.

## Known limitations, stated rather than hidden

- The propagator walks the fixed-edge graph from scratch on every call, where
  `circuit::Prevent` keeps its chain endpoints in backtrackable state and folds each new
  edge in in O(1). Nothing here needs that yet and the walk is far easier to read, but it is
  the obvious move if it ever shows up in a profile.
- It does not claim idempotence, unlike `circuit::Prevent`. Forcing a node to be a self loop
  fixes a successor and changes the chain structure the pass just walked, and one pass makes
  no attempt to reach the fixpoint of that cascade; the triggers cover it, since every
  inference here instantiates a successor.
- The evidence node is the lowest-numbered candidate, which is Francis and Stuckey's default.
  They report that choosing the one fixed highest in the search tree is usually a little
  better, and that is a knob worth having later rather than a choice worth guessing now.

### DIFF
--- a/gcs/CMakeLists.txt
+++ b/gcs/CMakeLists.txt
@@ -30,6 +30,7 @@ add_library(glasgow_constraint_solver
         constraints/circuit/circuit_base.cc
         constraints/circuit/circuit_prevent.cc
         constraints/circuit/circuit_scc.cc
+        constraints/circuit/subcircuit.cc
         constraints/comparison/comparison.cc
         constraints/count/count.cc
         constraints/cumulative/cumulative.cc
@@ -767,6 +768,20 @@ if(GCS_BUILD_TESTS)
     target_link_libraries(circuit_test PRIVATE glasgow_constraint_solver)
     add_test(NAME circuit_constraint COMMAND ${GCS_BASH} ${CMAKE_CURRENT_SOURCE_DIR}/../run_test_only.bash $<TARGET_FILE:circuit_test>)
     add_view_tests(circuit_constraint circuit_test 5)
+
+    add_executable(subcircuit_prevent_test constraints/circuit/subcircuit_prevent_test.cc)
+    target_link_libraries(subcircuit_prevent_test PRIVATE glasgow_constraint_solver)
+    add_test(NAME subcircuit_prevent COMMAND ${GCS_BASH} ${CMAKE_CURRENT_SOURCE_DIR}/../run_test_only.bash $<TARGET_FILE:subcircuit_prevent_test>)
+    add_view_tests(subcircuit_prevent subcircuit_prevent_test 8)
+
+    add_executable(subcircuit_test constraints/circuit/subcircuit_test.cc)
+    target_link_libraries(subcircuit_test PRIVATE glasgow_constraint_solver)
+    add_test(NAME subcircuit_constraint COMMAND ${GCS_BASH} ${CMAKE_CURRENT_SOURCE_DIR}/../run_test_only.bash $<TARGET_FILE:subcircuit_test>)
+    add_view_tests(subcircuit_constraint subcircuit_test 5)
+
+    add_executable(subcircuit_dup_test constraints/circuit/subcircuit_dup_test.cc)
+    target_link_libraries(subcircuit_dup_test PRIVATE glasgow_constraint_solver)
+    add_test(NAME subcircuit_dup COMMAND $<TARGET_FILE:subcircuit_dup_test>)
 
     add_executable(circuit_dup_test constraints/circuit/circuit_dup_test.cc)
     target_link_libraries(circuit_dup_test PRIVATE glasgow_constraint_solver)
--- a/gcs/constraints/circuit.hh
+++ b/gcs/constraints/circuit.hh
@@ -2,5 +2,6 @@
 #define GLASGOW_CONSTRAINT_SOLVER_GUARD_GCS_CIRCUIT_HH
 
 #include <gcs/constraints/circuit/circuit.hh>
+#include <gcs/constraints/circuit/subcircuit.hh>
 
 #endif // GLASGOW_CONSTRAINT_SOLVER_GUARD_GCS_CIRCUIT_HH
--- a/gcs/constraints/circuit/hints.hh
+++ b/gcs/constraints/circuit/hints.hh
@@ -17,6 +17,17 @@ namespace gcs::innards::hints
         ConstraintID originator;
         static constexpr std::string_view hint_name = "circuit";
     };
+
+    /**
+     * \brief SubCircuit's assertion hint: just the owning constraint.
+     *
+     * \ingroup Innards
+     */
+    struct SubCircuit
+    {
+        ConstraintID originator;
+        static constexpr std::string_view hint_name = "subcircuit";
+    };
 }
 
 #endif
--- a/gcs/constraints/circuit/subcircuit.cc
+++ b/gcs/constraints/circuit/subcircuit.cc
@@ -1,0 +1,481 @@
+#include <gcs/constraints/all_different.hh>
+#include <gcs/constraints/all_different/encoding.hh>
+#include <gcs/constraints/all_different/vc_all_different.hh>
+#include <gcs/constraints/circuit/hints.hh>
+#include <gcs/constraints/circuit/subcircuit.hh>
+#include <gcs/exception.hh>
+#include <gcs/innards/inference_tracker.hh>
+#include <gcs/innards/proofs/names_and_ids_tracker.hh>
+#include <gcs/innards/proofs/pol_builder.hh>
+#include <gcs/innards/proofs/proof_model.hh>
+#include <gcs/innards/propagators.hh>
+#include <gcs/innards/s_expr.hh>
+
+#include <util/enumerate.hh>
+#include <util/overloaded.hh>
+
+#include <algorithm>
+#include <any>
+#include <map>
+#include <memory>
+#include <optional>
+#include <string>
+#include <utility>
+#include <variant>
+#include <vector>
+#include <version>
+
+#if defined(__cpp_lib_print) && defined(__cpp_lib_format)
+#include <format>
+#else
+#include <fmt/core.h>
+#endif
+
+using namespace gcs;
+using namespace gcs::innards;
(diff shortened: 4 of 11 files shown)


ciaranm:
Merge review (Merger Claude).

Reviewed as a whole stack: I trial-merged all nine of #790 → #792 → #793 → #794 → #797 → #798 → #800 → #802 → #801 onto current `main` (`888d5433`, two merges past the `f16c67e2` this branch is based on). All nine merge clean; the merged tree builds with **zero warnings**, is clang-format 21 clean across all 14 touched C++ files, and passes **786/786** under ctest with VeriPB 3.0.2.

## One finding

`SubCircuitPosData`'s doc comment in `gcs/constraints/circuit/subcircuit_base.hh` describes the superseded numbering:

> `pos[i]` is node i's index along the tour, counting from whichever node is `first`; a node off the tour is given a position after every on-tour one, which is what makes the positions a permutation of 0..n-1 in every solution and so keeps them determined by the successors under unit propagation.

The code in this same commit writes `off_pos_<i>: pos[i] <= 0`, and `define_proof_model()`'s own comment explains at length why the positions are deliberately *not* a permutation — as does the PR description.

#797 rewrites this comment, so the end state of the stack is correct. But #790 through #794 each land a `main` whose header asserts the positions are a permutation of `0..n-1`, which is precisely the premise that makes a pigeonhole argument over positions look available when the whole point of the design is that it isn't. Cheaper to fix here than to leave four intermediate commits saying it.

## Confirming the certificate claim — with a caveat worth recording

I reproduced the mutation probe rather than taking it on trust. Replacing `derive_tour_at_most` with a plain `JustifyUsingRUP` makes VeriPB reject at `subcircuit_test`'s **n = 5** case (n = 3 and n = 4 both pass), exactly as described; mutating the `check` call site alone is already enough to be caught.

The caveat: it only reproduces with the seed pinned. `subcircuit_test` randomises its view-wrap configuration, and two of my probe runs appeared to **pass** before I pinned `--seed=1`. Anyone re-running the probe — which #801's testing notes rightly say to do after any encoding change — needs a fixed seed for the result to mean anything.

I also confirmed separately that #800's anchor auto-discovery does not quietly defeat `subcircuit_prevent_test`, whose whole purpose is the undetermined anchor: measured at the top of the stack it still has 0 `anchor_pos` rows and 8 `first_pos` rows, so posted `In` constraints do not narrow the declared domains `prepare()` reads and the test still exercises what it says it does.

ciaranm:
Thanks — all three points taken, and the whole stack is restacked with the fixes. `main` is unchanged as the base (`888d5433`), every commit still builds warning-free and passes its own tests, and the tip is 787/787.

## The doc comment: fixed here, as you suggested

You are right that it is cheaper here, and right about *why* it matters rather than just being stale: the superseded wording asserts the positions are a permutation of `0..n-1`, which is exactly the premise that makes a pigeonhole over positions look available — and the design's whole point is that it deliberately is not. Four intermediate commits saying so is four chances to build on it.

`SubCircuitPosData`'s comment in this commit now reads:

> `pos[i]` is node i's index along the tour, counting from whichever node is `first`; a node off the tour is given position **zero**, so the positions are not a permutation, and `pos[i] >= 1` already says node i is on the tour. Nothing needs them distinct: an off-tour node takes part in no position row at all, and all these rows have to do is leave `pos` determined by the successors under unit propagation, which is what solution checking needs. define_proof_model() says more about why, including what the permutation the stdlib decomposition does build would be good for.

#797 still updates it, but only to add the anchor clause — "counting from the anchor if there is one and otherwise from whichever node is `first`" — so the claim is right at every point in the stack now rather than only at the end. I did leave `EdgePosLines`' comment alone here, since "the anchor is whichever node is `first`" is simply true until #797 introduces the alternative.

## The unseeded probe

This is the more valuable of the two findings and I had not spotted it. It is now in #801's testing notes as part of the bullet rather than a footnote to it:

> **A mutation probe is the only thing that shows a certificate is load-bearing, and it needs a pinned seed.** [...] with `--seed=N`. `subcircuit_test` randomises its view-wrap configuration, and an unseeded probe can come out *passing*: the wrap it happens to pick decides whether unit propagation reaches the conflict unaided. A probe that can silently come out the wrong way is worse than no probe. With the seed pinned it rejects at `n = 5`, and `n = 3` and `n = 4` still pass, so the scenario size matters too.

I have also recorded it outside the repo, since it generalises past this constraint: a mutation probe on a test that randomises anything needs the seed pinned or it is not a probe.

Thank you for re-running the probe and the `subcircuit_prevent_test` check rather than taking either on trust — the second one is the specific thing I would have worried about after #800, and it is much better to have it measured at the top of the stack than argued from the source.

Issue #791, opened 2026-09-02
DAG: `dag` falls through to a distance labelling that writes 24.5 GB of proof for a twelve-node graph


ciaranm:
`dag` is the one member of the `globals.graph` ladder in #637 that was not built. #637 listed it as stage 5, noted it is "structurally different — a longest-path distance labelling with no tree — and can be done independently", and has now been closed as complete for the reachability–tree family (#784, #787). `network_flow` / `network_flow_cost` were declared out of scope there and remain so; `subcircuit` went to #788. So `dag` needs a home of its own, and this is it.

`minizinc/mznlib/` overrides no `fzn_dag`, and `minizinc/fzn_glasgow.cc` names it nowhere, so it falls through to the stdlib decomposition.

## What the fall-through costs

MiniZinc 2.10.1's `std/fzn_dag.mzn`, in full:

```
array [index_set(ns)] of var 0..length(ns) - 1: dist;   /* distance of longest path */
forall (n in index_set(ns)) (not ns[n] -> dist[n] = 0) /\
forall (e in EDGE) (es[e] -> dist[from[e]] + 1 <= dist[to[e]]) /\
% redundant constraint to ensure all distances are fixed
forall (n in index_set(ns)) (
  dist[n] = max([0] ++ [(dist[from[e]] + 1) * es[e] | e in EDGE where to[e] = n])
)
```

It is the same family of distance labelling that made `reachable` expensive, **plus a variable-by-variable product per edge**: the redundant `max` term multiplies a distance by an edge Boolean, so every edge brings an `int_times`. Flattening `dag` over 20 nodes and 400 edges through our own `mznlib/` gives 1300 variables and 1300 constraints:

| builtin | count |
|---|--:|
| `bool_clause` | 420 |
| `int_times` | 400 |
| `bool2int` | 400 |
| `array_int_maximum` | 20 |
| `int_lin_le_reif` | 20 |
| `int_lin_eq` | 20 |
| `int_eq_reif` | 20 |

(Under the default stdlib it is 1680/1680, the difference being 400 `int_max` that our library folds into 20 `array_int_maximum`. The products survive either way, and they are the part that matters.)

## Measured, on current `main` (`888d5433`)

Maximum acyclic subgraph posted *through the global* — every node selected, maximise selected edges — which is the shape the one corpus model below solves by hand. Release build, `--prove`, random digraphs (seed 7):

[...]

PR #792, opened 2026-09-02
Reach SubCircuit from MiniZinc

ciaranm:
Step 2 of #788, stacked on #790.

`fzn_subcircuit` fell through to the stdlib's position labelling — a variable-indexed element
into `order` per node, a reified guard on it, a second element for the not-in case, a second
`alldifferent`, and `firstin` as a min over *n* reified terms. Every subcircuit instance in
the Challenge corpus got that. Now they get the constraint.

## The override

Shifts successors to 0-based exactly as `fzn_circuit.mzn` does, and the shift is consistent
for the opt-out case too: a node off the tour has `x[i] = i`, which after shifting is "the
value equals the 0-based position".

One thing worth not tidying, and there is a comment saying so: the `min` stays *inside* the
comprehension. This override replaces the stdlib's own `length(xs) = 0` guard, and with an
empty index set the comprehension body — and so the `min` — never evaluates, so
`subcircuit([])` still flattens. Hoisting the `min` into a `let` would turn that into a
min-of-empty-set error, which is what the stdlib guard was there for. (`fzn_circuit.mzn` is
safe for the same accidental reason; Gecode's own `fzn_circuit` is not, but that is not ours.)

## Tests

Both carry `--fzn-pattern glasgow_subcircuit`, and that guard is load-bearing rather than
ceremonial: without it these tests pass against the decomposition, which agrees with the
reference solver and verifies its proof just as happily, and nothing else the harness checks
can tell that the builtin never ran. Verified by removing the override and confirming the
test then fails.

- `subcircuittest` — 0-based, four nodes, so it covers the index shift the way
  `circuittest.mzn` does. Four is the smallest size that says anything beyond
  all-different, since two disjoint cycles need four nodes between them; 21 of the 24
  permutations are solutions, the identity (the empty tour) among them.
- `subcircuitevidencetest` — 1-based, so the other half of the shift, and pins node 1 onto
  the tour. That pin is what gives the propagator an evidence node, without which no chain
  can be stopped from closing. It is the case `subcircuit_prevent_test.cc` drives directly,
  here made to survive the whole frontend.

## Support matrix

`subcircuit` gets its own row beside `circuit`, and loses its mention in the "rest of
`globals.graph`" row, where the reachability ladder had been carrying it. The XCSP3 column
says "not yet bound" and why — XCSP3's plain `<circuit>` is *this* constraint's semantics
rather than `Circuit`'s, which is #167, and that is the next PR.

## Testing

781/781 pass, including all 93 `minizinc-*` entries.

A measurement of this against the decomposition over all 27 `tpp`/`mario` instances is
running and will be added as a comment — it is the acceptance item #788 asks for, and it
belongs on this PR rather than the propagator one, since it is the binding that decides
which of the two the corpus actually gets.

### DIFF
--- a/dev_docs/frontend-support-matrix.md
+++ b/dev_docs/frontend-support-matrix.md
@@ -53,6 +53,7 @@ equivalent for that frontend's vocabulary).
 | binPacking | `BinPacking` (per-bin GAC)[^bp] | ✓ (`fzn_bin_packing` / `_capa` / `_load`) | ✓ (signatures 1/2/3; per-bin condition list `s UNSUPPORTED`) | ? |
 | knapsack | `Knapsack` | ✓ | ✓ (basic with two `XCondition`s; not yet exercised by a test) | ? |
 | circuit | `Circuit` | ✓ | ✓ (basic; sub-circuit with size param `s UNSUPPORTED`); semantics mismatch with XCSP3 spec, see #167 | ? |
+| subcircuit | `SubCircuit` | ✓ | not yet bound; XCSP3's plain `<circuit>` is *this* constraint's semantics, not `Circuit`'s, and its size-carrying forms are `s UNSUPPORTED` — see #167 and [#788](https://github.com/ciaranm/glasgow-constraint-solver/issues/788) | ? |
 | instantiation | `Equals` to constant | ✓ | ✓ | ? |
 | lex (ordered list) | `LexLessThan` / `LexLessThanEqual` / `LexGreaterThan` / `LexGreaterEqual` | ✓ | ✓ (lists; matrix as lex² over rows + columns) | ? |
 | slide (meta-constraint) | apply template per window | ? | ✓ (parser unfolds into per-window constraints) | ? |
@@ -80,7 +81,7 @@ addressed.
 | graph reachability (`reachable`, `dreachable`, `connected`, `dconnected`) | `Reachable` / `DReachable` | ✓ | n/a | ? | `connected` / `dconnected` ride the same override, being `reachable` / `dreachable` with an existential root; see [^reach] |
 | `subgraph` | `Subgraph` | ✓ | n/a | n/a | Two implications per edge, so the constraint exists for the C++ and `.scp` interfaces rather than to infer anything the decomposition would not |
 | graph trees and paths (`tree`, `dtree`, `path`, `dpath`) | `Tree` / `DTree` / `Path` / `DPath` | ✓ | n/a | n/a | Also reached by `steiner`, `dsteiner`, `bounded_path`, `bounded_dpath` and both `weighted_spanning_tree` spellings, which the stdlib defines in terms of these; see [^treefam] |
-| the rest of `globals.graph` (`dag`, `network_flow`, `network_flow_cost`, `subcircuit`) | – | decomposition | n/a | n/a | Not on the reachability ladder: `dag` is an acyclicity labelling, `network_flow` decomposes to plain flow conservation, and `subcircuit` belongs with `Circuit` ([#167](https://github.com/ciaranm/glasgow-constraint-solver/issues/167)). Open under [#637](https://github.com/ciaranm/glasgow-constraint-solver/issues/637) |
+| the rest of `globals.graph` (`dag`, `network_flow`, `network_flow_cost`) | – | decomposition | n/a | n/a | Not on the reachability ladder: `dag` is an acyclicity labelling and `network_flow` decomposes to plain flow conservation. Open under [#637](https://github.com/ciaranm/glasgow-constraint-solver/issues/637). `subcircuit` was listed here too until [#788](https://github.com/ciaranm/glasgow-constraint-solver/issues/788) gave it `SubCircuit`; it has its own row above, beside `circuit`, where it belongs |
 | `SmartTable` | `SmartTable` | ✓ | n/a | ? | Glasgow-specific extension |
 
 ## Solver gaps tracked elsewhere
--- a/minizinc/CMakeLists.txt
+++ b/minizinc/CMakeLists.txt
@@ -43,6 +43,16 @@ set_tests_properties(minizinc-cake PROPERTIES SKIP_RETURN_CODE 66)
 add_test(NAME minizinc-circuit COMMAND ${GCS_BASH} ${CMAKE_SOURCE_DIR}/minizinc/run_minizinc_test.bash $<TARGET_FILE:fzn-glasgow> ${CMAKE_SOURCE_DIR}/minizinc/ circuittest true true)
 set_tests_properties(minizinc-circuit PROPERTIES SKIP_RETURN_CODE 66)
 
+# --fzn-pattern is load-bearing here: without it these would pass against the
+# stdlib decomposition, which agrees with the reference solver and verifies its
+# proof just as happily, and nothing else the harness checks would notice that
+# the builtin never ran.
+add_test(NAME minizinc-subcircuit COMMAND ${GCS_BASH} ${CMAKE_SOURCE_DIR}/minizinc/run_minizinc_test.bash --fzn-pattern glasgow_subcircuit $<TARGET_FILE:fzn-glasgow> ${CMAKE_SOURCE_DIR}/minizinc/ subcircuittest true true)
+set_tests_properties(minizinc-subcircuit PROPERTIES SKIP_RETURN_CODE 66)
+
+add_test(NAME minizinc-subcircuit-evidence COMMAND ${GCS_BASH} ${CMAKE_SOURCE_DIR}/minizinc/run_minizinc_test.bash --fzn-pattern glasgow_subcircuit $<TARGET_FILE:fzn-glasgow> ${CMAKE_SOURCE_DIR}/minizinc/ subcircuitevidencetest true true)
+set_tests_properties(minizinc-subcircuit-evidence PROPERTIES SKIP_RETURN_CODE 66)
+
 # The reachability family. --fzn-pattern is the reachability guard: MiniZinc's
 # connected / dconnected wrappers go through reachable / dreachable, so without
 # it a test would still agree with the reference solver via the stdlib distance
--- a/minizinc/fzn_glasgow.cc
+++ b/minizinc/fzn_glasgow.cc
@@ -796,6 +796,15 @@ auto main(int argc, char * argv[]) -> int
                 const auto & vars = arg_as_array_of_var(data, args, 0);
                 problem.post(Circuit{vars});
             }
+            else if (id == "glasgow_subcircuit") {
+                // The mznlib redefinition has already shifted successors to be
+                // 0-based, as for glasgow_circuit. SubCircuit is the opt-out
+                // sibling: a node not on the tour takes its own index, and the
+                // empty tour is a solution, so unlike Circuit this must not
+                // assume every node is visited.
+                const auto & vars = arg_as_array_of_var(data, args, 0);
+                problem.post(SubCircuit{vars});
+            }
             else if (id == "glasgow_reachable" || id == "glasgow_dreachable") {
                 // The mznlib redefinition has already shifted the endpoints and
                 // the root to be zero-based, matching how Reachable numbers nodes.
--- a/minizinc/mznlib/fzn_subcircuit.mzn
+++ b/minizinc/mznlib/fzn_subcircuit.mzn
@@ -1,0 +1,15 @@
+predicate glasgow_subcircuit(array[int] of var int: x);
+
+% glasgow_subcircuit expects 0-based successors (a length-n array whose values are
+% in 0..n-1). FlatZinc subcircuit is defined over index_set(x) with successor
+% values in that same set, so shift by min(index_set(x)) rather than assuming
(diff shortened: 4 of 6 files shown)


ciaranm:
## Measured: a real challenge instance is now certifiable

The [step-0 comment](https://github.com/ciaranm/glasgow-constraint-solver/issues/788#issuecomment-5505574380)
found that with the decomposition we could not certify a single real subcircuit instance —
a *cut-down* 12-house `2013/mario` already needed a 537 MB proof and 19 minutes to check,
and the real instances start at 15 houses. With the binding in this PR:

**`2013/mario/mario_easy_2`, unmodified, 15 houses: 10,021 nodes, 0.056 s to solve, 0.53 s
with proof logging, an 894 KB `.opb`, a 194 MB `.pbp`, and `s VERIFIED BOUNDS -628 <= obj
<= -628` in 188 s.** Three minutes, end to end, for a certified optimum on a Challenge
instance.

The same cut-down curve as step 0, now with both configurations (same binary, same models,
the only difference being whether `mznlib/fzn_subcircuit.mzn` is present):

| houses | nodes decomp → native | `.pbp` decomp → native | verify decomp → native |
|--:|---|---|---|
| 6 | 59 → 17 | 357 KB → 114 KB | 0.03 s → 0.1 s |
| 8 | 509 → 47 | 3.4 MB → 419 KB | 0.62 s → 0.1 s |
| 10 | 6,681 → 331 | 52 MB → 3.2 MB | 45.0 s → **0.5 s** |
| 12 | 52,477 → 2,549 | 537 MB → 33.8 MB | 1,133 s → **13.1 s** |
| 14 | — | — | → 37.0 s |
| **15 (the real instance)** | 1,510,892 → **10,021** | not attempted → **194 MB** | not attempted → **188 s** |

The decomposition's 15-house proof was deliberately not attempted: at 12 houses it needed
1,133 s for 537 MB from 52,477 nodes, seconds-per-MB was still climbing (0.08 → 0.18 → 0.87
→ 2.11), and the real instance takes 1.5M nodes — some 29× the search, so tens of gigabytes
and many hours. The point of the row is that the comparison is not close enough to need
measuring.

### The `.opb` moves the other way, and that is worth knowing

#788 asks for `.opb` size, `.pbp` size and verify time separately "since those move
independently". They do, and here two of them move in opposite directions. On
`mario_easy_2` the encoding is *larger* with the native constraint:

| | rows | bytes |
|---|--:|--:|
| decomposition | 3,255 | 638 KB |
| `SubCircuit` | 3,607 | 893 KB |

40% more `.opb` for a 60× smaller `.pbp`. The reason is the case split #790 describes: not
knowing statically which edge wraps costs **four** rows per edge, where `Circuit` needs two,
so the position encoding is 4n² rather than 2n². A nominated mandatory-node anchor would
halve it — which is a second argument for the `with_required_node` shortcut, on top of the
cheaper per-firing certificate.

ciaranm:
## Measured: the whole corpus, decomposition against native

All 27 instances, 60 s cap, same binary both ways — the only difference is whether
`mznlib/fzn_subcircuit.mzn` is present. **This one is not a clean win, so read the
regressions before the headline.**

Two things to know before the table. First, on a capped-out instance the node count measures
*throughput*, not effort — the native configuration is roughly 4-7× faster per node on
`mario` because there are no `order` variables and no variable-indexed elements to propagate
through, so it visits more nodes in the same minute. Only the uncapped rows compare search
effort. Second, `mario` maximises and `tpp` minimises.

| instance | decomposition | native |
|---|---|---|
| 2012 tpp_3_3_30_1 | 60.1s, 18.17M n, obj 232 *(capped)* | 60.3s, 21.55M n, obj **226** *(capped)* |
| 2012 tpp_3_5_20_1 | 22.6s, 7.51M n, obj 127 | 21.3s, 7.50M n, obj 127 |
| 2012 tpp_5_3_20_1 | 21.9s, 7.06M n, obj 141 | 20.7s, 7.05M n, obj 141 |
| 2012 tpp_5_5_20_1 | 26.5s, 6.45M n, obj 115 | 24.3s, 6.45M n, obj 115 |
| 2012 tpp_7_3_20_1 | 60.2s, 17.93M n, obj 135 *(capped)* | 60.2s, 19.68M n, obj **132** *(capped)* |
| 2012 tpp_7_5_20_1 | 60.2s, 11.78M n, obj 105 *(capped)* | **54.2s, 11.81M n, obj 105** |
| 2012 tpp_7_5_30_1 | 60.2s, 16.09M n, obj 209 *(capped)* | 60.2s, 18.62M n, obj 209 *(capped)* |
| 2016 tpp_4_5_20_1 | 60.2s, 17.03M n, obj 138 *(capped)* | 60.2s, 18.75M n, obj 138 *(capped)* |
| 2016 tpp_5_3_30_1 | 60.2s, 15.72M n, obj 203 *(capped)* | 60.2s, 15.44M n, obj 203 *(capped)* |
| 2016 tpp_6_3_20_1 | 21.1s, 5.63M n, obj 131 | 19.4s, 5.63M n, obj 131 |
| 2016 tpp_7_3_20_1 | 60.2s, 18.17M n, obj 134 *(capped)* | 60.2s, 19.84M n, obj **132** *(capped)* |
| 2016 tpp_7_3_30_1 | 60.2s, 15.89M n, obj 221 *(capped)* | 60.2s, 17.57M n, obj 221 *(capped)* |
| 2013 mario_easy_2 | 23.2s, 1.51M n, obj 628 | **0.2s, 10K n, obj 628** |
| 2013 mario_easy_4 | 5.2s, 315K n, obj 545 | **0.5s, 53K n, obj 545** |
| 2013 mario_n_medium_2 | 60.2s, 1.58M n, obj 967 *(capped)* | 60.2s, 5.79M n, obj 861 *(capped)* |
| 2013 mario_n_medium_4 | **57.8s, 1.82M n, obj 832** | 60.2s, 6.76M n, obj 648 *(capped)* |
| 2013 mario_t_hard_1 | 60.2s, 46K n, **obj 0** *(capped)* | **17.7s, 1.13M n, obj 4783** |
| 2014 mario_easy_5 | 0.8s, 42K n, obj 445 | **0.3s, 16K n, obj 445** |
| 2014 mario_n_medium_3 | **30.8s, 894K n, obj 943** | 60.2s, 6.24M n, obj 897 *(capped)* |
| 2014 mario_n_medium_5 | 60.2s, 1.62M n, obj 771 *(capped)* | 60.2s, 5.94M n, obj 771 *(capped)* |
| 2014 mario_t_hard_2 | 60.2s, 46K n, **obj 0** *(capped)* | 60.2s, 4.27M n, obj **4514** *(capped)* |
| 2014 mario_t_hard_5 | 60.2s, 46K n, **obj 0** *(capped)* | 60.2s, 4.18M n, obj **4421** *(capped)* |
| 2017 mario_medium_1 | 60.2s, 1.62M n, **obj 0** *(capped)* | 60.2s, 6.35M n, obj **1095** *(capped)* |
| 2017 mario_medium_3 | 60.2s, 1.67M n, **obj 0** *(capped)* | 60.2s, 6.06M n, obj **1016** *(capped)* |
| 2017 mario_medium_4 | 60.2s, 1.71M n, **obj 0** *(capped)* | 60.2s, 6.26M n, obj **885** *(capped)* |
| 2017 mario_medium_5 | 60.2s, 1.70M n, **obj 0** *(capped)* | 60.2s, 5.53M n, obj **900** *(capped)* |
| 2017 mario_n_medium_1 | 60.2s, 1.62M n, obj 1006 *(capped)* | 60.2s, 6.13M n, obj 1006 *(capped)* |

### The regressions

**Two instances we used to solve, we now do not.** `2013/mario_n_medium_4` was proved
optimal at 832 in 57.8 s and now caps out at 648; `2014/mario_n_medium_3` was proved optimal
at 943 in 30.8 s and now caps out at 897. `n_medium_2` also gets a worse bound (967 → 861).
So on the `optimal within 60 s` count this is a wash: **9/27 either way** — native gains
`mario_t_hard_1` and `tpp_7_5_20_1`, and loses those two.

The `n_medium` family is visiting 4-7× more nodes and converging more slowly with the
constraint than with the decomposition, which means the decomposition's auxiliary structure
is doing something useful for the *search* there that we are not replacing — most likely its
`order` variables giving the brancher a second, better-ordered set of decisions. Worth saying
plainly that this is not a surprise and not specific to us: the same family got nothing from
Chuffed's own native subcircuit propagator in the
[step-0 experiment](https://github.com/ciaranm/glasgow-constraint-solver/issues/788#issuecomment-5505574380)
(822K vs 852K nodes), and on `2017/n_medium_1` Chuffed's native propagator was *worse* than
its decomposition too. Three solver configurations agreeing on which family this does not
help is a diagnosis, not noise.

### The wins

- **7 instances return nothing but the empty subcircuit with the decomposition. With the
  constraint, none do.** All three `t_hard` and all four 2017 `medium` instances go from
  objective 0 to a real solution, and `t_hard_1` from nothing at all to a proved optimum in
  17.7 s.
- Objective quality across the corpus: 10 instances better, 3 worse, 14 equal.
- `mario_easy_2` is 116× faster (23.2 s → 0.2 s) and `easy_4` 10×.
- `tpp` is flat on search — identical node counts on every uncapped instance — but ~7% faster
  per node, which is enough to flip `tpp_7_5_20_1` inside the cap. `tpp` gaining nothing is
  also what Chuffed shows (1.2-1.8× nodes, no time), and matches Francis and Stuckey: for
  subcircuit their `scc` arm is what moves the needle, and only when explained.

### What I would take from this

The case for merging does not rest on the search numbers, which are mixed. It rests on the
certification result in the [comment above](https://github.com/ciaranm/glasgow-constraint-solver/pull/792#issuecomment-5506756726)
— a real instance certified in three minutes where the decomposition could not be certified
at all — and on the coverage win, which is qualitative. #788's acceptance asks that both
directions stay reachable rather than the decomposition being deleted, and they do: removing
`mznlib/fzn_subcircuit.mzn` restores it exactly, which is how these numbers were produced.

The `n_medium` regression is the natural first target for the SCC arm, and a good reason not
to treat that arm as optional.

ciaranm:
## Correction to the `.opb` numbers above, and why it cannot be much smaller

Following [the discussion of whether the encoding should be the decomposition
instead](https://github.com/ciaranm/glasgow-constraint-solver/pull/790), #790 now pins
off-tour positions to zero rather than numbering them after the on-tour ones. That shaves
exactly n rows and 2% of bytes, so the table in the comment above should read:

| | rows (subcircuit's share) | bytes |
|---|--:|--:|
| decomposition | 1,531 | 638 KB |
| `SubCircuit`, as first pushed | 1,883 | 893 KB |
| `SubCircuit`, now | **1,868** | 873 KB |

Still 22% more rows than the decomposition, and that 22% is entirely the wrap row family.
Three things came out of chasing whether it could go:

**Dropping the wrap rows is not a free reformulation into the proof.** The remaining
encoding would be smaller than the decomposition's and still exact, but the counting fact the
wrap rows hand the propagator then has to be derived, and the derivation needs "the on-tour
positions are exactly `0..L-1`" — a pigeonhole induction conditional on the variable `L`.
That is the same obstacle that blocks porting `circuit_scc.cc`, so it is not available yet.
`define_proof_model()` now says this in a comment so the next reader does not have to
rediscover it.

**Circuit is small because its anchor is static, not because its encoding is cleverer.** It
writes one row family per edge — the `+1` form for every column but 0, the wrap form for
column 0 — and never both, because it knows node 0 wraps. A caller-nominated anchor collapses
`SubCircuit` to exactly that shape. MiniZinc cannot supply one, though, and I checked rather
than assumed: the flattener *does* narrow a domain before the redefinition runs, but only if
the pin appears earlier in the source, and both `tpp` and `mario` post theirs *after* the
`subcircuit` call. Moving the pin one line up makes it visible; leaving it where it is does
not.

**The change had a better payoff elsewhere than in bytes.** It moved the mutation-catching
property out of the hand-built scenario and into the plain enumeration test: with the old
off-tour numbering, a plain `JustifyUsingRUP` passed the entire enumeration at every n and
every view wrapping; now it fails at n = 5. Two percent of `.opb` for a check that runs on
every lane instead of one is the better half of this trade.

ciaranm:
Merge review (Merger Claude).

The `--fzn-pattern glasgow_subcircuit` guard on both new lanes is the right call, and the CMake comment explaining why it is load-bearing is worth having. The `min(index_set(x))` reasoning about the empty index set is also correct as written (the comprehension body never evaluates, so the `min` never does).

## One finding

This PR rewrites the `globals.graph` row in `dev_docs/frontend-support-matrix.md`, and the rewritten line still ends:

> Not on the reachability ladder: `dag` is an acyclicity labelling and `network_flow` decomposes to plain flow conservation. Open under [#637](https://github.com/ciaranm/glasgow-constraint-solver/issues/637).

#637 is now closed as completed, so that row points readers at a closed issue for three constraints that are still open. `dag` has since got its own tracking issue, [#791](https://github.com/ciaranm/glasgow-constraint-solver/issues/791) — with the measurements for why the stdlib decomposition is unusable there — while `network_flow` and `network_flow_cost` have no tracking issue at all.

Since you are already editing this exact row to move `subcircuit` out of it, it seems the natural place to retarget: `dag` → #791, and say outright that `network_flow` / `network_flow_cost` are untracked rather than implying #637 still covers them.

PR #795, opened 2026-09-02
Extensional GAC: hoist the reason, name the selector concretely

ciaranm:
Two small changes to `propagate_extensional`, measured separately.

`generic_reason(table.vars)` was rebuilt at each inference site. The factories
take their scope by value, so every call copied the whole scope vector into a
fresh `shared_ptr` — including with proofs off, where the reason is never
materialised. It is now built once in `ExtensionalData`'s constructor, which is
sound because the scope is fixed and `Reason` is declarative.

`ExtensionalData::selector` is now a `SimpleIntegerVariableID` rather than an
`IntegerVariableID`. It always was one — every caller allocates it via
`State::allocate_integer_variable_with_state` — so this stops the type lying and
lets `State`'s concrete overloads apply.

### Measurements

fataepyc-10 pinned, boost off, median of 3, over the table benchmark matrix
(18 instances, matched search trees):

| | speedup |
|---|---|
| reason hoist alone | 1.0–1.1× |
| plus the concrete selector | **1.2–1.3×** |

The selector's type is worth most of it: pass 2 asks `in_domain()` about the
selector once per (variable, value) whose residue is still valid, and the
`selector != i` prunings go through the concrete `infer_not_equal` path — which
is millions of calls on a mid-sized instance.

### What it does not buy

Nothing on the table variables. `State::in_domain` is 21% of the profile, and
`perf annotate` shows its hot block is a linear scan over the `IntervalSet`'s
interval array with no discriminant test and no indirect branch — GCC had
already lowered that visit away. Specialising the feasibility pass over a
concrete `vector<SimpleIntegerVariableID>` was tried and measures 1.0×; see the
`propagator-performance.md` note in the follow-up PR for the full write-up.

Groundwork for #508. Follows #786.

### Testing

`ctest` 645/645. `recursions`, `propagations` and solution counts unchanged on
every benchmark.

### DIFF
--- a/gcs/constraints/extensional_utils.cc
+++ b/gcs/constraints/extensional_utils.cc
@@ -22,6 +22,11 @@ using std::visit;
 using namespace gcs;
 using namespace gcs::innards;
 
+gcs::innards::ExtensionalData::ExtensionalData(SimpleIntegerVariableID selector, vector<IntegerVariableID> vars, ExtensionalTuples tuples) :
+    selector(selector), vars(move(vars)), tuples(move(tuples)), reason(generic_reason(this->vars))
+{
+}
+
 namespace
 {
     auto feasible(const State & state, const IntegerVariableID & var, const Integer val) -> bool
@@ -88,13 +93,13 @@ auto gcs::innards::propagate_extensional(
                 else if (logger && logger->get_assertion_level() != AssertionLevel::Off && state.has_single_value(table.selector))
                     // Last selector val so infeasible -> we need an explicit contradiction at higher assertion levels
                     // since there's no table for the implicit one.
-                    inference.contradiction(logger, JustifyUsingRUP{hint}, generic_reason(table.vars));
+                    inference.contradiction(logger, JustifyUsingRUP{hint}, table.reason);
                 else
                     inference.infer(logger, table.selector != Integer(tuple_idx), NoJustificationNeeded{}, NoReason{});
             });
             if (none_feasible && logger && logger->get_assertion_level() != AssertionLevel::Off)
                 // selector already empty on entry
-                inference.contradiction(logger, JustifyUsingRUP{hint}, generic_reason(table.vars));
+                inference.contradiction(logger, JustifyUsingRUP{hint}, table.reason);
         },
         table.tuples);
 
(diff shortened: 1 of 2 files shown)


ciaranm:
Pre-merge review of the table stack (#795 → #796 → #799 → #804), run as a whole. Nothing to raise on #795 or #796; findings are on #799 and #804.

### Stack shape

Clean four-link chain, no forks. None draft, all four `MERGEABLE`/`CLEAN`, and 40/40 CI checks green — including all three `windows-2022` lanes on every PR.

### Verified against current main, not just against each PR's own base

The stack sits on `888d5433`, 20 commits behind `main` (`6343e6f0` — the SubCircuit stack landed in between), so only #795's checks have ever seen `main`, and none of them have seen *today's* `main`. There is no file overlap with what `main` gained, which rules out a textual conflict but not a semantic one, so I trial-merged the tip onto `6343e6f0` and built it:

- merges clean (`ort`, no conflicts);
- GCC 15.2 release build, exit 0, reached 100%, 134 test binaries present, **0 compiler warnings** — the post-#710 baseline still holds;
- **788/788 ctest pass**, with VeriPB 3.0.2 and MiniZinc both present, so the `run_test_and_verify` lanes really did verify rather than skip.

### Merge-order pricing

The only other open PR sharing a file with this stack is #806, on `dev_docs/propagator-performance.md`; it inserts at ~L81 against this stack's ~L316. Trial-merged the two together and they auto-merge, so there is no rebase cycle to pay in either order.

### Mechanics I read rather than trusting the tests

Since #796 and #804 change how the propagator signals and how it backtracks, the parts CI would be least likely to catch:

- `contradiction` is `[[noreturn]]` on all three overloads, so pass 2 stays unreachable after #796's new `if (0 == live_count)` — the code it replaces relied on exactly the same thing.
- Both the live-set and compact paths return `PropagatorState::EnableButIdempotent`, so #804's dispatch does not change re-queueing.
- #804's "don't react to your own prunings" shortcut is sound: the filter prunes `v` at position `idx` precisely when no live tuple holds `v` there, so the support removal it skips is already a no-op. With a duplicated variable the *other* position's `previous_domain` still carries `v`, so that position does get the removal applied.
- Tail bits stay clear in both `previous_domain` and the rasterised words (rasteriser and both seeders bound on `n_values`), so `v * n_words` never indexes past a mask.
- The `limit_marker` sentinel both seeders push is what makes `0 == trail_mark` mean "the search has climbed above the switch node" rather than "nothing has been recorded yet".
- `n_tuples` is captured before `move(tuples)` at both `ExtensionalData` construction sites.

Merge checks 3, 4, 6, 7, 8, 9 and 10 all clean across the stack.

PR #797, opened 2026-09-02
The SCC arm's reachability core, and the anchored encoding it needs

ciaranm:
Step 5 of #788, stacked on #794. Two commits: a caller-nominated anchor, then the
connectivity core of Francis and Stuckey's `scc` on top of it.

**I told you twice this arm was blocked on the `L·d[r,i]` nonlinearity. It isn't, and the
reason is worth stating**: that term exists to shift positions to an *arbitrary* root, and
this arm never needs an arbitrary root. It cannot infer anything until some node is known to
be on the tour (F&S's own observation), so if the position encoding is anchored on that same
node, the reachability argument is rooted where the positions already start and there is
nothing to shift. The shift, and its variable modulus, simply do not arise.

## `with_required_node()`

Names a node already *declared* on the tour — its own index not in its successor's domain —
and throws if it is not. A precondition, not a strengthening: the constraint means the same
thing either way, so nothing about it reaches the `.scp`, exactly as the choice of algorithm
does not. Range is checked at the call; the domain has to wait for `prepare()`, since
`post()` only stores the constraint.

It buys the encoding `Circuit` gets for free. Unanchored, every edge carries a row family for
each case and every certificate splits over the cycle's nodes; anchored, only the edges into
the named node can wrap — one family per edge, no `first` flags, one `pol` per certificate.
At n = 5 the `.opb` drops from 252 rows to 208, and the saving is 2n(n-1) + 3n lines so it
grows. It also strengthens propagation as a side effect: a closed cycle that misses the
anchor is now an immediate contradiction with no evidence node to find, which is how
`Circuit` reasons.

## The certificate

Walks out from the anchor a layer at a time, deriving for each node `x` unreached within `t`
steps that `pos[x] = t → succ[x] = x`. Same proof every layer: somebody is `x`'s predecessor,
and each candidate `q` either *is* `x` (the conclusion) or sits at `t-1` by the step row —
where a `q` already reached cannot point at `x` (in the reason) and a `q` not reached carries
the previous layer's fact. At `t = 0` the step row puts `q` at position −1, which the
variable's own lower bound refuses, so the first layer stands alone.

Position `t ≥ 1` already means "on the tour", because off-tour nodes sit at zero — so nothing
counts how long the tour is. That is the other half of why this works.

"Somebody is `x`'s predecessor" is *not* in the encoding: the all-different rows are a
pairwise clique, which says at most one variable takes a value and nothing about at least
one. So it is derived by pigeonhole, once per value at `Top` and cached, the same way
`justify_all_different_hall_set_or_violator` does it — via `recover_am1_from_pairs()`, the
shared fold, rather than a fifth hand-rolled copy of the same staircase (see the comment
below and #805).

**That pigeonhole is the only cutting-planes step in the whole certificate**, and I had this
wrong at first in the other direction: I wrote a `pol` per candidate to carry "x is at t"
through the step row to "its predecessor is at t−1", on the assumption that unit propagation
could not do it. It can. The step row is written as two halves, `<=` and `>=`, over
`pos[j] - pos[i]` for the edge `i -> j`; the forward walk's edge is `q -> x`, so with
`pos[x]` fixed at `t` the `>=` half gives `pos[q] <= t-1` and the `<=` half gives
`pos[q] >= t-1`, and between them they pin every bit. It is the `>=` half that bounds the
predecessor from above, which reads the wrong way round — and which half does which flips
with the direction of the walk, since #798's backward walk follows `x -> y` instead. Both pols — the per-candidate one and
the one combining the candidates with the pigeonhole — were probed by deleting them, and
every test still verifies with neither, while a plain `JustifyUsingRUP` in place of the whole
function is still rejected and dropping the pigeonhole *does* make VeriPB reject. So the
lemma chain is what is load-bearing, not the arithmetic. Nothing has to cite a step row
either, since those are model rows; that works only because the encoding is anchored, as
without an anchor each half is guarded on a `first` flag that nothing here would have fixed.
The `.pbp` for `subcircuit_scc_test` drops 6.3% with the dead pols gone.

## On the honest value of the rule

**It saves two search nodes out of 558.** Check and prevent with an evidence node get to
nearly the same place by another route, so plain reachability is close to subsumed by them.
Whatever F&S measure for `scc` on subcircuit must come mostly from the four pruning rules and
from separating multiple components — none of which is here yet.

`subcircuit_scc_test` took three attempts to say anything at all. With the halves the other
way round, or the same size, both algorithms come out at **exactly** the same recursion count,
because the brancher never enters the unreachable half before the tour closes, and then
`check` forces it off for free. Giving that half the smallest domains is what makes the search
walk into it. The test therefore asserts the *root state* — six of six opted out with `SCC`,
zero of six with `Prevent` — rather than a recursion margin, because two nodes out of 558 is
not a number to hold a test to.

## Testing

784/784 pass. Anchored enumeration at n = 3, 4, 5 for the first and last node as anchor,
under both `Prevent` and `SCC`; both rejection paths for the precondition; the scenario above.
Clean under `clang++` with sanitizers. Swapping either certificate for a plain
`JustifyUsingRUP` makes VeriPB reject the proof, which is what keeps them honest.

## What is left of the arm

The four pruning rules, and the multiple-component contradiction. Those are where the value
looks likely to be, on the evidence above — and the `n_medium` regression from #792 is the
thing to point them at.

### DIFF
--- a/gcs/CMakeLists.txt
+++ b/gcs/CMakeLists.txt
@@ -774,6 +774,11 @@ if(GCS_BUILD_TESTS)
     add_test(NAME subcircuit_prevent COMMAND ${GCS_BASH} ${CMAKE_CURRENT_SOURCE_DIR}/../run_test_only.bash $<TARGET_FILE:subcircuit_prevent_test>)
     add_view_tests(subcircuit_prevent subcircuit_prevent_test 8)
 
+    add_executable(subcircuit_scc_test constraints/circuit/subcircuit_scc_test.cc)
+    target_link_libraries(subcircuit_scc_test PRIVATE glasgow_constraint_solver)
+    add_test(NAME subcircuit_scc COMMAND ${GCS_BASH} ${CMAKE_CURRENT_SOURCE_DIR}/../run_test_only.bash $<TARGET_FILE:subcircuit_scc_test>)
+    add_view_tests(subcircuit_scc subcircuit_scc_test 12)
+
     add_executable(subcircuit_test constraints/circuit/subcircuit_test.cc)
     target_link_libraries(subcircuit_test PRIVATE glasgow_constraint_solver)
     add_test(NAME subcircuit_constraint COMMAND ${GCS_BASH} ${CMAKE_CURRENT_SOURCE_DIR}/../run_test_only.bash $<TARGET_FILE:subcircuit_test>)
--- a/gcs/constraints/circuit/subcircuit.cc
+++ b/gcs/constraints/circuit/subcircuit.cc
@@ -5,6 +5,7 @@
 #include <gcs/constraints/circuit/subcircuit.hh>
 #include <gcs/exception.hh>
 #include <gcs/innards/inference_tracker.hh>
+#include <gcs/innards/proofs/am1_from_pairs.hh>
 #include <gcs/innards/proofs/names_and_ids_tracker.hh>
 #include <gcs/innards/proofs/pol_builder.hh>
 #include <gcs/innards/proofs/proof_model.hh>
@@ -44,7 +45,9 @@ using std::format;
 using fmt::format;
 #endif
 
+using std::cmp_equal;
 using std::cmp_greater_equal;
+using std::cmp_not_equal;
 using std::make_optional;
 using std::make_unique;
 using std::map;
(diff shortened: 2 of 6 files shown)


ciaranm:
Merge review (Merger Claude).

I measured the encoding claim directly rather than taking it on trust: at n = 12 the anchored model writes **242 `pos_step` + 22 `pos_wrap`** position rows against the unanchored 528, which is the exact halving claimed, and `(n-1)² = 121` step families plus `n-1 = 11` wrap families is where it comes from.

## Two findings

### 1. The bound directions in `derive_unreachable`'s comment are the wrong way round

The comment says:

> taking "x is at t" through the step row to "q is at t−1" is unit propagation on the two halves of that row, the `<=` half bounding the position from above and the `>=` half from below

`add_labelled_constraint(id, role_le, role_ge, expr == 1_i, guard)` returns the `<=` half first, so `step_le` is `pos[j] - pos[i] <= 1`. In the forward walk the edge is `q -> x` with `pos[x] = t`, so `j = x` and `i = q`:

- `step_le`: `pos[x] - pos[q] <= 1`, giving `pos[q] >= t - 1` — bounds `q` from **below**;
- `step_ge`: `pos[x] - pos[q] >= 1`, giving `pos[q] <= t - 1` — bounds `q` from **above**.

So it is the `>=` half that bounds from above, not the `<=` half. #801's design note has it the right way round — "the `≥` half bounds the neighbour's position from above while the `≤` half bounds it from below" — so the two currently contradict each other, and this is the half that is wrong. The same swap is in this PR's description.

Nothing about the proof or the code is affected; it is only the explanation, but it is an explanation the next reader will lean on precisely because the arithmetic is the non-obvious part.

### 2. `subcircuit_scc_test` is outside the view sweep

Every other scenario test in this family is in it. `circuit_disconnected`, `circuit_multiple_sccs`, `circuit_no_backedges`, `circuit_prune_root` and `circuit_prune_skip` all call `parse_view_wrap_config_from_argv` and get `add_view_tests`, and so does `subcircuit_prevent_test` from #790. This test uses neither, so the SCC propagation and both its certificates never run against a wrapped view.

Its assertions look view-invariant to me — 325 solutions, and root state 6 of 6 against 0 of 6, none of which a bijective remapping of values should move — so it reads as an omission rather than a decision. If it is a decision (the anchor's declared domain interacting with `create_integer_variable_or_constant_with_view` would be a plausible reason), a line in the test saying so would stop it looking like one. The same applies to `subcircuit_scc_reaches_anchor_test` in #798.

ciaranm:
Both findings are right, and both are fixed. Thank you especially for deriving the bound directions rather than reading past them — that is the half of the comment a reader is most likely to lean on, and it was wrong in the one direction that makes it useless.

## 1. The bound directions

Confirmed exactly as you set it out. `add_labelled_constraint(id, role_le, role_ge, expr == 1_i, guard)` returns the `<=` half first, the row is written over `pos[j] - pos[i]` for the edge `i -> j`, and the forward walk's edge is `q -> x` with `pos[x] = t`, so `step_ge` is the half giving `pos[q] <= t-1`. The comment now says so, and says the thing that makes it memorable — that it reads the wrong way round:

> taking "x is at t" through the step row to "q is at t-1" is unit propagation on the two halves of that row. The row is written over `pos[x] - pos[q]`, so with `pos[x]` fixed at t the `>=` half gives `pos[q] <= t-1` and the `<=` half gives `pos[q] >= t-1` -- the `>=` half is the one bounding q from above, which is the opposite way round from how it reads

One thing worth adding, since it is why the two documents could disagree without either being obviously wrong: **which half bounds which flips with the direction of the walk.** #798's backward walk follows `x -> y`, so there `i = x` and the `<=` half is the one bounding from above. #801's sentence was correct for the forward walk it was describing, but stated unconditionally, so I have made it explicit there rather than leaving it to be rediscovered as a contradiction:

> Which half does which flips with the direction of the walk, since the row is always written over `pos[j] - pos[i]` for the edge `i -> j`: chasing a *predecessor* it is the `≥` half that bounds from above, and chasing a *successor* it is the `≤` half.

The PR description above is corrected too.

## 2. The view sweep

An omission, not a decision — and your guess at what the decision *would* have been is the right thing to have checked, so I checked it rather than assuming. `subcircuit_scc_test` is now in the sweep: `parse_view_wrap_config_from_argv`, the out-of-range skip, `wraps_for_positions`, `create_integer_variable_or_constant_with_view`, the wrap label in the proof name, and `add_view_tests(subcircuit_scc subcircuit_scc_test 12)`. #798 does the same for `subcircuit_scc_reaches_anchor_test`.

The anchor survives wrapping for the reason that makes views the right tool everywhere in this stack: a view **is** the same variable seen through an offset, not a fresh one, so node 6's declared domain of `7..11` is still what `prepare()` and `with_required_node()` read. There is a comment in the test saying that, since it is the thing a future reader would reasonably worry about.

Measured rather than assumed, all 18 wraps at `--view-position=all` for both tests: 325 solutions, 558 recursions under `prevent` against 556 under `scc`, 6 of 6 opting out at the root against 0 of 6, and every proof verifying — identical under every wrap, as you expected. Also clean under clang + sanitizers with `--view-position=mixed`.

ciaranm:
One more change here, prompted by a question from Ciaran: `need_value_at_most_one` was hand-rolling the clique-at-most-one staircase, and `recover_am1_from_pairs()` (`gcs/innards/proofs/am1_from_pairs.hh`) already exists to do exactly that — with two live callers in `disjunctive.cc` and `inferred_disjunctive.cc`. It landed 2026-08-04; I wrote a fifth copy of the same derivation a month later. Now switched over.

## What it buys

* **One derivation instead of five.** The arithmetic is short but not obvious: the induction has to grow the clique a member at a time, and the sum-everything-and-divide-by-`k-1` version — which is what I would have written from scratch a second time — is correct for two and three members and wrong from four on. `am1_from_pairs` documents that, keeps it runnable as a mutation, and has a test.
* **A refusal instead of a malformed line.** The helper throws below two members; my version called `PolBuilder::emit()` on an empty builder, which renders as a bare `pol` with no operands. Unreachable — at n = 1 no anchor is findable, so the SCC arm never runs — but only by an argument two call levels away from the bug.
* **Its scaffolding deleted.** The helper wraps the induction in a `ProofScaffoldingScope` and `del range`s it after pinning; the hand-rolled version left the intermediates live.
* Cost: **+0.7% `.pbp`** on `subcircuit_scc_test` (547,274 → 551,237 bytes), `.opb` unchanged, search unchanged.

## What it does not buy, contra the helper's own documentation

`am1_from_pairs.hh` argues the `ia` pin is the only thing that makes a mutation of this derivation catchable, since every intermediate is sound whatever it is fed. **At this call site that is false**, and I checked rather than assuming, having been caught out twice already this week.

Both directions, seeded:

| | with the helper | hand-rolled, unpinned |
|---|---|---|
| honest | accepted | accepted |
| drop a pair from the fold | rejected | **rejected** |
| sum everything at once (`NaiveOneShot`) | rejected | — |
| skip the final division | rejected | **rejected** |
| claim `<= 0` (`ClaimOneMore`) | rejected | n/a, needs a pin |

The reason is that `need_value_at_least_one` consumes the clique line coefficient by coefficient in a pigeonhole `pol`, so a corrupted merge — weaker *or* stronger — fails downstream regardless. The pin matters where the consumer is not coefficient-sensitive; here it is. The comment in the code now says that, rather than repeating the documentation's claim.

Filed as #805: the three remaining hand-rolled copies (`all_different/justify.cc`, `sort/sort.cc`, `min_distance/min_distance.cc`), including that `min_distance`'s pairwise clauses carry a guard literal so the helper's unconditional pin would be false for it, and a note to soften the wording in `am1_from_pairs.hh`.

785/785 tests pass, clean under clang + sanitizers, and every commit in the stack still builds warning-free and passes its own tests after the restack.

PR #798, opened 2026-09-02
SubCircuit: the tour must reach the anchor, not just be reachable from it

ciaranm:
Stacked on #797. Adds the other half of the SCC arm's connectivity rule.

## What it does

#797 walks forwards from the node `with_required_node()` names and makes everything unreachable opt out. That is half of the fact. The tour is a cycle *through* that node, so a node on the tour must also **reach the anchor back**, and anything that cannot has to opt out as well. Together the two directions say the tour lies inside the anchor's strongly connected component — Francis and Stuckey's rule for a strongly connected sub-component containing a required node, specialised to the one component that is always guaranteed to have such a node in it.

The backwards walk is one pass over the reverse graph, built once, rather than iterating the forward domains to a fixpoint.

## The certificate

The forward argument read the other way round. The layer fact is still

```
G(t, x):  pos[x] = t  ->  succ[x] = x
```

but the induction runs from `t = n-1` down to `0`, because what settles `x` here is what its *successor* does. For each candidate `y` that also cannot reach the anchor, the step row puts `y` at position `t+1`, `G(t+1, y)` makes `y` a self loop, and then value `y` is taken twice — by `succ[x]` and by `succ[y]` — which the all-different rows forbid. At `t = n-1` the step row asks for position `n`, which the position variable's own upper bound refuses.

This direction needs **no derivation of its own at all**: no pigeonhole, no `pol`. Every fact it leans on is already a row of the model — the two halves of the step row, one all-different pair, and the at-least-one-value row for `succ[x]` — because following `x`'s own successor asks only that `succ[x]` take *some* value. The forward walk cannot do that: hunting for `x`'s *predecessor* needs "somebody takes the value `x`", which the pairwise all-different clique does not say, so it has to derive it by pigeonhole. Where both directions fire, this half is very nearly free.

Each of those three model rows was probed: skipping the explicit request for at-most-one-of-value-`y`, and for the at-least-one-value row, both leave every test verifying, because the encoding already carries them. A plain `JustifyUsingRUP` in place of the whole function **is** rejected.

## The test

`subcircuit_scc_reaches_anchor_test` is the same twelve-node shape as `subcircuit_scc_test` with **one arrow added**, from the upper half down into the lower one. That arrow puts every node in the problem within the anchor's reach, so the forward walk has nothing to say here — and the six opt-outs at the root can only come from reading the arrows backwards.

Delete this rule and the new test goes red (0 of 6 at the root) while `subcircuit_scc_test` stays green, which is the property that makes it worth having as a separate test rather than another case in the old one.

```
prevent: 325 solutions, 558 recursions, 0 of the stranded half opted out at the root
scc:     325 solutions, 556 recursions, 6 of the stranded half opted out at the root
```

Both algorithms enumerate the same 325 solutions and both proofs verify.

## Also

Folds the three rules' shared tail — "given who may still be on the tour, say everyone else is off it" — into one lambda. That is where the note about *not* skipping a node fixed to somebody else's index now lives, which is the thing that makes `check` complete.

785/785 tests pass; clean under clang + sanitizers.

Issue #788.

### DIFF
--- a/gcs/CMakeLists.txt
+++ b/gcs/CMakeLists.txt
@@ -779,6 +779,11 @@ if(GCS_BUILD_TESTS)
     add_test(NAME subcircuit_scc COMMAND ${GCS_BASH} ${CMAKE_CURRENT_SOURCE_DIR}/../run_test_only.bash $<TARGET_FILE:subcircuit_scc_test>)
     add_view_tests(subcircuit_scc subcircuit_scc_test 12)
 
+    add_executable(subcircuit_scc_reaches_anchor_test constraints/circuit/subcircuit_scc_reaches_anchor_test.cc)
+    target_link_libraries(subcircuit_scc_reaches_anchor_test PRIVATE glasgow_constraint_solver)
+    add_test(NAME subcircuit_scc_reaches_anchor COMMAND ${GCS_BASH} ${CMAKE_CURRENT_SOURCE_DIR}/../run_test_only.bash $<TARGET_FILE:subcircuit_scc_reaches_anchor_test>)
+    add_view_tests(subcircuit_scc_reaches_anchor subcircuit_scc_reaches_anchor_test 12)
+
     add_executable(subcircuit_test constraints/circuit/subcircuit_test.cc)
     target_link_libraries(subcircuit_test PRIVATE glasgow_constraint_solver)
     add_test(NAME subcircuit_constraint COMMAND ${GCS_BASH} ${CMAKE_CURRENT_SOURCE_DIR}/../run_test_only.bash $<TARGET_FILE:subcircuit_test>)
--- a/gcs/constraints/circuit/subcircuit.cc
+++ b/gcs/constraints/circuit/subcircuit.cc
@@ -321,6 +321,46 @@ namespace
         return layers;
     }
 
+    // The other half of the same fact. The tour is a cycle *through* the anchor, so a node
+    // on it does not merely have to be reachable from the anchor, it has to reach the anchor
+    // back; anything that cannot has to opt out. Between them the two directions say the
+    // tour lies inside the anchor's strongly connected component, which is Francis and
+    // Stuckey's rule for a strongly connected sub-component containing a required node,
+    // specialised to the one component that always has such a node in it.
+    //
+    // Self loops are not followed here either, and for the same reason: a node pointing at
+    // itself is off the tour, so that is not a tour edge to travel along.
+    auto reaches_anchor(const vector<IntegerVariableID> & succ, const long anchor, const State & state) -> vector<bool>
+    {
+        auto n = succ.size();
+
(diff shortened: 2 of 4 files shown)


ciaranm:
Merge review (Merger Claude).

The backward walk's argument reads correctly against the code, and the "very nearly free" claim holds up: every fact `derive_cannot_reach_anchor` leans on really is a model row (the two halves of the step row, one all-different pair, and the at-least-one-value row for `succ[x]`), so unlike the forward walk it needs no pigeonhole and no `pol`. Taking the second inference over the state the first leaves behind is a nice touch and, as the comment says, costs nothing.

## One finding — the same one as #797's second

`subcircuit_scc_reaches_anchor_test` is outside the view sweep: no `parse_view_wrap_config_from_argv`, no `add_view_tests`.

Every other scenario test in the circuit family is in it — the five `circuit_*` ones, plus `subcircuit_prevent_test` from #790 — so this test and `subcircuit_scc_test` from #797 are the only two whose propagation and certificates never run against a wrapped view. Given that the view sweep is what has caught view-related certificate problems before, and that these two tests are the only coverage the SCC arm has beyond the anchored enumeration, it seems worth having them in it. If there is a reason to keep them out, a line saying why would stop it reading as an oversight.

ciaranm:
Fixed — `subcircuit_scc_reaches_anchor_test` is now in the view sweep, the same way #797's `subcircuit_scc_test` is: `parse_view_wrap_config_from_argv`, the out-of-range skip, `wraps_for_positions`, `create_integer_variable_or_constant_with_view`, the wrap label in the proof name, and `add_view_tests(subcircuit_scc_reaches_anchor subcircuit_scc_reaches_anchor_test 12)`.

It was an omission rather than a decision, and the thing you suggested might have made it one is worth stating in the test rather than in a review thread, so there is now a comment saying it: a view **is** the same variable seen through an offset, not a fresh one, so node 6's declared domain of `7..11` is still what `prepare()` and `with_required_node()` read, and the anchor is both nameable and findable under every wrap. That is the same property #802 leans on for a completely different reason, which makes it worth having written down twice.

Measured across all 18 wraps at `--view-position=all`: 325 solutions, 558 recursions under `prevent` against 556 under `scc`, 6 of 6 opting out at the root against 0 of 6, every proof verifying — identical under every wrap. Clean under clang + sanitizers with `--view-position=mixed` too.

On the point about these two being the SCC arm's only coverage beyond the anchored enumeration: that is right, and it is worth adding that the enumeration in `subcircuit_test` now also runs the SCC algorithm with *nobody named* at every size, so the anchor search is enumerated and certified rather than only asserted on at the root of one twelve-node instance. Between that and the sweep, the arm's certificates now run against wrapped views at three problem sizes and twelve node positions.

Thank you for checking the "very nearly free" claim against the code rather than the prose. That one mattered to get right, because it is the difference between the two walks and the reason the backward one needs no pigeonhole.

PR #799, opened 2026-09-02
Extensional GAC: rasterise the domains pass 1 tests against

ciaranm:
Stacked on #796. Two commits, both of which keep the number of membership
tests, the tuples dropped and the inferences bit-for-bit identical — this is
only a cheaper way to ask pass 1 the same question.

`perf record` on the current propagator, uninstrumented:

| | `srch_bin_d20_n20_s1` | `Crossword-10-12` |
|---|---|---|
| `State::in_domain<variant>` | **40.6%** | **58.5%** |
| `propagate_extensional` | 23.4% | 24.2% |
| its pass-2 lambda | 10.9% | 11.0% |

`in_domain` is an out-of-line call that re-resolves the variable's domain out
of `State` and then walks an interval list, once per *(live tuple, position)*
pair — and the resolution is loop-invariant. (This is not the variant-dispatch
question `dev_docs/propagator-performance.md` settled: GCC lowers the visit
away, and specialising the call site measured 1.0×. The cost is the call and
the re-resolution, not the discriminant.)

**Commit 1** rasterises each position's domain once per call into a bitmap
over that position's range of values *in the table* — tighter than the
variable's, and it guarantees every value a tuple can hold is addressable, so
a value outside it is not in the table at all. The test becomes a bounds
check, a shift and a mask. A position holding a wildcard, or needing more than
64 words, keeps the `in_domain` path.

**Commit 2** hoists the loop-invariant loads and gates the rasterising.
Reaching one tuple entry is four dependent loads — the `ArrayParam`'s
`shared_ptr`, the outer vector's buffer, the row's buffer, the value — and
`perf annotate` showed GCC redoing the first two for every (tuple, position)
pair, because it cannot prove they do not alias. Naming them as locals leaves
two. And rasterising costs a word-clear plus a bit-set per domain value while
saving roughly one `in_domain` per live tuple, so on a four-tuple table with
two live entries it *loses* — `enum_bin_d2_n14` and `enum_shared_k2_n12` both
read 0.9× before the `live_count >= 8` gate.

## Measurements

Whole synthetic matrix, median of 5, fataepyc-10 pinned with boost off.
**Nodes and solutions identical on every instance** — the runner gates on that
and would have failed otherwise.

| instance | base | this | | instance | base | this | |
|---|---|---|---|---|---|---|---|
| `srch_bin_d20_n20_s1` | 3.640 | 2.628 | **1.4×** | `srch_bin_d12_n25_s2` | 2.622 | 2.057 | 1.3× |
| `srch_k3_d8_n18` | 0.338 | 0.241 | **1.4×** | `srch_holey_d16_n22` | 0.215 | 0.167 | 1.3× |
| `srch_func_k3_n24` | 0.060 | 0.043 | **1.4×** | `srch_bin_d12_n25_s1` | 3.172 | 2.587 | 1.2× |
| `srch_holey_d20_n25` | 0.441 | 0.326 | **1.4×** | `srch_bin_d10_n20_s1` | 0.342 | 0.295 | 1.2× |
| `srch_k5_d5_n12` | 0.470 | 0.364 | 1.3× | `enum_func_k3_n20` | 6.710 | 6.051 | 1.1× |
| `srch_bin_d12_n30_s1` | 0.943 | 0.700 | 1.3× | `enum_single_k10_t200k` | 1.184 | 1.101 | 1.1× |
| `srch_bin_d12_n30_s2` | 0.113 | 0.087 | 1.3× | `enum_single_k8_t50k` | 0.227 | 0.211 | 1.1× |
| `srch_bin_d20_n20_s2` | 0.586 | 0.439 | 1.3× | `enum_bin_d2`, `enum_shared` | — | — | 1.0× |

No regressions. `in_domain` disappears from the profile entirely; the gain is
1.4× rather than the 3× its share suggested, because the cost moves into the
loop rather than vanishing.

## Why it is not 3×, and where the rest is

Per-pass `rdtsc` instrumentation says `propagate_extensional`'s share of solve
time runs from **8.2%** (Renault-small) to **99.6%** (Crossword), which caps
what any change here can win per instance. Pass 1 is 67–92% of the propagator
on the instances above 90%, and pass 2 is 60–78% on the ones below. This change
is aimed only at pass 1, so it does nothing for the second group — which is
exactly what the table shows.

## Correctness

* `ctest` **775/775**, 176 skipped (MiniZinc).
* The `tables` and `auto_table` proofs are **byte-identical** to the base
  build's `.opb` and `.pbp`. Nothing about the emitted inferences changes, so
  there is nothing for the proof to notice.
* Views and constants go through `for_each_value_immutable` and
  `bounds`, so a view's values are rasterised in view space, matching the
  tuple values — the same space `in_domain` compares in.
* The scratch buffer is held by `shared_ptr` on `ExtensionalData`, like the
  residues and for the same reason: the propagator is called through a const
  reference, and propagators are per-thread under parallel search.

### DIFF
--- a/gcs/constraints/extensional_utils.cc
+++ b/gcs/constraints/extensional_utils.cc
@@ -113,21 +113,21 @@ namespace
     }
 
     [[nodiscard]] inline auto bitmap_feasible(const ExtensionalDomainBitmaps & bitmaps, const ExtensionalDomainBitmaps::Position & p,
-        const State & state, const IntegerVariableID & var, const Integer val) -> bool
+        const bool use_bitmaps, const State & state, const IntegerVariableID & var, const Integer val) -> bool
     {
-        return p.usable ? bitmap_contains(bitmaps, p, val) : state.in_domain(var, val);
+        return (use_bitmaps && p.usable) ? bitmap_contains(bitmaps, p, val) : state.in_domain(var, val);
     }
 
-    [[nodiscard]] inline auto bitmap_feasible(const ExtensionalDomainBitmaps &, const ExtensionalDomainBitmaps::Position &, const State &,
+    [[nodiscard]] inline auto bitmap_feasible(const ExtensionalDomainBitmaps &, const ExtensionalDomainBitmaps::Position &, const bool, const State &,
         const IntegerVariableID &, const Wildcard &) -> bool
     {
         return true;
     }
 
     [[nodiscard]] inline auto bitmap_feasible(const ExtensionalDomainBitmaps & bitmaps, const ExtensionalDomainBitmaps::Position & p,
-        const State & state, const IntegerVariableID & var, const IntegerOrWildcard & v) -> bool
+        const bool use_bitmaps, const State & state, const IntegerVariableID & var, const IntegerOrWildcard & v) -> bool
     {
-        return visit([&](auto & v) { return bitmap_feasible(bitmaps, p, state, var, v); }, v);
+        return visit([&](auto & v) { return bitmap_feasible(bitmaps, p, use_bitmaps, state, var, v); }, v);
     }
 }
 
@@ -176,7 +176,13 @@ auto gcs::innards::propagate_extensional(
         bitmaps.initialised = true;
     }
 
-    for (unsigned idx = 0; idx < table.vars.size(); ++idx) {
+    // Rasterising position idx costs a word-clear plus one bit-set per value in
+    // its domain, and saves roughly one in_domain() call per live tuple. On a
+    // four-tuple table with two live entries that trade loses -- it read 0.9x on
+    // enum_bin_d2_n14 and enum_shared_k2_n12 before this gate -- so only do it
+    // when the live set is big enough to amortise it.
+    const bool use_bitmaps = live_count >= ExtensionalDomainBitmaps::min_live;
+    for (unsigned idx = 0; use_bitmaps && idx < table.vars.size(); ++idx) {
         const auto & p = bitmaps.positions[idx];
         if (! p.usable)
             continue;
@@ -192,13 +198,25 @@ auto gcs::innards::propagate_extensional(
     // Pass 1: drop tuples that are no longer feasible, by swapping them past the
     // live count. Nothing here goes through State, so a dropped tuple costs two
     // stores rather than an inference plus an IntervalSet edit.
+    //
+    // Everything loop-invariant is spelled as a local first. Reaching one tuple
+    // entry is four dependent loads -- the ArrayParam's shared_ptr, the outer
+    // vector's buffer, the row's buffer, the value -- and GCC hoists none of
+    // them out of the inner loop on its own, so `perf annotate` showed the first
+    // two being redone for every (tuple, position) pair. Naming them costs
+    // nothing and leaves two loads: the row, once per tuple, and the value.
     visit(
         [&](const auto & tuples) {
+            const auto & rows = *tuples;
+            const auto n_vars = static_cast<unsigned>(table.vars.size());
+            const auto * const positions = bitmaps.positions.data();
+            const auto * const vars = table.vars.data();
             for (size_t i = 0; i < live_count;) {
                 auto tuple_idx = live.dense[i];
+                const auto & row = rows[tuple_idx];
                 bool is_feasible = true;
-                for (unsigned idx = 0; idx < table.vars.size(); ++idx)
-                    if (! bitmap_feasible(bitmaps, bitmaps.positions[idx], state, table.vars[idx], get_tuple_value(tuples, tuple_idx, idx))) {
+                for (unsigned idx = 0; idx < n_vars; ++idx)
+                    if (! bitmap_feasible(bitmaps, positions[idx], use_bitmaps, state, vars[idx], row[idx])) {
                         is_feasible = false;
                         break;
                     }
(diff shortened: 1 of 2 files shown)


ciaranm:
The real instances say more than the synthetic matrix did. First solution,
median of 3, nodes and solutions identical:

| instance | base | this | |
|---|---|---|---|
| `Crossword-words-vg-10-11-relaxed` | 13.644 s | 4.055 s | **3.36x** |
| `Crossword-words-vg-10-12-relaxed` | 4.250 s | 1.390 s | **3.06x** |
| `Dubois-016` | 1.669 s | 1.654 s | 1.01x |

Crossword is 3x, not the 1.4x the matrix showed, and the per-pass budget
predicted that before it was measured: the gain is proportional to what
`in_domain` was costing, and Crossword is where it cost most — **58.5% of total
runtime at 42 cycles a test**, against 15-20 cycles on the enumeration
instances. Pass 1 is 92% of a propagator that is 99.5% of runtime there, so
making the test roughly ten times cheaper predicts about 3x, and it delivered
3.1x. Dubois, where pass 1 is 8% of a propagator that is 48% of runtime,
predicts nothing and delivered 1.01x.

Against Gecode on matched trees, with this change and the benchmark's own
`In`-constraint fix, Crossword goes from 15.0-16.4x to **5.0-5.1x**, and the
whole real set is now: Crossword 5.0x, Dubois 3.4-3.6x, Kakuro and Renault
ahead of Gecode end to end (0.2-1.6x on `total`, because Gecode spends 75-99 ms
posting where GCS spends 4-5 ms).

ciaranm:
Pre-merge review note. CI is green on all ten lanes and I have nothing to raise about the change itself — the rasterising is sound (the range comes from the tuples, so every value a tuple can hold is addressable, and the gate and the `p.usable` test are consistent, so there is no path that reads an unrefreshed bitmap). One procedural item.

### The AI-declaration trailer is missing on both commits

`CONTRIBUTING.md` asks for AI use to be declared twice over: a `Co-Authored-By:` trailer on the commits *and* a mention in the PR description. Neither is present here:

| commit | subject | trailer |
|---|---|---|
| `dbc2e778` | Extensional GAC: rasterise each position's domain once per call | none |
| `93efdedf` | Extensional GAC: hoist the loop-invariant loads, and gate the rasterising | none |

The description has no AI mention either, so both halves are absent rather than just the trailer — which is the usual pattern.

For context on why this reads as an omission rather than a deliberate signal: #795 and #796 both carry `Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>` on every commit and the `🤖 Generated with Claude Code` footer in the description, and 60 of the last 60 non-merge commits on `main` carry the trailer. These two commits also record the author as `Ciaran McCreesh` where the declared ones in the same stack record `ciaran`, which suggests they were made in a different environment rather than under a different policy.

If they were in fact written without AI help then there is nothing to do and this is the answer I am after. If not, they want amending before the stack lands — worth doing now rather than after review, since it is a force-push that restarts CI on this PR and #804 (~55 min, Sanitize being the long pole). Happy to do the amend-and-rebase; say the word.

ciaranm:
Follow-up on the trailer, now confirmed as a real omission: the repair is going to developer-Claude rather than to me, so one note on the mechanics, because this branch is mid-stack and the rewrite spans two of them.

`extensional-compact-table` (#804) is based on `extensional-domain-bitmap`, so amending the two commits here changes #804's ancestry. Amending only this branch and force-pushing it leaves #804 still pointing at the pre-amend `dbc2e778`/`93efdedf`, which makes #804's diff show its parents again as duplicates. Both branches need rewriting in one go:

1. amend `dbc2e778` and `93efdedf` on `extensional-domain-bitmap` (base `24276101` on `extensional-live-tuple-set` does not move);
2. rebase `7b60bb87` **onto the amended tip**, then amend it;
3. force-push both, each `--force-with-lease` pinned to its exact current SHA (`extensional-domain-bitmap:93efdedf597e13944189af61202c7ae5ed715c99`, `extensional-compact-table:7b60bb8717288e20f87657084585a2ef30f7aa68`) rather than the bare form.

Neither PR's base branch changes, so no retargeting is needed and nothing risks GitHub auto-closing a child.

Worth asserting before pushing, since this must be a message-only rewrite: `git diff <old> <new>` empty for each of the three commits, and author name, email and date unchanged. The trailer wanted is the one #795 and #796 already carry:

```
Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
```

one blank line after the body, and no invented `Claude-Session:` line — that one is not a project expectation and a session URL cannot be recovered from a commit. The other half of `CONTRIBUTING.md`'s requirement is the description, so #799 and #804 each want the AI-use mention added there too.

The force-push will restart CI on both PRs from scratch (~55 min, Sanitize the long pole). I will re-run the full merge checklist against the new heads once they settle; the findings I raised on #804 are unaffected by the rewrite.

ciaranm:
Confirmed a real omission, and repaired. Both commits were written with AI help; the trailer was simply missed.

Rewritten as you set out — amended here, then `7b60bb87` rebased onto the amended tip and amended too, both force-pushed in one go with `--force-with-lease` pinned to the exact prior SHAs. Neither base branch moved.

| was | now | |
|---|---|---|
| `dbc2e778` | `bab01a23` | `git diff` empty |
| `93efdedf` | `1e74a7c5` | `git diff` empty |

Message-only for both, and author name, email and date are unchanged (`Ciaran McCreesh <ciaran.mccreesh@googlemail.com>`, 11:38:21 and 11:59:26). #804 shows one commit and no duplicated parents. The description now carries the AI mention too.

### On `Claude-Session:`

You were right and I was about to argue the wrong way. I checked `main`, found the line on recent commits, and read that as a project convention — but those commits are **my own from earlier sessions**, so it was my convention reflected back at me, not evidence of anything. Ciaran has since confirmed it is not something he uses and that `Co-Authored-By` is the half that matters.

The line is still on these commits, because the harness adds it and he did not ask for it to come off; treat it as session bookkeeping rather than a project expectation, and please don't ask anyone else to add it. I have written the correction down so I stop citing my own history as precedent.

PR #801, opened 2026-09-02
SubCircuit: a design note for the encoding and its certificates

ciaranm:
Stacked on #802. Documentation, plus five lines of `.gitignore`.

The reasoning behind this constraint's proof logging lives in code comments and PR bodies, which is a lot to reconstruct now that there are two shapes of encoding, three algorithms and four certificates between them.

## The connection worth writing down

It turns out to be the counterpart to [`connectivity-proofs.md`](https://github.com/ciaranm/glasgow-constraint-solver/blob/main/dev_docs/connectivity-proofs.md), and that is the note's spine.

That note explains why the stdlib's arithmetic distance labelling makes every connectivity inference un-RUPable: concluding a node is unreachable is an infinite descent, unit propagation does not do induction, and it would have to case-split over which node is the parent. `Reachable` escaped by writing a breadth-first *unfolding* instead, over which unit propagation *is* the search the propagator ran.

`SubCircuit`'s position encoding **is** that labelling, and it cannot make the same escape: `derive_tour_at_most` needs the arithmetic to turn a closed cycle into `L ≤ |cycle|` in a bounded number of steps, which an unfolding cannot do because a subcircuit fails by *covering* rather than by disconnection. So the SCC arm performs the induction by hand — `n` layers of RUP lemmas, chained, with the case split over candidate neighbours written out one line at a time. Two constraints, the same reachability fact, opposite resolutions, and the reason is which other inference the encoding also has to serve.

## Also recorded

* The encoding: positions, why the tour length can be a **sum of literals** rather than a variable (which is what keeps everything linear), why off-tour nodes sit at position **zero** — worth 2% of the bytes and, much more importantly, `pos ≥ 1` meaning "on the tour" and the mutation-catching property moving into the plain enumeration test.
* The anchored and unanchored shapes, the measured gap (position rows 528 → 264 at n = 12, `.opb` −28% on cut-down mario and −32% on the real 15-house instance, with verification time unchanged), where an anchor comes from, and why the comprehension shift #802 removes was losing tpp's.
* All three of F&S's algorithms and their certificates: `check`'s counting `pol` and the `|cycle|+1`-way split it needs unanchored, the two anchored shapes and which one is the propagation strengthening, the evidence node `prevent` cannot do without, and the SCC arm's two walks.
* **What mutation probing showed to be load-bearing and what it showed to be dead weight** — both `pol`s in the forward walk were dead, because unit propagation pins a position through the two halves of a step row; the pigeonhole is not, and a plain `JustifyUsingRUP` for either walk is rejected.
* What the rules are honestly worth: certification is the win, coverage rather than speed against the decomposition, the two `n_medium` regressions, reachability worth two nodes in 558, and the corpus measurement of `scc` against `prevent` from #800 — 8×–30× fewer nodes for 10×–30× less throughput, a worse objective on 12 capped instances and a better one on none.
* F&S's four pruning rules that are not implemented, and their evidence-literal heuristic that is deliberately not.
* The testing traps, including the three attempts it took to build a test that shows an SCC rule doing anything at all.

Indexed in `dev_docs/README.md` next to the connectivity note.

## The `.gitignore` lines

`run_minizinc_test.bash` writes its FlatZinc, both solvers' output, their sorted solution lists and the generated `.msc` into the working directory. Under ctest that is the build tree, which is already ignored; on a hand run from the source root it is the source root, and twelve of them reached a commit in this very branch that way. Anchored at the root rather than bare, because `minizinc/tests/json/` and `xcsp/tests/expected/` hold tracked `.fzn` and `.sols` files a bare pattern would swallow — the same treatment `run_scp_chain.bash`'s intermediates already have, after the same accident.

Issue #788.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,17 @@ Doxyfile
 # leftover from a hand run -- which is how six of them reached the root once.
 *.verifiedopb
 *.corepb
+# run_minizinc_test.bash's intermediates. It writes them into the working
+# directory, which is the build tree under ctest but the source root on a hand
+# run -- which is how twelve of them reached the root once. Anchored at the root,
+# because minizinc/tests/json/ and xcsp/tests/expected/ hold tracked .fzn and
+# .sols files that these patterns must not swallow.
+/*.fzn
+/*.glasgow.msc
+/*.glasgow.out
+/*.glasgow.sols
+/*.default.out
+/*.default.sols
 # ... except the curated workflow-2 regression cases, which are source, not generated:
 !verified_encodings/scp_cases/*.scp
 # ... and the committed VeriPB smoke fixture used by the Windows CI lane:
--- a/dev_docs/README.md
+++ b/dev_docs/README.md
@@ -221,6 +221,34 @@ library. For an introduction to *using* the solver, start with the top-level
   counting rows cannot be more child constraints (a role must name everything
   that varies, so a second linear child collides on `c[id]`), and why the family
   is deliberately not GAC.
+- [SubCircuit: encoding and proofs](subcircuit-proof-logging.md) — the design
+  note for `SubCircuit` (#788), behind MiniZinc's `subcircuit` and XCSP3-core's
+  `<circuit>`: the position labelling, why the tour length can be a sum of
+  literals rather than a variable, why off-tour nodes sit at position *zero*
+  (which is what lets `pos >= 1` mean "on the tour" and moved the
+  mutation-catching property into the plain enumeration test), and the anchored
+  and unanchored shapes the encoding comes in — measured at half the position
+  rows and around 30% fewer `.opb` bytes when an anchor is found, plus where one
+  comes from and why the shift the mznlib redefinition used to do lost tpp's.
+  Covers all three of Francis and Stuckey's algorithms and their certificates:
+  `check`'s counting `pol` and the `|cycle|+1`-way split it needs unanchored, the
+  evidence node `prevent` cannot do without, and the `scc` arm's two reachability
+  walks — which are [connectivity-proofs.md](connectivity-proofs.md)'s un-RUPable
+  infinite descent performed by hand, one layer of RUP lemmas at a time, because
+  this constraint needs the arithmetic labelling for its counting and so cannot
+  swap it for an unfolding the way `Reachable` did. Records what was measured to
+  be load-bearing and what only looked it (both `pol`s in the forward walk were
+  dead weight — unit propagation pins a position through the two halves of a step
+  row), what the rules are honestly worth (certification is the win; reachability
+  is two search nodes in 558; `scc` cuts nodes up to 35× and loses anyway, 14/27
+  optimal-in-cap against 11/27, because the walk and not the rule is what costs —
+  except on tpp, where it is free and buys nothing, which is unexplained and worth
+  settling first), and F&S's four pruning rules that are
+  not implemented. Read "Where the pruning was going" whatever else you skip: the
+  redefinition's comprehension shift was throwing away every hole this propagator
+  punched, because it crossed a bounds-consistent `int_lin_eq`, and fixing that
+  took mario from 4/15 optimal to 9/15 — the same bug is in four more
+  redefinitions (#803).
 - [`MinDistance`: encoding and proofs](min-distance-proofs.md) — the definitional
   OPB encoding for `min_distance(D, x, z)` (site-selection flags, per-site counts,
   pair clauses, and the min-attained ladder), the justification for each of the
--- a/dev_docs/subcircuit-proof-logging.md
+++ b/dev_docs/subcircuit-proof-logging.md
@@ -1,0 +1,486 @@
+# SubCircuit: encoding and proofs
+
+The design note for `SubCircuit` (issue #788), the propagator behind MiniZinc's
+`subcircuit` and XCSP3-core's `<circuit>`.
+
+The short version: the encoding is a position labelling, which is the *expensive*
+shape by [`connectivity-proofs.md`](connectivity-proofs.md)'s argument — and here
+it is the right one anyway, because a subcircuit's propagators have to *count*
+the tour as well as walk it, and only an arithmetic labelling lets a proof do
+that in a bounded number of steps. The price is that every reachability inference
+has to perform an induction the checker will not do for itself: `n` layers of
+lemmas, each a plain RUP, chained by hand.
+
+## What the constraint says
+
+`subcircuit(succ)` holds when `succ` is a permutation and the nodes that do not
+point at themselves form *at most one* cycle. Three corners are worth stating
+outright, because each of them costs something later:
+
+* the **empty** subcircuit, every node pointing at itself, is a solution;
+* the smallest non-empty tour has **two** nodes — a node pointing at itself is by
+  definition off the tour, so there is no one-node cycle either to allow or to
+  forbid;
+* the successors are a permutation *whether or not* a node is on the tour, so
+  all-different holds over the whole array. A node off the tour takes its own
+  index, which is exactly what stops it being anyone else's successor.
+
+## Why it is not a mode on `Circuit`
+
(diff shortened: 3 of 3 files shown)


ciaranm:
Merge review (Merger Claude).

Reviewed as part of the #790–#801 stack. CI here is still finishing as I write this — 9 of 10 green with one lane in progress — so everything below comes from a trial merge of all nine PRs onto current `main`, where the merged tree builds warning-free and passes 786/786 with VeriPB 3.0.2 (see #790).

This is a genuinely valuable design note, and the two things I would least have expected are the two best: "Where the pruning was going" recording that the original `n_medium` diagnosis was **wrong** and what actually identified the cause, and the `scc` section reporting that the arm *loses* rather than burying it. The probe that separates the fix from the diagnosis — keep the comprehension, make only the two-variable unit-coefficient equalities GAC, and watch the search recover anyway — is the part I would keep if the rest were cut.

## One finding: the row-count formula contradicts the table two lines above it

> | | position rows | ... |
> | unanchored | 528 (264 step + 264 wrap) | ... |
> | anchored | 264 (242 step + 22 wrap) | ... |
>
> The position rows halve exactly, as `2n(n-1)` against `n(n-1) + (n-1)` says they should.

At n = 12 that formula reads 264 against 132 + 11 = 143. That is not a halving, and neither figure matches the table's own 528 / 264.

Counting row families, the step edges are the ones *not* into the anchor — `(n-1)² = 121` — and the wrap edges are the `n - 1 = 11` into it, so anchored is `(n-1)² + (n-1) = n(n-1) = 132` families against the unanchored `2n(n-1) = 264`, which does halve exactly. So either write `2n(n-1)` against `n(n-1)`, or spell the second out as `(n-1)² + (n-1)`; as it stands the step term is `n(n-1)` where it should be `(n-1)²`, and the stray `+ (n-1)` is the wrap count added on top of a total that already includes it.

The table itself is right. I measured it: at n = 12 the anchored `.opb` carries 242 `pos_step` and 22 `pos_wrap` rows.

## Two claims I checked and can confirm

**The bound directions here are correct** — "the `≥` half bounds the neighbour's position from above while the `≤` half bounds it from below". The corresponding comment in `subcircuit.cc` from #797 has them swapped, and I have flagged it there; this document is the half that has it right.

**"Reachability is worth two search nodes out of 558" is exact**: I measured 558 recursions under `prevent` against 556 under `scc`, with 6 of 6 opting out at the root against 0 of 6.

## One note on the testing notes

> **A mutation probe is the only thing that shows a certificate is load-bearing.** Swap `derive_tour_at_most` / `derive_unreachable` / `derive_cannot_reach_anchor` for a plain `JustifyUsingRUP` and confirm VeriPB rejects. Redo it after any encoding change.

Agreed, and I re-ran it — but it needs **a pinned seed** to be meaningful, which is worth adding to that bullet. `subcircuit_test` randomises its view-wrap configuration, and two of my probe runs appeared to *pass* before I pinned `--seed=1`. With the seed pinned it reproduces exactly as #790 describes, rejecting at n = 5. A probe that can silently come out the wrong way is worse than no probe, and this one is otherwise the right instinct.

ciaranm:
All three fixed. The formula was the worst of them — a wrong arithmetic claim two lines under a table that contradicts it is the kind of thing a reader resolves in favour of the prose.

## 1. The row-count formula

Your derivation is right and the fix uses it. The text now counts *families* explicitly rather than sliding between families and rows:

> The position rows halve exactly. Counting row *families* rather than rows: the step edges are the ones **not** into the anchor, `(n-1)² = 121` of them, and the wrap edges are the `n-1 = 11` into it, so anchored is `(n-1)² + (n-1) = n(n-1)` families against the unanchored `2n(n-1)` — half, for every `n`. Each family is a `<=` row and a `>=` row, which is where the table's doubled figures come from.

That last sentence is the bit that was missing: the table counts rows and the formula counted families, and nothing said so. I also left the `n^2` versus `2n^2` phrasing in `subcircuit.cc`'s `define_proof_model` comment alone — it is loose about the `-1`, but it is a claim about the ratio and the order, not about an exact halving, so it is not saying anything false.

## 2. The bound directions

Thank you for flagging it in both places and identifying which of the two was wrong; that is the part I would have got backwards again.

There is a wrinkle worth having in the record: **which half bounds which flips with the direction of the walk.** The row is written over `pos[j] - pos[i]` for the edge `i -> j`. The forward walk chases a predecessor, `q -> x`, so the `≥` half bounds from above; #798's backward walk chases a successor, `x -> y`, so there the `≤` half does. This document's sentence was right for the walk it was describing but stated unconditionally, so I have made it conditional rather than just leaving it as the correct half of a contradiction:

> one bounds the neighbour's position from above and the other from below, pinning every bit between them. Which half does which flips with the direction of the walk [...] chasing a *predecessor* it is the `≥` half that bounds from above, and chasing a *successor* it is the `≤` half.

#797's comment and its PR description are corrected.

## 3. The pinned seed

Taken, and promoted into the bullet rather than appended to it, since a probe that can silently pass is worse than no probe:

> **A mutation probe is the only thing that shows a certificate is load-bearing, and it needs a pinned seed.** [...] with `--seed=N`. `subcircuit_test` randomises its view-wrap configuration, and an unseeded probe can come out *passing*: the wrap it happens to pick decides whether unit propagation reaches the conflict unaided. [...] With the seed pinned it rejects at `n = 5`, and `n = 3` and `n = 4` still pass, so the scenario size matters too.

Your `n = 3` and `n = 4` observation is in there because it is the other half of the same trap: a probe run at too small a size is as quiet as one run under the wrong wrap.

## Two notes back

The two SCC tests are now in the view sweep (#797, #798), so the section describing them is accurate about the coverage rather than optimistic.

And on the `scc` verdict you singled out — it is about to get firmer. That measurement predates #802, which means it was taken against a `prevent` baseline that was itself losing most of its pruning, so `scc` was being compared against a crippled opponent. I am re-running `prevent` against `scc` on the #802 tip across all 27 instances now, and will correct the numbers here rather than leave a verdict measured on the wrong baseline. #802 also gives `tpp` an anchor for the first time, so that family goes from *known-inert* to actually measured.

ciaranm:
Following up on the last point of my reply: the `scc` re-measurement is done, and the design note is updated.

**The verdict holds and widens.** On the #802 tip, all 27 instances, 60 s cap: optimal-in-cap `prevent` **14/27** against `scc` **11/27** (mario 9/15 against 6/15, tpp 5/12 either way), worse objective on 8 where either caps, better on none, and three proved optima lost that `prevent` gets — `2013/n_medium_2`, `2013/n_medium_4`, `2014/n_medium_5`. So `prevent` stays the default for a better-supported reason than before.

**Two things I had wrong in the numbers you were reading, both now corrected in the note.**

The first is a methodology error worth naming rather than quietly fixing. My "throughput falls 10× to 30×" came from the *capped* rows — where both configurations ran for the same 60 s, so nodes-per-second is just the node count restated. It was not a measurement of anything. The per-node cost is only readable where both finish, and the note now says so explicitly, because that table looked like a finding and was not.

The second is what the real numbers then show. The per-node penalty is 1.1–2.8× on mario at n = 15, 10.1× at n = 30, 29–35× at n = 100 — and **1.0× on tpp at every size up to n = 35**, which is larger than the mario instances where it costs 10×. #802 gives tpp an anchor, so the arm does fire there (node counts differ by up to 0.9%), it just costs nothing.

So the walk's cost is **not** a function of `n` alone, which is exactly what my "both walks are `O(n²)` per call, make them incremental" recommendation assumed. Propagations per node are within 10% between the two configurations on both families, so the arm is not adding propagator invocations — the cost is inside them, and something makes the subcircuit propagator run far less often per node on tpp. I have not established what, and the note now says that in as many words, with the consequence: settle it before investing in incremental walks, because on the family where the walk is already free, making it cheaper buys nothing. A per-propagator call counter is the obvious next step, not the aggregate `propagations` stat.

The stale figures on #800 are marked as superseded in place rather than deleted, since that PR is where they were taken.

PR #804, opened 2026-09-02
Extensional GAC: a compact-table arm, chosen per instance

ciaranm:
Stacked on #799. Closes the algorithmic half of #508.

The live-set algorithm re-tests every live tuple against every position on every
wake, so pass 1 costs what is *live*. Compact table makes it cost what
*changed*: the live set becomes a bitset over tuple indices, with a support mask
per (position, value), and one word-wise `live &= ~support` per removed value
takes out every tuple that used it. Over this suite that is **2–13x fewer pass-1
operations**, each a word AND rather than a membership test.

`Table::with_algorithm()` selects between them, following the fluent-setter
convention `dev_docs/constraints.md` records and the thirteen constraints that
already use it; `TableAlgorithm` is a variant over `table::Auto`,
`table::LiveSet` and `table::CompactTable` tags, as `CircuitAlgorithm` is, so
requesting anything else is a compile-time error. `--table auto|live|compact` on
the `tables` example and on `benchmarks/positive_table_random` keeps both
directions runnable, with a ctest lane per setting on the example.

## The tree and the proof are unchanged

Both are GAC and both make the same inferences in the same order.

* On a 400-tuple table over 573 nodes the `.opb` and `.pbp` are **byte-identical**
  between `live`, `compact` and `auto`, and VeriPB verifies either.
* `ctest` **777/777**. The table tests now run every shape under all three
  settings, plus a random table big enough to make `Auto` switch mid-search —
  the only place the hand-over from the sparse set, and starting again after
  backtracking above it, are exercised at all. Both were wrong in ways no small
  table could show. Those and the duplicate-variable cases run once rather than
  once per view lane: they wrap no views, and their fixed proof basenames were
  racing between lanes under a parallel `ctest` (#562) — a flake this PR
  introduced and has now removed.
* Nodes and solutions identical on all 33 benchmark instances, under all three.

## Auto is the default, and it needs four thresholds

Forced `CompactTable` is **3.7x on srch_k5 and 4.0x on Crossword but 0.74x on
Dubois and 0.82x on Kakuro**, so the choice has to be made per instance. Every
threshold is read off the measured suite rather than picked:

| | |
|---|---|
| `decide_after = 32` | Counted **per propagator**, not per search. Kakuro wakes a table one to three times in the whole search and Renault-megane about four; Crossword wakes one 1 898 times, `srch_k5` 3 146, `srch_bin_d20` 19 068. Nothing sits between. At 512 the wait itself cost `srch_k5` 36%. |
| `min_mean_live = 16` | Below this a call does too little work to cover an update's fixed cost. |
| `word bits * live >= tuples` | A bitset is only the right shape if its words are full: the support test costs a word per live word, where the live-set scan costs a step per tuple until it finds a witness. `enum_single_k10_t200k` keeps 46 tuples live in a 200 000-tuple table — one per word — and forced `CompactTable` measures 1.00x. |
| `min_tuples = one word` | A table that fits in one word never gets the state at all. |

Those last two, and every shift, mask and round-up over the bitsets, take the
width from `extensional_word_bits` — `numeric_limits<ExtensionalWord>::digits` —
rather than spelling 64 at each use. `max_words` keeps its literal: it is a count
of words, and matching 64 is a coincidence.

## Backtracking without a backtrack callback

`State::on_backtrack` is not reachable from a propagator's `const State &`, so
this follows `install_difference_propagator`: a propagator-owned undo trail plus
**one** trailed integer saying how much of it belongs to the current epoch,
unwound lazily at the top of the next call. The limit rides the trail rather
than taking a second slot — see the performance note below for what a slot
costs. `index` needs no trail at all: it uses the same sparse-set argument
`ExtensionalLiveTuples` uses for its size.

Because `Auto` switches mid-search it also has to cope with the search climbing
back above the node it switched at. That empties the trail, which is the signal:
the sparse set stopped being maintained there, so the whole table goes live
again and the next update filters it against the current domains.

## Measured

Median of 3, fataepyc-10 pinned, against the unmodified live-set build. Nodes
and solutions identical on every row.

| instance | live | compact | auto | |
|---|---|---|---|---|
| `Crossword-10-10` | 12.612 | 3.122 | 3.166 | **3.98x** |
| `Crossword-10-11` | 4.097 | 1.076 | 1.109 | **3.70x** |
| `srch_k5_d5_n12` | 0.364 | 0.098 | 0.101 | **3.61x** |
| `srch_k3_d8_n18` | 0.247 | 0.091 | 0.094 | **2.64x** |
| `srch_bin_d12_n30_s1` | 0.699 | 0.360 | 0.374 | 1.87x |
| `srch_bin_d20_n20_s1` | 2.619 | 1.474 | 1.491 | 1.76x |
| `srch_holey_d20_n25` | 0.325 | 0.207 | 0.212 | 1.53x |
| `srch_func_k3_n24` | 0.042 | 0.028 | 0.031 | 1.37x |
| `enum_single_k10_t200k` | 1.139 | 0.988 | 0.998 | 1.14x |
| Dubois, Kakuro, Renault, `enum_*` | | | | 0.97–1.02x |

The only loss is `srch_bin_d10_n20_s2` at **0.96x**, whose steady-state live set
is 7.7 tuples — below the threshold, but not yet at wake 32, when it still looks
like its sibling instance that gains 1.16x.

Against **Gecode 6.3.0** on matched trees, `Crossword` goes **5.0x -> 1.2-1.5x**
(it was 25-33x at this project's first baseline); the synthetic worst case is
4.7x on `enum_func_k3_n20`, which the per-pass budget says is engine-bound and
not a table problem. Renault and Kakuro stay ahead of Gecode end to end
(`total` 0.2-1.6x).

## One note for `dev_docs/propagator-performance.md`

Adding a second algorithm to a hot function cost the **first** one, on
instances that execute none of the new code, for three unrelated reasons —
separate costs on different instances, not three explanations of one number. The
inner membership test stopped being inlined once the function grew past GCC's
budget (27% of runtime as a suddenly-visible symbol; `always_inline` put it back,
worth 10% on `srch_k5` and 25% of the instruction count); instruction footprint
alone cost `Dubois` 7% before the new pass was moved out of line; and two extra
`add_constraint_state` slots cost `Dubois` 14% and `enum_shared` 24%, because
every slot is deep-copied into every search node even for a propagator that ends
up not using it. Written up there, because the
general shape — measure the *old* path against the *unmodified* build, not just
the new path against the old one — applies to any second implementation.

### DIFF
--- a/benchmarks/positive_table_random/positive_table_random.cc
+++ b/benchmarks/positive_table_random/positive_table_random.cc
@@ -4,8 +4,10 @@
 #include <examples/benchmark_cli.hh>
 
 #include <cstdlib>
+#include <iostream>
 #include <optional>
 #include <random>
+#include <string>
 #include <vector>
 
 #include <version>
@@ -14,15 +16,18 @@
 #include <print>
 #else
 #include <fmt/core.h>
+#include <fmt/ostream.h>
 #endif
 
 #include <cxxopts.hpp>
 
 using namespace gcs;
 
+using std::cerr;
 using std::mt19937;
 using std::nullopt;
 using std::optional;
+using std::string;
 using std::uniform_int_distribution;
 using std::vector;
 
@@ -33,11 +38,12 @@ using fmt::println;
 #endif
 
 // A random binary-CSP benchmark built entirely from Table (allowed
-// tuples), to exercise the negative-table propagator: n variables of domain
-// 0..d-1, and `constraints` random binary allowed-tuple tables
-// each of the d*d value pairs independently with probability `tightness`. Fixed
-// (deterministic) search, so the search tree is identical regardless of how the
-// propagator is triggered -- the interesting number is solve time / time-per-node.
+// tuples), to exercise the table propagator: n variables of domain
+// 0..d-1, and `constraints` random binary allowed-tuple tables, each allowing
+// each of the d*d value pairs independently with probability 1 - `tightness`.
+// Fixed (deterministic) search, so the search tree is identical regardless of
+// how the propagator is triggered -- the interesting number is solve time /
+// time-per-node.
 auto main(int argc, char * argv[]) -> int
 {
     cxxopts::Options options("positive_table_random", "Random allowed-tuple (Table) benchmark");
@@ -47,6 +53,7 @@ auto main(int argc, char * argv[]) -> int
         ("constraints", "Number of binary allowed-tuple tables", cxxopts::value<int>()->default_value("180"))                           //
         ("tightness", "Probability each value pair is forbidden (allowed = 1 - this)", cxxopts::value<double>()->default_value("0.34")) //
         ("seed", "Random seed", cxxopts::value<unsigned>()->default_value("1"))                                                         //
+        ("table", "Table algorithm: auto, live or compact", cxxopts::value<string>()->default_value("auto"))                            //
         ("first", "Stop at the first solution instead of enumerating all")                                                              //
         ("timeout", "Timeout in seconds (0 = none)", cxxopts::value<double>()->default_value("0"))                                      //
         ("help", "Display help");
@@ -62,6 +69,19 @@ auto main(int argc, char * argv[]) -> int
     auto tightness = options_vars["tightness"].as<double>();
     auto seed = options_vars["seed"].as<unsigned>();
     auto first_only = options_vars.contains("first");
+
+    // The in-repo way to reproduce the live-vs-compact comparison: the two are
+    // supposed to make identical inferences, so only the timings should move.
+    // Spelled as on the `tables` example.
+    auto algorithm = TableAlgorithm{table::Auto{}};
+    if (auto choice = options_vars["table"].as<string>(); "live" == choice)
+        algorithm = table::LiveSet{};
+    else if ("compact" == choice)
+        algorithm = table::CompactTable{};
+    else if ("auto" != choice) {
+        println(cerr, "Error: --table must be auto, live or compact");
+        return EXIT_FAILURE;
+    }
 
     mt19937 rng(seed);
     uniform_int_distribution<int> pick_var(0, n - 1);
@@ -82,7 +102,7 @@ auto main(int argc, char * argv[]) -> int
             for (int v = 0; v < d; ++v)
                 if (unit(rng) >= tightness)
                     allowed.push_back({Integer{u}, Integer{v}});
-        p.post(Table{{vars[a], vars[b]}, move(allowed)});
+        p.post(Table{{vars[a], vars[b]}, move(allowed)}.with_algorithm(algorithm));
     }
 
     unsigned long long solutions = 0;
--- a/dev_docs/propagator-performance.md
+++ b/dev_docs/propagator-performance.md
@@ -348,6 +348,39 @@ confident wrong answer first time:
   itself and reports exactly 1.00x for everything. `md5sum` the two binaries when
   a change reads 1.00x across the board.
 
+### Adding a second algorithm to a hot function slows down the first one
+
+The table propagator later grew a compact-table path alongside the live-set one,
+chosen per instance and dispatched on a bool at the top of
+`propagate_extensional`. That should cost the live-set path one predictable
+branch per call. It cost it far more than that, on instances that execute none
+of the new code, and for three unrelated reasons — each worth checking for
+whenever a second path lands in a function that was already hot. The three are
+separate costs on different instances, not three explanations of one number.
+
+- **The inner test stopped being inlined.** `perf record` showed
+  `bitmap_feasible` — pass 1's membership test, previously inlined away — as a
+  separate symbol at 27% of runtime. The function had not changed; the function
+  it was called from had grown past GCC's inlining budget. `[[gnu::always_inline]]`
+  on the test put it back, worth **10% on `srch_k5`** and 25% of the instruction
+  count. The symptom to look for is a helper appearing in the profile that did
+  not use to be there at all.
+- **Instruction footprint, even from code that never runs.** Before the new pass
+  was moved out of line with `[[gnu::noinline]]`, it cost 7% on `Dubois` — an
+  instance whose propagator does almost nothing per call, so its cost is
+  dominated by how much of the function is resident.
+- **Constraint state is not free to declare.** Every slot added with
+  `State::add_constraint_state` is deep-copied into every search node, so two
+  extra integers per table cost `Dubois` 14% and `enum_shared` 24% before the
+  second one was folded into the undo trail and the first was made conditional on
+  the table being large enough to want it. A propagator that ends up not using
+  the state still pays for it at every node.
+
+The general shape: a second algorithm is not free for the first one even when it
+is never executed, and the cost does not show up where you would look for it. Any
+such change needs the *old* path measured against the *unmodified* build, not
+just the new path measured against the old one.
+
 ## Is it the propagator, the strength, or the search?
 
 When a model is slow, "the propagator is slow" is only one of three
--- a/examples/tables/CMakeLists.txt
+++ b/examples/tables/CMakeLists.txt
@@ -5,5 +5,23 @@ target_link_libraries(tables PRIVATE glasgow_constraint_solver)
 target_link_libraries(tables PRIVATE cxxopts)
 target_link_libraries(tables PRIVATE gcs_fmt)
 
+# One lane per algorithm: the library paths are covered by table_test, so what
+# these add is that each algorithm solves this example correctly and writes a
+# proof VeriPB accepts. tables_compact is the only place outside table_test that
+# runs the compact algorithm at all -- this example's tables are all under
+# min_tuples, so auto never switches on them.
+#
+# They do not pin the example's own forwarding of --table, and cannot: all three
+# algorithms write byte-identical proofs by design, so run_test_and_verify.bash
+# has nothing to see. Deleting the forwarding, so that --table compact parses and
+# then selects auto, leaves all three lanes green (checked by mutation).
 add_test(NAME tables COMMAND ${GCS_BASH} ${CMAKE_SOURCE_DIR}/run_test_and_verify.bash $<TARGET_FILE:tables>)
+# --basename because these share a binary with the lane above, and without
+# distinct proof names a parallel ctest has them racing over one set of files
+# (issue #562).
+foreach(algorithm live compact)
+    add_test(NAME tables_${algorithm}
+        COMMAND ${GCS_BASH} ${CMAKE_SOURCE_DIR}/run_test_and_verify.bash --basename tables_${algorithm}
+        $<TARGET_FILE:tables> --table ${algorithm})
+endforeach()
 
--- a/examples/tables/tables.cc
+++ b/examples/tables/tables.cc
@@ -50,6 +50,8 @@ auto main(int argc, char * argv[]) -> int
             ("proof-files-basename", "Basename for the .opb and .pbp files", //
                 cxxopts::value<string>()->default_value("tables"))           //
             ("stats", "Print solve statistics")                              //
+            ("table", "Table algorithm: auto, live or compact",              //
+                cxxopts::value<string>()->default_value("auto"))             //
             ;
 
         options_vars = options.parse(argc, argv);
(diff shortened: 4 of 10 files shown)


ciaranm:
Pre-merge review. CI is green on all ten lanes, and the trial merge of the whole stack onto today's `main` builds with 0 warnings and passes 788/788 (details on #795). The trail and hand-over logic reads as sound — I have written up what I checked on #795 rather than repeating it here. Five things to raise, none of them about the algorithm.

### 1. The algorithm choice is a constructor parameter, where the house convention is a fluent setter

`gcs/constraints/table/table.hh:32` becomes:

```cpp
explicit Table(std::vector<IntegerVariableID> vars, ExtensionalTuples tuples, TableAlgorithm = TableAlgorithm::Auto);
```

Every other constraint that exposes an algorithm or consistency choice does it with a setter instead — 13 of them (`AllDifferent`, `BinPacking`, `Circuit`, `Divide`, `GlobalCardinality`, `Minus`, `Modulus`, `Multiply`, `NDimensionalElement`, `Plus`, `Power`, `ReifiedLinearEquality`, `SubCircuit`), and no constraint in the tree takes the choice as a constructor parameter. `dev_docs/constraints.md:141` records this as the convention families were deliberately migrated *to*:

> Families that used to expose several public classes behind a `using` alias — `AllDifferent`, `GlobalCardinality`, `Circuit` — are now a single class each, the choice moved onto the setter (`.with_consistency()`, or `.with_algorithm()` for `Circuit`'s `circuit::SCC`/`circuit::Prevent`).

`Circuit` is the direct analogue, and `gcs/constraints/circuit/circuit.hh:57` states the rule outright: "The constructor takes only the successor array; configure propagation with the fluent setters." A `.with_algorithm(TableAlgorithm)` would match that, and would leave the constructor signature and `clone()` untouched.

The *spelling* of the choice differs from the convention too, and the two points are worth taking together. All 13 of those are a `std::variant` over empty tag structs declared in the constraint's own header — `CircuitAlgorithm = std::variant<circuit::SCC, circuit::Prevent>`, `AllDifferentConsistency = std::variant<consistency::GAC, consistency::VC>`, and so on — where `TableAlgorithm` is a scoped enum in the shared `gcs/extensional.hh:99`. The variant form is what gets `Circuit` its "requesting anything else is a compile-time error" guarantee, and what lets `SubCircuit` hang per-algorithm knobs off its three tags; a `table::Auto` / `table::LiveSet` / `table::CompactTable` set would give `Table` the same, and would keep an enum that only `Table` accepts from sitting in front of `NegativeTable` and `SmartTable`.

Flagging rather than changing, since it is an API decision and outside what I should be touching.

### 2. `TableAlgorithm::Auto`'s doc claims more than the measurements support

`gcs/extensional.hh:106-107`:

> The default, and the only setting that never loses on this project's suite.

The PR body says otherwise, of `Auto` specifically:

> The only loss is `srch_bin_d10_n20_s2` at **0.96x**, whose steady-state live set is 7.7 tuples — below the threshold, but not yet at wake 32 [...]

So `Auto` does lose on one instance, by 4%. Worth fixing in the header rather than only in the body, because the comment is what a reader meets first and it will outlive the PR description — something like "the only setting without a significant regression" would be true and no less useful.

Two smaller things about that same instance: it does not appear in the measured table at all, and 0.96x falls outside the 0.97–1.02x band the grouped "Dubois, Kakuro, Renault, `enum_*`" row gives, so a reader working from the table alone cannot find it.

### 3. The new "10%" section does not add up against its own three items

`dev_docs/propagator-performance.md:356`:

> That should cost the live-set path one predictable branch per call. It cost it **10%**, on instances that execute none of the new code, and in three separate ways [...]

but L364, on the first of the three:

> `[[gnu::always_inline]]` on the test put it back and **recovered the whole 10%**.

and the other two are then given as 7% (`Dubois`, instruction footprint) and 14% / 24% (`Dubois` / `enum_shared`, the constraint-state slots). If item one accounts for the whole 10%, the headline should not be attributing that 10% to three causes; if the 10% is instead a total, then "recovered the whole 10%" is the wrong claim. I suspect the 10% is `srch_k5`'s figure for the inlining problem alone while the other two are `Dubois`/`enum_shared` figures, in which case it is purely a wording fix — but I cannot tell which from the document, which is the point.

### 4. The `tables` example's new `--table` flag has no ctest lane

`examples/tables/CMakeLists.txt:8` registers one lane with no arguments, so only `--table auto` is ever run:

```cmake
add_test(NAME tables COMMAND ${GCS_BASH} ${CMAKE_SOURCE_DIR}/run_test_and_verify.bash $<TARGET_FILE:tables>)
```

The *library* paths are covered thoroughly — `table_test` now runs every shape under all three settings plus the 120- and 400-tuple random tables — so the only uncovered thing is the example's own forwarding. Rather than leave that as a guess I ran it, and it is fine today: all three settings write byte-identical proofs, and veripb accepts them.

```
bcaa684e02411bd5675f6027880ab126  tab_auto.opb   bc98c4b914e05bef2e0c3e9e7b51f9ca  tab_auto.pbp
bcaa684e02411bd5675f6027880ab126  tab_live.opb   bc98c4b914e05bef2e0c3e9e7b51f9ca  tab_live.pbp
bcaa684e02411bd5675f6027880ab126  tab_compact.opb  bc98c4b914e05bef2e0c3e9e7b51f9ca  tab_compact.pbp

$ veripb tab_compact.opb tab_compact.pbp
s VERIFIED COMPLETE ENUMERATION OF 2 SOLUTIONS
```

`--table bogus` also exits 1 with the right message. Not a hole, then, just an unpinned one — and the precedent for pinning it is `regular_random`, which registers `regular_random-seeded-bacchus` for exactly this (`examples/regular_random/CMakeLists.txt:7`). Note the forced setting is the one worth a lane: the example's tables are all under `min_tuples`, so `auto` never reaches the compact path on them and only `--table compact` exercises it there.

### 5. The AI-declaration trailer is missing on `7b60bb87`

`CONTRIBUTING.md` asks for AI use to be declared both ways — a `Co-Authored-By:` trailer on the commits and a mention in the PR description — and this PR's single commit `7b60bb87` has neither. Same on #799's two commits, where I have written up the full context; in short, #795 and #796 both declare it on every commit and in their descriptions, and 60 of the last 60 non-merge commits on `main` carry the trailer, so this reads as an omission.

If these were written without AI help then there is nothing to do. If not, they want amending before the stack lands, and it is better done now than after review since the force-push restarts CI on both PRs.

ciaranm:
All five addressed. `3cfc962a` replaces `7b60bb87` (rebased onto #799's amended tip, then amended); ctest **777/777**, and the benchmark numbers are unchanged by the API move.

### 1. Fluent setter and variant tags — done

You were right that this was a convention violation, and it was mine to fix rather than just flag. `Table` now takes only the variables and the tuples, with `.with_algorithm()`, and `TableAlgorithm` is

```cpp
using TableAlgorithm = std::variant<table::Auto, table::LiveSet, table::CompactTable>;
```

over empty tag structs in the constraint's own `table::` namespace, matching `CircuitAlgorithm`. So the compile-time-error guarantee comes for free, the constructor signature and `clone()` are back to what they were, and the enum is no longer sitting in `gcs/extensional.hh` in front of `NegativeTable` and `SmartTable`. The innards do not see it at all now — `ExtensionalCompactTable` takes a plain `bool forced`, since the tag is a public API concern and the propagator only needs to know whether to skip the watching phase.

### 2. The `Auto` doc claim — fixed

Now "the only setting without a significant regression on this project's suite", with the exception named: within 2% everywhere except `srch_bin_d10_n20_s2`, where it is 4% behind. `min_mean_live`'s comment also now says *why* that instance slips through — the mean is taken over the first 32 wakes, which early in a search overestimates the steady state; that table settles at 7.7 live but does not look like it yet at wake 32. The commit message gives the instance its own line rather than hiding it in a grouped band.

### 3. The 10% — you read it right, and it was a wording bug

The 10% is `srch_k5`'s figure for the inlining problem alone; the 7% and the 14%/24% are `Dubois` and `enum_shared` figures for the other two. Three separate costs on different instances, not three explanations of one number. Reworded to say that explicitly, and the inlining bullet now names its own instance and adds the 25% instruction-count figure instead of claiming "the whole 10%".

### 4. ctest lanes — added, and they caught a flake I had introduced

`tables_live` and `tables_compact`, each with `--basename` so they do not race the existing lane over one set of proof files.

Adding them turned up something worse in my own test change: `run_big_random_table_test` wrote a fixed proof basename and was being called from inside the view-config loop, so `table_constraint` and `table_constraint_view_mixed` raced over `table_test_big.*` under a parallel ctest. It failed once for me and passed on rerun, which is exactly the shape of #562. The big-table and duplicate-variable tests wrap no views, so they now run once rather than once per view lane — which removes the race and the duplicated work together. Thank you for pushing on the uncovered lane; I would not have found this otherwise.

### 5. Trailer — real omission, repaired

Written with AI help; the trailer was missed, not withheld. Full mechanics on #799, including a correction about `Claude-Session:` where you were right and I was not.

ciaranm:
Re-reviewed after the force-push. The trailer is sorted on all three commits across #799 and #804, both descriptions now declare AI use, and author name, email and date survived the rewrite with the committer preserved as the author — so merge check 12 is satisfied. #799's rewrite was message-only (`git diff 93efdedf 1e74a7c5` empty); this one carried real changes, so I rebuilt and re-reviewed the delta rather than reusing the earlier result.

All four points are addressed, and the API one more thoroughly than I asked — `std::variant<table::Auto, table::LiveSet, table::CompactTable>` behind `with_algorithm()` matches `CircuitAlgorithm` down to the compile-time-error wording. I checked the dispatch conversion is truth-table equivalent for all three alternatives, and moving the innards struct to a plain `bool forced` is better than what I suggested: it drops the innards-to-public-enum coupling entirely. `clone()` carries the algorithm through the setter correctly.

**Verified on a trial merge onto today's `main`** (`6343e6f0`): merges clean, GCC 15.2 release build exit 0 at 100% with **0 warnings**, 134 test binaries, **790/790 ctest** with VeriPB 3.0.2. I also confirmed by running it that the big-random tests still run where they should after the move into the `run_dup` block — 12 invocations in the bare lane, 0 under `--view-position=mixed`, which is the de-duplication the comment describes. (That block reads oddly because `run_dup` now gates two unrelated things; `view_wrap_config_is_effectively_bare` is what it actually means, and it briefly convinced me the mid-search-switch coverage had been switched off.)

Three things from the delta.

### 1. The new lanes do not test what their comment says they test

`examples/tables/CMakeLists.txt` says:

> what these add is the example's own forwarding of `--table`, and that all three write a proof VeriPB accepts

The second half is true and the lane earns its place on it. The first half is not, and it is the #721 trap: since all three algorithms write byte-identical proofs, `run_test_and_verify.bash` cannot see which one ran. I mutation-tested both halves rather than guessing.

**Deleting the forwarding, keeping the validation** — one line after the `--table` parsing, so `--table compact` still validates but selects `table::Auto`:

```
1/3 Test #478: tables ...........................   Passed
2/3 Test #479: tables_live ......................   Passed
3/3 Test #480: tables_compact ...................   Passed
100% tests passed
```

All three green with the feature deleted, so no lane pins the forwarding.

**Throwing on entry to `propagate_compact_table`** instead:

```
1/3 Test #478: tables ...........................   Passed
2/3 Test #479: tables_live ......................   Passed
3/3 Test #480: tables_compact ...................***Failed
```

So `tables_compact` does reach the compact path, and `tables`/`tables_live` do not — which incidentally confirms that `auto` never switches on this example, since its tables are all under `min_tuples`. That is a genuinely useful lane: it is the only place outside `table_test` where the compact algorithm runs at all.

Two honest ways to close the gap: correct the comment to claim only what it pins, or make the flag observable — the example prints nothing naming the algorithm today, so a line it prints plus a `PASS_REGULAR_EXPRESSION` on the lane would make the forwarding claim true. Either is fine; the current pairing is the one thing that is not, because it reads as covered.

### 2. `overloaded{}` layout on the new `algorithm_label` (merge check 4)

`gcs/constraints/table/table_test.cc:188`, replacing the old `switch`:

```cpp
    return overloaded{[](table::Auto) { return "auto"; }, //
        [](table::LiveSet) { return "live"; },            //
```

First lambda up on the brace line, which is the shape the check exists to catch. I counted rather than asserting taste: 139 sites in the tree put `overloaded{` alone on its own line, 40 do not, so the switch-block form is the convention by a wide margin. The trailing `//` pins are currently holding the odd layout in place rather than forcing the switch-block one, which is the opposite of what they are for. clang-format is green on it either way — that is why this check is manual.

### 3. `--table` is not on any in-repo benchmark driver (merge check 11)

The body still says the flag is on "the `tables` example and on the benchmark driver". There are two in-repo table drivers, and `benchmarks/positive_table_random` posts

```cpp
p.post(Table{{vars[a], vars[b]}, move(allowed)});
```

with no `--table` option and no way to select an algorithm. The measurements are plainly from an out-of-tree harness — `Crossword-10-10`, `Dubois`, `Renault-megane` and the `srch_*`/`enum_*` families do not correspond to its `--vars/--domain/--constraints/--tightness` parameters — and a harness living outside the repo is normal and not a finding in itself. But a reader will go looking in `benchmarks/` on the strength of that sentence. Worth either saying it is out-of-tree, or giving `positive_table_random` the flag, since it is the in-repo way to reproduce the live-vs-compact comparison.

### One number moved that nobody asked to move

`extensional_utils.hh`'s `min_mean_live` note went from "0.57-0.93x ... Above it, 1.35x and upwards" to "0.74-0.96x ... 1.4x". Nothing I raised touched it and I cannot verify either version. Noting it only because `0.74` now agrees with the body's "0.74x on Dubois" where `0.57` matched nothing stated anywhere, so it looks like a tightening in the right direction rather than a slip.

CI is still running on both PRs (clang-format and Windows/VeriPB already green); I will confirm the full set once it settles. Nothing here blocks, and none of it needs a rebuild of the stack — 2 and 3 are a one-line tidy and a sentence.

PR #806, opened 2026-09-03
Say which constraint type the propagation time went to

ciaranm:
`GCS_PROPAGATOR_STATS=calls` adds a per-constraint-type block to the statistics — how many constraints of each type there are, how many propagators they installed, and how many times those propagators were called, changed a domain and contradicted. `GCS_PROPAGATOR_STATS=time` adds the elapsed time in each. It is a `ComponentStats` block, so it reaches `%%%mzn-stat:` lines from `fzn-glasgow -s` and `xcsp_glasgow_constraint_solver -s`, and the default `operator<<` on `Stats` prints the summary line, with no per-caller plumbing.

## Why

The aggregate `propagations:` count cannot tell you which constraint to look at, and #788 is the case for why that matters.

`SubCircuit`'s reachability arm looked free on the `tpp` family and crushing on `mario` at comparable `n`, which read as a per-call cost that was not a function of `n`. It is not. Per *call* it costs almost exactly the same on both families at the same `n` — 7.4 µs on `mario` at n=15 against 7.7 µs on `tpp` at n=15, 31.2 µs on `mario` at n=30 against 30.7 µs on `tpp` at n=35. What differs is how often it is woken: **1.4–2.5 times a node on `mario`, and 676 times in a 7.5-million-node search on `tpp`**, because `tpp`'s `seq_search` fixes all `n` successors in the first `n` levels and then spends the entire search on `purchaseLoc` below them. Calls-per-node and cost-per-call vary independently and only their product was visible.

The same run turned up a second thing on `tpp` that the aggregate count hides: the twenty `lin_not_equals` constraints are called 17.8 million times, **remove a value once in the whole search**, contradict 3.75 million times, and take half the propagation time. Their comment already says they only care when one variable is left unassigned, and they wake `on_change`. Filed separately.

## Reading the two rungs

`calls` is one increment per propagator run, so a run with it on is comparable with a run without — verified on a real instance: `GCS_PROPAGATOR_STATS=time` and an uninstrumented run report identical node counts on `mario_n_medium_3` and `mario_t_hard_5` under both algorithms. `time` reads the clock twice a run (about 13% on `tpp_3_5_20_1`, more where the propagators are individually tiny), so read shares and per-call figures off a timed run, never throughput.

## Design

- Grouped by constraint **type**, not by constraint: a model with ten thousand linear constraints wants one `lin_less_equal` row and not ten thousand, and the propagator and constraint counts say how many are behind each row.
- The labels are applied by `Constraint::install()`, the one function every install path runs through and so the only place that knows both a constraint's type and which propagators it installed. Doing it there is also what keeps a **child** constraint's propagators labelled with the child's own type: the child is installed from inside its parent's `prepare()`, so its own recursion through `install()` has already labelled them, and the parent labels only what is left.
- The block is registered only when it was asked for. An unasked-for component has not done nothing, it has not run, and a line on every solve naming an environment variable is noise rather than a finding.
- The default path pays one branch per propagator run and nothing else; the four per-propagator vectors are only allocated when the variable is set.

## Checks

- 789/789 tests pass; warning-free under GCC with `GCS_WERROR=ON`, and clean under clang + sanitizers (the two warnings clang reports are `variable_weighting.hh` and `element.cc`, both pre-existing on main).
- `propagator_call_stats_test` pins the exact figures for propagators whose behaviour is written out in front of them, and asserts the per-type counts sum to the aggregates. Two mutation probes were rejected: letting a parent overwrite a child's label, and dropping the per-propagator effectful increment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,6 +361,15 @@ autotuner) without recompiling. Each is read once, on first use.
   some backtrackable bookkeeping (default: 8). The test suite runs in both modes by setting
   this.
 
+* ``GCS_PROPAGATOR_STATS``: adds a per-constraint-type propagator breakdown to the statistics ---
+  how many constraints of each type there are, how many propagators they installed, and how many
+  times those propagators were called, changed a domain and contradicted. ``calls`` (or ``1``) is
+  one increment per propagator run, so a run with it on is comparable with a run without;
+  ``time`` (or ``2``) additionally times each run, which costs two clock reads per propagator call
+  and so makes the run's own wall clock and nodes-per-second incomparable with an untimed one ---
+  read shares and per-call figures off it, not throughput. An unrecognised value is ignored with a
+  warning. See ``dev_docs/propagator-performance.md``.
+
 * ``GCS_TABULATION_THRESHOLD``: under ``consistency::Auto``, a constraint tabulates itself for
   generalised arc consistency when the product of its variables' domain sizes is no bigger than
   this (default: 100). A variable that is functionally determined by the others (say, the result
--- a/README.md
+++ b/README.md
@@ -368,7 +368,9 @@ autotuner) without recompiling. Each is read once, on first use.
   ``time`` (or ``2``) additionally times each run, which costs two clock reads per propagator call
   and so makes the run's own wall clock and nodes-per-second incomparable with an untimed one ---
   read shares and per-call figures off it, not throughput. An unrecognised value is ignored with a
-  warning. See ``dev_docs/propagator-performance.md``.
+  warning. The breakdown appears as ``%%%mzn-stat:`` lines from ``fzn-glasgow -s``, as
+  ``d PROPAGATOR CALLS ...`` lines from ``xcsp_glasgow_constraint_solver``, and as a summary
+  line in the default ``Stats`` output. See ``dev_docs/propagator-performance.md``.
 
 * ``GCS_TABULATION_THRESHOLD``: under ``consistency::Auto``, a constraint tabulates itself for
   generalised arc consistency when the product of its variables' domain sizes is no bigger than
--- a/dev_docs/README.md
+++ b/dev_docs/README.md
@@ -46,6 +46,12 @@ library. For an introduction to *using* the solver, start with the top-level
   measuring the wall-time impact of a performance-sensitive change, the
   rationale for each pick, the harness pattern for comparing two builds,
   and what to capture. Use when quantifying a refactor's perf impact.
+- [Making a propagator faster](propagator-performance.md) — what to change
+  once a propagator is correct: the per-constraint-type breakdown
+  (`GCS_PROPAGATOR_STATS`) that says which constraint the time went to, then
+  the levers cheapest first, and the ground rules that keep a performance
+  change from quietly becoming a strength change. The companion to
+  `benchmarking.md`, which is about how to measure instead of what to change.
 - [Proof benchmarks](proof-benchmarks.md) — the counterpart set for when
   the *proof* is what is being measured: proof-writing cost, proof size
   and VeriPB checking time. Groups instances by whether they stress
--- a/dev_docs/propagator-performance.md
+++ b/dev_docs/propagator-performance.md
@@ -81,6 +81,49 @@ idea is worthwhile. Consistency has value: it removes a whole class of avoidable
 work and makes the next propagator's default the good one. Don't use this as a
 loophole for speculative single-constraint cleverness.
 
+### Find out which constraint is spending the time, before optimising one
+
+`GCS_PROPAGATOR_STATS=calls` adds a per-constraint-type block to the stats: how
+many constraints of that type there are, how many propagators they installed,
+and how many times those propagators were called, changed a domain, and
+contradicted. `GCS_PROPAGATOR_STATS=time` adds the elapsed time in each,
+totalled per type. It reaches `%%%mzn-stat:` lines from `fzn-glasgow -s` and
+`xcsp_glasgow_constraint_solver -s`, and the default `operator<<` on `Stats`
+prints the one-line summary, so a sweep harness can tabulate it without any
+per-constraint plumbing.
+
+Read the two rungs for what they are. `calls` is one increment per propagator
+run, so a run with it on is comparable with a run without. `time` reads the
+clock twice per run — around 10% on a propagation-heavy model, but far more if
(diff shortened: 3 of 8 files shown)


ciaranm:
Pre-merge review. CI is 10/10 green including all three Windows lanes, and since this branch sits on `6343e6f0` — 10 commits behind `main`, so it has never been built against today's table stack — I trial-merged it together with #810 and the #811/#813/#814 stack onto current `main`: no conflicts, **0 GCC 15.2 warnings**, **794/794 ctest**. `gcs/CMakeLists.txt` is shared with #814 but in a disjoint region, so this costs no rebase cycle in any merge order.

Four things, none of them a behaviour bug.

## 1. The `unlabelled` row names the wrong resident

`propagators.cc` says:

> A propagator installed by something that is not a Constraint --- the learned-nogoods store is the one today --- never passes through `Constraint::install()` and so is labelled by nothing.

The learned-nogoods store does pass through it. `Nogoods` is a `Constraint` with its own `constraint_type()` override, and `solve.cc:333` installs it the ordinary way (`std::move(nogoods_constraint).install(propagators, state, …)`). The actual unlabelled source is **`AutoTable`**, the only presolver that calls `propagators.install(...)` on the raw `Propagators` (`auto_table.cc:155`, with `CurrentlyUnnamedConstraint{}`).

Settled by running it rather than reading it — I patched `summary()` to list every row:

```
colour --stats --restarts 1   ->  not_equals=30/10p array_max=14/1p nogoods=1/1p
shikaku --autotable --stats   ->  lin_less_equal=286/20p multiply=70/5p disjunctive2d_strict=69/1p unlabelled=25/5p
```

So the bucket is right and live — it is only the example that is wrong, which matters because "your unlabelled row is dead code" would have been the wrong conclusion to draw. The claim is in three places: the comment above `rows[unknown_type].type = "unlabelled"`, the comment in `propagator_call_stats_test.cc`'s "A propagator no constraint claimed" case, and the description's "the one function every install path runs through" (`AutoTable` is the counter-example to the literal reading).

Two greps enumerate it, for next time:

```
grep -rn '\.install(\|->install(' --include=*.cc gcs/ | grep -v '^gcs/constraints/'
grep -rn 'propagators\.install' gcs/presolvers/
```

## 2. Nothing exercises the report through `Constraint::install()`

`GCS_PROPAGATOR_STATS` appears nowhere in the tree except README, dev_docs, the implementation and the new test — no CMakeLists, wrapper script or model sets it — and `fill_in_constraint_stats()` returns early when it is unset. So the report is invisible to all 794 lanes.

That would be fine on its own, except that `propagator_call_stats_test.cc` reaches the machinery by calling `Propagators::install()` and `Propagators::note_propagator_types()` **directly**, and never through `Constraint::install()`. `Constraint::install()` is the one line that actually wires the feature up, and it is where the whole design argument lives — the `propagators_before` window, the "`prepare()` returned false is included" case, and the child-keeps-its-own-type property, which the test reproduces by hand rather than through the production path. Both stated mutation probes land inside the tested half, so a mutation of `constraint.cc` cannot fail anything.

Not asking for a lane — noting that no lane could have told you. I did verify the description's headline non-interference claim myself, since nothing asserts it either: identical solutions, recursions and propagations with the knob off, at `calls`, and at `time`, on an 18-node SubCircuit enumeration (189,824 recursions). The `time` overhead there was ~2%, consistent with the doc's caveat that it is much worse where propagators are individually tiny.

## 3. `propagator-performance.md`: "growing about as `n²`" does not follow from the pair quoted

The doc gives 7.7 µs at n=15 and 30.7 µs at n=35 on `tpp`, "growing about as `n²`". That pair is `n^1.63` — n² would predict 41.9 µs, not 30.7. The *mario* pair in the PR description (7.4 µs at n=15, 31.2 µs at n=30) is `n^2.08`, almost exactly n², so the claim is true of the figures that did not make it into the doc. Either quote the mario pair alongside it or soften the exponent.

## 4. Minor: the timing overhead figure

The doc says `time` "reads the clock twice per run — around 10% on a propagation-heavy model"; the description says 13% on `tpp_3_5_20_1`. Same measurement, two numbers.

## AI declaration

The description declares AI assistance but none of the three commits carries the `Co-Authored-By:` trailer, which CONTRIBUTING.md asks for as the other half. 60 of the last 60 non-merge commits on `main` have it. Same on #810, #811, #813 and #814 — worth doing in one pass.

(For the record: these three commits were authored 09:43–10:16Z, i.e. *before* the same comment went on #799, so this one is not a repeat.)

ciaranm:

ciaranm:
## The `unlabelled` example was wrong, and the correction matters

You are right on both halves. `Nogoods` is a `Constraint` with its own `constraint_type()` and `solve.cc` installs it the ordinary way, so it gets a `nogoods` row; the actual resident is **AutoTable**, `auto_table.cc:151`, installing under `CurrentlyUnnamedConstraint` because a presolver-derived propagator has no posted-constraint identity. Fixed in all three places you list.

Worth saying why this was worth a finding rather than a nit: the bucket is right and live, and "your unlabelled row is dead code" would have been the wrong conclusion to draw from a wrong example. Your two greps are the enumeration; I have put the AutoTable case in the comment so the next reader has a live one.

## The coverage gap, and what it turned up about my own code

Added a case that goes through `Problem::post()` and `solve()`. `Power` is the vehicle for the child property and a neat one: `Power::prepare()` installs a `Table` child and `Power` installs no propagator of its own, so a posted `Power` shows up as a `table` row with **no `power` row at all** — the other way round if a parent overwrote its children's labels.

Then I mutated `constraint.cc` two ways, which is the thing your finding made possible:

- dropping the labelling call — **fails several lanes**, good;
- widening the window from `propagators_before` to `0` — **fails nothing**.

The second is not a test weakness, it is a fact about the code: labelling is first-writer-wins and installation is serial, so a wider window relabels nothing and only costs a longer scan. **`propagators_before` is a shortcut, not part of the correctness argument**, and my commit message for it implied otherwise. The comment now says which is which. That is the kind of thing only a production-path test could have surfaced, so the finding earned more than the test it asked for.

## The two figures

`n^1.63` is right and `n²` was not — it is the *mario* pair (7.4 µs at n=15 to 31.2 µs at n=30, `n^2.08`) that carries the claim, and it had not made it into the document. Both pairs are there now, mario first.

Overhead is 13%, on `tpp_3_5_20_1`; the doc's 10% was the loose one.

PR #810, opened 2026-09-03
Test the successor range guard Circuit and SubCircuit both rely on

### LINKED ISSUE #201: Circuit negative successor values should be a constructor-time InvalidProblemDefinitionException
`gcs/constraints/circuit/circuit_base.cc:70` currently does:

```cpp
if (*current_val < 0_i)
    throw UnimplementedException("Successor encoding for circuit can't have negative values");
```

This is caller-input validation in disguise — the caller built a `Circuit` whose successor variables have domains reaching negative values. Two problems with the current shape:

1. **Wrong exception class.** It isn't unimplemented functionality; it's a problem-definition error. Should be `InvalidProblemDefinitionException` (introduced in #194 / PR for that issue).
2. **Wrong place.** The check fires deep inside `output_cycle_to_proof` during proof emission. By then we've already done work that's about to be discarded. The check should hoist into `Circuit::prepare` (or the constructor), so callers get an immediate error rather than a delayed one that only surfaces when proof logging happens to trip it.

Spotted while surveying remaining `UnexpectedException` / `UnimplementedException` callers after #194.

ciaranm:
Closes #201.

## The gap

`Circuit::prepare()` and `SubCircuit::prepare()` both `define_bound()` every successor to `0..n-1`. Both then index per-node arrays with the fixed successor values through casts that check nothing — a `p[i]` position row per value in `Circuit`'s proof output, `has_fixed_pred` in `SubCircuit`'s fixed-edge walk. **Every existing test declares its successors as exactly `0..n-1`**, so that guard never did anything in the whole suite, and a no-op passes every check.

It is emphatically not a no-op. Deleting it:

- makes `Circuit` at n=1 report `succ = [-2]`, `[-1]`, `[1]` and `[2]` as one-node circuits;
- makes `SubCircuit` **segfault**, because a negative value indexes `has_fixed_pred` at a vast offset.

## The test

One wide-domain case per constraint, declaring two values either side of the range (so a negative value, `n` itself, and one beyond it are all in the declared domain) and asserting **the same solution set** the `0..n-1` runs get — the answer must not change. At n=1, the smallest size where an out-of-range value can reach the position encoding, and n=4; both algorithms; with proofs, so VeriPB checks the bound rows too (`s VERIFIED COMPLETE ENUMERATION`).

Run bare rather than through the view sweep: a view's own bounds are what a wrap would widen, and the guard is about the successor the constraint sees.

Both mutation probes above were run against these tests and are caught.

## What this does about #201

#201 asked for `circuit_base.cc`'s

```cpp
if (*current_val < 0_i)
    throw UnimplementedException("Successor encoding for circuit can't have negative values");
```

to become an `InvalidProblemDefinitionException` hoisted out of proof emission into `prepare()` or the constructor.

**The hoisted version already exists** and is the `define_bound` above — and it is a phase *earlier* than the constructor the issue suggested, because a variable's domain lives in the `State`, not in the `IntegerVariableID` handle the constructor is passed; only a constant successor could be range-checked at construction. So what was left at the emission site was never caller-input validation, it was a bug check standing in front of an unchecked `as_index()`. It now says that, as an `UnexpectedException`, matching the line directly above it. The issue's point 1 (wrong exception class) is fixed; its point 2 (wrong place) turns out already to have been done.

`SubCircuit`'s equivalent site is the per-call fixed-edge walk rather than proof emission, so a runtime check there would be paid on every propagator call. That one gets the invariant and its source written down instead.

## Checks

789/789 tests pass (788 on this branch, which does not carry #806's new test), warning-free under GCC with `GCS_WERROR=ON`.

### DIFF
--- a/gcs/constraints/circuit/circuit_base.cc
+++ b/gcs/constraints/circuit/circuit_base.cc
@@ -42,8 +42,16 @@ auto gcs::innards::circuit::output_cycle_to_proof(const vector<IntegerVariableID
     if (current_val == nullopt)
         throw UnexpectedException("Circuit propagator tried to output a cycle that doesn't exist");
 
+    // Cannot happen, and is checked because as_index() below is an unchecked cast: a
+    // negative value would index the position map at a vast offset rather than fail.
+    // Circuit::prepare() define_bound()s every successor to 0..n-1, which is one phase
+    // earlier than the constructor could manage (a variable's domain lives in the State,
+    // not in the handle) and long before any propagator runs, so nothing outside that
+    // range survives to be output here. Issue #201 asked for this to become a
+    // problem-definition error hoisted out of proof emission; the hoisted version is
+    // that define_bound, and what is left here is a bug check (#201).
     if (*current_val < 0_i)
-        throw UnimplementedException("Successor encoding for circuit can't have negative values");
+        throw UnexpectedException("Circuit propagator output a successor outside 0..n-1, which prepare() should have ruled out");
 
     PolBuilder pol;
     pol.add(pos_var_data.at(start).plus_one_lines.at(current_val->as_index()).geq_line);
--- a/gcs/constraints/circuit/circuit_base.cc
+++ b/gcs/constraints/circuit/circuit_base.cc
@@ -42,8 +42,11 @@ auto gcs::innards::circuit::output_cycle_to_proof(const vector<IntegerVariableID
     if (current_val == nullopt)
         throw UnexpectedException("Circuit propagator tried to output a cycle that doesn't exist");
 
-    // Cannot happen, and is checked because as_index() below is an unchecked cast: a
-    // negative value would index the position map at a vast offset rather than fail.
+    // Cannot happen, and only the negative side is checked here --- deliberately, because
+    // that is the side as_index() below cannot catch: it is an unchecked cast, so a
+    // negative value would index the position map at a vast offset rather than fail. A
+    // value >= n reaches plus_one_lines.at() instead, which is checked and throws
+    // std::out_of_range on its own.
     // Circuit::prepare() define_bound()s every successor to 0..n-1, which is one phase
     // earlier than the constructor could manage (a variable's domain lives in the State,
     // not in the handle) and long before any propagator runs, so nothing outside that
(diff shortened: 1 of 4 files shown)


ciaranm:
Pre-merge review. CI is 10/10 green including all three Windows lanes. This branch sits on `6343e6f0`, 10 commits behind `main`, so I trial-merged it together with #806 and the #811/#813/#814 stack onto current `main` — no conflicts (it shares `subcircuit.cc` and `subcircuit_test.cc` with #813/#814, but in disjoint regions), **0 GCC 15.2 warnings**, **794/794 ctest**.

## Both mutation probes reproduce, exactly as described

Commented out the successor `Bound::Lower` `define_bound` in `circuit.cc:126` and `subcircuit.cc:1010`, rebuilt, and ran all 28 circuit/subcircuit lanes:

```
circuit/scc [w0_pall] n=1: expecting 1 solutions      <- passes
circuit/scc [w0_pall] n=2: expecting 1 solutions      <- passes
circuit/scc wide domains n=1: expecting 1 solutions   <- FAILS here
```

```
subcircuit/check [w0_pall] n=2: expecting 2 solutions    <- passes
subcircuit/check wide domains n=1: expecting 1 solutions
run_test_only.bash: line 8: 3101718 Segmentation fault (core dumped)
```

Every pre-existing case passes under the mutation and 26 of the 28 lanes stay green, so the two new wide-domain cases are the only thing in the suite that catches it. That is the claim the PR makes, and it holds — including the segfault.

I also checked the premise behind the new comment in `subcircuit.cc`, since `walk_fixed_edges` sits in a file whose other rules are SCC-only: it is called unconditionally at the top of `propagate_subcircuit`, before the `if (scc_anchor)` block, so `check` and `prevent` do reach it and the wide-domain case in `subcircuit_test.cc` is the right test to point at. Comment is accurate.

## One nit, in the message rather than the code

`circuit_base.cc`'s guard tests only the lower bound:

```cpp
if (*current_val < 0_i)
    throw UnexpectedException("Circuit propagator output a successor outside 0..n-1, which prepare() should have ruled out");
```

The message says "outside `0..n-1`" but the condition catches only negatives. A value `>= n` reaches `plus_one_lines.at(...)` instead and comes out as a bare `std::out_of_range` — which is safe, since `.at()` is checked, and that is exactly why the negative case is the one worth its own throw. Either narrow the message to what it detects, or say in it that the upper side is caught by `.at()`; as it stands the next reader will assume both halves are tested here.

## AI declaration

The description declares AI assistance but the commit carries no `Co-Authored-By:` trailer, which CONTRIBUTING.md asks for as the other half; 60 of the last 60 non-merge commits on `main` have it. Same on #806, #811, #813 and #814, so it is one pass across all five. Worth noting this commit was authored at 14:01Z, after the same point went on #799 at 10:38Z.

ciaranm:

ciaranm:
Both fair.

The message now says "negative", and the comment says why the asymmetry is deliberate rather than leaving the next reader to assume both halves are tested there: the negative side is the one `as_index()` cannot catch, being an unchecked cast, where a value `>= n` reaches `plus_one_lines.at()` and throws `std::out_of_range` on its own.

Thanks for reproducing both mutation probes independently, segfault included — that is the claim the PR rests on and it is better for having been checked by someone other than the author.

PR #811, opened 2026-09-03
Work out what certifying F&S's scc pruning rules would take

ciaranm:
Documentation only. Expands `dev_docs/subcircuit-proof-logging.md`'s "Not implemented" section from three sentences into an analysis, for #788.

The performance case for these rules is weak — the arm they sit on is [measured losing](https://github.com/ciaranm/glasgow-constraint-solver/issues/788#issuecomment-5523737820), and the node reduction where it is measurable at all is 1.09x–1.38x — but whether they can be **certified** in this encoding is a separate question, and the answer is more positive than the encoding's shape suggests.

## The finding

F&S state the rules over the subtree structure of a depth-first traversal. `pos` here is a *tour-order* labelling pinned at exactly one place, `pos[anchor] = 0`, so none of that structure exists in the encoding. Working each of the seven explanation clauses in §5.3.2 back to what makes it true, the DFS tree turns out to be how the propagator *finds* these inferences rather than why they hold:

- **Rules 1 and 5, and the conflict cases of 3 and 4, are already ours.** They all rest on one cut fact — if both sides of a cut hold an evidence node and no edge crosses it, then whichever side the anchor is on, the other is stranded — and stranding is exactly what the two existing walks derive.
- **Prune root (6) and prune within (7) are strictly stronger, and both are the existing induction run under one assumed edge.** Neither fires on the live domains, which is why they add strength. In this encoding they are **singleton arc consistency on the successor variables**, over the candidates a DFS pass identifies cheaply.

## Probed, not just argued

`~/claude/tmp/788-subcircuit-step0/prune_root_probe.cc` — six nodes, anchor 0 with candidates `{1,4}`, a strongly connected `{1,2,3}` with no edge to `{4,5}`, node 4 required on the tour and reachable only via `0 -> 4`. Every successor in `{1,2,3}` keeps two values under the hypothesis, so all-different fixes nothing and no cycle closes; that is what stops `check` and `prevent` reaching the contradiction first.

| | result |
|---|---|
| `succ[0]` free | satisfiable; `scc` finds it in **1 recursion**, `prevent` in 5 |
| `succ[0] = 1` posted after the constraint | `scc` contradicts at the root (**1 recursion**, `prevent` 3), proof **`s VERIFIED UNSATISFIABLE`** |
| same, `derive_unreachable` swapped for plain `JustifyUsingRUP` | VeriPB **rejects**, at `.pbp:95` |

So the existing certificate survives being run under a hypothesis — **no new row family, no new proof-only variable, no new cutting-planes step** — and the induction is doing the work rather than unit propagation. For these two rules the proof-logging question is answered; what is left is propagator work.

## What would actually be hard, also written down

1. **Proof size is the binding constraint.** `O(n²)` rows per inference, `O(n³)` per call once it runs per candidate value. The real 15-house instance already writes a 194 MB `.pbp` with the walks firing once. Measure before building.
2. **Forced edges: shave, don't derive the cut row.** Shaving yields `succ[c] = b` as a by-product at `O(n³)` literals; deriving the cut row needs one induction but carries the whole crossing set as guards on every row, `O(n⁴)`. The cheaper route is the one that looks more wasteful.
3. **Where the hypothesis lives** — a guard on every derived row, or a deleted proof level. Decide before writing any of it.
4. **Rule 1 with no anchor is genuinely out of reach**: the induction has to start somewhere and `pos` is pinned only at the anchor. This is the one place the arbitrary-root position-offset problem is real — and unreachable for us anyway, since the arm does nothing without an anchor.

Plus a sign error in the paper worth knowing before implementing from it: §5.3.2's definition of `in(a)` has a parenthetical (`a ∈ D(x_a)`) that contradicts both its own prose and every clause body in the section.

### DIFF
--- a/dev_docs/subcircuit-proof-logging.md
+++ b/dev_docs/subcircuit-proof-logging.md
@@ -431,12 +431,107 @@ The same shape is in four more redefinitions — the three `bin_packing` ones an
 `fzn_regular` — and is [issue #803](https://github.com/ciaranm/glasgow-constraint-solver/issues/803).
 The graph family shifts *parameters*, which costs nothing.
 
-## Not implemented
+## Not implemented: F&S's pruning rules, and what certifying them would take
 
-F&S's four extra `scc` pruning rules — prune root, prune skip, fix required edges
-and prune within — with the evidence-node guard each one needs for `subcircuit`.
-Those need the subtree structure of a depth-first traversal, where the two
-reachability walks need only the component. Their explanations are all `O(n²)`.
+F&S's remaining `scc` rules — prune root, prune skip, fix required edges and prune
+within — with the evidence-node guard each one needs for `subcircuit`. Their
+explanation clauses are §5.3.2 of the paper, numbered 1 to 7 there and referred to
+by those numbers below.
+
+They are stated over the **subtree structure of a depth-first traversal**: which
+subtree was visited before which, and the lowpoint of a node's first child. Our
+encoding knows nothing about that. `pos` is a *tour-order* labelling, not a DFS
+one, and it is pinned at exactly one place, `pos[anchor] = 0`.
+
+That sounds like an obstacle and mostly is not, because the DFS tree turns out to
+be how the propagator *finds* these inferences rather than why they hold. Working
+each clause back to what makes it true:
+
+**Rules 1 and 5, and the conflict cases of 3 and 4, are already ours.** Rule 1 is
+a strongly connected sub-component `S` with an evidence node `a ∈ S`. Given an
+anchor there are only two cases and the walks cover both: if the anchor is in `S`
+then nothing outside `S` is forward-reachable from it, and if it is not, then `a`
+cannot reach the anchor. Rule 5 (an edge skipping subtrees) is the same argument
+over the partition its reason describes — with no `A → B ∪ C` and no `B → C`
+edges, whichever of the three sets holds the anchor leaves one of the other two
+unreachable in one direction or the other. The general shape is a **cut**: if
+`A` and `V \ A` both contain an evidence node and no edge crosses `A → V \ A`,
+then whichever side the anchor is on, the other side is stranded, so one of the
+two walks fires. That cut fact is what all of 2, 3, 4 and 5 rest on.
+
+**Prune root (rule 6) and prune within (rule 7) are strictly stronger than what we
+do, and both are the existing induction run under one assumed edge.** Prune root:
+assume `succ[anchor] = e`; if the last subtree is only reachable through the
+anchor, it is now unreachable, so its evidence node is forced off the tour, which
+contradicts the evidence literal. Prune within: assume `succ[p] = c`; the subtree
+at `c` then has no way back out, so the backward walk strands it, and `succ[p] = c`
+with `c` a self loop takes value `c` twice. Neither fires on the live domains,
+which is precisely why they add strength.
+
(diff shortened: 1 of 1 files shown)


ciaranm:
Pre-merge review, bottom of the #811 → #813 → #814 stack. CI is 10/10 green including all three Windows lanes. Documentation only, so nothing builds it — which is what Merge Check 14 exists for: I read the document against itself rather than re-running anything.

The substantive content reads well and the probe is the right kind of evidence. Three things.

## 1. The `.pbp` size contradicts the same document

The new text says:

> The real 15-house instance already writes a 194 MB `.pbp` with the walks firing once

Line 309 of the same file, in the results section, says:

> Real `mario_easy_2`, 15 houses unmodified: 10,021 nodes, a 594 KB `.opb` and a **186 MB** `.pbp`, verified in 176 s.

A file size is deterministic, so one of the two is stale. It is not the MB/MiB factor either — 186 MiB is 195.0 MB and 194 MB is 185.0 MiB, so no single byte count rounds to both. I cannot tell which is right, so flagging rather than picking.

Worth saying that this one self-heals: **#813 deletes the sentence containing "194 MB"**, so it exists on `main` only in the window between the two merges. Still worth fixing here if the stack is going to land one PR at a time.

## 2. "recorded elsewhere" has no antecedent

> This is the one place the "arbitrary root" position-offset problem (recorded elsewhere as *not* a real blocker, correctly, for the arm as it stands) is real

On this branch that phrase appears exactly once in the whole repository:

```
$ git grep -n -iE 'arbitrary root|arbitrary-root|position.offset|pos\[a\]' <this branch>
dev_docs/subcircuit-proof-logging.md:523:   place the "arbitrary root" position-offset problem
```

So the back-reference points at nothing a reader of the repo can find — presumably an issue comment or a scratch note. Either name the source (#788's thread, if that is where it is) or drop the parenthetical. #813 does in fact drop it, so this is the same transient as above.

## 3. Merge-order note, not a defect

#813 replaces this entire section — of the ~99 lines added here, only the "sign error in the paper" subsection survives into #814's tree. That reads as the intended analysis-then-implementation narrative, and the analysis is worth having in the history on its own account, but it does mean #811's lasting contribution to the file is small and that both findings above are transients. Flagging so the sequencing is deliberate rather than a surprise.

## AI declaration

The description declares AI assistance but the commit carries no `Co-Authored-By:` trailer, which CONTRIBUTING.md asks for as the other half; 60 of the last 60 non-merge commits on `main` have it. Same on #806, #810, #813 and #814 — one pass across all five. This commit was authored at 14:12Z, after the same point went on #799 at 10:38Z.

ciaranm:

ciaranm:
Both real, both fixed here even though #813 makes them transient — the stack may land one PR at a time, which is exactly your point.

**The `.pbp` size.** 186 MB is right and 194 was mine. The document settles it two hundred lines above the sentence I wrote: line 121 records 193.6 MB → 186.5 MB as the *unanchored* → *anchored* pair from #800, so I quoted the pre-anchor figure. Nothing to re-measure, just a number to read off the file I was editing.

**The back-reference.** Correct that it pointed at nothing in the tree — it is the residue of a wiki-style link into my own notes that I stripped without giving it an antecedent. It now names #788's thread, which is where the argument actually lives.

On the merge-order note: agreed and deliberate. The analysis is what made #813 buildable in a day rather than a week, and it is worth having in the history for that reason even though most of the text does not survive into the final tree.

Issue #812, opened 2026-09-03
SubCircuit's SCC certificate throws UnimplementedException when the successor array contains a constant, which every mario instance's does


ciaranm:
Found while measuring the proof cost of the pruning rules (#788). **This predates that work** — plain `subcircuit::SCC` hits it — and it means the SCC arm's certificate has never been exercised on a real instance of the only family where the arm can run at all.

## What happens

```
fzn-glasgow: error: unexpected problem: unimplemented at
  gcs/innards/proofs/names_and_ids_tracker.cc:470 in
  NamesAndIDsTracker::need_constraint_saying_variable_takes_at_least_one_value(IntegerVariableID)
    ::<lambda(const ConstantIntegerVariableID&)>
```

`derive_unreachable`'s pigeonhole — "somebody is x's predecessor", the one cutting-planes step in that certificate — sums *this variable takes at least one value* over every successor. That function is `throw UnimplementedException{}` for a `ConstantIntegerVariableID`.

## Why every mario instance trips it

`2013/mario` posts `constraint succ[LuigiHouse] = MarioHouse`, and the flattener folds it into the array as a literal:

```json
"succ" : { "a": ["X_INTRODUCED_0_", 1, "X_INTRODUCED_2_", "X_INTRODUCED_3_", ...] }
```

That constant is *also* what gives mario an anchor (#800 finds it there), and the SCC arm does nothing without an anchor. **So the one thing that lets the arm run is the one thing that stops it writing a proof.**

## Repro, cut-down `2013/mario` (`cut_mario.py`, k houses)

| k | `scc` | `scc` + prune within |
|---|---|---|
| 6 | proof written, verified | **throws** |
| 8 | proof written, verified | **throws** |
| 10 | **throws** | **throws** |
| 12 | **throws** | **throws** |

Plain `scc` survives k=6 and k=8 only because its walk never fires there — nothing is unreachable, so the pigeonhole is never needed. The pruning rules shave under a hypothesis, so they reach it at every size.

The truncated `.pbp` left behind by the throw also makes VeriPB report a parse error at EOF rather than anything about the real problem, which is how this stayed hidden.

## Why the tests do not catch it

Every enumeration test builds its successors with `create_integer_variable`, so a constant only appears through `create_integer_variable_or_constant_with_view` in `subcircuit_test` — and on those instances the reachability walk does not fire, so the pigeonhole is never reached. Both halves have to line up, and no fixture lines them up.

[...]

PR #813, opened 2026-09-03
Implement Francis and Stuckey's scc pruning rules, expensively

ciaranm:
Stacked on #811, which is the analysis this implements. For #788.

`with_prune_root()` and `with_prune_within()`, both off by default. They cost between 1x and 95x the proof for at most a ten per cent reduction in search, and that is the point: they are *certifiable*, and a corpus of expensive-but-correct rules a checker can be pointed at is worth having on its own account.

## The rules are one operation

Assume an edge, run the reachability rules against the domains that leaves, and drop the value if a node that must be on the tour would be stranded in either direction. Prune root is that over the anchor's successor (F&S rule 3 / explanation 6), prune within over every other node's (rule 4 / explanation 7); they are split at the anchor so they do not overlap, and both are one function, `shave_by_reachability`. In this encoding the pair is **singleton arc consistency on the successor variables**.

Neither fires on the live domains — while a successor is free the anchor reaches everything any of its values reaches — which is exactly why they add strength and why they cost a walk per candidate. F&S read the condition off the depth-first traversal they already have, which is a cheap *sufficient* condition for the same thing, so asking the question directly of each value prunes at least as much and needs no traversal.

## The certificates needed nothing new

No new row family, no new proof-only variable, no new cutting-planes step — as #811 predicted and probed. Three small pieces of mechanism:

- **`AssumedEdge`**, and walks that take one, carried as data rather than read back out of the `State`: the walk's result and the justification that reproduces it must be over the same edge set, and a justification may read only the reason and the model.
- **The hypothesis as a guard** on every emitted row, so each says "*if* that edge is taken, then x at position t is off the tour" and the conclusion is "that edge is not taken". One extra literal per row. The alternative — a proof level deleted once the conclusion is drawn — saves those literals and costs the level discipline.
- **Candidates snapshotted** before the loop infers anything. Sound, not just convenient: a walk only ever uses the *absence* of a value to rule a neighbour out, so a domain that has since lost more values makes the reason a stronger premise.

## The cost, measured

One random instance per `n`, the same one across configs, complete enumeration with proofs on.

| n | recursions (scc / +root / +within) | `.pbp` scc | +prune_root | +prune_within | verify scc → +within |
|---|---|---|---|---|---|
| 6 | 16 / 16 / 17 | 11 KB | 1.0x | **13.1x** | 0.10 → 0.11 s |
| 7 | 119 / 119 / 117 | 139 KB | 1.5x | **15.4x** | 0.11 → 0.10 s |
| 8 | 492 / 492 / 478 | 212 KB | 1.0x | **51.6x** | 0.11 → 0.51 s |
| 9 | 1281 / 1281 / 1199 | 586 KB | 1.6x | **72.2x** | 0.21 → 2.11 s |
| 10 | 8057 / 8057 / 7461 | 4.3 MB | 1.0x | **94.9x** | 1.71 → 24.3 s |
| 11 | 25469 / 25420 / 22846 | 72.9 MB | 1.7x | **25.5x** | 12.8 → 108 s |
| 12 | 104134 / 104044 / — | 73.0 MB | 3.7x | *abandoned at 10.8 GB* | 93.1 s → — |

**Every proof verified**, including the 1.86 GB one at n=11. The multiplier is not monotone (95x then 25x) because the denominator moves too — `scc`'s own proof jumps 17x between those sizes — so the absolute figures are the honest ones.

Measured on constant-free instances, because the `mario` family cannot be used: **#812**, a constant in the successor array makes the certificate throw, predating this work (plain `scc` hits it at k≥10 on cut-down mario).

## Tests

- The anchored enumeration sweep runs prune root, and both rules together, at every size and anchor with proofs on — pinning that they cannot change the solution set and verifying their certificates over a whole enumeration. A rule that inferred nothing would pass that, so:
- `subcircuit_prune_root_test` and `subcircuit_prune_within_test` assert the **root state**. Building scenarios where the rule is the only route to its conclusion took several attempts: the plain backward walk gets there free if a component has one edge back to the anchor and all-different removes it, and `check` gets there free if any domain is a singleton, since one fixed successor cascades into a closed cycle. The prune-within instance was ultimately *generated* rather than built — hand-building kept producing unsatisfiable structures — so that fixture is deliberately asymmetric and untidy.
- Mutation probes caught by the tests: dropping the hypothesis guard, and replacing the induction with a plain `JustifyUsingRUP`.

790/790 tests pass, warning-free under GCC with `GCS_WERROR=ON`, clean under clang + sanitizers.

### DIFF
--- a/dev_docs/subcircuit-proof-logging.md
+++ b/dev_docs/subcircuit-proof-logging.md
@@ -435,7 +435,7 @@ The graph family shifts *parameters*, which costs nothing.
 
 `with_prune_root()` and `with_prune_within()`, both off by default. Together they are
 Francis and Stuckey's remaining `scc` rules, and they cost between 1x and 95x the
-proof for at most a ten per cent reduction in search. They are here because they are
+proof for about a ten per cent reduction in search. They are here because they are
 *certifiable*, and a corpus of expensive-but-correct rules that a checker can be
 pointed at is worth having on its own account.
 
@@ -491,8 +491,11 @@ small pieces of mechanism:
   uses the *absence* of a value to rule a neighbour out, so a domain that has since
   lost more values makes the reason a stronger premise, not a weaker one.
 
-Both mutation probes are caught by `subcircuit_prune_root_test`: dropping the guard,
-and replacing the induction with a plain `JustifyUsingRUP`.
+Both mutation probes are caught: dropping the guard, and replacing the induction with a
+plain `JustifyUsingRUP`. The guard one is caught three times over --- by
+`subcircuit_prune_root_test`, by `subcircuit_prune_within_test`, and by the anchored
+enumeration sweep in `subcircuit_test.cc`, which reports the signature worth recognising:
+the solution set is still right and VeriPB refuses the certificate.
 
 ### The cost, measured
 
@@ -523,8 +526,9 @@ this machine means it does not fit: the scratchpad is a 16 GB tmpfs and therefor
 *shares the guest's RAM*. Complete enumeration with prune within has a practical
 ceiling somewhere around `n = 11`.
 
-And what it buys: **at most ten per cent fewer search nodes** (8057 → 7461, 25469 →
-22846, 1281 → 1199), and for prune root essentially nothing at all (104134 → 104044).
+And what it buys: **about ten per cent fewer search nodes** (8057 → 7461 is 7.4%,
+25469 → 22846 is 10.3%, 1281 → 1199 is 6.4%), and for prune root essentially nothing at
+all (104134 → 104044).
 Which is consistent with everything else measured on this arm --- the node reduction
 where it is measurable at all is 1.09x to 1.38x --- and is the argument for treating
 the proof-size constant as the thing to attack rather than the propagator.
--- a/dev_docs/subcircuit-proof-logging.md
+++ b/dev_docs/subcircuit-proof-logging.md
@@ -431,99 +431,128 @@ The same shape is in four more redefinitions — the three `bin_packing` ones an
 `fzn_regular` — and is [issue #803](https://github.com/ciaranm/glasgow-constraint-solver/issues/803).
 The graph family shifts *parameters*, which costs nothing.
 
-## Not implemented: F&S's pruning rules, and what certifying them would take
+## The pruning rules: implemented, certified, and expensive
 
-F&S's remaining `scc` rules — prune root, prune skip, fix required edges and prune
-within — with the evidence-node guard each one needs for `subcircuit`. Their
-explanation clauses are §5.3.2 of the paper, numbered 1 to 7 there and referred to
-by those numbers below.
+`with_prune_root()` and `with_prune_within()`, both off by default. Together they are
+Francis and Stuckey's remaining `scc` rules, and they cost between 1x and 95x the
+proof for at most a ten per cent reduction in search. They are here because they are
+*certifiable*, and a corpus of expensive-but-correct rules that a checker can be
+pointed at is worth having on its own account.
 
-They are stated over the **subtree structure of a depth-first traversal**: which
-subtree was visited before which, and the lowpoint of a node's first child. Our
-encoding knows nothing about that. `pos` is a *tour-order* labelling, not a DFS
-one, and it is pinned at exactly one place, `pos[anchor] = 0`.
+### Why the encoding takes them at all
 
-That sounds like an obstacle and mostly is not, because the DFS tree turns out to
-be how the propagator *finds* these inferences rather than why they hold. Working
-each clause back to what makes it true:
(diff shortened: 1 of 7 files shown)


ciaranm:
Pre-merge review, middle of the #811 → #813 → #814 stack. CI is 10/10 green including all three Windows lanes; trial-merged with #806 and #810 onto current `main` (this stack sits 10 commits behind it), no conflicts, **0 GCC 15.2 warnings**, **794/794 ctest**.

The certificate work is the interesting half and it stands up — see the mutation result below. One real finding, on the half that is not about proofs at all.

## 1. This makes the DEFAULT `scc` path ~15% slower, with both flags off

Plumbing the assumed edge through the walks replaced a lazy iteration with a materialised `vector`:

```diff
-   for (const auto & v : state.each_value_immutable(succ[i])) {        // no allocation
-       auto j = static_cast<size_t>(v.raw_value);
+   for (const auto & v : walkable_values(succ, state, assumed, i)) {   // heap vector per call
+       auto j = static_cast<size_t>(v);
```

In `reachable_layers` that is inside the `t` × `i` double loop, so it is `O(n²)` allocations per walk — and `reachable_layers` is on the plain `subcircuit::SCC` path, which pays it whether or not either new flag is set.

A/B, interleaved, patching only that one loop back and rebuilding just the library. n = 18, proofs off, workload lifted from `subcircuit_scc_test.cc`'s two-halves scenario and scaled up:

| | elapsed (ms) |
|---|---|
| as pushed | 2436, 2447, 2450, 2454 |
| that one loop reverted | 2131, 2139, 2123, 2121 |

**1.15x, non-overlapping ranges**, and `recursions` (189,824) and `propagations` (898,290) are *identical* in both arms — so this is pure per-call cost, not a change in what the rule infers.

The same helper in `reaches_anchor` is only `O(n)` per walk and costs nothing measurable (if anything a small win — a flat vector is cheaper to re-walk than the domain range), so the whole difference is that one loop and the fix is local: branch on `assumed` once outside the loop rather than per iteration.

Worth weighing against this document's own claim that "the walk alone is already 87--98% of all propagation time" on the challenge families. Nothing in CI times anything, so no lane could have said this.

## 2. The guard mutation is caught by three lanes, not one

The doc says:

> Both mutation probes are caught by `subcircuit_prune_root_test`: dropping the guard, and replacing the induction with a plain `JustifyUsingRUP`.

I made `guarded()` a no-op in both `derive_unreachable` and `derive_cannot_reach_anchor`:

```
345 - subcircuit_prune_root      (Failed)
346 - subcircuit_prune_within    (Failed)
349 - subcircuit_constraint      (Failed)
```

— and the signature is exactly right: `expected eight solutions from each, got 8 and -1`, i.e. the solution set is still correct and VeriPB refuses the certificate ("The constraint is not implied by reverse unit propagation"). So the anchored enumeration sweep in `subcircuit_test.cc` catches it too, which the doc undersells. Coverage is stronger than advertised; only the sentence needs widening.

## 3. Minor: "at most a ten per cent reduction"

Said twice (header paragraph and the closing "what it buys"). The n=11 row is 25469 → 22846, which is 10.3%. Trivially over, but it is the claim the section is built around, so "about ten per cent" would be safer.

## AI declaration

The description declares AI assistance but none of the three commits carries the `Co-Authored-By:` trailer, which CONTRIBUTING.md asks for as the other half; 60 of the last 60 non-merge commits on `main` have it. Same on #806, #810, #811 and #814 — one pass across all five. These were authored 14:36–15:08Z, after the same point went on #799 at 10:38Z.

ciaranm:

ciaranm:
## The 15% default-path regression is real, and fixed

Reproduced your workload exactly before touching anything — same n=18 scenario, same 189,824 recursions and 898,290 propagations, 2417–2434 ms. So we were measuring the same thing.

`walkable_values` is now a template visitor rather than a returned vector, so the allocation leaves the `t`×`i` loop entirely:

| | elapsed (ms) |
|---|---|
| as pushed | 2417, 2423, 2426, 2434 |
| fixed | 2094, 2101, 2101, 2120 |
| #811 tip, before the rules existed | 2070, 2071, 2137 |

Non-overlapping against "as pushed", and back inside the pre-#813 baseline rather than merely near it — `reaches_anchor` got the same treatment, which is where the extra few ms come from. `recursions` and `propagations` identical in all three arms.

A template parameter and not a `std::function`, since the point is to keep an indirect call out of that loop.

## The other two

Coverage claim widened: you are right that the guard mutation is caught by three lanes, and the anchored sweep catching it is worth saying explicitly, along with the signature — solution set still right, VeriPB refusing the certificate.

"At most ten per cent" → "about ten per cent", with each figure given (7.4%, 10.3%, 6.4%). It was wrong by 0.3 percentage points on the row the section is built around, which is exactly the sort of claim not worth being loose about.

PR #814, opened 2026-09-03
Count the pigeonhole over the non-constant successors

### LINKED ISSUE #812: SubCircuit's SCC certificate throws UnimplementedException when the successor array contains a constant, which every mario instance's does

ciaranm:
Closes #812. Stacked on #813.

## The bug

A constant successor has no encoded atoms — no "takes at least one value" row to cite, and no equality literal a `pol` can name an operand for. `derive_unreachable`'s pigeonhole summed that row over *every* successor, so it threw `UnimplementedException` on any array containing one.

**Every `mario` instance contains one**: its `succ[LuigiHouse] = MarioHouse` pin folds into the array as a literal. That pin is also what gives mario its anchor — so **the one thing that let the SCC arm run was the one thing that stopped it writing a proof**. Plain `scc` threw on cut-down mario at k ≥ 10, surviving k = 6 and 8 only because its walk never fires there.

## The fix

Both halves of the pigeonhole now range over the non-constant successors only, with the constants accounted for separately. With `m` non-constant successors out of `n`, and `v` not pinned by any constant:

```
    sum over the m of "takes at least one value"        =  m
    for each w != v not pinned:  at most one takes w    <= 1   (n - pinned - 1 of them)
    for each pinned c:           none of the m takes c  <= 0
    ------------------------------------------------------------------------------
    so  sum over the m of [succ_i = v]  >=  m - (n - pinned - 1)  =  1
```

since `m = n - pinned`. "None of them takes `c`" is one RUP row: assuming `succ[i] = c` forces both halves of that pair's all-different rows, whose guards are each other's negation, so unit propagation closes it. The at-most-one had to stop including constants too, for a different reason — `recover_am1_from_pairs` needs a `pol` operand naming each member.

When `v` *is* pinned there is nothing to count, and the conclusion differs: `x` at position `t` would put the pinned node at `t-1`, where its own `E(t-1, ·)` says it is a self loop, which its pin contradicts. So the row is the stronger `pos[x] != t`. Emitting the weaker `E(t, x)` there also verifies — a contradictory antecedent implies anything — so that choice is not load-bearing, and the comment says so rather than implying otherwise.

## It now works on the family it is for

Cut-down `2013/mario`, proofs on, every size tried:

| k | `scc` | `scc` + both pruning rules |
|---|---|---|
| 6 | 130 KB, VERIFIED | 199 KB, VERIFIED |
| 8 | 633 KB, VERIFIED | 3.1 MB, VERIFIED |
| 10 | 6.1 MB, VERIFIED | 43.3 MB, VERIFIED |
| 12 | 26.8 MB, VERIFIED | 545 MB, VERIFIED |

All eight previously either threw or were unreachable. This is the first time the SCC arm has been proof-checked on an instance of the family it can actually run on.

## What the tests can and cannot pin

This took several attempts and is worth recording. The anchored enumeration sweep now runs with a constant in the array at every size and anchor with proofs — 72 runs — and that covers the *branch structure*. It does **not** cover the arithmetic: at n ≤ 6 unit propagation reaches `E(t, x)` from the candidate rows and the reason alone, so the pigeonhole line is not load-bearing at those sizes and deleting the new at-most-zero row still verifies. Two hand-built fixtures passed the mutation before I established that, and complete enumeration cannot be run at a size where it bites.

So `subcircuit_constant_test` carries a second, **generated** eight-node scenario, searched for by exactly the property that its proof fails under that mutation — which is why that fixture is arbitrary and asymmetric. The same mutation is rejected on real cut-down mario at k = 10.

791/791 tests pass, warning-free under GCC with `GCS_WERROR=ON`, clean under clang + sanitizers.

### DIFF
--- a/dev_docs/subcircuit-proof-logging.md
+++ b/dev_docs/subcircuit-proof-logging.md
@@ -549,14 +549,51 @@ the proof-size constant as the thing to attack rather than the propagator.
    anchor, so an arbitrary component would need layers indexed relative to the unknown
    `pos[a]`. It is the one place the "arbitrary root" position-offset problem is real
    --- and unreachable for us anyway, since the arm does nothing without an anchor.
-4. **A constant in the successor array breaks the certificate**, which is
-   [issue #812](https://github.com/ciaranm/glasgow-constraint-solver/issues/812) and
-   not these rules' fault: the pigeonhole sums "this variable takes at least one value"
-   over every successor and that is `UnimplementedException` for a constant. Every
-   `mario` instance has one, because its `succ[LuigiHouse] = MarioHouse` pin folds into
-   the array as a literal --- so the one thing that gives the arm an anchor is the one
-   thing that stops it writing a proof, and the cost table above is measured on
-   constant-free instances for that reason.
+4. **A constant in the successor array used to break the certificate**, which was
+   [issue #812](https://github.com/ciaranm/glasgow-constraint-solver/issues/812) --- not
+   these rules' fault, since plain `scc` hit it too, and the cost table above is measured
+   on constant-free instances for that reason. Fixed; see below.
+
+### Constants in the successor array (#812)
+
+A constant successor has no encoded atoms: no "takes at least one value" row to cite, and
+no equality literal a `pol` can name an operand for. The pigeonhole summed that row over
+*every* successor, so it threw an `UnimplementedException` on any array containing one ---
+and **every `mario` instance contains one**, its `succ[LuigiHouse] = MarioHouse` pin having
+folded into the array as a literal. That pin is also what gives mario its anchor, so the
+one thing that let this arm run was the one thing that stopped it writing a proof.
+
+Both halves of the pigeonhole now range over the non-constant successors only, and the
+constants are accounted for separately. Writing `m` for the number of non-constant
+successors out of `n`, and taking `v` not pinned by any constant:
+
+```
+    sum over the m of "takes at least one value"        =  m
+    for each w != v not pinned:  at most one takes w    <= 1   (n - pinned - 1 of them)
+    for each pinned c:           none of the m takes c  <= 0
+    ------------------------------------------------------------------------------
+    so  sum over the m of [succ_i = v]  >=  m - (n - pinned - 1)  =  1
+```
+
+since `m = n - pinned`. "None of them takes `c`" is one RUP row: assuming `succ[i] = c`
+forces both halves of that pair's all-different rows, whose guards are each other's
+negation, so unit propagation closes it.
+
+When `v` *is* pinned there is nothing to count --- that constant is the predecessor --- and
+the conclusion is different too: `x` sitting at position `t` would put the pinned node at
+`t-1`, where its own `E(t-1, .)` says it is a self loop, which its pin contradicts. So the
+row emitted is the stronger `pos[x] != t` rather than `E(t, x)`.
+
+**What the tests could and could not pin.** The anchored enumeration sweep now runs with a
+constant in the array, at every size and anchor, with proofs --- 72 runs. That covers the
+branch structure, but **not the arithmetic**: at `n <= 6` unit propagation reaches
+`E(t, x)` from the candidate rows and the reason alone, so the pigeonhole line is not
+load-bearing at those sizes and deleting the at-most-zero row still verifies. Complete
+enumeration cannot be run at a size where it does bite. So
+`subcircuit_constant_test` carries a second, *generated* eight-node scenario, searched for
+by the property that its proof fails if the pinned value is counted with an ordinary
+at-most-one --- which is why that fixture is arbitrary and asymmetric. On real cut-down
+mario the same mutation is rejected at k = 10.
 
 ### A sign error in the paper, for whoever implements from it
 
--- a/dev_docs/subcircuit-proof-logging.md
+++ b/dev_docs/subcircuit-proof-logging.md
@@ -585,15 +585,17 @@ the conclusion is different too: `x` sitting at position `t` would put the pinne
 row emitted is the stronger `pos[x] != t` rather than `E(t, x)`.
 
 **What the tests could and could not pin.** The anchored enumeration sweep now runs with a
-constant in the array, at every size and anchor, with proofs --- 72 runs. That covers the
-branch structure, but **not the arithmetic**: at `n <= 6` unit propagation reaches
+constant in the array, at every size and anchor --- 48 runs, 24 of them with proofs. (72 is
+the figure a careless `grep -c` gives, the proofs-on runs each printing a `$ veripb …` echo
+line as well.) That covers the branch structure, but **not the arithmetic**: at `n <= 6` unit propagation reaches
 `E(t, x)` from the candidate rows and the reason alone, so the pigeonhole line is not
 load-bearing at those sizes and deleting the at-most-zero row still verifies. Complete
 enumeration cannot be run at a size where it does bite. So
 `subcircuit_constant_test` carries a second, *generated* eight-node scenario, searched for
-by the property that its proof fails if the pinned value is counted with an ordinary
-at-most-one --- which is why that fixture is arbitrary and asymmetric. On real cut-down
-mario the same mutation is rejected at k = 10.
+by the property that its proof fails under exactly one mutation: counting a pinned value
+with the ordinary at-most-one instead of the at-most-zero row, which is what deleting
+`need_no_variable_takes` amounts to. That is why the fixture is arbitrary and asymmetric.
+On real cut-down mario the same mutation is rejected at k = 10.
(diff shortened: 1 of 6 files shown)


ciaranm:
Pre-merge review, top of the #811 → #813 → #814 stack. CI is 10/10 green including all three Windows lanes; trial-merged with #806 and #810 onto current `main` (this stack sits 10 commits behind it), no conflicts, **0 GCC 15.2 warnings**, **794/794 ctest**, and all four new lanes confirmed registered and running.

The pigeonhole rework reads correctly and the caller obligation is honoured — `need_value_at_least_one` has exactly one call site and it is behind the new `consts.value_is_pinned[x]` guard. Three things.

## 1. The new code landed inside an existing doc comment

`struct ConstantSuccessors` and `survey_constants()` were inserted between the second and third lines of `need_value_at_most_one`'s comment, so a sentence is now split across ~30 lines of code. As the file reads now:

```cpp
    // This was a hand-rolled staircase until it turned out to be the fifth copy of one
    // derivation, four of which predate the shared version. Not for the reason the shared
    // Which successors are constants. A constant successor has no encoded atoms at all ---
    ...
    struct ConstantSuccessors final { ... };
    auto survey_constants(...) -> ConstantSuccessors { ... }

    // version's own documentation gives, though, which is worth recording because the next
    // caller to switch over will read it: recover_am1_from_pairs pins its result with an
```

So two things are broken at once: the "At most one of the successors takes value v / recover_am1_from_pairs's job" intro is stranded above `ConstantSuccessors` and trails off mid-sentence, and `need_value_at_most_one` now carries an orphan block that *opens* mid-sentence with "version's own documentation gives, though…". Comment-only fix — move the two halves back together above `need_value_at_most_one` and put `survey_constants`'s own comment above `survey_constants`. clang-format cannot see this in either direction, which is why CI is happy.

## 2. "72 runs" is 48 runs (24 of them with proofs)

> The anchored enumeration sweep now runs with a constant in the array, at every size and anchor, with proofs --- 72 runs.

```
$ ./build/subcircuit_test 2>&1 | grep -c 'scc-const'
72
$ ./build/subcircuit_test 2>&1 | grep -o 'subcircuit/anchored/scc-const[a-z-]*' | sort | uniq -c
     24 subcircuit/anchored/scc-const
     24 subcircuit/anchored/scc-const-pruning-rules
$ ./build/subcircuit_test 2>&1 | grep 'scc-const' | grep -vc '^subcircuit/anchored/'
24
```

72 is the first of those, and the extra 24 lines are the `$ veripb …` echoes the proofs-on runs print. The loop bounds agree with 48: `for proofs {false,true}` × `for n {3,4,5}` × `for anchor {0,n-1}` × 2 constants × 2 calls, with neither `continue` firing for n ≥ 2. Both 48 and "24 with proofs" are defensible figures — but the sentence says "with proofs", so 24 is probably the one meant.

## 3. Minor: where the distinctness assumption lives

The derivation's last step is `since m = n - pinned`. That identity holds only if the constant successors are pairwise distinct — `pinned` counts distinct *values* (`value_is_pinned` is a `vector<bool>` keyed by value) while `m = n − (number of constant successors)`. Two constants pinned to the same value would make `m > n − pinned` and the `>= 1` conclusion would not follow, so the `pol` would emit something weaker and a later RUP would fail.

It cannot arise in practice, because all-different contradicts at the root on such a model and the walk never runs — which is exactly the reasoning already written out for the neighbouring out-of-range case ("such a constant makes the model unsatisfiable at search start and this walk never runs. Guard anyway rather than index out of range"). Worth one clause saying all-different is what supplies distinctness, so the identity does not read as unconditional.

Also, a small mismatch between the description and the doc on which mutation the generated fixture pins: the description says "deleting the new at-most-zero row", the doc says "if the pinned value is counted with an ordinary at-most-one". Probably the same intent, but they are different edits.

## AI declaration

The description declares AI assistance but the commit carries no `Co-Authored-By:` trailer, which CONTRIBUTING.md asks for as the other half; 60 of the last 60 non-merge commits on `main` have it. Same on #806, #810, #811 and #813 — one pass across all five. This one was authored at 15:37Z, after the same point went on #799 at 10:38Z.

ciaranm:

ciaranm:
All three good, and the first one is embarrassing in a useful way.

**The comment surgery.** Exactly as described — the insertion landed between lines two and three of `need_value_at_most_one`'s comment, leaving a stranded intro trailing off mid-sentence and an orphan block opening at "version's own documentation gives, though…". Both halves are back together above the function with the constants note folded in where the intro used to end, and `survey_constants` carries its own comment above itself. Your observation that clang-format cannot see this in either direction is the part worth remembering: a green formatting lane says nothing about whether a comment still parses as English.

**48, not 72.** A careless `grep -c` counting the `$ veripb …` echo lines the proofs-on runs print. It is 48 runs, 24 with proofs, and the loop bounds agree. Fixed, with a parenthetical naming the trap so the number does not get "recounted" the same way later.

**The distinctness assumption.** Right that `m = n - pinned` is not unconditional — `pinned` counts distinct values, `m` counts successors. It now says so and names all-different as what supplies it, with the same reasoning the out-of-range constant already gets a few lines up: two constants sharing a value contradict at the root, so the walk never runs on such a model.

And the description/doc mismatch on the mutation: same edit, two vocabularies. The doc now gives both — counting a pinned value with the ordinary at-most-one *is* deleting `need_no_variable_takes`.

Issue #825, opened 2026-09-04
Recovering LessThanEqual from two-term int_lin_le is blocked by find_makespan_links, not by difference logic


ciaranm:
Follow-up to #824, which recovers `Equals` / `NotEquals` from the two-term `int_lin_eq` / `int_lin_ne` MiniZinc hands us. The third shape is the most common of the three — two-term unit-coefficient `int_lin_le` is **37,850 constraints in 30 of 75 sampled challenge instances**, against 25 `int_le` — but recovering it as `LessThanEqual` is blocked, and not where I first guessed.

## Difference logic is fine

`DifferenceLogic::run` already enumerates `ReifiedCompareLessThanOrMaybeEqual` donors alongside `ReifiedLinearInequality` ones, and `lift_row` deviews both operands, so `x <= y + d` lifts to exactly the edge the two-term linear form did. Measured with the rewrite in place:

| | `EdgesLifted` | `ComparisonEdgesLifted` | `Nodes` | `SkippedNotTwoTerms` | `SkippedCoefficients` |
|---|---|---|---|---|---|
| before | 4 | 0 | 4 | 1 | 1 |
| after | 4 | **4** | 4 | 1 | 1 |

on `minizinc/tests/differencelogic.mzn`, and 6/6 on a purpose-built chain. Only the family classification moves. (`minizinc-differencelogic` pins `differenceLogicComparisonEdgesLifted=0`, so it fails on the rewrite and would need that one expectation updated — the counter is doing its job.)

Non-difference rows are still declined, just for a different recorded reason: `x + y <= d` becomes `LessThanEqual{x, -y + d}`, which the presolver rejects as a negated view rather than for its coefficients. Same decision, and arguably the more accurate label.

## `find_makespan_links` is not fine

`gcs/presolvers/innards/makespan_links.cc` enumerates **only** `ReifiedLinearInequality`, and wants exactly the shape the rewrite consumes: two terms, coefficients +1 and -1, `MustHold`, i.e. `start - makespan <= -bound`. With the rewrite it finds nothing, so `InferredCumulative` and `InferredDisjunctive` lose every makespan link and the certified makespan bounds of #672/#673 quietly stop being derivable.

[...]

PR #826, opened 2026-09-04
Teach find_makespan_links the comparison family

ciaranm:
Half of #825. A makespan row can be spelled two ways and this only knew one.

`makespan - start >= length` as a two-term linear row it found. `start + length <= makespan`, and the `start <= makespan - length` the FlatZinc reader recovers from a two-term `int_lin_le` (#824), it did not — so on such a model `InferredCumulative` and `InferredDisjunctive` got **no makespan links at all**, every task's energy went uncounted, and the certified bound of #672/#673 was quietly not derivable. Nothing said so: neither presolver is installed by `fzn_glasgow.cc`, and every test that reaches this posts its donors through the C++ API, where the class is whatever the test named.

## The scan

`ReifiedCompareLessThanOrMaybeEqual` as well, `MustHold` rows only, as the linear branch does. `deview_difference_operand` — the helper the difference-logic presolver already uses on its own donors — reduces each operand to variable-plus-offset and declines a negated view, which is not this shape at all. `left <= right` states `left - right <= 0` (`<= -1` for the strict spelling), so with `left = start + cl` and `right = makespan + cr` the length wanted out of `makespan - start >= length` is `cl - cr + (or_equal ? 0 : 1)`. Degenerate and aliased rows are declined exactly as the linear branch declines them, and the strongest row still wins where a variable has several — now across both families.

## The proof side, which is why this was left undone

The header said a comparison spelling's "bits would not cancel against the plain variable's in the `pol`", and that was right. A view has its own `BinEnc` in the OPB — checked, not assumed:

```
@c[view_of_start0_plus_2][viewle] ... -1 p[0_view_of_start0_plus_2][b0] ... 1 i[start0][b0] ... >= -2;
@c[_1] -1 p[0_view_of_start0_plus_2][b0] -2 p[...][b1] -4 p[...][b2] 1 i[makespan][b0] ... >= 0;
```

so the row is over `p[0_view_of_start0_plus_2][b*]`, not `i[start0][b*]`, and the confine `pol` in `makespan_energy` adds that row to the start's and the makespan's own order-literal definitions expecting both to cancel.

The machinery for fixing that already exists: `ProofModel::add_labelled_constraint` derives a deview-form for every row with view terms, and `PolBuilder` has a deview mode that substitutes it in. One call, a no-op for rows without view terms, so the linear family's proofs are byte-for-byte what they were.

## Fixtures

The makespan demo re-run under each comparison spelling — one with the view over the start, one over the makespan. Both reach the same bound of **six**, match brute force, agree with the proofs-off run, and verify: 11 proofs VERIFIED, and the two deliberate mutations (`MakespanClaimHigher`, `MakespanOmitRow`) still rejected. Drop the comparison loop and the test fails with

```
makespan demo, offset_on_start: the derived constraint inferred nothing, not the six its
tasks' energy supports --- the model's rows were not recognised as makespan links
```

which is exactly the silent failure this is meant to prevent.

**Deleting the deview call fails nothing, and the comment now says so** rather than implying a check that does not exist. The framework pre-derives the atom-level `[V >= v] <=> [X >= k]` clauses for every view atom a proof uses, so the lemma's own RUP crosses the V/X boundary in one step and reaches the deadline literals without the confine clause; I confirmed the confine `pol` does fire on a view row in these fixtures, and that VeriPB accepts the non-cancelling line as a valid consequence that nothing then needs. The clause is what the `pol` claims to derive, and without deviewing it derives a valid line that is not that clause — so it stays, documented as correct rather than as verified. If someone wants it load-bearing, the fixture needed is one where the lemma asks for a deadline literal the atom-level links cannot reach.

Full `ctest` 794/794; GCC `-DGCS_WERROR=ON` and clang 21 clean.

Leaves the other half of #825 open: the `int_lin_le` recovery itself, which measured neutral, and a lane that posts a donor scan's donors through the FlatZinc reader so a class change fails loudly rather than silently.

### DIFF
--- a/gcs/constraints/cumulative/derived_cumulative_test.cc
+++ b/gcs/constraints/cumulative/derived_cumulative_test.cc
@@ -1,3 +1,4 @@
+#include <gcs/constraints/comparison.hh>
 #include <gcs/constraints/cumulative.hh>
 #include <gcs/constraints/cumulative/derived_cumulative.hh>
 #include <gcs/constraints/innards/constraints_test_utils.hh>
@@ -359,6 +360,24 @@ namespace
         }
     };
 
+    /// How a model spells the rows that say every task finishes by the makespan.
+    /// find_makespan_links has to recognise all three, and the energy
+    /// derivation cites whichever row it found --- which for the two comparison
+    /// spellings is stated over an offset view, whose BinEnc the confine `pol`
+    /// deviews before it can cancel. So each spelling is its own path through
+    /// the proof, not a restatement of the same one.
+    enum class MakespanRows
+    {
+        /// `LinearGreaterThanEqual{makespan - start, length}`, which is what a
+        /// scheduling model writes and what MiniZinc's int_lin_le flattens to.
+        Linear,
+        /// `LessThanEqual{start + length, makespan}`: the view is over the start.
+        OffsetOnStart,
+        /// `LessThanEqual{start, makespan - length}`: the view is over the
+        /// makespan. This is the shape the FlatZinc reader recovers.
+        OffsetOnMakespan
+    };
+
     struct Instance
     {
         vector<pair<int, int>> start_ranges;
@@ -370,6 +389,8 @@ namespace
         /// --- which is the entailment the makespan bound's derivation rests
         /// on, and the only thing that makes those demos legal.
         optional<int> makespan_horizon = nullopt;
+        /// Which spelling those rows get. Ignored without a horizon.
+        MakespanRows makespan_rows = MakespanRows::Linear;
         /// When non-empty, one per task: the domain that task's presence
         /// variable is declared over, and the constraint is posted with the
         /// optional-task constructor. A solution then carries every presence
@@ -438,7 +459,12 @@ namespace
         if (inst.makespan_horizon) {
             makespan = p.create_integer_variable(0_i, Integer{*inst.makespan_horizon}, "makespan");
             for (size_t i = 0; i < starts.size(); ++i)
-                p.post(LinearGreaterThanEqual{WeightedSum{} + 1_i * *makespan + -1_i * starts[i], lengths[i]});
+                switch (inst.makespan_rows) {
+                    using enum MakespanRows;
+                case Linear: p.post(LinearGreaterThanEqual{WeightedSum{} + 1_i * *makespan + -1_i * starts[i], lengths[i]}); break;
+                case OffsetOnStart: p.post(LessThanEqual{starts[i] + lengths[i], *makespan}); break;
+                case OffsetOnMakespan: p.post(LessThanEqual{starts[i], *makespan - lengths[i]}); break;
+                }
         }
 
         vector<IntegerVariableID> presences;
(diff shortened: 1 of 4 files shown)


ciaranm:
Filed the two things this PR leaves deliberately undone, so they don't only live here:

- **#828** — the linear branch still declines a view on the task side, a guard whose stated reason ("a view's bits are not those") is exactly what the deview mode added here undoes. The comparison branch relies on deviewing and accepts views on either side, so the two branches now disagree about the same question. Not done here because relaxing it finds *more* donors, which moves the bound and the search — a strength change, and mixing it in would have spoiled "the same bound of six on both spellings", which is the assertion that says this PR works.
- **#829** — no fixture makes the deview call load-bearing, with the three checks that establish that (the pol does fire on a view row; removing deview still verifies; VeriPB accepts the non-cancelling line as a valid consequence nothing needs), why (the atom-level `[V >= v] <=> [X >= k]` clauses let RUP cross the boundary unaided), and the two instance shapes that failed to make it bite. It also names the general hazard: a `pol` whose terms don't cancel is silent, and a `PolBuilder` check on the emitted line's shape would catch that class at the point of the error.

ciaranm:
Pre-merge check finding: the two comments this PR adds about the deview mode disagree with each other, and the one in `derived_cumulative_test.cc` is the wrong one.

`makespan_energy.cc` says:

> Removing this does not currently fail any fixture, and that is worth knowing rather than hiding

`derived_cumulative_test.cc`, on the same mechanism, says:

> The first failure shows up as a wrong bound here, the second as a proof VeriPB rejects.

I tested both halves by mutation, against the seven open PRs trial-merged onto `b11586eb` (clean merge, 0 warnings, 801/801).

**The deview half is not pinned.** Removing `confine.enable_deview_mode(logger.names_and_ids_tracker());` leaves both new lanes passing, with VeriPB accepting both proofs:

```
1/2 Test #10: inferred_cumulative_presolver ....   Passed
2/2 Test #86: derived_cumulative ...............   Passed
```

The new cases do run — `offset_on_start` and `offset_on_makespan` each invoke veripb (22 invocations in `derived_cumulative_test` overall), so this is not a skipped fixture. `makespan_energy.cc` is right, including about why: the framework pre-derives the atom-level `[V >= v] <=> [X >= k]` clauses, so the lemma's own RUP crosses the V/X boundary without the clause.

**The recognition half is pinned, and strongly.** Disabling the new `ReifiedCompareLessThanOrMaybeEqual` loop in `find_makespan_links` fails both lanes with exactly the predicted messages:

```
inferred cumulative test failure: certified bound, comparison rows: derived a makespan
bound of 0, not the eleven the two-term linear spelling derives --- the model's rows
were not recognised as makespan links

derived cumulative test failure: makespan demo, offset_on_start: the derived constraint
inferred nothing, not the six its tasks' energy supports --- the model's rows were not
recognised as makespan links
```

So this is a wrong description, not wrong behaviour — the `enable_deview_mode` call is correct and the argument for keeping it stands. Only the test comment over-claims what its lanes catch, which matters because the note in `makespan_energy.cc` says the first lemma that needs the clause rather than the atoms "would fail somewhere far from the cause": a reader trusting the test comment would think that call is already protected.

Suggested precise form for the test comment: the lanes pin `find_makespan_links` recognising the shape (miss it and the bound is wrong), but not the deview mode, which no current fixture distinguishes.

Nothing else came up on this PR. Note for merge order: #827 contains an explicit merge of this branch (`c2a2b55c`) as well as #824, though it declares only #824 as its base — so this needs to land before #827.

ciaranm:
You're right, and the test comment was the wrong one — fixed in 5e4ac368.

It claimed the missing deview would show up "as a proof VeriPB rejects", which is exactly what the note at the call site says, with the mutation run to confirm it, does *not* happen. Two comments about the same mechanism, contradicting each other, and the one a reader would reach for first was the false one — worse than either being absent, given that note says the first lemma needing the clause rather than the atoms "would fail somewhere far from the cause".

The comment now says what the lanes pin (the recognition half: miss it and the bound is wrong), names the half they don't (the deview mode, which no current fixture distinguishes, and why — the atom-level `[V >= v] <=> [X >= k]` clauses let the lemma's RUP reach the deadline literals unaided), and points at `makespan_energy.cc` rather than restating it.

Your two mutation results match mine exactly, including the bound of 0 and the six-versus-nothing message, so I've left the behaviour and the `enable_deview_mode` call alone. #829 tracks what a fixture that does distinguish it would need, and the more general point that a `pol` whose terms don't cancel is silent.

The `inferred_cumulative_test` comment #827 adds only claims the recognition half ("post a class it does not know and nothing errors — every task simply loses its deadline and the bound below goes to zero"), so it needed no change; I re-read it against your finding rather than assuming.

Merge order noted and agreed — #827's merge of this branch is why it has to. I've merged the correction into #827 as well (6cc4abad) so it isn't carrying the stale text: 798/798 there, 794/794 here.

PR #827, opened 2026-09-04
Recover x - y <= d from the two-term linear form too

### LINKED ISSUE #825: Recovering LessThanEqual from two-term int_lin_le is blocked by find_makespan_links, not by difference logic

ciaranm:
Closes #825. Stacked on #824 (whose `recover_two_variable_form` this sits beside) and on **#826, which it needs**: without #826's comparison-family scan this would silently stop `find_makespan_links` finding anything. `makespan-links-comparison-family` is merged in so the tests here are honest; review base is `fzn-two-term-linear-rewrite`.

The third and most common of the shapes MiniZinc hands us as a two-term unit-coefficient linear constraint: **37,850 across 30 of 75 sampled challenge instances**, against 25 `int_le`.

## Normalisation

An inequality does not survive being divided by a negative, so unlike the equality case the `a = -1` patterns hand the operands over the other way round rather than reaching for `GreaterThanEqual`. All four sign patterns are in the test model, taken from what the flattener actually produces rather than from what the source looks like:

| source | coefficients |
|---|---|
| `x - y <= d` | `[1, -1]` |
| `x - y >= d` | `[-1, 1]` |
| `-x - y <= d`, `x + y >= d` | `[-1, -1]` |

Without a negative-first-coefficient row, a normalisation that never swaps passes the test — which is how the first draft of this model was wrong, and the sabotage caught it.

## Difference logic keeps every edge

It enumerates both donor families and deviews each operand, so `x <= y + d` lifts to the edge the linear row lifted. On `minizinc/tests/differencelogic.mzn`: 4 edges and 4 nodes either way, both non-difference rows still declined — only the family classification moves, which is what `differenceLogicComparisonEdgesLifted` counts and what that lane's one flipped expectation now says.

On real scheduling models, whose precedences are exactly this shape:

| instance | edges lifted | dl nodes | search nodes |
|---|---|---|---|
| `2016/rcpsp-wet` | 58 → 58 | 32 → 32 | 30 → 30 |
| `2019/rcpsp-wet-diverse` | 174 → 174 | 96 → 96 | 92,703 → 92,703 |

## It measures neutral

`LessThanEqual` and a two-term `LinearLessThanEqual` are both bounds reasoning over the same two variables. First-solution runs on the twelve `le`-heaviest instances, eight of which found one inside 45 s: **node counts identical on all eight**, answers identical, times 0.50→0.50, 0.40→0.40, 0.40→0.30, 2.70→2.60, 21.92→22.62, 0.20→0.10 (`2022/roster-sickness`, 33,761 of these constraints and four nodes), 0.10→0.10, 0.10→0.10. The four that timed out are not comparable and are not counted.

So this lands for uniformity with #824 and to stop the frontend being the only place these shapes go unrecognised — not as a speedup.

## The coupling is now pinned from both ends, loudly

That was the other half of #825, and it needed no new mechanism, just the guards pointed at both sides:

* **What the reader posts.** The two new lanes require `propagatorCallsLessEqualConstraints=8`, so a change of class there fails instead of passing quietly, and `--fzn-pattern int_lin_le` fails if MiniZinc ever starts emitting `int_le` and makes the recovery dead code.
* **What the scan accepts.** `inferred_cumulative_presolver_test` now runs its makespan fixture a second time with the rows spelled the way this reader spells them, and asserts the same certified bound of eleven. Drop #826's comparison loop and it fails with *"derived a makespan bound of 0 … the model's rows were not recognised as makespan links"* — the silent failure that started all this, now loud, in the test for the presolver that actually consumes those links.

38 proofs VERIFIED in that test; `ctest` 798/798; GCC `-DGCS_WERROR=ON` and clang 21 clean.


ciaranm:
Merge-order note for whoever lands these, since this one is not a plain `main` PR:

* Base is `fzn-two-term-linear-rewrite` (#824), because this sits beside its `recover_two_variable_form`.
* It also carries an explicit merge of `makespan-links-comparison-family` (#826), which it *needs*: without #826's comparison-family scan this rewrite silently stops `find_makespan_links` finding any makespan links.

So: **#826, then #824, then this**. If #824's branch is deleted on merge, GitHub will retarget this to `main` on its own — check that happened before merging rather than after.

Issue #833, opened 2026-09-04
Constraints need a large-domain policy: 14 of 25 MiniZinc-reachable globals wedge or run out of memory on a 0..10^9 domain


ciaranm:
## Summary

Nothing in the solver bounds the work a constraint does as a function of how *wide* a
variable's domain is. A propagator that reasons value by value over a 10^9-value domain
does 10^9 units of work per call, an initialiser that precomputes per value allocates per
value, and a scheduling propagator that indexes by time allocates over the whole horizon.
None of that is a bug in any one constraint; it is a missing policy.

I propose we adopt one, and that the default for **every** constraint is no stronger than
"fall back to something cheap once a domain is large" — the policy the arithmetic family
already implements as `consistency::Auto` (tabulate only if the enumeration fits
`default_tabulation_threshold()`, else BC). This should come **before** the finer
"bounds plus interval-level pruning" work discussed on #815 and #816, as a safety measure.

Below is an audit of where we stand, because the size of the problem turned out to be
larger than I expected and it changes what the policy has to say.

## Impact: 14 of 25 MiniZinc-reachable globals misbehave on `var 0..1000000000`

One-constraint models, flattened through our own mznlib, run with `-t 2000` under
`ulimit -v 2000000`. No domainless variables are involved — this is an ordinary declared
domain of a billion values.

| outcome | constraints |
|---|---|
| **wedged** — ignored the budget until a 25 s `SIGKILL` | `all_different`, `all_different_except`, `array_int_maximum` |
| **`std::bad_alloc`** | `among`, `cumulative`, `disjunctive`, `element`, `global_cardinality`, `nvalue`, `table`, `regular`\* |
| **overran the budget** | `count` (20.6 s), `member` (5.9 s), `knapsack` (5.2 s) |
| behaved | `all_equal`, `arg_sort`, `bin_packing`, `diffn`, `increasing`, `inverse`, `lex`, `seq_precede_chain`, `sort`, `subcircuit`, `value_precede` |

\* `regular` is fine when its *alphabet* is small and its variables are narrow; give the
variables the wide domain and it dies at 1.98 GB in 7.0 s.

Not probed, but the same shape in the source: `mdd`, `smart_table`, `min_distance`,
`at_most_one` (`at_most_one.cc:103` iterates the value variable's domain), and
`all_equal`'s hole path (`all_equal.cc:122`; my probe had no holes, so it never ran).

[...]

PR #847, opened 2026-09-05
A large-domain development tripwire, and a per-constraint audit lane (#833)

ciaranm:
Nothing in the solver bounds the work a constraint does as a function of how *wide* a
variable's domain is (#833). This is the measuring device the rest of that work needs: an
opt-in build that turns work proportional to a domain's width into an attributable test
failure, and a lane that runs every constraint class against it.

**First of four.** This one is a pure addition — no behaviour changes with the option off.

## The guard is not a safety net, deliberately

`GCS_LARGE_DOMAIN_GUARD` is **OFF by default and not user-facing**. What protects a user is
a constraint having somewhere cheap to fall back to, not an exception thrown from the
middle of propagation; the guard only helps us find the places that do not have one yet.
With it off the checks compile to nothing that even evaluates the size being checked, and
the only trace left in the library is the exception class's message string, because the
type is compiled unconditionally so the lane can name it (measured: `strings` on the
guard-off `.a` finds that one and none of the per-iteration ones).

Two kinds of check, and the difference is load-bearing:

* **Iterations count values as they are handed out** (`LargeDomainIterationCounter`, in
  `State`'s four value-iterating entry points).
* **Span-sized allocations check up front** (`GCS_CHECK_LARGE_DOMAIN`, at the cumulative,
  disjunctive and extensional residue sites).

Counting work done is the right measure and checking the domain's *width* is not: a
branching heuristic legitimately asks for a generator over a billion-value domain and
reads one value from it. The first version of this checked the width and condemned `Plus`,
`Abs`, `LessThan` and `LinearEquality`, all of which are fine.

`GCS_CHECK_LARGE_DOMAIN` is a macro rather than a function on purpose: its argument must
not be *evaluated* when the guard is off. That is the only reason; the iteration side
needs no macro because an empty class does the job.

## The audit lane

`large_domain_audit_test` posts each constraint class once over `0..10^9`, installs it and
propagates at the root and nowhere else. Every row pins its expected outcome **in both
directions**, so a row that stops tripping fails too — that is good news needing the table
updated, and pinning both ways is what stops the table drifting from what the code does.
Built always so it cannot rot, registered as a ctest case only when the guard is on, since
otherwise every probe passes trivially.

67 probes: 21 `KnownTrip`, 27 `Clean`, 14 `NoWidePosition`, 5 `HazardNotReached`. (The
next PR in the stack sharpens the last group away; the counts there are 68 / 24 / 30 / 14 /
0.)

Three results were not what #833 predicted, and they change what the later stages have to
do:

1. **`AllDifferent` under `consistency::VC{}` and `Element` under `consistency::BC{}` are
   both clean.** Those fallbacks really are width-independent, so adopting them is a policy
   decision with no propagator work behind it.
2. **`Power` and `LexSmartTable` trip**, and are on neither of #833's lists — what auditing
   per constraint class rather than per MiniZinc-reachable global buys.
3. **A "survives" is only worth as much as the probe behind it**, which is what
   `HazardNotReached` exists to say, and why the 21 is a lower bound.

## Testing

802/802 with the guard off, byte-identical behaviour. Guard on: builds clean and the lane
is green with its table pinned. Also checked that no legitimate work in the suite trips
the guard.

`dev_docs/large-domains.md` carries the rule (*every bounds-consistency path must be
independent of domain width*), the hazard taxonomy, the guard's use and the audit table.

---

Written with the assistance of Claude Opus 5.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -74,6 +74,17 @@ cmake -S . -B build -DGCS_ENABLE_VIEW_WRAP_SWEEP=ON
 The harness and `--view-wrap=N` / `--view-position=K` argv flags on each
 test are always built; only the ctest registrations are gated.
 
+Enable the large-domain guard, a development tripwire that turns work
+proportional to a variable's domain width into a test failure (issue #833):
+```shell
+cmake -S . -B build-guard -DGCS_LARGE_DOMAIN_GUARD=ON
+```
+Off by default, and deliberately not user-facing: what protects a user is a
+constraint having somewhere cheap to fall back to, not an exception thrown from
+the middle of propagation. Turning it on also registers `large_domain_audit_test`,
+which probes every constraint class over a `0..10^9` domain against a table of
+expected outcomes. See `dev_docs/large-domains.md`.
+
 By default the data-driven constraint tests run with generous per-solve
 solution/recursion caps (`GCS_TEST_CAP_DEFAULTS=ON`) so a pathological
 random instance can't dominate the parallel suite. A capped run checks
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,6 +122,22 @@ unset(_gcs_enable_executables_default)
 option(GCS_BUILD_TESTS "Build GCS tests" ${PROJECT_IS_TOP_LEVEL})
 option(GCS_ENABLE_VIEW_WRAP_SWEEP "Register the view-wrap proof-verification sweep (many tests, most failing today; opt in when working on view proof logging)" OFF)
 
+# The large-domain guard (issue #833). A development tripwire, not a user-facing
+# safety net: it turns work proportional to a variable's domain width -- a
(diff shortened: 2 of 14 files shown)


ciaranm:
CI was red on `ubuntu-26.04` only — the **Build** step, not tests. GCC 15 emits a false-positive `maybe-uninitialized` inside `std::vector<gcs::Integer>`'s destructor, walking the inactive alternative of a `SmartEntry` variant, from the nested braced list this file's SmartTable probe used to build its `SmartTuples`. Only that lane fails because it is the one with `GCS_WERROR=ON`, which is off in a normal build — so nothing local could have caught it.

Reproduced here with `-DGCS_WERROR=ON`, and fixed in 9f0d7d37 by building the tuple a step at a time, which is the spelling `smart_table_test.cc` already uses. The -Werror build is clean now.

The three PRs above have been rebased onto this.

Issue #815, opened 2026-09-03
fzn-glasgow ignores --timeout, SIGINT and SIGTERM on the 2010 challenge filters instance: 14 minutes of CPU on 33 variables with domains of 101


ciaranm:
`fzn-glasgow` on the MiniZinc Challenge 2010 `filters` instance never stops: it ignores
`--timeout`, `SIGINT` and `SIGTERM`, and has to be `SIGKILL`ed. The model is small — **33
integer variables, all domains ≤ 101 values, 37 constraints** — so this does not look like
"the instance is just hard".

## Reproduction

```shell
# MiniZinc Challenge 2010, filters/filter.mzn with ar_1_1.dzn (first instance alphabetically)
minizinc --solver minizinc/glasgow.msc -c --fzn filters.json filter.mzn ar_1_1.dzn
./build/fzn-glasgow -a -s -t 5000 filters.json      # asks for five seconds
```

It prints the pre-search statistics block and then nothing. Observed **14 minutes of CPU
time at 100%** against a requested 5 s, with no solution, no `=====UNKNOWN=====`, and no
final statistics. `timeout -s TERM 30` does not end it — `timeout` sits in `do_wait`
forever, because the `SIGTERM` never takes effect.

The signals *are* installed: `/proc/<pid>/status` shows `SigCgt: 0000000100004002`, which
has the bits for both `SIGINT` (2) and `SIGTERM` (15), and `SigIgn` is empty. Sending
`SIGINT` by hand changes nothing — still running, still `R`, 15 s later. So a handler is
setting a flag that nothing on this code path ever reads.

## Where it is spinning

`perf record` (frame pointer, 31 876 samples over 8 s). Everything here is propagation;
there is no search or backtracking symbol above 1%:

| % | symbol |
|---|---|
| 19.2 | `ArrayMinMax::install_propagators`'s propagator lambda |
| 17.9 | `cfree` |
| 8.9 | `small_vector_base<…Literal…>` copy constructor |
| 8.9 | `__memmove_avx_unaligned_erms` |
| 8.9 | `State::change_state_for_not_equal` |
| 5.6 | `State::in_domain` |
| 5.5 | `each_value_generator` |
| 2.5 | `State::infer_not_equal` |

The flattened model is 8.6 KB: 30 `int_lin_le`, 4 `int_lin_eq`, 2 `glasgow_diffn`, 1
`array_int_maximum`.

## Why this looks like a loop rather than slowness

[...]

PR #851, opened 2026-09-05
Prune ArrayMinMax's result to the union of the array at interval level (#833, #815)

ciaranm:
Follows #850. First of the interval rewrites, and the first thing in this stack that makes
a wide-domain model actually work.

## What it does

`ArrayMinMax`'s "result in union(vars)" sweep walked result's whole domain, so it was
O(|D(result)|) **even when it removed nothing** — and it is also the only thing bounding
result from the loose side, so on a domainless FlatZinc objective it had 9.2x10^18 values
to get through and never returned (#815). Compute the unsupported set as result's domain
minus the union of the array's, and remove it a range at a time.

Same removals, so the same search; only the number of inferences changes.

On #815's shape — array `1..101`, result `-10^9..10^9` — this goes from **never
terminating** to **0.004 s and five nodes**.

## The proof, which is the interesting part

My first attempt was wrong, and the way it was wrong is worth recording because it decides
which other interval rewrites are cheap.

A range conclusion is not RUP here, but **not** because "a range atom pins no bits". The
conclusion side is fine: two opposing bounds on `result` do contradict through the bit
rows. **The reason side is the blocker.** A reason literal `var NOT IN [lo, hi]` is the
disjunction `~ge_lo(var) \/ ge_hi+1(var)`, and RUP cannot case split — propagation can
*refute* "result >= lo and var <= lo-1 and f_i" but cannot *derive* "var >= lo" from it.
The per-value form never meets this, because `result = val` pins every bit of `result` and
so pins `var`'s through the linear row, deciding both halves of the disjunction at once.

So each refutation is restated as a clause the checker can propagate — the two ge-layer
bound lemmas — and only then is the conclusion drawn. Per variable:

```
rup  ~f_i \/ ~ge_lo(result) \/ ge_lo(var_i)          [crosses the half-reified row]
rup  ~ge_hi+1(var_i) \/ ge_hi+1(result)              [crosses the unconditional row]
rup  ~f_i \/ ~in(result,lo,hi) \/ in(var_i,lo,hi)
```

then RUP the conclusion. **3n+1 lines per interval, independent of the range width** —
the property that matters, since the whole point is replacing 9.2x10^18 steps with a
constant number.

Which lemma carries the selector depends on which row it crosses: for `max` the model has
`result >= var_i` unconditionally and `result <= var_i` under `f_i`, so the *lower* lemma
needs it; for `min` it is the upper. This is `justify_not_in_range_across_equality()`
generalised from an unconditional equality to a guarded one.

The procedure was derived and verified independently against real solver OPB output before
any of it was written into the propagator — including a signed 63-bit result, a 2^40-wide
gap, and negative controls confirming that dropping either lemma, or dropping the selector
from the one that needs it, makes VeriPB reject.

## What it does not do

Views keep the per-value path: their atoms are spelled through the view and the lemmas have
not been shown to bridge that. Same restriction, and the same reason, as the single-support
range path fifty lines below in the same file.

The audit row for `ArrayMinMax` **stays `KnownTrip`**, correctly. It has a second per-value
sweep — the full-GAC pass over the array variables — which a probe with *every* variable
wide still reaches. That one spins at a flat 6 MB rather than allocating, which is how the
two are told apart.

The hull bound #815 proposes as its own first fix stays unfixed, and is now recorded as a
separate problem rather than a small one: there the selector has to be *derived* rather
than assumed, so these lemmas do not apply and it needs `pol` over OPB line numbers
`define_proof_model` does not keep. This rewrite makes it unnecessary for the case that
motivated it.

## Testing

802/802 guard off, with both `min_max` lanes verifying under VeriPB including the
view-mixed one; audit lane green guard on.

---

Written with the assistance of Claude Opus 5, and of a second model consulted specifically
on the proof procedure.

### DIFF
--- a/dev_docs/large-domains.md
+++ b/dev_docs/large-domains.md
@@ -52,31 +52,58 @@ The tree already has the machinery: `IntervalSet::each_interval_minus()`,
 that removes "everything not in this small given set" one value at a time is an
 interval operation written out longhand, and rewriting it keeps GAC.
 
-### A trap: a bound that is obvious is not necessarily RUP
+### Getting a range removal past the checker
 
-`ArrayMinMax` has no hull bound on the loose side — for a max it posts
-`result >= lb(var_i)` for every i but never `result <= max_i ub(var_i)`, so the
-per-value union sweep is what establishes it (issue #815, which proposes adding
-the bound as a small, obviously-correct first fix).
+Replacing a per-value removal with a range one is not free in the proof, and the
+reason it is not is worth understanding once, because it decides which
+interval rewrites are cheap.
 
-It is obviously correct, and it is **not a small fix**, because it does not
-verify. Negating it gives `result > max_i ub(var_i)`; each selector asserts
-`result <= var_i`, so every selector must be false and the al1 row fails. That
-argument needs to carry an order atom on `result` across the half-reified row
-relating `result` and `var_i`, which is a linear row over the *bits* — and unit
-propagation does not cross a bit sum on a wide domain, where nothing fixes an
-individual bit. Adding the bound with `JustifyUsingRUP` fails VeriPB on
-`min_max_constraint` and `min_max_constraint_view_mixed`. Making it verify needs
-an explicit `pol` per selector, over line numbers `define_proof_model` does not
-currently keep.
+The conclusion side is fine. Negating `result NOT IN [lo, hi]` gives two opposing
+bounds on `result`, and those *do* contradict through the bit rows — that is the
+configuration `justify_not_in_range_across_equality()` already relies on, and it
+holds even for a signed 63-bit variable.
 
-The interval rewrite of the union sweep is the better target anyway: it fixes the
-general case rather than only the narrow-array one, and the range-removal proof
-pattern it needs already exists and verifies fifty lines further down the same
-file (`min_max.cc:199`, `infer_not_in_range` with a value-independent
-justification emitted once per interval).
+**The blocker is the reason side.** A reason literal `var NOT IN [lo, hi]` is the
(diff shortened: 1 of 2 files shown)


ciaranm:
CI caught a **soundness bug** in this, on macOS, Ubuntu 24.04 and the clang lane: `min (1,3) [(0,4),(0,5),(0,6)]` lost the solution `(1, [3, 3, 1])`.

The cause is mine and unambiguous. I built the union of the array's domains with

```cpp
for (auto [lo, hi] : state.copy_of_values(var).each_interval())
```

and `IntervalSet`'s class documentation says plainly that the iteration methods hand out generators which *borrow* the set's storage, and the set must outlive the generator. That one is a temporary. The generator dangles, reads garbage intervals, and so fails to erase the values that really are supported — leaving supported values in the unsupported set and removing them from the result. Lost solutions, not merely lost pruning. Being UB it is platform-dependent, which is why it passed on Windows and on my machine.

Fixed in dc7e2e41 by binding the copy to a named local. Confirmed rather than argued: built with clang and run at CI's own seed, the unfixed code loses exactly `(1, [3, 3, 1])` and the fixed code does not. Nothing else in the tree does this; `abs.cc` already uses named locals.

**Why it reached you.** Every check I ran was capped. `GCS_TEST_CAP_DEFAULTS` puts per-solve caps in each test's environment, and a capped run verifies soundness and the partial proof but *not* completeness — so a propagator that loses solutions passes it. The Ubuntu CI lanes build with caps off, which is what caught this. My "802/802" reports on this PR were therefore worthless for the one risk this change actually carried. Re-verified uncapped, on GCC and clang, over four seeds, plus the full suite and the audit lane.

PR #853, opened 2026-09-05
Cap declared domains at a quarter of the machine range, and diagnose rows that will not fit (#852)

### LINKED ISSUE #852: Proof model emission overflows on variables as wide as a domainless FlatZinc int, leaving a truncated OPB
## Summary

With proof logging on, a constraint over variables spanning the full range
`fzn-glasgow` gives a domainless FlatZinc `var int` — `[Integer::min_value()/2,
Integer::max_value()/2]`, i.e. `[-2^62, 2^62-1]` — aborts while *writing the OPB*, before
search, with

```
unexpected problem: Integer overflow: --9223372036854775808
```

The OPB file is left **truncated** (18 866 bytes in my probe), so the failure is partway
through emission rather than a clean up-front rejection.

## It is not one constraint

Every one I tried fails, over two variables with that domain:

| constraint | outcome |
|---|---|
| `ArrayMax` | `Integer overflow: --9223372036854775808` |
| `Max` (two-argument) | same |
| `LessThan` | same |
| `AllDifferent` under `consistency::VC{}` | same |
| `Plus` | `Integer overflow: 9223372036854775807 + 1` (different site, same trigger) |

So this is the shared OPB emission path, not a quirk of any one constraint.

## Where

`gcs/innards/proofs/emit_inequality_to.cc:38`, inside `append_or_fold_term_to`, which
renders a term "with negated weight". The negation is unary `operator-` on `Integer`
(`gcs/integer.hh:180`) applied to `LLONG_MIN`, which has no positive counterpart:

```
throw_integer_overflow ()                       integer_overflow.cc:31
gcs::operator- ()                               integer.hh:180
operator()<gcs::SimpleIntegerVariableID> ()     emit_inequality_to.cc:38
... append_or_fold_term_to ...
```

## Proof logging only

**Without** `ProofOptions` the same models post and propagate fine — they just search a
huge space. So this does not affect ordinary solving; it means a proof cannot be written
at all.

## Why it matters

`fzn_glasgow.cc:464` gives a domainless FlatZinc `var int` exactly this domain, and that is
deliberate: the large-domain policy (#833) settled on keeping wide domains and having
constraints fall back to bounds reasoning rather than refusing them. So **any FlatZinc
model with two domainless ints in one constraint cannot be proof-logged**, and the failure
mode is a truncated OPB rather than a diagnostic.

Reproduction (library API, since MiniZinc rejects these bounds at flatten time):

```cpp
Problem p;
auto lo = Integer{-(1LL << 62)}, hi = Integer{(1LL << 62) - 1};
p.post(LessThan{p.create_integer_variable(lo, hi), p.create_integer_variable(lo, hi)});
solve_with(p, /* callbacks */, ProofOptions{"probe"});
```

## Possibly related, not verified

#833 records that the Challenge 2025 `work-task-variation` instance "does not even load",
with `Integer overflow: 4611686018427387904 += -4611686018427387904`. That constant is
`2^62`, the same bound as here, so the two may share a root cause — but I have not checked
that instance, and the operator differs (`+=` rather than unary `-`), so treat it as a lead
rather than a duplicate.

---

Found by an AI agent (Claude) while working on #833. Measured on `fataepyc-10`, Release
build of `7d014207`.

ciaranm:
Fixes #852. Independent of the #833 stack — this is proof emission, not large-domain
propagation — so it is based on `main`.

## The measurement

The boundary is measured, not chosen. Same three constraints, varying only the width:

| declared domain | `LessThan` | `Plus` | `AllDifferent`/VC |
|---|---|---|---|
| `[-2^61, 2^61-1]` | writes its model | writes its model | writes its model |
| `[-2^62, 2^62-1]` | aborts mid-OPB | aborts mid-OPB | aborts mid-OPB |

`2^62` is exactly what `fzn_glasgow.cc` gave a domainless FlatZinc `var int`, so any model
with two domainless ints in one constraint could not be proof-logged, and the failure left
a **truncated OPB** rather than a diagnostic.

Two bits of headroom rather than the obvious one, because emission costs more than a
single carry: a half-reified row's reification constant is the sum of the positive
contributions of *every* term, so a row relating two variables needs room for both — and
that constant is negated again when the row is rendered in `>=` form, and the most negative
`Integer` has no negation.

## The cap

`Integer::min_bounded_value()` / `max_bounded_value()` name a quarter of the machine range.
`State::allocate_integer_variable_with_state` refuses anything wider, and `fzn-glasgow`
uses the constants.

The check is on the `State` primitive rather than on `Problem` because constraints mint
auxiliaries there directly (`power.cc:185`, `divide_modulus.cc:134`) and would otherwise
bypass it. **It refuses rather than narrowing** — silently shrinking a declared domain
would change what the model means.

`fzn_glasgow.cc`'s comment already said the bounds were halved "so there is headroom": the
intent was right and the amount was one bit short, so this makes the existing comment true
rather than contradicting it.

Arithmetic is deliberately **not** held to the cap. Intermediates are expected to exceed
it — that is what the headroom is for. This bounds what may be *declared*, so the ordinary
routes cannot reach the cliff by accident; nothing stops a caller computing a wider value
and using it anyway, which is why the emitter is hardened too rather than trusting the cap.

## The hardening

Three emission sites can overflow, and each now restates the arithmetic failure as a
`ProofError` naming what did not fit:

* sizing the reification constant (`reification_shape`);
* that constant landing on the most negative `Integer` — caught there, where something
  useful can still be said, rather than downstream at the negation in `emit_inequality_to`
  where nothing knows which variable is at fault;
* the coefficient and right-hand-side folding, in **both** row renderers. I found this one
  only by trying to reach the first two, which is why it is here and not in my first pass.

This makes the failure legible; it does not make wide rows work. With the cap in place no
ordinary route reaches it — **large coefficients still do**, which is what the regression
test uses: a reified linear with `2^60` coefficients now names the reification constant
instead of failing as bare arithmetic.

## Deliberately not fixed

A fourth overflow site turned up while testing: `tidy_up_linear` (`linear/utils.cc:38`)
overflows while normalising a linear constraint whose constant does not fit, **with proofs
off entirely**. Refusing that is correct — the constraint genuinely does not fit — and only
its message is poor, so it is a separate change.

## Worth a second opinion on review

* The cap binds **internal auxiliaries** too, not just user variables. Nothing in the suite
  trips it, but a constraint wanting a legitimately wide auxiliary is now refused rather
  than producing a broken proof. I think that is right, since it *would* be broken, but it
  is a behaviour change beyond the FlatZinc path.
* `min_bounded_value()` / `max_bounded_value()` are public API on `Integer`, since
  `fzn-glasgow` needs them and callers should be able to see the limit they are held to.

## Testing

802/802. New coverage: the cap's boundary in `problem_test.cc` (both overloads), and the
emission diagnostic in `solve_test.cc`.

---

Written with the assistance of Claude Opus 5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,6 +388,13 @@ This returns an ``IntegerVariableID``, which is a light-weight handle which you 
 value. For type-safety and avoiding overflow problems, the solver does not accept raw ``int`` and
 similar types directly, and anywhere you use a numerical value you must create an ``Integer``.
 
+For the same reason there is a limit on how wide a domain you may declare: bounds must lie within
+``Integer::min_bounded_value()`` and ``Integer::max_bounded_value()``, which is a quarter of the
+range of the underlying type in each direction, and creating a variable outside it throws an
+``InvalidProblemDefinitionException``. The headroom that leaves is what lets the solver's internal
+arithmetic, and the coefficients it writes into a proof, stay within range; arithmetic itself is
+deliberately not held to the same limit.
+
 You will also want some constraints. These can be found in the ``gcs/constraints/`` directory. Once
 you construct a constraint, you can add it to a problem instance using ``Problem::post``.
 
--- a/gcs/innards/proofs/emit_inequality_to.cc
+++ b/gcs/innards/proofs/emit_inequality_to.cc
@@ -1,5 +1,7 @@
+#include <gcs/innards/integer_overflow.hh>
 #include <gcs/innards/proofs/emit_inequality_to.hh>
 #include <gcs/innards/proofs/names_and_ids_tracker.hh>
+#include <gcs/innards/proofs/proof_error.hh>
 #include <gcs/innards/proofs/simplify_literal.hh>
 
 #include <util/overloaded.hh>
@@ -88,20 +90,42 @@ namespace
     }
 }
 
+namespace
+{
+    // Writing a row folds constant terms into the right-hand side and negates
+    // every weight to convert <= into >=, so a row whose coefficients and bounds
+    // together do not fit in an Integer fails here -- part-way through building
+    // the string, with the OPB already half written and nothing in the arithmetic
+    // failure naming what was being emitted. Restate it as something a caller can
+    // act on. Variable domains are capped so the ordinary routes cannot reach
+    // this (Integer::max_bounded_value()); large coefficients, or views that
+    // offset a domain outwards, still can. See issue #852.
+    [[noreturn]] auto rethrow_as_proof_error() -> void
+    {
+        throw ProofError{"cannot write a pseudo-Boolean row whose coefficients and variable bounds together do not fit in an Integer; "
+                         "the variables involved may have domains near Integer::max_bounded_value(), or the coefficients may be too large"};
+    }
+}
+
 auto gcs::innards::emit_inequality_to(NamesAndIDsTracker & names_and_ids_tracker, const SumLessThanEqual<Weighted<PseudoBooleanTerm>> & ineq,
     string & out, EnsureNames ensure_names) -> void
 {
     // build up the inequality, adjusting as we go for constant terms,
     // and converting from <= to >=.
-    Integer rhs = -ineq.rhs;
-    for (auto & [w, v] : ineq.lhs.terms) {
-        if (0_i == w)
-            continue;
-        append_or_fold_term_to(names_and_ids_tracker, w, v, out, rhs, ensure_names);
-    }
+    try {
+        Integer rhs = -ineq.rhs;
+        for (auto & [w, v] : ineq.lhs.terms) {
+            if (0_i == w)
+                continue;
+            append_or_fold_term_to(names_and_ids_tracker, w, v, out, rhs, ensure_names);
+        }
 
-    out += ">= ";
-    append_number_to(out, rhs);
+        out += ">= ";
+        append_number_to(out, rhs);
+    }
+    catch (const IntegerOverflow &) {
+        rethrow_as_proof_error();
+    }
 }
 
 auto gcs::innards::emit_reified_inequality_to(NamesAndIDsTracker & names_and_ids_tracker, const SumLessThanEqual<Weighted<PseudoBooleanTerm>> & ineq,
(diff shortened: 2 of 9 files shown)


ciaranm:
Two documentation items from the pre-merge review. Nothing here is about the code: CI is green on all
ten lanes, and a local trial merge of this branch together with the #847-#861 stack builds clean (0
warnings, GCC 15.2, against a baseline of 0) and passes 804/804 including all 100 minizinc lanes.

## 1. The cap is user-facing, and undocumented

`Integer::min_bounded_value()` / `max_bounded_value()` carry a good explanation of *why* the boundary
sits where it does in `integer.hh`, but a user currently meets the limit for the first time as a
thrown `InvalidProblemDefinitionException`. It should be documented in both of these places:

* **`README.md`** — the "Navigating the Source Code" section (~L384-389) is where
  `Problem::create_integer_variable` is introduced, and it already explains the `Integer` wrapper as
  being there "for type-safety and avoiding overflow problems". A limit on what a domain may declare
  belongs in the same breath.

* **The doxygen on the creation entry points**, in `gcs/problem.hh`. All four routes reach
  `State::allocate_integer_variable_with_state` and so can now throw, and none of their comments
  mentions it:
  - `create_integer_variable(Integer lower, Integer upper, name)` — L132-141
  - `create_integer_variable(const std::vector<Integer> & domain, name)` — L143-152, which checks the
    `minmax_element` of the vector; that is the path `problem_test.cc`'s vector case pins
  - `create_integer_variable_vector(...)` — L154-161, forwards to the first
  - `create_n_integer_variables<n_>(...)` — L163-176

## 2. The fzn-glasgow figure in the stack's new doc goes stale when this lands

`dev_docs/large-domains.md` (added by #847) states as present-tense fact at L19 that "`fzn-glasgow`
gives a domainless FlatZinc `var int` a domain of about 9.2x10^18 values". This PR changes that bound
from `min_value()/2 .. max_value()/2` to `min_bounded_value() .. max_bounded_value()` — a span of
4.61x10^18, i.e. half. So whichever of the two lands second should update that number.

Neither branch's CI could see this, which is why it is worth writing down: the stack's checks never
had this branch, and this branch's checks never had the doc. Two further occurrences of the same
figure (`large-domains.md` L114, and `min_max.cc:112`) read as past-tense or illustrative, so they
are probably fine as they stand, but worth a glance while you are in there.

Incidentally, `problem_test.cc:65` asserting that `min_value()/2 .. max_value()/2` now throws is a
neat confirmation that outlawing the previous fzn-glasgow bound is deliberate rather than incidental.

*(From the automated pre-merge checklist review; not merged, pending these and your own review.)*

ciaranm:
Both done, in **23e5d7d6** here and **f92f9213** on #862.

## 1. The cap is now documented on the way in

You are right that the first a user learns of it is the exception. All four creation routes reach `allocate_integer_variable_with_state` and so can throw, and all four now carry a `\throws` line. The vector overload's is worded separately, since it is the smallest and largest value given that are checked and the holes in between are unconstrained — which is exactly what `problem_test.cc`'s vector case pins.

`README.md` gets it in the paragraph you pointed at, next to the existing "for type-safety and avoiding overflow problems" sentence, together with the reason it is *not* a limit on arithmetic: the headroom is what lets intermediate results and proof coefficients stay in range. That distinction seemed the part most likely to be misread.

## 2. The fzn-glasgow figure — unpinned rather than updated

Rather than change 9.2×10^18 to 4.6×10^18 and leave the same trap for the next person, I removed the figure. The sentence only needs the span to be astronomically large, which it is either way, so it now says that and says why it gives no number. That decouples the two branches, so neither has to land first and nobody has to remember.

Your framing of *why* this was invisible is the useful part and I have put it in the commit message: this document does not exist on your branch and your change does not exist on the stack, so no CI run anywhere could have seen the two together. A literal in prose is a dependency that nothing checks.

The other two occurrences I have left, and I agree with your reading: `large-domains.md` L131 is about the number of proof *steps* an interval rewrite replaces rather than a declared domain, and `min_max.cc`'s is a past-tense account of what #815's instance did when it was measured.

802/802 here with 0 warnings.

And agreed on `problem_test.cc:65` — asserting that the *old* fzn-glasgow bound now throws is a nice way to pin that outlawing it is deliberate. Worth keeping.

Issue #857, opened 2026-09-06
Proof-scaling survey: no VeriPB feature candidate survives once the propagators are read


ciaranm:
The proof-scaling survey added under #833 (`large_domain_audit_test "[.proofscaling]"`)
exists to answer one question: which constraints write a proof whose *steps* grow with a
variable's domain while the *encoding* stays fixed. That is the copy-paste a VeriPB feature
could collapse in one rule, and the survey is where the evidence for asking for such a
feature was supposed to come from. (The other half — encodings that grow — is #846, and is
a different problem with different answers.)

Re-measured across all 69 probes, **the survey currently supports no such feature
request.** Every row whose steps grow at a fixed encoding turns out to be a propagator that
has already computed an interval and is spelling it out one value at a time.

## What changed

The table as written named `Among` and `Table` as the two clean candidates, on the grounds
that their encodings do not grow at all so every extra step is the same derivation with a
different constant in it. That was true, and it was still the wrong conclusion: both were
also H1a sites, and the interval rewrites (#854 for `Table`, #856 for `Among`) collapsed
both to flat.

|  | steps @1e3 → @1e4 | growth |
|---|---|---|
| `Among`, before | 32996 → 329996 | 10x |
| `Among`, after | 98 → 98 | **1.0x** |
| `Table`, before | 50788 → 509788 | 10x |
| `Table`, after | 93 → 93 | **1.0x** |

The rows that grow now are different ones, and they had been growing unnoticed: #849
sharpened three probes from `HazardNotReached` into real hazards without the survey being
re-run, so the published table was stale from that point on.

| constraint | opb (fixed) | steps @1e3 → @1e4 |
|---|---|---|
| `AllEqual/holes` | 22 | 16987 → 169987 |
| `GlobalCardinality` | 30 | 18024 → 180024 |
| `Element` | 32 | 27981 → 279981 |

## Why none of the three is evidence

The discriminating question is not "are the steps repetitive" — they always are — but **is
the removed set an interval?** If it is, an interval rewrite deletes the volume at full
strength and needs nothing from the checker. All three fail that test as candidates:

* **`AllEqual/holes`** is the starkest. `all_equal.cc` already computes the values to remove
  with `each_interval_minus`, and then expands each interval a value at a time — the
  interval is literally in hand when the per-value loop starts. The real obstacle is the
  *witness* in the reason: the difference is taken against the intersection of every
  domain, so which variable is missing a value can change within one interval. Taking the
[...]

Issue #846, opened 2026-09-05
Tracker: OPB encodings that grow with a variable's domain width


ciaranm:
Tracker for the constraints whose **OPB encoding** grows with a variable's domain width.
This is the encoding half of #833: that issue is about propagation and install-time work,
and a consistency level or a weaker propagator does nothing here, because
`define_proof_model` runs whenever proofs are on regardless of how the constraint chooses
to propagate.

Filed as a group so the shape is visible; individually they are quite different problems.

## The cases

| issue | constraint | rows | what varies per row | re-encodable? |
|---|---|---|---|---|
| #841 | `Regular`, `RegularLegacy`, `RegularBacchus` | O(layers x states x D) | a value with no transition at that state | **yes** |
| #842 | `MDD` | O(layers x nodes x D) | same | **yes** |
| #843 | `NValue` | O(D) | a value of the union of the domains | only by changing encoding family |
| ~~#844~~ | ~~`Cumulative`~~ | ~~O(tasks x horizon)~~ | ~~a time point~~ | **already done in #781** |
| #845 | `Power` / `PowerTable` | O(D) | a row of an enumerated relation | it *is* a table; wants a budget |

Everything else is flat. Of 67 constraint-class probes run at widths 10^3 and 10^4, 59
show no growth in OPB rows at all; these were the whole list.

**#844 is withdrawn**: `Cumulative`'s horizon-free start-checkpoint encoding is already
open as **#781** (for #780), and takes the same instance from 1 945 OPB rows to 46 at
`0..100`, and from 190 045 to 46 at `0..10000` — flat in the horizon. It is opt-in behind
`GCS_CUMULATIVE_ENCODING=start-checkpoint` and #781 says the `TimeIndexed` default is
deliberate pending a measurement over #777, so the remaining decision lives there.

## What the rows actually are, which is not what the counts suggest

Breaking down the OPB rather than counting them turned out to matter, because the majority
of the per-value cost is usually **not** the constraint:

| constraint (3 vars, `0..100`) | total | variable `geN`/`eqN` atom definitions | the constraint's own rows |
|---|---|---|---|
| `Regular` | 1 841 | 1 218 (66%) | 600 |
| `NValue` | 1 431 | 1 218 (85%) | 213 |
| `Cumulative` | 1 945 | **0** | 1 836 |

For `Regular` and `NValue` the bulk is the **variable's own direct encoding**, materialised
[...]

Issue #780, opened 2026-08-25
Cumulative's OPB encoding is O(n x horizon); a start-checkpoint encoding is O(n^2), and the proof only wants the time points the search cites


ciaranm:
`Cumulative::define_proof_model`
(`gcs/constraints/cumulative/cumulative.cc:351`) emits three fully reified flags
per **(task, time point)** over each task's possible-active window, plus one
`cap_t` row per time point. That is `O(n × horizon)`, paid unconditionally,
before search, whether or not anything cites it.

There is another encoding that is `O(n²)` and horizon-free: check capacity only
at time points that are the **start time of some task**. In MiniZinc's
vocabulary that is the task decomposition,

```
forall (j in TASK) (
    sum (i in TASK) ( (s[i] <= s[j] /\ s[i] + d[i] > s[j]) * r[i] ) <= L )
```

and in PB form, for every ordered pair `(i, j)`:

```
before_{i,j} <-> s_i <= s_j
after_{i,j}  <-> s_i + l_i >= s_j + 1
active_{i,j} <-> before_{i,j} /\ after_{i,j}   [ /\ present_i ]
```

with one row per task `j`:

```
C^start_j :  Sum_i h_i * active_{i,j}  <=  capacity
```

It is sound and complete because the load profile is a step function that only
rises at start times, so any over-capacity point is dominated by the last start
at or before it. Capacity, lengths and heights are all `>= 0` already
(`cumulative.cc:146-157`, and the constructor for constants), so "nothing active implies satisfied" holds.

## Why it is worth wanting: the horizon is not the problem's size

Measured through `fzn-glasgow --prove` on a generated `cumulative`, `n = 6`,
capacity 5, heights 2, durations `H/8`:

| horizon | OPB lines | OPB bytes |
|---|---|---|
| 50 | 1,865 | 296 KB |
| 100 | 3,715 | 653 KB |
| 200 | 7,415 | 1.4 MB |
| 400 | 14,815 | 3.1 MB |
| 800 | 29,615 | 6.7 MB |

Exactly linear, and linear in `n` too (2,511 / 4,919 / 9,735 lines for
`n = 4/8/16` at `H = 100`) — about six lines per (task, time point), which is
the three flags at two reification halves each. The proposed encoding is about
six lines per *ordered pair* plus `n` rows, so **`~6n²` lines, flat in the
horizon**. Crossover is at `H ~ n` on lines, `~2n` on bytes, since a pairwise
[...]

PR #861, opened 2026-09-06
Prune AllEqual to the common domain by range (#833)

ciaranm:
Part of #833 (stage 4), stacked on #860.

The values to remove were **already computed as intervals** with `each_interval_minus` and
then written out one at a time, so the interval was in hand when the per-value loop
started. What stopped that being a one-line change is the *witness*: the reason names some
variable whose domain lacks the value, and because the difference is taken against the
intersection of *every* domain, which variable that is can change part-way through an
interval.

So the difference is taken against one variable at a time instead. Every range that yields
against `vars[j]` has `vars[j]` as a witness for the whole of it, by construction, and
ranges already accounted for are struck off as we go, so each is removed once rather than
once per variable that happens to witness it. The union over `j` of `D_i \ D_j` is `D_i`
minus the intersection of the others, which is `D_i \ common` — so the removal set is
exactly what it was, confirmed by identical node counts on every benchmark shape.

Domains are snapshotted up front: the old code took `common` from the state on entry to the
block, so recomputing a domain mid-loop would use one this block had already shrunk, and
could remove more.

## The proof

`equals.cc`'s, generalised from two variables to n:
`justify_not_in_range_across_equality` carries the bounds across to `vars[j]`, and then the
reason's range literal is a clause with every literal falsified.

The model here is a **chain** of consecutive-pair equalities, so for non-adjacent `i` and
`j` unit propagation has to walk the chain between them. That verifies — and it is checked
where the two interesting axes meet rather than where the suite happens to go: the suite
reaches witness gaps of 1–3 but only ranges of **width 1**, so the wide case is checked
separately, at widths up to **4999** with gaps up to **5**. A negative control confirms the
lemmas are load-bearing: suppress them and VeriPB rejects.

## Two performance findings, both from the benchmark rather than the tests

- **A width-1 range literal *is* the eq atom** — the proof machinery never makes an interval
  of one — so saying it as `infer_not_equal` is the identical inference at no
  set-construction cost. Over holey domains most removed intervals are single values, and
  routing them through the interval path cost 2%.
- **The witness-splitting machinery almost never has anything to split.** A scan for a
  single covering witness is a few allocation-free merge-walks, where building a leftover
  set to subtract from allocates whatever happens. Checking the easy case first takes most
  of the rest back.

Together those bring a rewrite that was 2.5% slower on holey-domain search back to level on
two-interval domains, the common shape. **On four-interval domains it is still about 1.7%
behind**, which is the honest cost of width-independence here.

## A branch with no coverage at all

The splitting branch was reached by nothing — not the suite, not the probes — because
everywhere they go, one variable's hole happens to cover the whole run. `all_equal_test`
gains a case that forces it: `x` full, `y` missing the lower half of the middle band and `z`
the upper half, so the band leaves `x` as one interval that neither alone witnesses. It
reaches the branch, and verifies at widths up to 1000.

## Measurements

Pinned, three repeats, node counts identical throughout:

| | before | after |
|---|---|---|
| wide (5 wide vars vs a two-value partner) | 198 ms | **0 ms** |
| narrow2 (two-interval holey domains) | 355 ms | **352 ms** |
| narrow (four-interval holey domains) | 365 ms | 371 ms |
| audit probe at `0..10^9` | 131 s, 16 GB | **6 ms, 4.7 MB** |
| proof survey | 16987 → 169987 steps | **flat 37** |

## Testing

804/804. `all_equal_test` uncapped on GCC and clang, plain and view-wrapped. Audit lane
green, and the row flips to `Clean`.

That leaves **`GlobalCardinality/hall` as the only row in the survey whose steps still
grow** — see #857.

### DIFF
--- a/dev_docs/large-domains.md
+++ b/dev_docs/large-domains.md
@@ -439,7 +439,7 @@ previous version of it went stale, see below.
 | **Both** grow | 10x / 10x | `Power`, `PowerTable`, `NValue`, `Regular`, `RegularLegacy`, `RegularBacchus`, `MDD` |
 | **OPB only** | 10x / 1.0x | `Cumulative` (19046 → 190046 rows; one capacity line per time point, so it is H3 on the encoding side) |
 | **Steps only** | 1.0x / 10x | `GlobalCardinality/hall` (34-row OPB fixed, 33984 → 339984 steps) |
-| neither | 1.0x / 1.0x | everything else, 60 of 70 |
+| neither | 1.0x / 1.0x | everything else, 61 of 70 |
 
 `Multiply`, `Divide` and `Modulus` sit in the last row but are not flat: their OPB
 grows 1.8x for a 10x width, which is the bit-width of the product, not a per-value
--- a/dev_docs/large-domains.md
+++ b/dev_docs/large-domains.md
@@ -124,6 +124,25 @@ Views keep the per-value path: a view's atoms are spelled through the view and
 the lemmas have not been shown to bridge that. Same restriction, and same reason,
 as the single-support range path in the same file.
 
+**Two things that make an interval rewrite pay, learned on `AllEqual`.**
+
+*Keep the per-value spelling for a width-1 interval.* A range literal of one value
+*is* the eq atom — the proof machinery never makes an interval of one — so saying
+it as `infer_not_equal` is the identical inference, and finding its reason costs
+no set construction. Over holey domains most removed intervals are single values,
+and routing them through the interval path instead cost 2% on a search.
+
+*Do not pay for machinery the common case does not need.* Where the reason names a
+witness that can vary within an interval, the general answer is to split the
+interval by which variable accounts for which part. But almost always one variable
+explains the whole of it, and a scan for a single covering witness is a few
+allocation-free merge-walks, where building a leftover set to subtract from costs
+an allocation whatever happens. Check for the easy case first and only then split.
+
+Together these take a rewrite that was 2.5% slower on holey-domain search back to
+level, while keeping the wide case (200 ms of work, and 131 seconds in the audit
+probe) at nothing.
+
 **Where the equality is guarded by a conjunction, the same lemmas take a longer
 guard.** `Element`'s model is `result - array[i] == 0` half-reified on the
 conjunction of index conditions, so it is `ArrayMinMax`'s shape with the selector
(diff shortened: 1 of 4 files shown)


ciaranm:
Findings from the pre-merge review of the #847-#861 stack. The first one wants fixing before this
lands; the rest are documentation counts that have drifted as the stack progressed.

## Blocking: `run_mixed_witness_test` introduces a reproducible CI flake

`run_mixed_witness_test` is the only test in `all_equal_test.cc` that is **not** gated on
`view_wrap_config_is_effectively_bare(view_cfg, n_positions)`, so it runs in *both* registered lanes
— `all_equal_constraint` and `all_equal_constraint_view_mixed` — and both use the same fixed proof
basename `all_equal_test_mixed_witness`. The two lanes share a working directory, and
`dispose_of_proof_files()` removes the proof once veripb has run, so under `ctest -j` (which is how
CI runs it: `ctest -j $(nproc)` in cmake.yml, `-j 4` on Windows) one lane deletes or truncates the
other's proof mid-verification.

Reproduced on a trial merge of the whole stack:

* `ctest -R all_equal -j 8` — **5 of 6 rounds fail**
* the two binaries launched concurrently by hand — **11 of 12 fail**

Both halves of the race show up. Losing the file outright:

```
all_equal mixed witness with proofs: expecting 11 solutions
$ veripb all_equal_test_mixed_witness.opb all_equal_test_mixed_witness.pbp
Error: Failed to open file!
Caused by:
    No such file or directory (os error 2)
terminate called after throwing an instance of 'gcs::UnexpectedException'
  what():  unexpected problem: veripb verification failed
```

and reading one that is still being written:

```
$ veripb all_equal_test_mixed_witness.opb all_equal_test_mixed_witness.pbp
Error: Syntax error while parsing proof file!
Caused by:
    Expected a constraint ID (label or signed integer), `:` or `;` but found `-` at line 329 col 8.
```

Either of the file's own two conventions fixes it: gate the call on
`view_wrap_config_is_effectively_bare(...)` the way `run_holes_test` is (it is a bare-variable test,
so the view lanes add nothing but a duplicate run), or fold the view label into the proof name the
way `run_test` does with `"all_equal_test_" + view_wrap_config_label(view_cfg)`. Gating looks right
here. This is also the hazard the `disjunctive_precedences_mutation_*` block already calls out in its
own comment — "Distinct basenames so the lanes do not race over one set of proof files".

Worth stressing that **this PR's CI is green and a full 804/804 run of the trial merge passed** — the
overlap window depends on ctest's scheduling, so what this buys is intermittent redness on `main`
rather than a deterministic failure. That is why it is worth fixing now rather than after it bites.

## `dev_docs/large-domains.md`: three counts left behind by the stack's own edits

All three are the same shape — a number updated in one place and not in the sentence next to it — and
the cheapest fix is here at the tip, since this branch holds the final text and fixing it lower down
would rebase everything above.

1. **"69 probes" twice, where it is 70** — L432 ("Re-measured over all 69 probes") and L499 ("none of
   the 69 probes produces one today"). 70 is confirmed four independent ways: the source registers 70
   `add("...", Expect::...)` probes; the audit lane asserts 140 = 70 x 2; the proof survey prints 70
   rows; and this document's own table agrees (19 + 37 + 14, and "61 of 70"). History: 05f50f80 (#856)
   wrote "69" when it was correct, 01138efe (#860) added the 70th probe (`GlobalCardinality/hall`),
   and 735c9a7e here fixed "60 of 70" -> "61 of 70" but left both "69"s. L432 also now under-names the
   rewrites it postdates — it credits `ArrayMinMax`, `Table` and `Among`, but `AllEqual`,
   `GlobalCardinality` and `Element` have since landed too.

2. **Six constraints listed, "the five" in the next sentence** — L322-326. b28ca4dd here inserted
   `AllEqual/holes` into the KnownTrip-to-Clean list, making it six; "`GlobalCardinality` mattered
   most of the five" was written by #860 and not updated.

3. **"Two of the three are left" (L491-492) contradicts all three bullets being marked Done** — and
   also L494's "the survey currently supports **no** VeriPB feature request at all", and the results
   table's single steps-only row. 63454f52 (#858) added it when `Element` was the first of the three
   done; #860 and this PR completed the other two.

Everything quantitative in the document that I could check does hold, for what it is worth — I re-ran
the survey on the trial merge and got `Among` 98, `Table` 93, `AllEqual/holes` 37,
`GlobalCardinality` 63, `GlobalCardinality/hall` 33984 -> 339984, `Element` 108, `Cumulative` 19046 ->
190046, `Regular` 18042 -> 180042, and the `Regular`/`NValue` row breakdowns add up exactly
(1218 + 600 + 23 = 1841; 1218 + 213 = 1431). So it is only the prose counts that drifted.

Three much smaller things, take or leave:

* L509-511: "Eight constraints ... All five are filed and parked under tracker #846" — five issues,
  eight constraints, and the referent of "All five" is the sentence before last.
* L614 gives `Cumulative` as 190045 rows at `0..10000` where the survey and L440 both say 190046 at
  10^4. Probably just the off-by-one between `0..10000` and 10^4 rather than an error.
* `Modulus`'s proof steps are not flat either — 77 -> 85, 1.1x — which the "neither | 1.0x / 1.0x" row
  and the `Multiply`/`Divide`/`Modulus` paragraph below it don't mention (that paragraph covers only
  the 1.8x OPB growth).

The audit lane itself is green on the trial merge: 140 assertions, 19 `KnownTrip` / 37 `Clean` / 14
`NoWidePosition`, matching the table exactly.

*(From the automated pre-merge checklist review; not merged, pending the flake fix and your own review.)*

ciaranm:
Both fixed. Thank you — the flake is a real one and I had no way of seeing it.

**The flake.** Reproduced exactly as you describe before touching anything: `ctest -R all_equal -j 8` failed **5 rounds in 6**. Gated `run_mixed_witness_test` on `view_wrap_config_is_effectively_bare` next to `run_holes_test`, which is the option you preferred and I agree — the test builds bare variables, so the view lanes only ever ran it a second time with nothing wrapped, and folding the label into the proof name would have kept a pointless duplicate run. **0 rounds in 6 after the gate**, and 0 in 8 while iterating. Coverage is unchanged: the plain lane still runs it, and I checked it still reaches the witness-splitting branch it exists for (2 hits).

That fix is in **1603d880 on this branch**, not at the tip, since the test is this PR's own and I would not want #861 landing alone with it. #862 was the only branch above and rebased cleanly; #861's push was a fast-forward, so nothing was rewritten. Verified standalone on this branch too: 804/804 and 0-in-4 on the race check.

The part worth recording is why I never saw it: the suite passed 804/804 on every run this stack made, because among 804 tests the two `all_equal` lanes rarely overlap. `-R all_equal -j 8` makes them collide almost every time. A green parallel suite is not evidence that two lanes sharing a filename are safe — that is in the commit message.

**The doc counts.** All five fixed at the tip as you suggested (**7ad4a452** on #862), including the two you offered as take-or-leave. The `Multiply`/`Divide`/`Modulus` paragraph now also names `Modulus`'s 77 → 85 steps, and says explicitly that the `neither` row's header is a threshold rather than a measurement — that was the actual bug in how it read.

**One I did not change: `Cumulative`'s 190045 at `0..10000`.** You guessed it was the `0..10000` versus 10^4 off-by-one, and I think that is right, so I reverted my first attempt at `fixing` it. They are different instances: the same passage gives 1945 rows at `0..100`, which does not scale to the survey's 19046 at 10^3 either. It is #780's figure on #780's instance and I have no evidence it is wrong — changing a cited number I cannot verify seemed worse than leaving it.

Every count now agrees with a fresh run: 70 probes registered, 140 assertions, 70 survey rows, 19/37/14, and 61 + 7 + 1 + 1 = 70.

Also glad the survey figures reproduced on your trial merge — that is the check I most wanted someone else to run independently.

PR #862, opened 2026-09-07
Reuse Among's variable partition instead of recomputing it (#833)

ciaranm:
Part of #833, stacked on #861. Improves code introduced in #856 — squash it into that if you
would rather, it is independent of everything between.

This started as a question about whether `domain_is_inside_values_of_interest` could be made
faster with an `IntervalSet` and a set difference. **It can, asymptotically, and it makes no
measurable difference at all** — but profiling to establish that found something that is
worth 4x.

## What the profile said

`domain_is_inside_values_of_interest` is **0.75%** of the propagator's time.

What dominates is that two places were recomputing what the variable partition at the top of
the propagator had just worked out:

- a variable is not wholly inside the value set **exactly when it is not a `must_match` one**;
- it still has a value of interest **exactly when it is not a `must_not_match` one**.

Both were recomputed per variable per call by walking the value set, so the propagator paid
`O(|voi|)` twice over for facts already in hand. Iterating the partition's own subranges
costs nothing.

## No inference changes, in the strong sense

Not merely "the same removals": at a fixed seed the **OPB, the proof and the test output are
byte-identical** before and after. Skipping the `must_not_match` variables skips only
inferences that were no-ops — those variables have no value of interest left to remove — and
a no-op inference emits no proof line.

## Measurements

Pinned, three repeats, node counts identical throughout:

| value set | before | after | |
|---|---|---|---|
| 40 values | 7157 ms | **3829 ms** | 1.9x |
| 400 values | 47546 ms | **17391 ms** | 2.7x |
| 5000 values | 11316 ms | **2890 ms** | 3.9x |
| wide domain | 12 ms | 12 ms | — |

The third commit then takes those further: 3795 → **2281** ms at 40 values, 17370 → **5878**
at 400, 2909 → **1816** at 5000.

## The interval version, and a correction

This PR originally said the `IntervalSet` version measured "exactly neutral" and left it
out. **That was wrong, and the measurement was worthless.** The change carried a compile
error — `move` on a `gcs::` type is not found by ADL, where `move` on a `std::vector` is —
and the build check in use was `grep -c " error "`, which matches neither compiler's output.
So the library never rebuilt, the benchmark linked the old one, and comparing a binary
against itself gave the answer it gives for any change: no difference.

Done properly, in the third commit here, it is **1.7x / 3.0x / 1.6x** at 40 / 400 / 5000
values. `State` now answers both of Among's domain questions at interval level —
`domain_intersects_with()` and a new `domain_is_subset_of()`, built on a new
`IntervalSet::contains_all_of()` — and `among.cc` uses none of `State`'s per-value iterators
at all. Proofs stay byte-identical.

## Testing

804/804. `among_test` uncapped on GCC and clang, plain and view-wrapped. Audit lane green.

### DIFF
--- a/dev_docs/large-domains.md
+++ b/dev_docs/large-domains.md
@@ -61,6 +61,23 @@ Neither the audit lane nor the test suite can tell these two apart — both are
 the loop for its exit condition before reaching for an interval rewrite, and put
 any rewrite of a hot query through a before/after benchmark.
 
+**And before optimising a predicate, check it is the cost.** `Among`'s
+`domain_is_inside_values_of_interest` looks like the obvious thing to speed up, and
+a profile puts it at **0.75%** of the propagator's time. What the profile does show
+is that the two places asking it were recomputing what the variable partition had
+just worked out: a variable is not wholly inside the value set exactly when it is
+not a `must_match` one, and it has some value of interest exactly when it is not a
+`must_not_match` one. Iterating the partition's own subranges instead is **1.9x**
+at a 40-value set, **2.7x** at 400 and **3.9x** at 5000, for no change to a single
+inference — the proof is byte-identical.
+
+The rewrite that did *not* pay is worth recording next to it, because it was the
+first idea and it is a reasonable one: taking the surviving predicate against an
+`IntervalSet` instead, whether by set difference or by `State::domain_intersects_with`'s
+no-copy merge-walk. It measures as *exactly* neutral — at 40 values, at 5000, and
+on a shape built so the `none_of` cannot short-circuit at all. Asymptotics are not
+a reason on their own; the early exit already bounds it in practice.
+
 **The cheapest H1a of all is where the conclusions are already intervals and only
 the search for them is per-value.** `In`'s constant-set filter emitted one
 `infer_not_in_range` per maximal run of forbidden values, and found those runs by
--- a/dev_docs/large-domains.md
+++ b/dev_docs/large-domains.md
@@ -340,7 +340,7 @@ table, which is a snapshot for orientation.
 as `KnownTrip` and are now `Clean`, by the interval rewrites rather than by a
 weaker arm: all of them still propagate at their original strength.
 
-`GlobalCardinality` mattered most of the five, because it was the rule's own
+`GlobalCardinality` mattered most of the six, because it was the rule's own
 counterexample: already the bounds arm, and still enumerating, so it had nothing
 weaker to fall back to. Its just-met-demand branch forces a variable to a value,
 which is two range removals however wide the domain is. **A second per-value site
@@ -446,8 +446,9 @@ separately, because they mean different things:
 
 ### Results at 10^3 → 10^4
 
-Re-measured over all 69 probes after the interval rewrites for `ArrayMinMax`,
-`Table` and `Among` landed. The figures move, so re-run the survey rather than
+Re-measured over all 70 probes after the interval rewrites landed for
+`ArrayMinMax`, `Table`, `Among`, `Element`, `In`, `GlobalCardinality` and
+`AllEqual/holes`. The figures move, so re-run the survey rather than
 quoting this table after touching any propagator's removal loop — that is how the
 previous version of it went stale, see below.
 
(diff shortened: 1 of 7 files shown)


ciaranm:
Pre-merge review, at `39607acb`. The refactor holds up and I could not fault the reasoning; one
documented claim does not survive being measured, plus a small tidy.

First, confirming this PR's own claims on a trial merge of the whole stack: full build **100%, 0
warnings** (GCC 15.2, baseline 0), **804/804** ctest, and the **audit lane green** — 140 assertions,
19 `KnownTrip` / 37 `Clean` / 14 `NoWidePosition`.

I also checked the part that could have gone wrong, since the refactor's correctness rests entirely
on the partition subranges being equivalent to the predicates they replace. They are, and the
argument holds for a reason worth stating: between the partition and either loop only `how_many` is
inferred on, so no variable's domain has moved, and the `must_match_count == at_most_how_many` branch
returns `DisableUntilBacktrack` before the second loop can run. Dropping `must_not_match_vars` from
the first loop drops only no-ops, and inside the second loop the removals delete only values outside
`voi`, so `might_match` cannot flip part-way through even when the same variable appears twice in
scope.

## `dev_docs/large-domains.md:75` — "the proof is byte-identical" is not true

Measured rather than argued, which seems the right register given the sentence. Fixed seed,
`GCS_PRESERVE_PROOF_FILES=all`, `among_test --seed=424242`, comparing the stack at #861's tip against
this PR's tip, 124 proof artefacts across 31 instances:

* test output (`run.log`) — **identical**
* all 31 `.opb` — **identical**
* all `.varmap` / `.scp` — **identical**
* **4 of 31 `.pbp` — differ**

The differences are pure ordering, not content. `among_test_w0_pall.0001.pbp` is 509416 bytes and
10073 lines on *both* sides; a four-line block moves:

```
< rup 1 ~i[_2][eq2] ... >= 1;
< rup 1 ~i[_2][eq6] ... >= 1;
< del id -3;
< pol @c[_1][le] -1794 + ;
---
> pol @c[_1][le] -1794 + ;
> rup 1 ~i[_2][eq2] ... >= 1;
> rup 1 ~i[_2][eq6] ... >= 1;
> del id -3;
```

and one relative constraint-id reference shifts as a consequence (`pol @c[_1][le] -1935 +` becomes
`-1934`). Cause: both loops now iterate the *partitioned* subranges where they used to iterate `vars`,
and `std::ranges::partition` is not stable, so which variable's block is emitted first can change.
Every proof still verifies, and the same four files differ whether I stop at `34f8ffff` or go to the
tip — so `39607acb` neither adds nor removes any difference.

So the substantive half of that sentence is **confirmed**: no inference changes, the encoding is
untouched, the search is unchanged. It is only "byte-identical" that overreaches. Worth correcting
rather than shrugging at, because this document uses byte-identical proofs as a *technique* — a
false instance of it makes the true ones harder to trust. Either soften the wording to something like
"the same proof steps, though the partition reorders which variable's block is emitted first", or
switch to `stable_partition` if byte-identity is actually the property wanted.

**Scope note:** this is only the claim at L75, about the partition reuse. The separate byte-identical
claim at **L104**, about `In`'s constant-set filter, is a different rewrite and I have **not** tested
it — nothing here casts doubt on it.

## Tidy: two leftover brace blocks

Removing the `if`s left a redundant nested block at `among.cc:232-233` and `294-295`:

```cpp
for (const auto & var : can_be_either_vars) {
    {
        vector<Literal> inferences;
```

Legal, and clang-format is happy either way, but it reads as though a condition were still there.

## Worth saying out loud

`39607acb` correcting its predecessor's "exactly neutral" measurement — and diagnosing *why* the
measurement lied, rather than just replacing the number — is the most useful thing in this PR. A
build check of `grep -c " error "` matching neither compiler's output, so the library never rebuilt
and the benchmark compared a binary against itself, is exactly the trap our own checklist records as
"gate every run on the build having actually succeeded", and it is one that returns a plausible
answer rather than an obvious failure. Checking build exit status is the right fix.

*(From the automated pre-merge checklist review; not merged.)*

ciaranm:
Both fixed, in **f3511392**. You were right on the substantive point and I had believed it on evidence that could not have shown otherwise.

## The byte-identical claim

Reproduced your result exactly before changing anything: 4 of 31 `.pbp` differ, all `.opb`/`.varmap`/`.scp` identical, test output identical. I looked at the content too — the only *non*-reordering differences are the `pol ... -N +` relative back-references, which count lines backwards and shift when a block moves. Byte and line counts are unchanged on every file.

**I tried `stable_partition` rather than assuming, and it does not help** — still 4 files, a slightly different four. The reason is that the grouping *is* the reordering: the loops used to walk `vars` and skip, which interleaves must_match and can_be_either variables, where any partition puts one group entirely before the other. Stability within each group does not undo that. So I reverted it rather than pay for it, and took your first option: the wording now says what holds — same OPB, same varmap, same scp, same test output, same node counts, every proof still verifying — and says why byte-identity is the wrong property to ask for here.

**Why the check believed it is the part worth having.** `GCS_PRESERVE_PROOF_FILES=1` keeps only the *last* instance's files, since every case writes to the same basename — so a run over 31 instances left six artefacts out of 124 for me to compare, and the four that moved were not among them. Same shape as the stale-binary error one commit earlier: a check that looked decisive while covering a fraction of the ground. The document now says to use `=all`.

Your scope note prompted the other thing worth reporting. The **`In` claim at L104 rested on exactly the same weak evidence**, so I re-checked it the strong way rather than leave it standing on a method I now distrust: **byte-identical across all 212 artefacts**. It holds, and the document now says how it was checked.

## The brace blocks

Removed, both of them. You are right that they read as though a condition were still there.

## On the equivalence argument

Thank you for working through it independently — that between the partition and either loop only `how_many` is inferred on, and the first branch returns `DisableUntilBacktrack` before the second can run, is the part I would most have wanted a second pair of eyes on, and I had not spelled out the aliasing case (`might_match` cannot flip part-way through because the removals only delete values outside `voi`).

804/804, 0 warnings on all four builds, `among_test` uncapped on GCC and clang plain and view-wrapped, audit lane green at 140 assertions.

Issue #819, opened 2026-09-04
NotEquals wakes on_change through a shared Triggers, but its enforce pass only reads optional_single_value


ciaranm:
The tail of #807 asks the same question of the other propagators that can only act on a fixed variable. `ReifiedEquals`'s must-not-hold pass is the next one, and it is the same shape:

```cpp
auto enforce_constraint_must_not_hold = [...](const State & state, auto & inference, ...) -> PropagatorState {
    auto value1 = state.optional_single_value(v1);
    if (value1) { ... infer_not_equal_or_stop(logger, v2, *value1, ...); return DisableUntilBacktrack; }
    auto value2 = state.optional_single_value(v2);
    if (value2) { ... infer_not_equal_or_stop(logger, v1, *value2, ...); return DisableUntilBacktrack; }
    return PropagatorState::Enable;
};
```

It reads nothing but `optional_single_value`, and disables itself until backtrack once it has acted — so, exactly as in #807, `on_instantiated` cannot miss its wake. But `install_propagators` registers

```cpp
Triggers triggers;
triggers.on_change = {_v1, _v2};
```

for all four reification branches at once, and the undecided pass genuinely needs `on_change` (it reads `in_domain` and `domains_intersect`).

## Why this one is likely to matter more than #807's

`NotEquals` is what the graph-shaped examples post per edge — `colour`, `ortho_latin`, `sudoku`, `hitori`, `seat_moving`, `crystal_maze`, `tour`, `circuit_random` — and the wasted wakes there are quadratic in the degree, not linear in the constraint count. Fixing a vertex removes one value from each of its `d` neighbours; every one of those removals is an interior change that wakes all of *that* neighbour's `NotEquals` constraints, each of which finds neither endpoint fixed and returns having done nothing. With `on_instantiated` the whole cascade disappears, since the only wake that can matter is the instantiation itself.

Unlike #807's `lin_not_equals` on `tpp`, this one has no contradiction-heavy path to hide behind: `NotEquals` contradicts only when both sides are fixed and equal, which is a leaf.

## What has to change

`install_reified_dispatcher` already splits at install time — when the condition is decided it installs a propagator that runs *only* the relevant enforce pass — so the split is there, it is just that the one `Triggers` argument is shared. It needs a way for the caller to hand a narrower trigger set to the decided must-not-hold branch. The undecided branch must keep the union: its propagator can take any branch at runtime.

[...]

Issue #479, opened 2026-07-09
cake validation: fully-reified (_iff) linear/comparison fail to chain-verify


ciaranm:
## Goal

Direct closer investigation. Fully-reified (`_iff`) linear and comparison
instances fail the cake-validation chain. This records the failing cases and how
to reproduce; the aim is to determine specifically what diverges and where the
fix belongs.

## Workflow context

The solver writes `.scp`/`.pbp`, `cake_pb_cp` re-derives cake's OPB from the
`.scp`, and `veripb --elaborate` checks our proof against cake's OPB. Encodings
are not byte-identical; the chain relies on our proof lining up with cake's OPB.
(See the harness README.)

## Observed failures

`veripb` aborts while parsing our proof because it cites a constraint **label**
that cake's re-derived OPB does not assign to a constraint:

```
The label `@c[_1][lt]` is not assigned to a constraint ID … pol @c[_1][lt]
```

Representative case (`lin_less_equal_iff`):
```
( ((_1 0 2) (_2 -2 2) (_3 0 5) (c 1 1))
  ((_1 lin_less_equal_iff (c = 1) (1 _1 2 _2 3 _3) 6))
  (enumerate) )
```

Also on `linear_equality_test` modes `ge_if`, `ge_iff`, `le_if`, `le_iff`,
`ne_iff`, `ne_iff_gac`, and on `comparison` `_iff` modes. The non-reified and
half-reified (`_if`) forms of the same constraints chain-verify — it is specific
to the fully-reified `_iff` encoding.

## Reproduce

```
GCS_TEST_CAKE=1 ./build/linear_test le_iff     # CAKECHAIN … FAIL_ELAB
# inspect a preserved triple:
veripb <name>.<n>.cakeopb <name>.<n>.pbp --elaborate /tmp/core.pbp
```

## Scope

All fully-reified `_iff` linear/comparison instances.

ciaranm:
Investigation results (verified on minimal API-built instances plus authentic `linear_test`/`comparison_test` runs; probe artifacts under `~/claude/tmp/issue479/`).

## The failing set

Exactly: linear **`le_if`, `ge_if`, `le_iff`, `ge_iff`, `ne_iff`** (and the gac variant, which shares the encoding). Everything else this issue implicates passes. Authentic failure rates: `le_if` 31/34, `le_iff` 32/34, `ne_iff` 68/68 provable instances (the few OKs are proofs that happen never to cite the constraint's labels).

[...]

PR #863, opened 2026-09-07
Constraint family documents: the template, equals as the pilot, and its re-audit

ciaranm:
Three commits: the per-family document template, the first family document written to it, and — new — the re-audit of that document against the code that the first pass's findings changed. Plus a fourth reframing the mutation-coverage policy, after review, and a fifth answering @mmcilree's review.

**Review round 1 is addressed** — see the review thread for the full response. In short: 14 of the 16 rules across the two audited families asserted a bare `RUP` with no argument. The arguments turned out to be published already, in @mmcilree's thesis, so a new `dev_docs/justification-techniques.md` collects the four unit-propagation facts (Thm 2.6–2.9), the completeness invariant behind them (`Inv1`, Thm 3.3), the `B ∈ {0,1}` precondition and its counterexample (Example 2.15), and the procedures we instantiate (JP 3.1, 3.2, 3.12) — and every rule now cites its procedure and theorem instead. Where a rule *departs* from a published procedure it says so outright, which is the distinction an external justifier needs; `equals`'s disjointness witness is the arc's first. `⟦v⟧` became `BinEnc(v)` to match the thesis, the at-least-one rows are written as implications, and "lemma" stays with the template now recording why it is deliberately undefined.

Tracker #871 has been brought into line too.

Rebased onto `main` at `76bfd836`.

## What changed since you last looked

The first pass filed seven issues (#864–#870) and this document was the argument for them. **Six are fixed**, one more arrived from outside the audit (#889), and four of the seven changed what the document *says* rather than only what the code does — so it was out of date in substance, not just in its numbers.

| Fixed | What it changed in the document |
|---|---|
| #864 → #873 | the no-overlap reason is guarded on `want_reasons()` |
| #865 → #883 | `clone()` carries `_neq`, so all six variants write their own `.scp` keyword |
| #866 → #884 | the interval bridge has its own subhint, so the wire inventory is three forms, not two |
| #867 → #881 | the no-overlap certificate is **interval-wise** |
| #869 → #886 | five mutation lanes and a control |
| #870 → #885 | no justification in the family reads `state` |
| #889 → #894 | a constant-operand reified equals wakes on a refined watch |

Two of those are load-bearing rather than tidying. The **interval-wise witness** removed the family's only large-domain failure and its only super-logarithmic cost, which is most of what "Robustness and limits" used to be about, and it is the one thing in this family that is a proof *technique* rather than an application of one — so it is now argued in Prior art as novel. The **trigger change** is the family's one real performance lever pulled in the opposite direction from #819: finer than any coarse trigger can express, rather than coarser.

The rule catalogue took most of the weight, which is the structural bet holding up — three of the seven fixes are changes to one rule's derivation and each landed in that rule's entry plus at most one section of shared context. The document says where it did *not* hold up, too: `clone()` touched four sections, because a flag deciding how a constraint describes itself is not a property of any rule.

## Everything was measured again

At `76bfd836`, Release + `GCS_WERROR=ON`, g++ 15.2.0, Ryzen 9 9950X3D, verified idle, `taskset`-pinned, min of three. Where a figure moved, the document says so and why.

**The no-overlap rule, which is what #881 changed.** One `EqualsIff` over two bounds-disjoint domains, root propagation only:

| width | proofs off | proof lines | proof size | VeriPB |
|---|---|---|---|---|
| 10³ | 0 ms | 13 | 965 B | verified |
| 10⁴ | 0 ms | 13 | 1.2 KB | verified |
| 10⁶ | 0 ms | 13 | 1.7 KB | verified |
| 10⁹ | 0 ms | 13 | 2.6 KB | verified |

Thirteen lines at every width, of which **two are the witness** — one lemma and the conclusion — six define the two order literals they mention, and five are proof preamble. Against `std::bad_alloc` at 10⁹ before.

**Its per-run cost**, which nothing had measured: `v1` the even values of `[0, 2n)` and `v2` the odd ones, so every run is one value long and every second run costs lemmas — the walk's worst case.

| runs (n) | total lines | witness | shared literal layers |
|---|---|---|---|
| 2 | 25 | 4 | 21 |
| 8 | 97 | 16 | 81 |
| 32 | 385 | 64 | 321 |
| 128 | 1,537 | 256 | 1,281 |

Exactly `2n` against `10n + 1`, so **five sixths of the proof is not this rule** even in the shape that costs it most. The first pass asserted that the shared layers dominate; this is the number, and wanting it is one of the template changes.

**CPU.** `ortho_latin --all 6` is 20.64 s against 20.49 s, at an identical search tree to the digit. That 0.7% is not noise and is accounted for: #894 put the watch-index test on the inference replay path, which its own `perf stat` measured at +0.69% of this model's instructions. Hoisting it out is #895.

**Proof.** `ortho_latin --all 5` is byte-identical at all three assertion levels — none of the six fixes touched what this model emits. VeriPB 1.83 s against the first audit's 2.30 s for the same proof; only this run was pinned on a verified-idle machine, so it is the one quoted and the discrepancy is stated rather than smoothed over.

**And one finding that came out of reading those assertions back.** All 13,323 equals-hinted assertions are the bare wire form, and it is not because no subhint exists: every literal in all of them is an `eq` atom, 13,309 being a two-literal clause with both literals negated at the same value and 14 a single positive literal. That is rule 5 and rule 1-against-a-constant, and nothing else. **`ortho_latin` exercises two of the family's nine rules in its proof** — no bound push, no interval bridge, no disjointness walk. That was true at the first audit too and went unsaid, and it is exactly what a cross-family pivot over these documents would silently average over.

## Cross-solver CPU performance is no longer `Not measured.`

#888 built the harness #868 asked for, and Gecode is measured on both arms — `queens` for unreified (0.93x–0.97x, level with Gecode) and `nmseq` for reified (6.2x–6.4x before #894, 2.6x after). Both models come from `minizinc-benchmarks` rather than being posted by us, which is the part worth reusing: the natural gcs benchmark for this family is a disequality clique, and no other solver would be given one. #868 stays open for Choco and ACE.

## Three template changes

All things this rewrite wanted and could not find, which is the same test the pilot applied.

- **A re-audit is a distinct event**, and the status line carries both dates plus an `issue → PR → what it changed here` table. A reader needs to know which commit each figure was taken at.
- **A rule entry gains a `Tightness` field.** The template asked under Tests "whether the derivation has been shown to be tight", which is the right question at the wrong altitude — a mutation lane corrupts *one rule's* step. At family altitude this one could answer "yes, tight" and stop, where three of its five lanes are one rule and six of its nine rules have no lane at all. The field is there so someone changing a rule can see whether a lane will catch them; `Not shown.` is an ordinary answer and **not** a coverage target, which the template now says outright.
- **The own-versus-shared proof-size split wants a measurement**, not an estimate. "2n against 10n + 1" is a finding; "the shared layers dominate" is a guess that happens to be right.

Plus, under CPU performance, a requirement to say what the chosen benchmark does **not** exercise, read off the proof's assertions rather than guessed.

## One finding, recorded rather than actioned

`with_proof_mutation` reaches `enforce_equality`, so `fixed_operand_reason` corrupts rule 1's reason but **not rule 5's** — the busiest rule in the solver, and 13,309 of the 13,323 equals-hinted assertions in the family's own proof benchmark. The corruption already exists; only the lambda's capture list and one lane registration are missing.

Not filed and not on the ranked list, per your call that mutation coverage is a development tool rather than something to have everywhere for its own sake. It sits under "Not to do, having been considered" with enough detail to act on if that rule is ever touched, and the third commit reframes the template the same way: `Not shown.` is an ordinary answer, partial coverage is the expected state, and the remainder is not carried as outstanding work.

## Checks

Docs only; no code touched. `ctest` **810/810 with nothing skipped**, including the six `equals_mutation` lanes and their control. `equals_test --seed=1` is **296 proofs verifying in 1.13 s**. Every table caption carries its commit, date and machine, and the two sets of figures measured elsewhere (#819's and #889's pre-merge A/Bs, and the cross-solver snapshot at `238c7836`) are in their own labelled paragraphs rather than beside fresh ones.

Two errors in #894's own PR body are corrected here rather than propagated: `magic_series` lives in `minicp_benchmarks/`, not `examples/`, and takes its size positionally — it has no `--size` and no `--stats`. Verified by running it: 1,193 recursions, 15,926,749 propagations.

### DIFF
--- a/dev_docs/constraints/equals.md
+++ b/dev_docs/constraints/equals.md
@@ -1,0 +1,905 @@
+# `Equals`: two operands are (or are not) the same value
+
+> **Maturity** production ·
+> **Audited** 2026-09-07 at `7d014207` ·
+> **Open issues** none open specific to this family; see [Next
+> steps](#next-steps) for the three this audit would file
+
+Six posted classes over one implementation: an equality between two operands,
+optionally reified, optionally negated. It is the smallest interesting
+constraint in the solver and one of the busiest — 71% of all propagator calls
+on `ortho_latin --all 6`, and 44% of the assertions in its proof — so its cost
+per call and its trigger set matter more than its algorithm does.
+
+## What it is
+
+### Semantics
+
+`Equals(v1, v2)` holds iff `v1 = v2`. `NotEquals(v1, v2)` holds iff `v1 ≠ v2`.
+The four reified forms take an `IntegerVariableCondition` `cond`:
+
+| Class | Semantics |
+|---|---|
+| `Equals(v1, v2)` | `v1 = v2` |
+| `NotEquals(v1, v2)` | `v1 ≠ v2` |
+| `EqualsIf(v1, v2, cond)` | `cond → v1 = v2` |
+| `EqualsIff(v1, v2, cond)` | `cond ↔ v1 = v2` |
+| `NotEqualsIf(v1, v2, cond)` | `cond → v1 ≠ v2` |
+| `NotEqualsIff(v1, v2, cond)` | `cond ↔ v1 ≠ v2` |
+
+All six are the one class `ReifiedEquals`, distinguished only by the
+`ReificationCondition` its constructor passes down. The `NotEquals*` variants
+do **not** flip any propagation logic; they negate the condition at
+construction, so `NotEqualsIff(v1, v2, c)` is stored as `Iff{¬c}` and reaches
+the same equality-enforcing code. Read the concrete variant's constructor to
+see what a given class actually installs; the `_neq` member is not the place to
+look (see [Known limitations](#known-limitations)).
+
+Degenerate cases:
+
+- **Aliased operands.** `NotEquals(x, x)` on a non-constant `x` throws
+  `InvalidProblemDefinitionException` at construction. The other five accept
+  aliasing; `EqualsIff(x, x, c)` correctly forces `c`, and `NotEqualsIf(x, x,
+  c)` correctly forces `¬c`, via the alias check in the undecided pass.
+- **Constant operands.** Two constants are a valid model, including two equal
+  constants under `NotEquals` — trivially infeasible, but a model, so it is not
+  rejected. Only genuine variable aliasing is.
+- **Views.** Accepted in either position, at a cost — see [Variable kinds and
+  views](#variable-kinds-and-views).
+
+### Concrete constraints and frontend coverage
+
+| C++ class | MiniZinc / FlatZinc | XCSP3 | CPMpy | SCP `s_expr` | Notes |
+|---|---|---|---|---|---|
+| `Equals` | ✓ `int_eq`, `bool2int`, `bool_eq`; also `decompose`[^lin] | ✓ `intension` eq (n-ary, chained pairwise), `instantiation`, top-level `iff` | ? | ✓ `equals` | |
+| `NotEquals` | ✓ `int_ne`, `bool_not`; also `decompose`[^lin] | ✓ `intension` ne | ? | ✓ `not_equals` | |
+| `EqualsIf` | frontend gap — no FlatZinc predicate maps to a half-reified equality | ✓ `ifThenElse` arms | ? | ✓ `equals_if` | |
+| `EqualsIff` | ✓ `int_eq_reif`, `bool_eq_reif`, `int_ne_reif`[^neg] | ✓ `intension` reified EQ, `iff` | ? | ✓ `equals_iff` | |
+| `NotEqualsIf` | frontend gap — as `EqualsIf` | unsupported | ? | ✓ `not_equals_if` | |
+| `NotEqualsIff` | ✓ via `EqualsIff` with a negated condition | ✓ `intension` reified NE | ? | ✓ — but writes as `equals_iff`[^flip] | |
+
+[^lin]: MiniZinc rewrites `x = y` and `x != y` between two variables into a
+    two-term `int_lin_eq` / `int_lin_ne`, emitting `int_eq` / `int_ne` only for
+    reified forms. Across 75 sampled challenge instances: 13,605 two-term
+    unit-coefficient `int_lin_eq` in 37 of them and 24,636 `int_lin_ne` in 6,
+    against **zero** `int_eq` and 125 `int_ne`. `fzn_glasgow.cc` therefore
+    recovers the two-variable shape from any two unit coefficients and posts
+    `Equals` / `NotEquals` instead of `LinearEquality` / `LinearNotEquals`.
+    Without that recovery this family would be almost unreachable from
+    MiniZinc. For `_eq` the recovery is a strength gain as well as a speed one:
+    the generic linear equality is bounds(Z) and cannot carry a hole across,
+    where `Equals` intersects the domains. For `_ne` both are GAC, so it is
+    purely the cheaper propagator.
+
+[^neg]: `int_ne_reif` posts `EqualsIff{v1, v2, reif != 1}` — the negation goes
+    into the condition rather than into a different class.
+
+[^flip]: The `not_equals_iff` keyword exists in the reader and in
+    `verified_encodings/scp_cases/not_equals_iff_sat.scp`, but the writer never
+    emits it. See [Known limitations](#known-limitations).
+
+### Options
+
+`None.` The family has no tunables: no consistency-level knob, no algorithm
+selection, no incrementality threshold. Everything the six classes differ by is
+a constructor argument, not an option.
+
+This is worth stating rather than omitting, because the neighbouring linear
+family does have `with_consistency` and `with_incremental_threshold`, and a
+reader coming from there will look for the equivalent here.
+
+### Variable kinds and views
+
+Both operands are `IntegerVariableID`, so plain variables, constants and views
+are all accepted, and the propagator is instantiated over the `visit` of both
+operand kinds.
+
+Views are where the proof and the propagator diverge:
+
+- **Propagation** treats a view like anything else; the inferences go through
+  the same `infer_*` calls and the view layer translates them.
+- **Proof.** One rule — the interval-wise symmetric difference — is only
+  available when *both* operands are a bare `SimpleIntegerVariableID`. It is
+  guarded on exactly that (`both_simple`), and a view or constant in either
+  position falls back to a per-value rule that is correct but emits one
+  inference and one proof line per removed *value* instead of per removed
+  *interval*. The reason is a proof-layer limitation, not a propagation one: the
+  bridge lemmas that carry a range literal across the equality need a plain
+  variable on the far side.
+
+The view sweep in the test binary exercises both positions
+(`add_view_tests(equals_constraint equals_test 2)`), so the fallback path is
+covered; it is not free, and a model built entirely out of views pays the
+per-value cost.
+
+### Reification
+
+All four reified forms are first-class here rather than bolted on: the class
+*is* `ReifiedEquals` and `Equals` / `NotEquals` are the degenerate
+`MustHold` / `MustNotHold` cases. The machinery is
+`install_reified_dispatcher` (see `dev_docs/reification.md`), which installs a
+single propagator and dispatches on the condition's current state.
+
+Two things are specific to this family:
+
+- The condition's own literal is passed into every pass and appears in every
+  reason. The propagator never consults the reification *policy* — which
+  literal to infer for which verdict is the dispatcher's decision, driven by
+  `evaluated_reif::Undecided`'s three flags.
+- When the condition is already decided at install time, the dispatcher installs
+  a propagator that runs one enforce pass directly, with no per-call
+  `test_reification_condition`. That is the path `Equals` and `NotEquals`
+  always take, and it is also what a FlatZinc-generated constraint with an
+  already-fixed reification literal takes.
+
+### Relation to other families
+
+**Decomposes into this family.** The not-equals clique encoding of an
+all-different — one `NotEquals` per pair — is the source of essentially all the
+call volume in [CPU performance](#cpu-performance). Note that this is the
+*model's* choice, not `AllDifferent`'s: `ortho_latin --all-different
+not-equals` posts the clique itself in a double loop, and `AllDifferent` has no
+such mode. `gcs::AllEqual` likewise has its own propagator and does not
+decompose here.
+
+What does land here: XCSP3's `instantiation` (one `Equals` to a constant per
+variable), its n-ary `intension` eq (chained pairwise), its `intension` ne and
+reified eq/ne, its `ifThenElse` arms and top-level `iff`; FlatZinc's `int_eq` /
+`int_ne` / `bool2int` / `bool_eq` / `bool_not` and the two reified forms; and
+the two-term linear recovery in `fzn_glasgow.cc`, which is where almost all of
+the MiniZinc traffic arrives.
+
+**Posts as children.** Nothing. `ReifiedEquals::prepare` allocates nothing and
+posts nothing.
+
+**Shares code with.** `enforce_equality` is exported from `gcs::innards` and
+called by `Element` (`element.cc:699`) to tie a result variable to a selected
+array entry. Any change to its inference set or its return value is an
+`Element` change too — this is the one cross-family coupling in the file.
+
+**Presolvers.** None rewrite or target this family. `DifferenceLogic` scans for
+`x - y ≤ d` shapes and does not recognise an `Equals`.
+
+**Nearly, but not, the same family.** `gcs/constraints/comparison/` is a
+parallel hierarchy of twelve classes over `ReifiedCompareLessThanOrMaybeEqual`,
+with the same reified-dispatcher shape and the same cosmetic-negation-flag
+design — but a separate encoding, separate propagator and separate `hints.hh`.
+The template's family list flags `comparison`+`equals` as a candidate merge;
+**this audit says keep them separate.** They share a pattern, not code, and
+merging the documents would bury two distinct OPB encodings in one section.
+
+## The proof model
+
+### OPB encoding
+
+The operands are compared through their **bit** (binary) encodings, not their
+order literals. Writing `⟦v⟧` for the bit sum of `v`:
+
+```
+MustHold      (Equals)          ⟦v1⟧ - ⟦v2⟧ = 0
+                                  split into two rows, le and ge
+
+MustNotHold   (NotEquals)       ne                     ->  ⟦v1⟧ - ⟦v2⟧ >= 1
+                                ¬ne                    ->  ⟦v1⟧ - ⟦v2⟧ <= -1
+                                  with one fresh flag ne per constraint
+
+If            (EqualsIf)        cond                   ->  ⟦v1⟧ - ⟦v2⟧ = 0
+                                  split into le and ge, each half-reified
+
+NotIf         (NotEqualsIf)     gt                     ->  ⟦v1⟧ - ⟦v2⟧ >= 1
+                                ¬gt                    ->  ⟦v1⟧ - ⟦v2⟧ <= 0
+                                lt                     ->  ⟦v1⟧ - ⟦v2⟧ <= -1
+                                ¬lt                    ->  ⟦v1⟧ - ⟦v2⟧ >= 0
+                                lt + gt + ¬cond        >=  1
+
+Iff           (EqualsIff,       cond                   ->  ⟦v1⟧ - ⟦v2⟧ = 0
+               NotEqualsIff)      split into le and ge, each half-reified
+                                gt                     ->  ⟦v1⟧ - ⟦v2⟧ >= 1
+                                lt                     ->  ⟦v1⟧ - ⟦v2⟧ <= -1
+                                lt + gt + cond         >=  1
+```
+
+The encoding is **definitional**: every row states part of what the constraint
+means, and no row is a consequence. There is no scaffolding, no auxiliary
+integer variable, and no per-value row — so the encoding is **logarithmic in
+domain size**, which is why this family is one of the few that is comfortable
+at a domain width of 10⁹.
+
+Two details that only show up in the emitted file. The half-reification
+coefficient is the worst-case violation, so it scales with the domain width
+(`8 ~b[_1][ne]` for a width-4 domain); this is the reification layer's
+choice, not the constraint's, but it means the OPB's *coefficients* grow with
+domain size even though its *row count* does not. And `NotIf` reifies its two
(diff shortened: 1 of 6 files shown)


ciaranm:
@mmcilree I'm going to get Claude to do one of these for every constraint group. Partly it's to help the upcoming big "VeriPB can proof log all of CP" paper, partly it's so we actually audit every constraint carefully, and partly it's in the hopes that it'll be useful for your assertion-rewriter. Could you take a look and see if there are any additions / structural changes you'd like before I start on the rest?

ciaranm:
## Issues filed from the pilot

Tracked under **#871**, all labelled `family-docs (#871)`. The document now cites them where it previously said "unfiled", and its status line lists them.

**Specific to this family:**

| | | |
|---|---|---|
| #864 | `bug` | A reified `Equals` over wide non-overlapping domains builds a reason per value with proofs off, and dies at 10⁹ |
| #865 | `bug` | `ReifiedEquals::clone()` drops `_neq`, so the `not_equals_iff` s-expression fix from #479 never took effect |
| #866 | `enhancement` | The not-in-range bridge shares a wire hint with a bare RUP pruning, so a justifier must key off literal shape |
| #867 | | The no-overlap witness is per-value, so the rule is O(domain width) even when its proof *is* wanted — a research question, not a fix |
| #870 | `code cleanup` | `EqualsNoOverlap` holds a `State` pointer its justification does not need |

**Audit-wide, filed here because this is where they first bit:**

| | |
|---|---|
| #868 | No identical-search-tree comparison against Gecode, Choco or ACE. *Every* family document will have to leave that section unmeasured until this exists, so it is a prerequisite for the arc rather than an equals task. |
| #869 | No recorded mutation showing VeriPB refuses a mangled equals derivation. Also raises the standing question of whether tightness evidence is expected per family or gathered opportunistically. |

#864 and #865 are small and self-contained. #864 is the one worth doing first: it is a one-line guard whose pattern is already in the file's history (`fea9508d`), and it removes a `std::bad_alloc`.

## Tracker

**#871** carries what a clean session needs to pick this up: the goal (the solver in good shape for the paper rather than a fast pile of documents), the decisions already settled, which family is next and why `cumulative` is deferred, the method that produced the pilot — including that three of its four most useful findings came from *running* things rather than reading them — the probe and env-var recipes, and the traps.

## Where this PR stands

Ready for review, and paused there pending Matthew's comments on the template and the pilot. Nothing further planned on this branch until then; the next family (`element`) would be a separate PR.

dev_docs/constraints/equals.md
```diff
@@ -0,0 +1,1366 @@
...
+The template's family list flags `comparison`+`equals` as a candidate merge;
+**this audit says keep them separate.** They share a pattern, not code, and
+merging the documents would bury two distinct OPB encodings in one section.
+
+## The proof model
+
+### OPB encoding
+
+The operands are compared through their **bit** (binary) encodings, not their
+order literals. Writing `⟦v⟧` for the bit sum of `v`:
```
mmcilree: Hmm.. `⟦v⟧` is repurposing a standard notation for one thing (boolean condition evaluation) and using it to mean something else (bit sum evaluation). It's also awkward for humans to type. I would suggest `bits(v)` should be the convention when defining encodings like this.

dev_docs/constraints/equals.md
```diff
@@ -0,0 +1,1366 @@
...
+                                  with one fresh flag ne per constraint
+
+If            (EqualsIf)        cond                   ->  ⟦v1⟧ - ⟦v2⟧ = 0
+                                  split into le and ge, each half-reified
+
+NotIf         (NotEqualsIf)     gt                     ->  ⟦v1⟧ - ⟦v2⟧ >= 1
+                                ¬gt                    ->  ⟦v1⟧ - ⟦v2⟧ <= 0
+                                lt                     ->  ⟦v1⟧ - ⟦v2⟧ <= -1
+                                ¬lt                    ->  ⟦v1⟧ - ⟦v2⟧ >= 0
+                                lt + gt + ¬cond        >=  1
```
mmcilree: I would write this as cond -> lt + gt >= 1

mmcilree (changes requested):
Structure mostly good, but I would argue it is missing the most useful point: which is *why* the stated proof techniques will always work. E.g. the fact that X = Y and [x = 1] propagates [y = 1] with our careful literal setup is something we use all over the place, so it be worth stating and proving this everywhere its used rather than just asserting "RUP". 

Also I feel like 'lemma' is being used in a slightly inconsistent way in the explanation of "why it works". I would avoid using this term unless it is clearly defined somewhere what a lemma is (either in a VeriPB proof or as part of solver internals).

ciaranm:
## Review round 1 (@mmcilree)

All four points taken. The main one turned out to have a better answer than writing proofs, so it is worth saying what changed and why.

### "Missing *why* the stated proof techniques will always work"

Agreed, and it was the majority of the catalogue: **14 of the 16 rules across the two audited families asserted a bare `RUP`** with no argument, including your example.

Going looking for the argument, it is already proved — in your thesis. So rather than write our own, the documents are now an **index into the published ones**, which should also be more useful to the rewriter, since its job is to replay those procedures in place of `a` lines.

New **`dev_docs/justification-techniques.md`** collects, in the final accepted version's numbering:

- what the arguments rest on: the two kinds of atomic variable, the binary backbone, `Inv1`, and **Thm 3.3** (complete propagation of implied atomic literals within one variable) — which is what "our careful literal setup" actually names;
- the four facts — **Thm 2.7** (opposing bounds on one bit sum), **Thm 2.8** (an equality fixes every bit — your example), **Thm 2.9** (a bound crossing a difference row), **Thm 2.6** (a reified step reduces to its consequent);
- **where 2.9 stops**: the `B ∈ {0,1}` precondition, with **Example 2.15** as the counterexample to relaxing it. An equality half has degree 0 and a comparison row 0 or 1, so both are inside; a general two-term linear row is not, which is why the linear family needs JP 3.15 instead;
- the procedures we instantiate — **JP 3.1, 3.2, 3.12** — and the one we deliberately do not.

Each rule now cites its procedure and theorem. Two examples:

> **Proof technique** — `RUP`, by **JP 3.12 (equality propagation)**, whose correctness proof is **Theorem 2.8** applied twice: the negated conclusion propagates `v1 = v` and `¬(v2 = v)`, the first of which fixes every bit of `BinEnc(v1)`, the equality rows then fix every bit of `BinEnc(v2)`, and that contradicts the definition of `v2 = v`. This is the inference the thesis uses to introduce that procedure.

> **Proof technique** — `RUP+hints`. **No published procedure: this rule is ours.** The thesis states the same fact as **JP 3.13 (equality infeasibility)**, one RUP line per surviving value, which is what gcs emitted until #881 and is O(domain width). The interval walk below replaces it …

That second kind of entry is the one I think matters most for your tool: a rule citing a procedure can be rebuilt by something that knows the procedure, and a rule that departs cannot. So the template now requires the field to say which, and `equals`'s disjointness witness is the arc's first departure. `comparison` (#896) has none — all seven rules are JP 3.2 or a Thm 2.6 reduction to it — and its Prior art says so, with your own statement of why a comparison's justification is simple quoted in place of the argument the document had been making from scratch.

The template changed to match: **Appendix A** stops being a list of names and gains a column saying what licenses each; the **Proof technique** field must cite the procedure, the theorem, and the *preconditions*; and **Why it is true** gains an explicit prohibition on proof-line counts and VeriPB vocabulary.

Two small things fell out that are worth flagging:

- The `B ∈ {0,1}` boundary shows up empirically as well. The `equals` mutation lane that removes the bridge lemmas is **accepted** by VeriPB at 2 of the 13 widths tried — both putting an interval endpoint on a bit boundary, where the bound is one literal rather than a sum. A narrow fixture would have concluded the lemmas were unnecessary.
- **One thing I'd like a second opinion on rather than change.** `range_literals_spec.md` §5 says "a lone bound still cannot cross a bit-sum equality — the thesis Example 2.15 limit is real". Read strictly that looks wrong to me: an equality half is a difference row of degree 0, so Thm 2.9 covers it, and `equals`'s bounds-intersection rule is precisely a lone bound crossing an equality and verifies in 296 proofs. Example 2.15's middle row has degree 3, so what it rules out is *relaxing* `B ∈ {0,1}`, not this. My guess is the sentence means "one endpoint bound is not enough, you need the pair", which is true and is why the bridge emits two lemmas. I have **not** touched it — it is your paragraph and I would rather ask.

### "`⟦v⟧` is repurposing a standard notation"

Fixed, and we went with **`BinEnc(v)`** rather than your suggested `bits(v)` — same objection answered, and it matches the thesis, so anyone reading both sees one notation. Say the word if you'd rather have `bits`.

### "`lt + gt + cond >= 1`" → "`cond -> lt + gt >= 1`"

Fixed, both rows (`cond ->` for `NotIf`, `¬cond ->` for `Iff`). The surrounding rows in the same block were already implications, so the raw-sum spelling was inconsistent within one code block.

### "'lemma' is used inconsistently … avoid unless clearly defined"

**Keeping the term, and the template now records why it is deliberately undefined**: the thing it would name — a nameable, reusable derived statement — is a feature VeriPB cannot soundly have, so there is nothing formal to point at, and a local definition would only compete with the informal usage in ten other `dev_docs` files and several headers (`comparator_network.hh` most of all). If VeriPB ever grows a `lemma`, we change the terminology across the tree in one go rather than per document.

Your "inconsistent" did land on something real, though, and that is fixed: rule 9's **Why it is true** carried a "Lemmas" column, and that field is specified to be readable by someone auditing soundness without knowing what VeriPB is. The emission costs moved to **Proof size**, and the template now forbids the leak explicitly.

dev_docs/constraints/equals.md
ciaranm: Agreed on both counts. Changed — but to `BinEnc(v)` rather than `bits(v)`: same objection answered, and it is the thesis's own notation, so anyone reading the family documents alongside Chapter 3 sees one spelling. It is now a stated convention in `TEMPLATE.md` so later families do not diverge. Happy to switch to `bits` if you would rather have the shorter one — it is your notation complaint.

dev_docs/constraints/equals.md
ciaranm: Done. That one is the `NotIf` row, so it is `cond -> lt + gt >= 1` exactly as you wrote it; the `Iff` row alongside it is the same shape with the condition the other way round, so it is now `¬cond -> lt + gt >= 1`.

Worth noting the inconsistency this fixes was inside a single code block: every other row in it was already written as an implication, and only these two were left as raw PB sums with a negated literal.

ciaranm:

**Separate derivation technique from assertion reconstruction hints.**

[Appendix A](https://github.com/ciaranm/glasgow-constraint-solver/blob/a52527e6505e646376706c3692b3a4ee7d464a10/dev_docs/constraints/TEMPLATE.md#appendix-a-proof-technique-vocabulary) defines `RUP` as having no hints and `RUP+hints` as having constraint-id hints. That mixes two different things: a checked derivation (one RUP, several RUP steps, cutting planes, etc.) and the annotation used to reconstruct an asserted inference. It also risks confusion with actual VeriPB RUP antecedent line hints (`RUPProofRule.lines`).

A one-step comparison can carry an assertion annotation, while the equality interval bridge needs several RUP steps. Please describe the derivation in **Proof technique**, and put the reconstruction annotation in the existing **Hint** field; no extra per-rule field is needed. If a separate vocabulary entry is useful, `RUP sequence` would describe the latter technique more directly.

The comparison document's statement that a subhint distinguishes derivations of different *length* is correspondingly too narrow. Its purpose is to select the reconstruction procedure and any witness information that procedure needs; different procedures can have the same length.

This fits the intended pipeline: emit compact assertions and hints, trim, then reconstruct retained assertions and introduce the required encoding literals. The documentation can describe that evolving interface without requiring the external justifier to be complete, or requiring it to reproduce the solver's exact original derivation.

ciaranm:

**Give presolver-installed runtime propagators an explicit documentation home.**

[The presolver variant](https://github.com/ciaranm/glasgow-constraint-solver/blob/a52527e6505e646376706c3692b3a4ee7d464a10/dev_docs/constraints/TEMPLATE.md#the-presolver-variant) drops the propagator inventory, mutable state, idempotence and consistency sections on the premise that a presolver has no propagator. A presolver can install runtime machinery directly: [`ParitySystemGathering`](https://github.com/ciaranm/glasgow-constraint-solver/blob/d3f3f1aa1498f65afb8c4c26572a218a4d9c7221/gcs/presolvers/parity_system_gathering/parity_system_gathering.cc#L225) calls `install_parity_system_propagator` and can disable its donors. Its comment identifies AutoTable and DifferenceLogic as analogous cases.

Please make the omission conditional: a presolver document should either cover the runtime machinery it installs, including its proof rules and state, or link to the family document that does. Otherwise those parts can fall between the rewrite catalogue and the family catalogue.

The curated work list also needs explicit homes for `Dag` and `ParitySystemGathering`, which are currently absent. `ParitySystem` itself could be covered by `parity.md`; the important part is to state the allocation. This is a request to complete the curated list, not to replace it with a directory listing.

ciaranm:

**Clarify uncapped audit runs versus the default CI caps.**

The family documents say “No runtime caps” (`equals`, `comparison`, `element`, and the new `all_different`). Following the discussion about why the random-test caps exist, this should describe the configuration of the reported runs rather than imply that the test harness has no caps.

Suggested wording, where it matches the recorded run:

> The test results reported here were obtained with solution and recursion caps disabled. CTest enables caps by default to prevent occasional random instances with very many solutions from dominating CI runtime.

The template's Tests prompt could ask for these two facts separately: the default CI policy and the configuration used for the reported audit results, with the uncapped invocation or setting recorded once.

This is a provenance/wording clarification. I have no evidence that the reported figures were produced with caps enabled, and am not recommending a change to the default CI policy.

ciaranm:

**Correct the self-disabling description for non-singleton equality.**

[Self-disabling and Mutable state](https://github.com/ciaranm/glasgow-constraint-solver/blob/a52527e6505e646376706c3692b3a4ee7d464a10/dev_docs/constraints/equals.md#propagator-inventory) say rules 1, 2 and 4 return `DisableUntilBacktrack`, and that remembering equal domains is subsumed by disabling. The interval-intersection path in [`enforce_equality`](https://github.com/ciaranm/glasgow-constraint-solver/blob/d3f3f1aa1498f65afb8c4c26572a218a4d9c7221/gcs/constraints/equals/equals.cc#L207) instead returns `Enable` at line 290. The singleton fast paths return `DisableUntilBacktrack`.

Equal non-singleton domains do not entail equality: after bringing both domains to `{1,2}`, a later restriction of one operand still has to propagate to the other. Please limit the disabling claim to the cases that actually entail the relation, and remove the suggestion that equal non-singleton domains permit disabling. The implementation already keeps this path active; this is a documentation correction.

ciaranm:

**Scope the tightness claim to the mutations actually exercised.**

The [Tests section](https://github.com/ciaranm/glasgow-constraint-solver/blob/a52527e6505e646376706c3692b3a4ee7d464a10/dev_docs/constraints/equals.md#tests) says “The derivations are known to be tight” and calls the tested corruptions “the family's whole proof surface”. Those are stronger claims than the selected mutation experiments establish, and sit awkwardly beside the per-rule `Tightness: Not shown` entries.

Please describe the evidence as selected reason/derivation mutations being rejected on the named fixtures, with the unmodified controls accepted. The rule catalogue can remain the place that records which rules have that evidence. The fact that the steps are RUP does not make a finite selection of mutations exhaustive over their premises, literals, or applicable cases.

This does **not** ask for complete mutation coverage or turn the remaining `Not shown` entries into outstanding work. The policy in #871 is sensible; the family-level conclusion should match it.

ciaranm:

**Carry shared-helper changes through to the shared justification note.**

The [`JP 3.12` row](https://github.com/ciaranm/glasgow-constraint-solver/blob/a52527e6505e646376706c3692b3a4ee7d464a10/dev_docs/justification-techniques.md#the-procedures-we-use) still lists equality's “per-value symmetric-difference fallback” as a current caller, although the family document records its deletion after #904. The downstream [Element departure paragraph](https://github.com/ciaranm/glasgow-constraint-solver/blob/113c1f62766b43c22f5782a93220efc32a222f40/dev_docs/justification-techniques.md#where-we-depart-from-a-published-procedure) likewise restricts interval result-union to bare variables and describes a retained per-value fallback, despite the view support recorded in the updated Element document.

Please reconcile those descriptions with the current implementation, retaining the old behavior only as explicitly dated history. For future re-audits, a short reminder to follow shared-helper changes into the shared technique note and caller families would help: updating the status/fix table alone does not update these present-tense claims. No new benchmark is needed to correct these descriptions.

ciaranm:

Thanks — all six points checked against `main` at d3f3f1aa and agreed. Addressed in 59d19355bd81623561f21ed13797bc00c1afc564, except where noted:

1. **Derivation vs. assertion annotation.** Appendix A retires `RUP+hints`. `RUP sequence` names lemmas-then-conclusion; `hinted RUP` means VeriPB's antecedent list (`RUPProofRule::lines`) and nothing else, matching the term `veripb-facts.md` and the code already use. The Proof technique and Hint fields now say which of the two each one describes, and the Hint field describes the subhint as selecting a reconstruction procedure. equals.md's two uses are renamed. The comparison document's "length" sentence is fixed in #896 (7704220599eb1e688fae1ea2dab0a37ff4db78e3).
2. **Presolver-installed machinery.** The omission is now conditional. It turned out not to be only `ParitySystemGathering`: every presolver in the tree installs propagators directly. The variant now carries a table of what each one installs, whether it can disable donors, and which family document owns the machinery. The work list gains `dag.md`, and `parity_system_gathering` with its GF(2) propagator allocated to `parity.md`.
3. **Caps.** The Tests prompt now asks for the default policy and the reported configuration separately, and whether the default caps actually fire. They do: over three unseeded runs, `equals_constraint` and `equals_constraint_view_mixed` truncated 42 and 44 solves. equals.md now says that, and says that all eight `equals_*` lanes pass with `-DGCS_TEST_CAP_DEFAULTS=OFF`.
4. **Self-disabling.** Only rules 1 and 4 disable; rules 2 and 3 return `Enable` for the reason given. The "subsumed" claim is gone.
5. **Tightness.** Scoped to the five named corruptions of rules 1, 2 and 8 on their fixtures, in Tests, Proof-logging gaps, rule 8 and the status table; the template says why a finite set of lanes cannot establish tightness.
6. **Shared note.** The JP 3.12 row no longer lists the deleted fallback. While checking the table against the thesis PDF, JP 3.2 had also lost its overline and now reads `y≥v ∧ x<u`; a footnote records how the JP 3.12 box differs from its proof. The template now asks a re-audit to carry shared-helper fixes into this note and into the helper's other callers. The Element departure paragraph is fixed in #898 (0588f83eb7259dd126451f637a880fd012c2e7a9).

The stack above was rebased onto this commit.

Issue #882, opened 2026-09-07
Views have no range literal, so every interval-shaped fact carries a view detour (eight and counting)


ciaranm:
## Summary

A `ViewOfIntegerVariableID` has no range ("in") literal. Every part of the
solver that wants to say something interval-shaped therefore carries its own
detour for views, and there are now **eight** of them — three that throw and
five that silently degrade to per-value, two of which are the same code written
twice. `dev_docs/range_literals_spec.md` §9.1 deferred this ("folding views in =
deview the interval onto the underlying variable; deferred, do not block on
it"), which was the right call at the time; the cost of the deferral is now
visible enough to be worth pricing.

The immediate prompt is #867 / #881, which added the eighth. But the issue is
the general one, not that constraint's.

## The inventory

Throws `UnimplementedException`:

1. `simplify_literal.cc:75` — a range condition on a *registered* view.
2. `simplify_literal.cc:113` — a range condition on a view taking the deview
   fallback.
3. `names_and_ids_tracker.cc:558, 577` — `need_pol_item_defining_literal` for
   `InRange` / `NotInRange` on a view, in both the registered and deview arms.

Silently degrades to per-value:

4. `proof_logger.cc:419` (`ProofLogger::infer`) — a `NotInRange` *conclusion* on
   a view becomes one line per value.
5. `infer_explicitly.hh:105` (`infer_explicitly`) — the same test and the same
   loop, duplicated, for the `JustifyExplicitly` path.
6. `reason.cc:30` (`materialise_generic`) — a view's hole runs become one
   literal per value.
7. `equals.cc`'s `enforce_equality` holes path — `both_simple` gates the whole
   interval pruning arm, so a view operand takes the per-value path even for
   the parts that have nothing to do with the view.
8. `equals.cc`'s no-overlap witness (#881) — a run of values a view cannot take
   is spelled per value.

Sites 1–3 mean a range literal on a view is not merely slower, it is a crash, so
every producer has to know not to make one. That is what turns a deferred
feature into a rule every caller must remember: the failure mode for forgetting
is an exception from inside proof logging, not a slow proof.

## Why it matters more than it used to

Three things have changed since §9.1 was written.

- **The interval vocabulary is no longer optional.** `generic_reason` states
  hole runs as range conditions, branching yields range guesses, `In` batches
  initial-domain gaps, and `infer_not_in_range` is the ordinary conclusion path.
  A view is now the one variable kind that opts out of the common vocabulary.
[...]

Issue #864, opened 2026-09-07
A reified Equals over wide non-overlapping domains builds a reason per value with proofs off, and dies at 10^9


ciaranm:
## Summary

`ReifiedEquals`' no-overlap rule builds a reason containing **one literal per value
in the first operand's bounds range**, and does not guard that assembly on
`inference.want_reasons()`. With proofs off nothing will ever read the reason, so
the whole vector is built and thrown away. On a wide domain that is not merely
wasteful, it is fatal.

In `gcs/constraints/equals/equals.cc`, `no_overlap_justification`:

```cpp
auto v1_bounds = state.bounds(v1);
ReasonLiterals reason{{v1 >= v1_bounds.first, v1 <= v1_bounds.second}};

for (Integer val = v1_bounds.first; val <= v1_bounds.second; ++val)
    if (state.in_domain(v1, val))
        reason.emplace_back(v2 != val);
    else
        reason.emplace_back(v1 != val);
```

It is called unconditionally from `infer_cond_when_undecided`'s
`! state.domains_intersect(v1, v2)` branch.

## Measurements

`EqualsIff{x, y, b == 1}` with `x` in `[0, w/2]` and `y` in `[w/2+1, w]`, so the rule
fires once at the root. **Proofs off.** Release build at `7d014207`, g++ 15.2.0,
Ryzen 9 9950X3D.

| `w` | wall time |
|---|---|
| 10^4 | 0 ms |
| 10^6 | 29 ms |
| 10^7 | 350 ms |
| 10^9 | `std::bad_alloc` after 2.6 GB |

Linear in domain width in both time and memory.

## Why this is the guard's own documented case

`InferenceTrackerBase::want_reasons()` in `gcs/innards/inference_tracker.hh` names
this situation:

> A propagator whose reason is expensive to assemble (a `ConcatReason` allocation,
> **a long extra-literal walk**) should guard that assembly on this query, so it
> optimises away whenever nothing will read it.

`SimpleInferenceTracker::materialises_reasons` is `false`, so with proofs off the
reason is dead on arrival. The must-not-hold pass **in the same file** already
guards for exactly this reason — added in `fea9508d`, "equals: guard the reified
must-not-hold reason on want_reasons". This call site was missed.

## What to do

1. Guard the walk on `want_reasons()`. The pattern is already in the file.
2. Add a wide-domain case to `gcs/constraints/equals/equals_test.cc`. **Nothing in
[...]

Issue #598, opened 2026-07-28
Difference-logic presolver: lift MustNotHold / NotIf donors in both the linear and comparison families


ciaranm:
## Summary

The difference-logic presolver lifts `MustHold` and `If` donors from both the linear and
comparison families, but skips the **`MustNotHold` / `NotIf`** spellings, counting them in
`skipped_reified`.

Since #596 labelled the comparison rows, those forms are now perfectly citable (same empty
role, `@c[<id>]`), so this is no longer a proof-side limitation — it is a deliberate scope
decision.

## Why they were skipped

`MustNotHold`/`NotIf` assert the *negated* inequality, so lifting them means inverting the
edge (`x - y <= d` becomes `y - x <= -d-1`, with the usual care over strictness). That is
straightforward, but:

- nothing in gcs currently constructs a comparison with `MustNotHold` or `NotIf`;
- the **linear** family leaves its own equally-citable `MustNotHold` row on the floor in
  exactly the same way.

Lifting one family and not the other would introduce an asymmetry with no consumer. Better
to close both families together, once something actually posts these forms.

## Scope if picked up

- Invert the edge for `MustNotHold` (unconditional) and `NotIf` (half-reified), in both the
  linear and comparison detection paths.
- Get the strictness flip right — an off-by-one here is *unsound*, not merely incomplete,
  so it wants a fixture with a margin of exactly one unit.
- Extend the lifted/skipped counters and their "fix the presolver, don't update the number"
  assertions, and mutation-test them, per the discipline in #588.
- Reject negated views as everywhere else.

The implementation existed briefly during #596 and was deliberately backed out, so this is
a small piece of work rather than a design question.

PR #896, opened 2026-09-08
Constraint family documents: comparison

ciaranm:
> **Stacked on #863.** Its base is `constraint-family-doc-template`, not `main` — the document references `TEMPLATE.md`, which only exists there. **Retarget this to `main` before merging #863**, not after: merging with `--delete-branch` closes every PR targeting that branch, and GitHub then refuses both `gh pr reopen` and `gh pr edit --base`. Review the top commit only.

The second family document, and the first written *to* the template rather than shaping it. The skeleton fitted without modification.

**Two commits now**: the audit, and a second carrying @mmcilree's review of #863 through to this family. Every rule cites the procedure that licenses its `RUP` rather than asserting the name — and this is the easy case: all seven are **JP 3.2 (comparison)** or a **Theorem 2.6** reduction to it, nothing departs from a published procedure, and the theorem's `B ∈ {0,1}` precondition *is* exactly this family's two row shapes (`right − left ≥ 0` and `≥ 1`), so it cannot emit anything outside its own licence. That is also the sharpest answer to why a two-term `LinearLessThanEqual` is not interchangeable on the proof side whatever the CPU numbers below say: a linear row with a constant outside `{0,1}` is outside Theorem 2.9 and needs JP 3.15. Prior art now quotes the thesis's own statement of the strength claim in place of the argument the document had been making from scratch, and this is the first family in the arc with an empty novelty column.

## Why `comparison`

Picked for having no open interval-performance issue, and that turns out to be structural rather than lucky — it is the document's first observation:

**The family's whole vocabulary is bounds.** Every fact it states is a single order literal (`v1 ≥ k`) or a reification condition. A view *has* order literals; it is a **range** literal a view lacks, which is what forces the eight detours #882 counts, two of them in `equals`. So a family that never says anything interval-shaped is immune by construction — no `both_simple` guard, no per-value fallback, no entry in #882, and nothing to fix. Likewise `comparison.cc` contains **no loop of any kind**, so its three `Clean` rows in the large-domain audit are a theorem rather than a measurement. That is the opposite of the `equals` lesson about `Clean` rows, and worth having both cases written down.

## The shape

Twelve posted classes over one propagator; seven inference rules; **one** definitional OPB row (two for `Iff`); **zero proof flags, zero auxiliary variables, zero scaffolding**; one wire hint with no subhint; and every rule a bare RUP, so every **Proof size** is one line and every **Offline reconstructibility** verdict is `offline`.

That makes it the emptiest proof model audited and a useful baseline for the paper: it is what a certificate costs when the constraint is trivial. The substantive contrast with `equals` — and the reason this audit keeps the two documents apart, as the pilot decided — is that `Iff` needs no flag here, because the negation of `v1 ≤ v2` is a single inequality where the negation of `v1 = v2` is a disjunction.

## Measured

At `76bfd836`, Release + `GCS_WERROR=ON`, g++ 15.2.0, Ryzen 9 9950X3D, verified idle, `taskset`-pinned, min of three.

The right benchmark already existed and is purpose-built: `difference_chain` is Kletzander et al.'s Example 8, and `--donor` spells each edge either as a comparison or as a two-term linear inequality **over an identical search tree**. At n=500 that is 63.0M propagator calls with **99% of propagation time in this family** — much more family-dominated than `ortho_latin` is for `equals` (71% of calls, 18% of time).

| Arm | Time | Recursions | Propagations | Effectful |
|---|---|---|---|---|
| `--donor=comparison` | 4.01 s | 503 | 63,004,752 | 375,751 |
| `--donor=linear` | 1.48 s | 503 | 63,253,752 | 376,251 |

**The comparison spelling costs 2.7x the linear one.** The example's own comment says `--donor` "is what measures whether the second route costs anything"; as far as I can tell nobody had answered it.

`perf` says where it goes: 51% of cycles in the enforce lambda, and **30% of all cycles in one `memcpy`** — the `small_vector` inside `ExplicitReason{ReasonLiterals{...}}`. All **nine** reason constructions in `comparison.cc` are unconditional, where `gcs/constraints/linear/propagate.cc` routes every one of its through `inference.want_reasons()`.

Guarding the four on the enforce passes:

| | n=500 | Recursions | Propagations | Effectful |
|---|---|---|---|---|
| today | 4.01 s | 503 | 63,004,752 | 375,751 |
| guarded | **2.30 s** | 503 | 63,004,752 | 375,751 |

**1.74x at a byte-identical search tree**, 63/63 relevant ctest lanes green. Not in this PR — it is docs only — but measured, and ranked first in Next steps. `equals`'s #864 was the same defect in a more dramatic form; here it does not crash, it just costs 43% of the cycles.

Guarded, the arm is still 1.55x the linear one, and the residual is IPC-shaped rather than instruction-shaped (2.73 against 4.25). That is consistent with the offset view's indirection on every bounds read, and I did **not** isolate it — the document says so rather than guessing, and it is item 2 in Next steps because if a view costs that on a bounds read it is a fact about the view layer, not about comparisons.

### Proof

`difference_chain --mode=refute -n 40`, a root refutation so the proof is all propagation:

| Assertion level | Size | Lines | Assertions | VeriPB |
|---|---|---|---|---|
| `Off` | 4.64 MB | 48,680 | 0 | 0.86 s (**30x** the solve) |
| `Links` | 1.95 MB | 32,305 | 26,556 | — |
| `Inferences` | **0.44 MB** | 5,752 | 5,745 | — |

The 10.5x drop at `Inferences` (against 44% for `equals`) has one cause, and it is the own-versus-shared split the re-audit made the template demand a number for:

| Line kind | Count | Whose |
|---|---|---|
| `red` | 16,332 | shared bit/order-literal layer |
| `pol` | 16,127 | same layer |
| `rup` | 10,472 | **5,742 this family's**, rest the literal layer |
| `del` | 5,743 | shared |

**11.8% of the proof is this family** — 8.5 full-proof lines per inference, of which one is ours. 5,742 of 5,745 assertions carry the `comparison` hint, all one bare form.

And, as for `equals`, reading the assertions back shows the family's own benchmark reaches **one of its seven rules** (all `ge` atoms; 5,661 two-literal, 81 unit — rule 1 alone). The other six are covered by `comparison_test`'s **1,356 proofs verifying in 3.35 s** across twelve modes.

## Two findings, neither filed pending your view

1. **The unguarded reasons**, above. Largest single win the audit found, a handful of lines, pattern already in the linear family.
2. **`MustNotHold` / `NotIf` throw from `s_expr()` and leave a truncated `.scp`.** Both are reachable from the public base class, both propagate correctly (verified against brute force — 6 pairs for `¬(x ≤ y)`, 10 for `¬(x < y)` over `[0,3]²`), and both have a correct labelled OPB row. Posting one with proofs on writes the `.opb`, a 33-byte `.pbp` and an **88-byte `.scp` cut off mid-`(constraints`**, then throws. It is known and worked around — `constraint_row_test.cc` sets `s_expr_file = nullopt` with a comment saying exactly why — so what is new is that it is publicly reachable and that the failure is not clean. The spelling is mechanical (`MustNotHold` with `or_equal` emits precisely the row `LessThan{right, left}` does), and the linear family has the identical gap, so #598 is the natural place for both.

Say the word on either and I will open the issue, or the fix.

## No template changes

First time. Two observations went into Developer commentary instead:

- The per-rule **`Tightness`** field answers `Not shown.` seven times here and reads as a fact about the family rather than a to-do list — which is what the policy note added in #863 intended, so this is evidence for it.
- The family wants a **sixth "relation" direction**: not shared code, but a *published read interface a presolver depends on*. `DifferenceLogic` lifts every comparison into a global propagator through `left_variable()` / `or_equal()` / `reification_condition()` / `primary_row_role()`, and cites the row that last one names in `pol` steps that `cake_pb_cp` re-derives at the other end. Rather than add vocabulary on one example, the document says so in prose and leaves the closed list alone; the next family that consumes or is consumed by a presolver should settle it.

## Checks

Docs only; no code touched. All cross-referenced `dev_docs/` files exist and every internal anchor resolves. Every table caption carries its commit, date and machine. Structural claims re-verified by grep against the source (no loops, nine unguarded reasons, no `JustifyExplicitly`, twelve classes).

### DIFF
--- a/dev_docs/constraints/README.md
+++ b/dev_docs/constraints/README.md
@@ -19,6 +19,15 @@ constraints and do not restate it.
   this template, and now also the first **re-audit**: it filed seven issues,
   six are fixed, and four of the fixes changed what the document says rather
   than only what the code does.
+- [`comparison.md`](comparison.md) — `LessThan`, `LessThanEqual`, their
+  `Greater*` mirrors and their eight reified forms: twelve posted constraints
+  over one propagator whose whole vocabulary is bounds, so one definitional OPB
+  row, **no proof flags and no view detour anywhere**, seven inference rules
+  each a single RUP, and 11.8% of its own proof. Nearly unreachable from the
+  frontends, which turn a binary ordering into a two-term linear inequality;
+  its real consumer is the difference-logic presolver. The audit's finding is a
+  reason assembled on every call whether or not anything reads it — 43% of the
+  cycles on a family-dominated benchmark.
 
 Everything else is still to write; the family list in
 [`TEMPLATE.md`](TEMPLATE.md#provisional-family-list) is the work plan, and #871
--- a/dev_docs/constraints/comparison.md
+++ b/dev_docs/constraints/comparison.md
@@ -2,10 +2,10 @@
 
 > **Maturity** production ·
 > **Audited** 2026-09-08 at `76bfd836` ·
-> **Open issues** none specific to this family. #868 is the audit-wide
-> cross-solver prerequisite; #598 would close a deliberate presolver gap that
-> is half about this family. Two findings from this audit are unfiled — see
-> [Next steps](#next-steps). Tracked under #871.
+> **Open issues** #907 (every reason built unguarded) and #908 (`MustNotHold`
+> and `NotIf` throw from `s_expr()`), both filed by this audit. #868 is the
+> audit-wide cross-solver prerequisite; #598 would close a deliberate presolver
+> gap that is half about this family. Tracked under #871.
 
 Twelve posted classes over one implementation and one propagator: an
 inequality between two operands, optionally reified, in either direction, with
@@ -948,7 +948,7 @@ performance](#cpu-performance) and [Next steps](#next-steps).
 ### Known limitations
 
 **`MustNotHold` and `NotIf` throw when an `.scp` is written, and leave a
-truncated file behind.** Both kinds propagate correctly — verified against
+truncated file behind** (#908). Both kinds propagate correctly — verified against
 brute force, `¬(x ≤ y)` giving exactly the 6 pairs of `x > y` and `¬(x < y)`
 the 10 of `x ≥ y` over `[0,3]²` — and both have a correct, labelled OPB row.
 What they do not have is a cake spelling, so `s_expr()` throws
@@ -1001,7 +1001,7 @@ reified forms behave the other way — `LessThanIff(x, x, c)` cheerfully forces
 Ranked. Two are findings of this audit and neither is filed yet; the third is
 the audit-wide prerequisite.
 
-1. **Guard the reason assembly on `want_reasons()`.** Not filed. Nine reason
+1. **#907 — guard the reason assembly on `want_reasons()`.** Nine reason
    constructions in `comparison.cc`, none guarded; the four on the enforce
    passes are the hot ones and guarding just those is **1.74x on
    `difference_chain -n 500`** (4.01 s to 2.30 s) at a byte-identical search
@@ -1011,7 +1011,8 @@ the audit-wide prerequisite.
    form. This is the largest single win the audit found and it is a handful of
    lines; the remaining five sites belong to the reified verdicts and the
    both-constant initialiser, which are not hot but should go the same way for
-   uniformity.
+   uniformity. A working patch is preserved at
+   `~/claude/tmp/famdocs-findings/`.
 2. **Find out what the rest of the comparison-versus-linear gap is.** Not
    filed. Guarded, the comparison spelling is still 1.55x the two-term linear
    one on an identical tree, and the difference is IPC-shaped rather than
--- a/dev_docs/constraints/comparison.md
+++ b/dev_docs/constraints/comparison.md
@@ -250,20 +250,22 @@ else.
 
 ### OPB encoding
 
-One inequality over the operands' **bit** encodings. Writing `⟦v⟧` for the bit
-sum of `v`, and `d` for `0` in the `<=` form and `−1` in the `<` form:
+One inequality over the operands' **bit** encodings. Writing `BinEnc(v)` for
+the bit-sum encoding of `v` --- the notation
+[`justification-techniques.md`](../justification-techniques.md) and the thesis
+both use --- and `d` for `0` in the `<=` form and `−1` in the `<` form:
 
 ```
-MustHold      ⟦left⟧ - ⟦right⟧ <= d
+MustHold      BinEnc(left) - BinEnc(right) <= d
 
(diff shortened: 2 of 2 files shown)


ciaranm:

**Correct the threshold in the must-hold assertion and its proof argument.**

In [bounds-from-must-hold](https://github.com/ciaranm/glasgow-constraint-solver/blob/6a4013274521eab6c7fb58789b289d10bb24b149/dev_docs/constraints/comparison.md#rule-bounds-from-must-hold), the first inference defines `k = ub(right) + [or_equal]`, but the displayed assertion uses `right ≤ k − 1 + [or_equal]` as its reason. It should use `right ≤ k − [or_equal]`:

```text
left < k ∨ ¬(right ≤ k − [or_equal]) ∨ ¬cond
```

For `≤`, the printed clause is false at `cond = true` and `left = right = k`, even though that assignment satisfies the comparison. For `<`, it also misstates which right-hand bound licenses the documented inference. The same threshold needs correction in the preceding **Proof technique** paragraph.

The [implementation](https://github.com/ciaranm/glasgow-constraint-solver/blob/d3f3f1aa1498f65afb8c4c26572a218a4d9c7221/gcs/constraints/comparison/comparison.cc#L219) uses the actual `ub(right)` in the reason. This is a mathematical error in the document, not a claim that the current propagation or emitted proof is unsound.

ciaranm:

**Update the present-tense serialization claim after #915.**

The paragraph “Two reification kinds no class posts” near the start of [comparison.md](https://github.com/ciaranm/glasgow-constraint-solver/blob/6a4013274521eab6c7fb58789b289d10bb24b149/dev_docs/constraints/comparison.md#L99) still says `MustNotHold` and `NotIf` have no `.scp` spelling and that `s_expr()` throws. The status/limitations discussion in the same document records the fix in #915, and the current implementation handles both cases.

Please describe their supported serialization in the current-behavior section and keep the old exception/truncated-file behavior in the dated audit history. A reader should not have to reach the fix history to discover that the earlier API description is obsolete.

ciaranm:

Both points agreed and addressed in 7704220599eb1e688fae1ea2dab0a37ff4db78e3:

1. **Threshold.** Now `k − [or_equal]` in both the Proof technique paragraph and the assertion; the code puts `right ≤ ub(right)` in the reason. The measured example line in the document only matches the corrected form.
2. **Serialization after #915.** The reification-kinds paragraph now describes the current spelling: each kind is written as the comparison it enforces. The throw stays as dated history under Known limitations.

Also here, from the #863 review: the subhint sentence now says a subhint selects a reconstruction procedure rather than a derivation length. "No runtime caps" is replaced by a measurement: the default caps truncate solves on ten of the 28 lanes, all half-reified, and all of them plus `constraint_row_test` pass uncapped at d3f3f1aa.

PR #898, opened 2026-09-08
Constraint family documents: element

ciaranm:
> **Stacked.** Base is `comparison-family-doc` (#896), which is itself on `constraint-family-doc-template` (#863) — the document links to `TEMPLATE.md`, `justification-techniques.md` and `comparison.md`, none of which exist on `main` yet. **Review the top commit only**, and retarget children before merging any parent.
>
> The whole stack has been rebased onto `main` at `57a6dace`, so it is now **docs-only**: #897 has merged (as `977532a6`), and the commits this branch was carrying for it are gone. The audit's measurements were taken at exactly that code — `element.cc`, `equals.cc`, `comparison.cc` and `interval_set.hh` are all unchanged between `977532a6` and current `main`, so nothing needed re-measuring. `ctest` re-run on the rebased state: **810/810, nothing skipped**.

The third family document, and the first non-binary one. Almost everything interesting about `element` follows from that.

## The encoding is per-cell

One half-reified equality per array cell, so **linear in the array** where `equals` and `comparison` are logarithmic in everything. Measured, counting the constraint's own OPB block:

| | |
|---|---|
| 1-D, `n` cells | `2n` rows of its own, plus `4n + 2` of index-literal definitions it pulls in |
| domain width | **no effect on the row count** — 50 rows at width 7, 15, 31 and 63 |

So the own-versus-shared split the template now demands a number for comes out at about a third, and the shared part is the index variables' atomic literals being defined lazily on first use inside the block.

## It labels nothing

The only family so far that calls `add_constraint` rather than `add_labelled_constraint` and has no `ConstraintProofModelData` specialisation. Verified against a real `.opb`: of 50 rows in a 3-cell instance's element block, the 34 labelled ones are all `@i[…]` variable-literal definitions and the 16 unlabelled ones are the constraint's.

Two consequences, only one of which is a problem. Nothing can cite an `Element` row, so it cannot be a presolver donor the way `comparison` is — nothing wants to. But an external justifier cannot *name* the row it is RUPing against and has to identify the cell positionally, which is why every reconstructibility verdict here is `hinted` rather than `offline`.

## Its dominant wire hint belongs to another family

This is the finding most likely to bite the assertion rewriter, so it is worth the space.

Rule 5 delegates to the exported `enforce_equality`, which hints `equals` with its **caller's** constraint id. On `langford --size=8` at `AssertionLevel::Inferences`:

| Wire hint | Assertions | Owning family |
|---|---|---|
| `equals` | 6,545 | **`element`, all 6,545** |
| `element` | 485 | `element`, all 485 |

`langford` posts `AllDifferent`, `Element` and `Plus` — **no `Equals` at all**. Every one of the 6,545 resolves, through the OPB's own provenance comments, to an `Element`. So the family's real proof footprint is 7,030 of 39,593 assertions (17.8%), of which **93% wears another family's hint**.

That is sound and the attribution is right — the derivation genuinely is `equals`-shaped and the owner genuinely is the `Element`. But **the hint names the procedure, not the family**, and a tool that reads `equals:((constraint_id _23))` and looks up `_23` expecting `Equals`'s two equality rows will find an `Element`'s per-cell half-reified pair. What it wants is the cell, and the reason's index literals name it. @mmcilree — flagging this one specifically.

## Three of six rules are published; two are ours

`element` is the best-covered family in the arc: **JP 3.9** (empty intersection), **JP 3.10** (missing value) and **JP 3.11** (single value) are rules 1, 4 and 5, resting on Thm 2.8 per line and **Thm 3.2** (an emptied defined domain unit-propagates to conflict) for the collapse.

Thm 3.2 is a fifth fact `justification-techniques.md` did not have, and it is the one to reach for whenever a derivation has the shape *"rule out each value in turn, then conclude"* — which is most of the non-binary families. Added there, with the three procedures.

The two departures: the interval result-union rule states JP 3.10's fact over a *run* at the cost of two bridge lemmas per index tuple (the same construction as `equals`'s interval bridge, JP 3.2 / Thm 2.9), and the bounds arm is JP 3.10's walk with a bound as the conclusion — better called a gap in the published list than a new technique, since nothing in the argument depends on which literal is concluded.

## Three other firsts

- **A real option.** `with_consistency(GAC | BC)` changes propagation strength and *provably not* the OPB encoding, which is what makes it an option rather than a model choice — and why the `.scp` not recording it is correct rather than a gap.
- **An idempotence claim.** Three propagators claim, all conditioned on a hand-computed `scope_has_aliasing`. The engine's own trigger-scope downgrade cannot substitute for it, precisely *because* the index propagators deliberately omit the variable they write from their triggers — the engine cannot see aliasing involving a variable it was never told about. The range propagator deliberately does *not* claim, and its comment gives a worked counterexample rather than an argument, which is the right way round.
- **Propagators that are not all GAC**, and an asymmetry worth knowing: the index side is GAC under both settings, because `_bounds_only` only chooses between the two *result* propagators.

## Measured

At `76bfd836` plus #897, idle machine, `taskset`-pinned, min of three.

| Benchmark | Arm | Time | Calls | Share of propagation time | Per call |
|---|---|---|---|---|---|
| `qap 12` | constant array, 2-D, BC | 1.72 s | 23,773,491 of 24,507,109 (97%) | **96% of 1,825 ms** | ~74 ns |
| `langford 11` | variable array, 1-D, GAC | 3.87 s | 25,663,053 of 29,205,903 (88%) | **39% of 3,750 ms** | ~57 ns |

`qap` is the most family-dominated benchmark in the arc (against `ortho_latin`'s 18% for `equals`); `langford` is the only one reaching rule 5, so it is the one to use for anything touching `enforce_equality`.

Proofs, `langford --size=8`:

| Assertion level | Size | Lines | Assertions | VeriPB |
|---|---|---|---|---|
| `Off` | 3.92 MB | 71,243 | 0 | 1.39 s (**35×** the solve) |
| `Links` | 5.00 MB | 49,498 | 38,465 | — |
| `Inferences` | 4.77 MB | 45,093 | 39,593 | — |

**Asserting inferences makes the proof bigger here** — 3.92 MB to 4.77 MB — where it cut `equals`'s by 44% and `comparison`'s by 10.5×. A family whose justifications are already one line per inference gains from that switch; one whose justifications are a product over index domains, under a wide `generic_reason`, does not. That is worth knowing before anyone reads the hints-only mode as a uniform saving.

`element_test` is 96 proofs verifying in 7.14 s across five modes — about **30× the checking cost per proof** of `comparison`'s 1,356 in 3.35 s. `langford11` is the slowest proof in the whole curated set at 662.9 s.

And, for the third family running, **the benchmark exercises a minority of the rules**: two of six, with not a single range literal in the proof. `langford`'s result is a root singleton, so the union sweep has nothing to prune.

## Two issues filed from it

**#900 — the index-support rule's justification is per-value.** It emits one proof line per value of the selected entry's domain, which is JP 3.9 as published, so not a divergence — but measured on a 1-D instance where the rule fires once at the root:

| entry width | proof lines | proof size | to write |
|---|---|---|---|
| 10³ | 8,016 | 653 KB | 4 ms |
| 10⁴ | 80,016 | 8.0 MB | 40 ms |
| 10⁵ | 800,016 | 92.9 MB | 421 ms |
| 10⁶ | **8,000,016** | **1,010.8 MB** | 4,505 ms |

Exactly `8w + 16` lines, **1.01 GB for a single inference** at 10⁶, against 0 ms with proofs off at every width to 10⁷. And only one line in eight is the rule's own — the other seven are the variable-encoding layer defining an `eq` atom for every value of the entry, because the rule's lines name `array_var != v`. So an interval-wise witness would remove both halves at once, and there is a candidate: this is disjointness between the entry's domain and the result's, which is the fact `equals`'s rule 9 states, and #881 gave that one a 13-line-at-any-width witness.

**#901 — nothing can select the consistency level from a model.** `with_consistency` is a C++ call and neither frontend exposes it, so every model gets whichever default its class carries. I A/B'd it on both constant-array benchmarks before proposing anything, and it changed what the issue asks for:

| Benchmark | Arm | Time | Recursions | Propagations | Effectful |
|---|---|---|---|---|---|
| `qap 12` | `BC` (default) | **1.71 s** | 123,333 | 24,507,109 | 4,679,021 |
| `qap 12` | `GAC` | 3.37 s | 123,333 | 25,452,372 | 7,587,766 |
| `tsp` | `BC` (default) | **11.19 s** | 4,285,745 | 83,372,056 | 41,243,802 |
| `tsp` | `GAC` | 13.67 s | 4,285,745 | 57,702,481 | 38,036,662 |

**Bit-identical search trees both ways**, while `GAC` makes **62% more effectful inferences** on `qap` — every one unobserved. `tsp` is the clean case from the other side: `GAC` does *fewer* propagations (57.7M against 83.4M) and is still slower.

The owner supplied the right framing for why, which is more general than my first attempt at it: **what matters is which consumers of a variable can observe an interior value at all.** In both models the result is consumed only by a bounds-consistent `LinearEquality` and is not branched on, so nothing outside the constraint can tell a hole from its absence.

Following that through gave a sharper result for the constant-array arm, which I then checked: **the element cannot observe it either.** The index-support rule tests `in_domain(result, array[i])`, and `GAC` removes from `result` exactly the values no live entry supplies — never of the form `array[i]` for a live `i`. And the equality rule, which is what would push `result`'s domain into an entry, is `if (array_has_nonconstants)` and is not installed for a constant array. Measured on a constant-array instance whose result has interior values no entry supplies:

```
BC : index {0 1 2 3}  result {0 1 2 3 4 5 6}
GAC: index {0 1 2 3}  result {0 2 4 6}
```

Different result domain, **identical index domain**. So for a constant array, `GAC` on the result is observable only outside the constraint — which makes the two differing defaults look principled rather than arbitrary, since for a *variable* array the equality rule is installed and a tighter result does reach the entries.

**#902** is the general form, raised by the owner: if every constraint declared per variable whether it might care about interior values, a bounds-only variable could be detected per *model* rather than guessed per class. Filed separately with the caveats — the brancher is a consumer (`dom` and `dom_wdeg` read domain size), so is the objective, it is a fixpoint rather than one pass, and it must not change the OPB encoding.

## No template changes

Third document, second time the skeleton fitted. Two stretches recorded in Developer commentary instead: **Reification became `None.`** for the first time (kept as a short section rather than dropped — "no reified form" is a fact a reader wants confirmed, especially in a family whose *encoding* is full of half-reification), and the wire-form inventory now has an entry that is another family's, which needed no template change but does mean a cross-family pivot keyed on hint name would misattribute 6,545 assertions.

## Checks

Docs only; no code of mine. Every cross-referenced file exists on this branch and every internal anchor resolves. Structural claims re-verified by grep and against real `.opb`/`.pbp` output. One claim was wrong in draft and is corrected rather than shipped: "the three largest proofs in the curated set are `Element` proofs" — they are not (`rcpsp_dl21` is 1,242 MB); `langford11` is the *slowest to verify*, and `Element`-posting instances hold four of the top six by size, of which two post several other constraints too, so the document says which part is evidence.

### DIFF
--- a/dev_docs/constraints/README.md
+++ b/dev_docs/constraints/README.md
@@ -28,6 +28,16 @@ constraints and do not restate it.
   its real consumer is the difference-logic presolver. The audit's finding is a
   reason assembled on every call whether or not anything reads it — 43% of the
   cycles on a family-dominated benchmark.
+- [`element.md`](element.md) — `Element`, `Element2D` and their constant-array
+  forms: four posted classes over one templated implementation and **four**
+  propagators, and the first family in the arc that is not binary. One
+  half-reified equality **per array cell**, so its encoding is linear in the
+  array and the slowest proof to verify in the curated set is its; the first with
+  a real option (`with_consistency`), the first to claim idempotence, the first
+  whose propagators are not all GAC, and the only one whose dominant wire hint
+  belongs to another family — 93% of the assertions it is responsible for
+  arrive labelled `equals`, because it reuses `enforce_equality`. Three of its
+  six rules are published justification procedures; two are ours.
 
 Everything else is still to write; the family list in
 [`TEMPLATE.md`](TEMPLATE.md#provisional-family-list) is the work plan, and #871
--- a/dev_docs/constraints/element.md
+++ b/dev_docs/constraints/element.md
@@ -1,0 +1,1157 @@
+# `Element`: a result equals the array entry an index selects
+
+> **Maturity** production ·
+> **Audited** 2026-09-08 at `76bfd836` plus #897, since merged as `977532a6` ·
+> **Open issues** #900 (the index-support rule's per-value proof), #901
+> (nothing can select the consistency level) and #902 (the general form of
+> #901), all filed by this audit. #878 is
+> closed by #897, which this branch carries. #868 is the audit-wide cross-solver
+> prerequisite. Tracked under #871.
+
+Four posted classes over one templated implementation and **four** propagators:
+`result = array[index]`, in one or two dimensions, over an array of variables or
+of constants. It is the first family in the arc that is not binary, and almost
+everything interesting about it follows from that.
+
+Three things to know before touching it. Its OPB encoding is **one half-reified
+equality per array cell**, so unlike [`equals`](equals.md) and
+[`comparison`](comparison.md) it is *linear in the array* rather than
+logarithmic in anything — and the slowest proof to verify in the curated
+benchmark set is an `Element` one. It is the first family with a real **option**
+(`with_consistency`), the first that **claims idempotence**, and the first whose
+propagators are not all GAC. And it is the only family so far whose dominant
+wire hint belongs to **another family**: 93% of the assertions it is responsible
+for on its own benchmark arrive labelled `equals`, because it reuses
+`enforce_equality`.
+
+## What it is
+
+### Semantics
+
+`NDimensionalElement<EntryType_, dimensions_>` enforces
+`result = array[index₁ - start₁]…[index_d - start_d]`. Four concrete classes:
+
+| Class | Array | Dims | Default consistency |
+|---|---|---|---|
+| `Element(result, index, array)` | variables | 1 | `GAC` |
+| `Element2D(result, index₁, index₂, array)` | variables | 2 | `GAC` |
+| `ElementConstantArray(result, index, array)` | constants | 1 | `BC` |
+| `Element2DConstantArray(result, index₁, index₂, array)` | constants | 2 | `BC` |
+
+Each index is given either bare (0-based) or as a `pair{var, start}`, where
+`start` is subtracted before indexing. That offset exists for the frontends:
+FlatZinc's `array_int_element` is 1-based, and XCSP3 carries an explicit
+`startIndex` per dimension.
+
+The index variables' bounds are **narrowed at `prepare()` time** to
+`[start, start + size - 1]` via `define_bound`, so an out-of-range index is
+removed before search rather than propagated away.
+
+Degenerate cases:
+
+- **A zero-length dimension.** No assignment can satisfy the constraint. This
+  is a valid model, not a rejected one: `prepare()` records the flag,
+  `define_proof_model` emits a trivially-false `0 ≥ 1` row so the OPB stays
+  self-describing, and `install_propagators` installs a contradiction
+  initialiser rather than a propagator. The index-bound definitions are skipped
+  for that dimension, because `[start, start - 1]` would label an empty
+  interval as if it were an ordinary trim.
+- **A constant-valued array of variables.** Accepted, and detected:
+  `_array_has_nonconstants` is computed in `prepare()`, and when every entry is
+  already fixed the array is not watched and the equality propagator is not
+  installed at all.
+- **Aliasing anywhere in the scope** — result also an index, an index repeated,
+  or result or an index appearing inside the array. Accepted, and it has one
+  consequence that is easy to miss: it turns off every idempotence claim in the
+  family. See [Propagator inventory](#propagator-inventory).
+- **Views.** Accepted everywhere, at a cost in one rule; see [Variable kinds
+  and views](#variable-kinds-and-views).
+
+### Concrete constraints and frontend coverage
+
+| C++ class | MiniZinc / FlatZinc | XCSP3 | CPMpy | SCP `s_expr` | Notes |
+|---|---|---|---|---|---|
+| `Element` | ✓ `array_var_int_element`, `array_var_bool_element` (1-based) | ✓ `element` over a variable list, and the value-form with a constant result | ? | ✓ `element` | |
+| `ElementConstantArray` | ✓ `array_int_element`, `array_bool_element` (1-based) | ✓ `element` over an int list | ? | ✓ `element`[^same] | the arm MiniZinc reaches most |
+| `Element2D` | frontend gap — no FlatZinc builtin for a 2-D element | ✓ `element` matrix form, both variable-result and constant-result | ? | ✓ `element_2d` | |
+| `Element2DConstantArray` | frontend gap — as `Element2D` | ✓ `element` matrix form over an int matrix | ? | ✓ `element_2d`[^same] | |
+
+[^same]: `constraint_type()` returns `element` or `element_2d` and nothing else,
+    so the constant-array forms describe themselves exactly as the variable
+    forms do, and `read_scp` rebuilds them as the variable form over
+    singleton-domain variables. That is semantically the same constraint and a
+    deliberate consequence of the `.scp` recording the *model* rather than the
+    solving strategy — see [Options](#options).
+
+**The consistency default differs by arm, and no frontend overrides it.**
+`ElementConstantArray` is bounds-consistent by default, which is what MiniZinc's
+`array_int_element` gets. Nothing in the frontends, the examples or the
+benchmarks calls `with_consistency` on this family, so the constant-array GAC
+arm is reached only by `element_test`'s `constgac` lane — which exists for
+exactly that reason.
+
+### Options
+
+**`with_consistency(consistency::GAC{} | consistency::BC{})`** — the first real
+option in the arc, and the only thing this family is tunable by.
+
+It selects **propagation strength only and never changes the OPB encoding**,
+which is the property that makes it an option rather than a model choice: the
+proof model is the same either way, so a proof written under one setting is a
+proof of the same problem as a proof written under the other. That is also why
+the `.scp` does not record it, and why that is correct rather than a gap — the
+`.scp` describes the constraint, not the strategy used to propagate it.
+
+What it changes is which of the two result propagators is installed:
+
+- `GAC` installs the union propagator: `result` keeps only values some live
+  array entry can take.
+- `BC` installs the range propagator: `result`'s *bounds* are narrowed to the
+  range spanned by live entries, and interior values with no support survive.
+
+**The index side is `GAC` either way.** `_bounds_only` is only ever read when
+choosing between the two result propagators; the index propagator watches full
+domains and tests against `result`'s whole domain, so an interior hole in
+`result` still removes a now-unsupported index. That asymmetry is deliberate and
+is stated in the code.
+
+### Variable kinds and views
+
+Result, indices and array entries are all `IntegerVariableID`, so plain
+variables, constants and views are accepted throughout, and the test binary
+sweeps views over **six** positions in the variable-array mode
+(`add_view_tests(element_constraint_var element_test 6 var)`).
+
+One rule degrades, and it is the same underlying cause as in
+[`equals`](equals.md): **a view has no range literal** (#882). The GAC union
+propagator's interval path is guarded on `result` *and* every array entry it
+considered being a bare `SimpleIntegerVariableID`; if any is a view, the whole
+propagation falls back to removing one value at a time. The guard is not about
+propagation — the values removed are identical — but about the proof: the range
+conclusion has to carry an order atom from `result` across the model's
+half-reified equality to the entry, and a view's atoms are spelled through the
+view, so that crossing has not been shown to bridge one.
+
+The code names its two neighbours on this point, which is worth repeating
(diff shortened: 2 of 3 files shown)


ciaranm:
@mmcilree I guess the question here is what we should be doing for hints when we are delegating. This probably comes up for all-different too, when it's used by other constraints.

ciaranm:

**Fix the Element deletion polarities in the shared procedure table.**

The [JP 3.9–3.11 rows](https://github.com/ciaranm/glasgow-constraint-solver/blob/113c1f62766b43c22f5782a93220efc32a222f40/dev_docs/justification-techniques.md#the-procedures-we-use) currently display positive equality literals where the documented inferences remove values. With `y` the index, `z` the result, and `x_i` an entry, the deletion shapes should read:

- JP 3.9: per entry value `w`, `R ⇒ (y ≠ v ∨ x_v ≠ w)`, then `R ⇒ y ≠ v`.
- JP 3.10: per index `i`, `R ⇒ (z ≠ v ∨ y ≠ i)`, then `R ⇒ z ≠ v`.
- JP 3.11, in the stated entry-pruning direction: `R ⇒ x_i ≠ v` for an unsupported entry value, with the selected index in the reason.

Equivalently, retain the PB `≥ 1` notation and complement the corresponding equality atoms. The current positive forms state the opposite of the pruning described in the Element rule catalogue. Please check the table against those concrete inferences before it becomes the shared reference for the paper and external justifier.

ciaranm:

**Include the delegated `equals/not_in_range` form in the wire inventory.**

The [catalogue preamble](https://github.com/ciaranm/glasgow-constraint-solver/blob/113c1f62766b43c22f5782a93220efc32a222f40/dev_docs/constraints/element.md#L734) says there are two wire forms: bare `element` and bare `equals`. Rule 5 also names only `hints::Equals`. The shared helper's interval path emits [`hints::EqualsNotInRange`](https://github.com/ciaranm/glasgow-constraint-solver/blob/d3f3f1aa1498f65afb8c4c26572a218a4d9c7221/gcs/constraints/equals/equals.cc#L244), with the owning Element's constraint id.

A probe with index fixed to 0, a single variable entry in `[0,7]`, and result domain `{0,3,4,7}` emitted:

```text
a 1 ~i[entry][in1_2] 1 ~i[index][eq0] 1 i[result][in1_2] >= 1::equals:((constraint_id _2) (subhint not_in_range));
```

Here `_2` owns the Element; the `[5,6]` gap gives the same additional form. The fully justified version verified and enumerated four solutions.

Please add this third form and describe how the delegated interval bridge uses the selected-cell guards. A small selected-entry interval fixture would also make the inventory exercise the shared helper's bridge path; the Langford inventory does not establish that this path is absent.

This adds a concrete case to the existing delegation discussion. It does not require freezing the hint format or completing the external justifier before merging the document.

ciaranm:

**Remove the claimed disabling through `enforce_equality`.**

The [propagator table and self-disabling paragraph](https://github.com/ciaranm/glasgow-constraint-solver/blob/113c1f62766b43c22f5782a93220efc32a222f40/dev_docs/constraints/element.md#propagator-inventory) give selected-entry equality a disabling exception “via `enforce_equality`”. The [call site](https://github.com/ciaranm/glasgow-constraint-solver/blob/d3f3f1aa1498f65afb8c4c26572a218a4d9c7221/gcs/constraints/element/element.cc#L925) ignores that helper's return value and then returns `Enable` or `EnableButIdempotent`. Thus the helper does not disable this Element propagator.

Please change that table cell to “no” and remove the exception from the paragraph. This is a description of the current implementation, not a request to add a disabling optimization.

There is one other table/prose mismatch in the same inventory: the result-union row lists `result` under Triggers, while its Holes affect cell and the following prose correctly say it does not watch `result`. The installed union triggers contain the indices and nonconstant entries; please remove `result` from that trigger cell too.

ciaranm:

**Revise the explanation for the larger assertions file.**

The [proof-performance discussion](https://github.com/ciaranm/glasgow-constraint-solver/blob/113c1f62766b43c22f5782a93220efc32a222f40/dev_docs/constraints/element.md#proof-performance) attributes the increase to replacing short intermediate lines with an assertion carrying the full `generic_reason`. But [`infer_explicitly`](https://github.com/ciaranm/glasgow-constraint-solver/blob/d3f3f1aa1498f65afb8c4c26572a218a4d9c7221/gcs/innards/proofs/infer_explicitly.hh#L115) emits the same conclusion under the same reason in both modes: the full derivation ends with a RUP under that reason, while inference mode emits the assertion and its annotation.

For a separate `langford --size=8` run at `6b220c79`, I measured:

| Whole proof file | Bytes |
|---|---:|
| Full justification | 3,972,370 |
| Inference assertions | 4,766,288 |
| Annotation text within the assertions file | 1,241,992 |
| Assertions file excluding annotation text, for accounting only | 3,524,296 |

Thus the annotation text alone exceeds the net size increase in that run. These are measurements from that snapshot, not replacements for the document's different-snapshot full-proof figure, and the file also contains AllDifferent and Plus contributions.

Please retain the observed size increase but correct its causal explanation and avoid generalising that nested derivations intrinsically lose from assertion mode. Annotation removal here is only an accounting experiment, not a proposed format change. The intended benefit should ultimately be assessed over trimming, reconstruction and checking; initial file size is an intermediate metric, and that end-to-end measurement can wait for the external tool.

ciaranm:

**Describe the Auto test's actual comparison, rather than claiming exact tree equality.**

Both [Interior values and optional pruning](https://github.com/ciaranm/glasgow-constraint-solver/blob/113c1f62766b43c22f5782a93220efc32a222f40/dev_docs/constraints/element.md#interior-values-and-optional-pruning) and [Tests](https://github.com/ciaranm/glasgow-constraint-solver/blob/113c1f62766b43c22f5782a93220efc32a222f40/dev_docs/constraints/element.md#tests) say `element_auto_test` establishes that Auto explores exactly GAC's tree.

The current [`Outcome`](https://github.com/ciaranm/glasgow-constraint-solver/blob/d3f3f1aa1498f65afb8c4c26572a218a4d9c7221/gcs/constraints/element/element_auto_test.cc#L128) contains recursion count, solution count and best objective; `CHECK(automatic == gac)` compares those aggregates. For uniform arm selections there are additional propagation-count comparisons. Equal aggregates are useful regression evidence, but do not establish node-for-node equality of the visited trees.

Please say exactly what is checked. If exact tree equality becomes an experimental claim that needs certification, a node trace or digest could support it; adding one is not necessary merely to make the present description accurate. The discriminating fixtures and mutation remain useful evidence for the policy.

ciaranm:

All five agreed and addressed in 0588f83eb7259dd126451f637a880fd012c2e7a9:

1. **JP 3.9–3.11 polarities.** Checked against the thesis PDF: every literal there is negated, and the rows now spell them as `≠`. A footnote notes that the JP 3.10 box ends with `¬(y=i)`, where its precondition means `¬(z=v)`.
2. **Third wire form.** Your probe reproduces exactly at d3f3f1aa, both gaps, and the fully justified instance verifies with 4 solutions. The inventory is now a three-form table, and rule 5's Hint, Assertion and reconstructibility fields include `not_in_range`. Rule 5 now explains how the bridge gets its guard: through the reason's index literals, which switch on the cell's half-reified rows, since `enforce_equality` does not pass the helper's `under` flag. A selected-entry fixture is listed under Next steps; this arc stays docs-only.
3. **Disabling and triggers.** The selected-entry cell is now "no", the paragraph says why, and `result` is removed from the union propagator's triggers.
4. **Assertions file.** The causal explanation is corrected, and the generalisation about nested derivations is gone. Your figures reproduce byte for byte. A separate by-line-kind table at d3f3f1aa adds one thing: about 0.80 MB of derivation is removed, against 1.24 MB of annotation text and 0.36 MB of solution lines that are spelled in bits at that level. It is labelled as accounting only, and it defers the real assessment to trimming, reconstruction and checking.
5. **Auto test.** Both places now say it compares recursion count, solution count and best objective, plus propagation counts for uniform arms. The test's own name overclaims, and that is listed as a next step.

Also here, from the #863 review: the Element departure paragraph in `justification-techniques.md` now describes the width gate from #925, and `RUP+hints` becomes `RUP sequence`. "No runtime caps" is replaced by a measurement: the default caps truncate solves on all five variable-array lanes, and all 16 `element_*` lanes pass uncapped.

Issue #903, opened 2026-09-09
`divide_modulus` does not achieve GAC when an operand is a view: 68 of the view-wrap sweep's tests fail


ciaranm:
Found while establishing a baseline for #882. This is **not** a proof-logging
failure and has nothing to do with that work — it is a propagation-strength
failure, and it reproduces on unmodified `main`.

## Reproducing

```shell
cmake -S . -B build -DGCS_TEST_CAP_DEFAULTS=OFF -DGCS_ENABLE_VIEW_WRAP_SWEEP=ON
cmake --build build --parallel $(nproc)
./run_test_only.bash ./build/divide_modulus_test --view-wrap=1 --view-position=0
```

```
divide auto gac-checked (-4, 4) (-2, 3) (-4, 4) views (1, 0, 1, 0, 1, 0) : expecting 45 solutions
gac not achieved in test
var varidx 2 value -4 has no support at the current node
values of varidx 2 that do have a support here: [-1, 0, 1, 2]
var varidx 0 has values [-4, -3, -2, -1, 0, 1, 2, 3, 4]
var varidx 1 has values [-2, 2, 3]
var varidx 2 has values [-4, -3, -1, 0, 1, 2, 3, 4]
terminate called after throwing an instance of 'gcs::UnexpectedException'
  what():  unexpected problem: consistency not achieved
```

The propagator leaves `-4`, `-3`, `3` and `4` in the result's domain when only
`[-1, 0, 1, 2]` have support. No proof is written before the failure — the test
aborts on the consistency check.

## Scope

A full sweep run (`ctest -j 96 -R _view_`, `GCS_TEST_CAP_DEFAULTS=OFF`,
7314 tests) on `main` @ 977532a6 gives **68 failures, all of them
`divide_modulus_constraint_view_*`**, spread across wraps w1–w18 and positions
pall/p0/p1/p2. Every other constraint in the sweep passes under every wrap. So
this is the whole of the sweep's current red, and it is one constraint.

## Why it is worth a ticket rather than a note

`dev_docs/view-proof-logging.md` says under Status that "all constraints
registered in the sweep verify under all wraps", and `CLAUDE.md` said the sweep
was mostly red. Both were wrong in different directions, and the second made the
sweep useless as a gate: if you believe most of it fails anyway, a genuine
regression in it does not stand out. Fixing the two statements is part of #882;
this issue is the thing they were obscuring.

[...]

PR #904, opened 2026-09-09
Give views their own range literals, linked to the underlying variable's (#882)

### LINKED ISSUE #882: Views have no range literal, so every interval-shaped fact carries a view detour (eight and counting)

ciaranm:
Closes #882.

A registered view now owns its range ("in") literals, exactly as it already
owned its equality and order atoms, and they are joined to the underlying
variable's by a pair of link clauses. All ten view detours the issue inventoried
are deleted rather than adjusted — the three that threw `UnimplementedException`
and the seven that silently degraded to one proof line per value (`min_max` and
`all_equal` had grown their own since the issue was filed).

Full write-up in the new `dev_docs/view-range-literals.md`. The short version,
and the three things that were not obvious.

## 1. Design B, and why not A

Design A of the issue — deview a view's range conditions onto the underlying
variable, joining the four scalar operators in `simplify_literal`'s deview arm —
was rejected. It would make ranges the one atom kind that is *not* in `V`-form,
so a `pol` step resolving a range literal against a `V`-form constraint body
would strand: that is invariant 1 of `dev_docs/view-proof-logging.md`, and it is
what the whole view design rests on. Design B keeps `V` indistinguishable from a
bare variable, which is the property that makes the generic machinery work.

The deview arm still handles range conditions, but only on the path it already
handled the other operators: an *unregistered* view, first seen during proof
logging, which has no encoded variable to be consistent with.

## 2. The issue's premise about linking was half right

The issue recorded #881's observation that "a range literal never has to cross an
equality: it is reified against its variable's own two order cuts, so an interval
fact decomposes into cross one bound, move within one variable, cross back", and
suggested B might therefore need no linking at all.

That is exactly true of the two *implications* `F_V → F_X` and `F_X → F_V`, both
of which are one unit-propagation pass with no new clauses. But UP uses a clause
in whichever direction has a unit, and the directions that matter are the
negative ones, `¬F_X ⟹ ¬F_V` and `¬F_V ⟹ ¬F_X`. Neither is derivable. A negated
interval literal is a unit on *neither* cut — its reverse reification leaves a
binary clause — so it can only descend its own variable's containment edges, and
bottoms out at cells that are themselves unlinked range literals unless they are
width 1. The argument decomposes a conjunction of two bounds; a negated interval
is a disjunction of two, and UP cannot carry a disjunction one disjunct at a time.

Both directions are ordinary rather than exotic: search branches on real
variables only, while `generic_reason` states hole runs over the *operand*, which
under a wrap is the view.

So the cost is two rup clauses per interval literal per registered view, plus the
mirrored literal — the same constant factor eq and ge atoms already pay, and
nothing proportional to a domain width.

## 3. Linking has to be triggered by *naming*, and that is where this first broke

A range literal comes into existence two ways: some caller asks for it, which
reaches `need_invar`; or the partition machinery creates it as a **cell**, from
`ensure_partition_cut` / `init_interval_partition`, straight through
`define_plain_invar`. A cell is therefore never "requested" — and both
`need_proof_name` and `xliteral_for_ensuring` short-circuit on a condition that
already has a proof name, so a decision, reason or conclusion that happens to
name one of those cells never reaches `need_invar` at all.

Linking only what is requested therefore leaves every cell unlinked. That is a
silent unit-propagation dead end, and it is what the first version of this branch
did: `equals` under `--view-position=mixed` failed a backtrack clause with
`[view in 11..12]` and `[x in 1..2]` both defined and nothing joining them.
`mirror_invar_across_view_link` is now called from `need_invar` outside its
"already defined" guard, and again from `xliteral_for_ensuring` whenever a range
literal reaches a proof line.

One consequence: `need_invar` is now genuinely re-entrant, because resolving a
literal while emitting a covering can re-enter it. `ensure_partition_cut`
publishes a partition boundary *before* defining the two halves it creates, so
during that window the partition claims a cell that does not exist yet.
`define_invar_with_covering`'s single-cell path defines it in that case, and
`define_plain_invar` is idempotent so the cell is not then defined twice.

## Evidence

**Four new witnesses**, gate-on and veripb-verified, in the style of
`dev_docs/range_literals_spec.md` §8. W6 and W7 are the two directions of the
crossing; W8 and W9 are the trigger, a literal a partition cell already brought
into existence. Validated by ablation with a temporary knob, since removed:

| ablation                | W6     | W7     | W8     | W9     |
|-------------------------|--------|--------|--------|--------|
| none                    | passes | passes | passes | passes |
| both link clauses gone  | FAILS  | FAILS  | FAILS  | FAILS  |
| L1 removed              | FAILS  | passes | FAILS  | FAILS  |
| L2 removed              | passes | FAILS  | passes | FAILS  |
| link on request only    | passes | passes | FAILS  | FAILS  |

Every clause and every rule is separately load-bearing, which is stronger than
the spec records for W2 and W5, both of which are structural.

`invar_view_test` covers what the witnesses and the sweep between them cannot:
`need_pol_item_defining_literal`'s range arms, which have no caller at all — every
constraint using that function passes a bound or an equality, so the arms
replacing the throws would otherwise ship untested — and the endpoint arithmetic
in both directions, asserted with `find_xliteral_for` so that "the mirror exists
at exactly these bounds and not the neighbouring ones" is a real claim.
Mutation-validated: an off-by-one and a dropped endpoint swap are both caught.
It explicitly does *not* witness the link clauses, and says so in its header —
every RUP line in it still verifies with the pair deleted, which is §1's P1/P2
trap, and an earlier draft of it claimed otherwise until I ablated it.

**Why the sweep is not the evidence.** The negative crossing happens to work with
no linking clause at all whenever the range touches a definition bound (the
boundary pin makes the reverse reification a unit), is the whole range, or has
every cell inside it shattered to width 1. Measured two ways: over domain widths
4–16 with 0–7 random requests, **80% of instances cross correctly with no link
present**; and over 987 randomly generated instances that actually reach a
conflict, ablating the links makes only **37** of them fail. The sweep is about
96% coincidence.
A green randomised sweep is therefore close to no evidence, which is why the
witnesses above are constructed to sit strictly inside both variables' bounds,
with an interior cell of width ≥ 2 that nothing shatters. The control that gets
this wrong — decide `¬[b in 3..4]` instead of `¬[b in 3..5]` — verifies with
either clause removed.

**Proof size**, one `Equals` with a single wide interior hole, root propagation
only, every row VeriPB-verified:

| width | plain | view before | view after | negview after | mixed after |
|---|---|---|---|---|---|
| 10^3 | 40 | 36,012 | **106** | 106 | 109 |
| 10^4 | 40 | 360,012 | **106** | 106 | 109 |
| 10^5 | 40 | *timed out at 300 s* | **106** | 106 | 109 |
| 10^6 | 40 | *timed out at 300 s* | **106** | 106 | 106 |

`before` is `2 + 36·w` and stops being writable between 10^4 and 10^5. `after` is
flat; 106 rather than 40 is the view's own encoding and the link pairs, a
constant.

**Suites**, all with `GCS_TEST_CAP_DEFAULTS=OFF`:

- GCC 15.2, full suite **plus** the `--view-wrap` sweep, run **twice**: 8050
  tests each time, **68 failures each time, a byte-identical set** to the 68 that
  fail on unmodified `main` — every one a `divide_modulus_constraint_view_*`,
  failing on propagation strength ("gac not achieved in test") with no proof
  written. The two runs are worth having separately because these tests reseed,
  so the second pass is an entirely different set of instances. Zero
  regressions. That pre-existing failure is now filed as #903; it is also the
  reason `CLAUDE.md`'s "most fail today" and `dev_docs/view-proof-logging.md`'s
  "all constraints verify under all wraps" — which contradicted each other, and
  were both wrong — are corrected here.
- clang 21, full suite **plus** the sweep: **8054 tests, 68 failures**, the same
  byte-identical `divide_modulus` set — a third independent instance stream. No
  warnings in any changed file.
- `GCS_LARGE_DOMAIN_GUARD=ON`: `large_domain_audit_test` green, full suite
  **815/815**. Worth running because `all_equal.cc` loses a
  `LargeDomainIterationCounter` here along with the per-value walk it guarded,
  and because that guard is the tripwire for reintroducing work proportional to a
  domain width.

## Completeness

The obligation was formalised rather than assumed, in three parts: **mirror
closure** (an interval exists on one side iff its image exists on the others,
with the link pair present — because `link` requests the far side, which mirrors
onward, so a literal first seen on `V₁` reaches `V₂` through `X` in the same
call); **unit transport** (any literal that is a unit on one side has its image a
unit on every other side — ge-links for bounds, eq-links for equalities, L1/L2 for
intervals); and **per-side UP-completeness**, by induction on (cells inside the
literal, width), which works because every in-bounds endpoint of every interval
and equality is a partition boundary, so a cell is either wholly inside a literal
or disjoint from it. The root covering turns out not to be needed, matching the
observation already recorded in §8.1 of the range spec.

What it depends on — and so what would break it — is written into
`dev_docs/view-range-literals.md` §5: that every non-`red` emission names its
literals, that domain changes are logged eagerly, and that in-bounds endpoints are
partition boundaries.

That last one supersedes how §3 originally put it. There is no such thing as an
unnamed range literal in a definitions-on proof: every creation path writes a
covering or an edge about the literal it just created, in the same call, and
those render with `EnsureNames::Yes`. So a cell is linked when its own covering
is written, not when a later solver fact happens to mention it.

## Consultation

Two Fable sessions were used as proofs consultants on narrow questions. The first
adjudicated the negative-crossing claim with a closure argument plus an
independent Python re-implementation of the clause families driven by a PB unit
propagator, and produced the 80% figure. The second found the unlinked-cell
defect independently, with its own VeriPB counterexamples; W8 and W9 are its
instances. One claim of its I checked and did not
confirm: it reported that the two sides' literal families are not affine images
of each other. That was true of the intermediate build it reviewed, but is not
true of the finished one — measured across 258 proofs, all 300 registered
(variable, view) pairs match exactly, because coverings name their cells and
naming is now what links them. The dev_doc records the corrected version, and
also records that the correspondence must not be used as the correctness
argument: nothing maintains it as an invariant, and no test checks it.

https://claude.ai/code/session_018s5Rxd5qtz8CRHe53cFzLy

### DIFF
--- a/dev_docs/view-proof-logging.md
+++ b/dev_docs/view-proof-logging.md
@@ -10,8 +10,7 @@ machinery.
 
 The short version: every constraint currently registered in the view-wrap
 sweep verifies under every wrap. Abs and AllDifferent — historically the
-hard cases — are now in that set. (One constraint, `divide_modulus`, does fail
-that sweep, but on propagation strength rather than on its proof — issue #903.)
+hard cases — are now in that set.
 
 ## Part 1 — The idea, on top of the thesis
 
--- a/dev_docs/view-proof-logging.md
+++ b/dev_docs/view-proof-logging.md
@@ -10,7 +10,8 @@ machinery.
 
 The short version: every constraint currently registered in the view-wrap
 sweep verifies under every wrap. Abs and AllDifferent — historically the
-hard cases — are now in that set.
+hard cases — are now in that set. (One constraint, `divide_modulus`, does fail
+that sweep, but on propagation strength rather than on its proof — issue #903.)
 
 ## Part 1 — The idea, on top of the thesis
 
@@ -159,7 +160,13 @@ You only need to think about views when your justification emits **explicit
 
 Reasons over a view (e.g. `{v ≥ lb, v ≤ ub}`) are fine: they are logged in
 the view's atoms and UP crosses to `X` through the eq/ge links if the rest
-of the proof needs it.
+of the proof needs it. The same now goes for range ("in") conditions — a view
+has its own range literals, linked to the underlying variable's — but the
+linking there is less forgiving than for eq and ge atoms, because a *negated*
+range literal cannot cross on the atom links alone. If you are touching that
+machinery, read `dev_docs/view-range-literals.md` first; in particular, do not
+assume the two sides' literal families are affine images of each other, because
+they are not.
 
 ### If you are touching the view machinery itself
 
@@ -203,6 +210,12 @@ actually exercised:
    link (`derive_deviewed_form_for`). Runtime emissions do *not*
    auto-derive the `X`-form (size); callers opt in with
    `_then_deview` / `PolBuilder` deview-mode.
+5. **Range-literal links `[V in a..b] ⇔ [X in a'..b']`** —
+   `NamesAndIDsTracker::mirror_invar_across_view_link` mirrors the interval onto
+   the other side and emits both rup clauses, triggered whenever a range literal
+   is *named* rather than when it is defined. See
+   `dev_docs/view-range-literals.md`, which explains why both clauses are needed
+   and why naming is the right trigger.
 
 The reification big-M lives in `NamesAndIDsTracker::reify`; the operand-form
 choice lives in the per-constraint `justify_*` helpers.
--- a/dev_docs/view-proof-logging.md
+++ b/dev_docs/view-proof-logging.md
@@ -164,9 +164,10 @@ of the proof needs it. The same now goes for range ("in") conditions — a view
 has its own range literals, linked to the underlying variable's — but the
 linking there is less forgiving than for eq and ge atoms, because a *negated*
 range literal cannot cross on the atom links alone. If you are touching that
-machinery, read `dev_docs/view-range-literals.md` first; in particular, do not
-assume the two sides' literal families are affine images of each other, because
-they are not.
+machinery, read `dev_docs/view-range-literals.md` first; in particular, the two
+sides do hold the same literals, but not in the same *roles* — the same interval
+can be a partition cell on one side and a request with its own covering on the
+other — so argue per side rather than by assuming the structures match.
 
 ### If you are touching the view machinery itself
 
--- a/dev_docs/view-range-literals.md
+++ b/dev_docs/view-range-literals.md
@@ -296,25 +296,44 @@ nothing here is proportional to a domain width.
 
 Measured on one `Equals` whose operands are a bare `0..w` and a two-value
 `{0, w}`, so root propagation prunes exactly one wide interior run; proofs on,
-root propagation only, `.pbp` lines, every row VeriPB-verified. `view` wraps both
-operands as `x + 10^6`, `negview` as `-x + 10^6`, `mixed` gives the two operands
-different wraps of opposite sign.
+root propagation only, `.pbp` lines, every `after` row VeriPB-verified with
+`--force-checked-deletion`. `view` wraps both operands as `x + 10^6`, `negview`
+as `-x + 10^6`, `mixed` as `x + 10^6` against `-y + 10^6 + w`.
 
 | width | plain | view before | view after | negview after | mixed after |
 |---|---|---|---|---|---|
-| 10^3 | 40 | 36,012 | **106** | 106 | 109 |
-| 10^4 | 40 | 360,012 | **106** | 106 | 109 |
-| 10^5 | 40 | *timed out at 300 s* | **106** | 106 | 109 |
-| 10^6 | 40 | *timed out at 300 s* | **106** | 106 | 106 |
+| 10^3 | 68 | 71,017 | **200** | 200 | 200 |
+| 10^4 | 68 | 710,017 | **200** | 200 | 200 |
+| 10^5 | 68 | *timed out at 300 s* | **200** | 200 | 200 |
+| 10^6 | 68 | *timed out at 300 s* | **200** | 200 | 200 |
 
-The `before` column is `2 + 36·w` lines and stops being writable at all between
-10^4 and 10^5. The `after` columns are flat, and 106 rather than 40 because the
+The `before` column is `17 + 71·w` lines and stops being writable at all between
+10^4 and 10^5. The `after` columns are flat, and 200 rather than 68 because the
 view carries its own bit-vector encoding, its own order and equality atoms and
 the link pairs — a constant, not a function of the width. That constant is the
-price of the design; the slope was the point of the issue.
+price of the design; the slope was the point of the issue. `plain` is the
+control that makes the two builds comparable: it is the same 68 on either side
+of the change, as it should be, since nothing here touches a bare variable.
 
-VeriPB checks every `after` row in 0.01–0.02 s, against 5.4 s for the largest
-`before` row that finishes at all.
+VeriPB checks every `after` row in about 0.012 s. The `before` row at 10^3 takes
+42 s, and the one at 10^4 — ten times the lines — was still going after an hour,
+when it was stopped. So writability is the generous half of the `before`
+column's problem: the widths that can still be written are already past the
+widths that can be checked.
+
+The `mixed` offsets are not free to choose, and getting them wrong is quiet.
+The two wraps have to put both operands over the *same* interval; otherwise they
+stop overlapping, the instance is unsatisfiable at the root, and the row measures
+a refutation rather than the hole prune. An earlier version of this probe paired
+`x + 10^6` with `-y + 2·10^6`, which overlap only at `w = 10^6` — so three of the
+four `mixed` rows were the wrong shape, and the fourth was the right one. The
+only tell was a three-line step in a column that should have been flat.
+
+These figures are against `main` at 7b582656, and are about twice the ones this
+branch first reported. #914 puts each of the tracker's definition lines into
+VeriPB's core set, which writes an extra `core id` line for each; both columns
+grow by that factor. Flat against linear is unchanged, and so is everything the
+table is used for.
 
 ## 9. The one residue
 
--- a/dev_docs/view-range-literals.md
+++ b/dev_docs/view-range-literals.md
@@ -201,7 +201,7 @@ the range spec already assumes; and that in-bounds endpoints are partition
 boundaries — which is why `need_direct_encoding_for` cuts at `v` and `v+1` when a
 partition exists, and dropping that would break the base case above.
 
-Corroboration, which is not the argument but is worth having: 2998 randomly
+Corroboration, which is not the argument but is worth having: 3000 randomly
 generated view proofs verified with zero mirror-closure violations, over Equals,
 NotEquals, AllEqual, Min, Max, In, LessThan, LessThanEqual and AllDifferent with
 holey domains, three-view bases, and intervals pushed outside the definition
@@ -227,7 +227,7 @@ Measured two ways, both damning for the sweep as evidence. Over domain widths
 4–16 with 0–7 random requests per instance: **80% of instances cross successfully
 with no linking clause at all** — abut-high 39%, abut-low 23%, whole-range 12%,
 shattered 5.6% — and of the 20% that stall, a single wide cell accounts for 88%.
-And over 985 randomly generated instances that actually reach a conflict, with
+And over 987 randomly generated instances that actually reach a conflict, with
 the link clauses ablated: only **37 of them notice**. The sweep is about 96%
 coincidence.
 
--- a/dev_docs/view-range-literals.md
+++ b/dev_docs/view-range-literals.md
@@ -20,8 +20,9 @@ own order atoms and its own equality atoms, joined to `X` by a single
 definitional link axiom plus lazily emitted atom-level biconditionals. Range
 literals join that list: `[V in a..b]` is an ordinary range literal over `V`,
 reified against `V`'s own two order cuts, with `V`'s own partition, coverings
-and containment edges. Every interval request on either side is mirrored onto
-the other, and the two literals are joined by **two** rup clauses. Nothing
+and containment edges. Every interval named on either side is mirrored onto the
+other — *named*, not requested, which §3 is entirely about — and the two
+literals are joined by **two** rup clauses. Nothing
 above the encoding layer knows any of this happened: a propagator says
 `infer_not_in_range(operand, lo, hi)` and states hole runs as range conditions
 over the operand, whether or not the operand is a view.
@@ -111,9 +112,10 @@ failure. `mirror_invar_across_view_link` is therefore called from `need_invar`
 whenever a range condition that already has a name reaches a proof line. It is
 idempotent, and after the first time it is a set probe.
 
-Literals that are never named need nothing, and cost nothing: a fact only ever
-enters a variable's literal family through a named literal or through a bound,
-and both of those cross.
+A literal that is never named needs nothing, because a fact only ever enters a
+variable's literal family through a named literal or through a bound, and both
+of those cross. As it happens that set turns out to be empty — §4 — but the rule
+does not depend on it being, and should not be changed to.
 
 One consequence worth knowing about: `need_invar` is now genuinely re-entrant,
 because resolving a literal while emitting a covering can re-enter it.
--- a/dev_docs/view-range-literals.md
+++ b/dev_docs/view-range-literals.md
@@ -208,11 +208,24 @@ fact has to compose `view1 → b → view2`, needs both clauses. Every clause an
 every rule here is separately load-bearing, which is a stronger result than the
 spec records for W2 and W5, both of which are structural.
 
-Note for anyone adding to this set: the obvious control for W8 — decide
+`invar_view_test` sits alongside them and covers something different: the
+`need_pol_item_defining_literal` range arms, which no propagator reaches, and the
+sign arithmetic of the endpoint map in both directions — checked with
+`find_xliteral_for`, which does not introduce what it fails to find, so "the
+mirror exists at exactly these bounds and not at the neighbouring ones" is a real
+assertion. Both an off-by-one and a dropped endpoint-swap mutant are killed by
+it. It deliberately does **not** witness the link clauses, and says so: it states
+the crossings as RUP lines, and every one of them still verifies with the link
+pair deleted, because a RUP check is strictly stronger than the single
+unit-propagation pass the links exist for. That is §1's P1/P2 distinction, and it
+caught this document claiming otherwise in an earlier draft.
+
+Two notes for anyone adding to this set. The obvious control for W8 — decide
 `¬[b in 3..4]` instead of `¬[b in 3..5]`, so the decision names a fresh interval
 rather than a cell — verifies with *either* link clause removed, because the
 interval then sits at the bottom of a covering-forced block and the bound
-crosses. Ablate before claiming a witness bites.
+crosses. And a witness stated as a RUP line of the fact you care about tests
+nothing at all here. Ablate before claiming a witness bites.
 
 ## 7. Cost
 
--- a/dev_docs/view-range-literals.md
+++ b/dev_docs/view-range-literals.md
@@ -1,0 +1,249 @@
+# Range ("in") literals on views
+
+How the two mechanisms in `dev_docs/range_literals_spec.md` and
+`dev_docs/view-proof-logging.md` combine. Neither document covered this, because
+until issue #882 they did not combine: a `ViewOfIntegerVariableID` had no range
+literal at all, and every part of the solver that wanted to say something
+interval-shaped about a view carried its own detour — three that threw, and five,
+later seven, that silently degraded to one literal per value.
+
+**Audience.** Anyone touching `need_invar`, `simplify_literal`, the view
+machinery in `NamesAndIDsTracker`, or a propagator that emits explicit `pol`
+over a range literal. Read both of the documents above first; this one assumes
+the partition/covering/containment apparatus of §3 of the range spec and the
+three view invariants of the view document.
+
+## 1. The design in one paragraph
+
+A registered view `V = sX + c` already owns its own bit vector `BinEnc(V)`, its
+own order atoms and its own equality atoms, joined to `X` by a single
+definitional link axiom plus lazily emitted atom-level biconditionals. Range
+literals join that list: `[V in a..b]` is an ordinary range literal over `V`,
+reified against `V`'s own two order cuts, with `V`'s own partition, coverings
+and containment edges. Every interval request on either side is mirrored onto
+the other, and the two literals are joined by **two** rup clauses. Nothing
+above the encoding layer knows any of this happened: a propagator says
+`infer_not_in_range(operand, lo, hi)` and states hole runs as range conditions
+over the operand, whether or not the operand is a view.
+
+The alternative — rewriting a view's range conditions onto `X`, as
+`simplify_literal`'s deview arm does for the four scalar operators — was
+rejected. It would make ranges the one atom kind that is *not* in `V`-form,
+and a `pol` step resolving a range literal against a `V`-form constraint body
+would then strand (view invariant 1). The deview arm still handles range
+conditions, but only on the path it already handled the others: an
+*unregistered* view, seen for the first time during proof logging, which has no
+encoded variable to be consistent with.
+
+## 2. The link pair
+
+For a registered view `V = sX + c` the interval `[a, b]` on `V`'s value scale
+corresponds to
+
+    s = +1:  [a - c,  b - c]
+    s = -1:  [c - b,  c - a]
+
+on `X`'s — the endpoints swap for a negated view, because it reverses the
+order, and the map is then its own inverse. Width is preserved either way, which
+is why a width-1 request is the eq atom on both sides and never reaches this
+machinery at all. `interval_on_underlying` and `interval_on_view` are the two
+directions.
+
+Writing `F_V` for `[V in a..b]` and `F_X` for its image, the link is
+
+    L1:  ¬F_V ∨ F_X
+    L2:  ¬F_X ∨ F_V
+
+Each is RUP at emission: assert one and the negation of the other, and the
+forward reification gives that side's two cuts, the ge-links carry each cut
+across (for `s = -1` the two cuts swap roles, which is exactly the endpoint
+swap above), and the far side's reverse reification fires.
+
+**Both clauses are needed, and each for its contrapositive rather than for its
+implication.** As implications, `F_V → F_X` and `F_X → F_V` are both already
+derivable by unit propagation — one forward reification, two ge-links, one
+reverse reification, no new clauses. That is what issue #882 observed, and it
+is true. But UP uses a clause in whichever direction has a unit, and the
+directions that matter are the negative ones:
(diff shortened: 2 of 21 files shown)


ciaranm:
`min_max.cc` loses its step-3 `both_simple` here, but keeps the step-1 one at
`min_max.cc:118` — the `result in union(vars)` rule. The same argument looks like
it covers that site too, so I think it is a miss rather than a deliberate
exception.

Measured on this branch at `0af4d0fc`, with that one gate forced true and nothing
else changed (`min_max_test --prove`, seed 1 unless said otherwise):

| configuration | verified | rejections |
|---|---|---|
| bare | 88 | 0 |
| `--view-wrap=3` / `9` / `11`, `--view-position=all` | 80 each | 0 |
| `--view-position=mixed`, seeds 1, 2, 3 | 80 each | 0 |

The runs are not passing by failing to reach the path: with the gate off, the
interval branch fires **1058** times under `--view-wrap=9 --view-position=all` (a
negated view in all six positions), against **0** with it on. And the derivation
is doing real work over the view rather than being carried by something else —
dropping just the two bound lemmas in that same configuration is rejected by
VeriPB after 45 instances. So the crossing does happen on the view's own encoded
variable, as the new header paragraph says it should.

Why it seems worth doing rather than leaving: while the gate is there, a view'd
`result` or array entry keeps the per-value walk over `dom(result)`, which is
both the #833 hazard and, by that site's own comment, the only thing bounding
`result` from the loose side (#815).

For what it is worth, #912 gives `In`'s equivalent of step 1 the same treatment
— with both bound lemmas guarded, since both halves of its link are
half-reified — and I get the same answer there: with this branch merged in and
`In`'s guard removed, views on the pruned variable and on the sources both
verify, hole-shaped domains included, and the lemmas stay load-bearing.

CLAUDE.md
```diff
@@ -66,8 +66,11 @@ Disable XCSP or MiniZinc support to reduce dependencies:
 cmake -S . -B build -DGCS_ENABLE_XCSP=OFF -DGCS_ENABLE_MINIZINC=OFF
 ```
 
-Enable the view-wrap proof-verification sweep (~3000 extra tests, most
-fail today; opt in when working on view proof logging):
+Enable the view-wrap proof-verification sweep (~7200 extra tests; opt in when
+working on view proof logging). Measured 2026-09-09 on `main`: all pass except
+the 68 `divide_modulus_constraint_view_*` entries, which fail on propagation
+strength ("gac not achieved in test") rather than on any proof — that is issue
+#903, and it is the only known failure in the sweep:
```
ciaranm: **Fix 1 of 3 (site 1 of 2): this is now false — #909 closed #903 yesterday.**

#903 was closed at 2026-09-09T21:22:34Z by #909, which landed on `main` along with #911. Those 68 `divide_modulus_constraint_view_*` lanes now pass, so this text would ship telling developers about a failure that no longer exists, and would send anyone who hits a red sweep looking at the wrong issue.

Measured on a trial merge of this branch onto current `main` (`9f7dabb9`), with `-DGCS_ENABLE_VIEW_WRAP_SWEEP=ON`:

```
ctest -R '_view_' --parallel 12
100% tests passed, 0 tests failed out of 7315
```

So the sweep now has **no known failures at all**. The `~7200 extra tests` figure is right (I count 7315 sweep lanes), it is only the failure claim that has expired.

Suggested: drop the `Measured ... only known failure in the sweep` sentence, or restate it as "all pass as of 2026-09-10". Same claim needs fixing in `dev_docs/view-proof-logging.md:13` — see the other comment.

ciaranm (commented):
Pre-merge review (merger-Claude). Three things to fix, then this is good to land as far as I can tell.

Everything substantive checks out, and I confirmed the central evidence rather than taking it on trust: I patched `mirror_invar_across_view_link` myself and **reproduced the §7 ablation matrix exactly, cell for cell** on all three rows that are reachable without the removed knob (both clauses gone; L1 only; L2 only). W1/W2/W3/W5 stayed green, so the ablation was targeted. §7's subtlest claim also holds — `invar_view_test` really does *not* witness the link clauses; it passes with both ablated.

Also verified on a trial merge onto current `main`: 0 compiler warnings from 0 sites (GCC 15.2, 145 binaries, build reached 100%); 815/815 ordinary suite; **7315/7315 view-wrap sweep**; guard lane green at 186 assertions / 93 rows with `2 x 93 = 186`. The five new tests are all wired through `run_test_and_verify.bash`, so VeriPB does check them. `need_pol_item_defining_literal`'s range arms genuinely have no production caller, so adding `invar_view_test` for them was the right call. No default-path cost: a plain variable pays one `get_if` plus one failed hash lookup, and only with proofs on.

Two things that would trip a naive scan and are **not** problems: the new `AssertProofRule` site is this file's existing convention (4 instances -> 5) and is gated on a non-default level, with `Off` the default and `Off < Links` taking the RUP branch; and every new `overloaded{}` site is in correct switch-block layout.

Three fixes below. The first is the one that matters — it is pure bad luck from #909 landing while this sat.

dev_docs/view-proof-logging.md
```diff
@@ -10,7 +10,8 @@ machinery.
 
 The short version: every constraint currently registered in the view-wrap
 sweep verifies under every wrap. Abs and AllDifferent — historically the
-hard cases — are now in that set.
+hard cases — are now in that set. (One constraint, `divide_modulus`, does fail
```
ciaranm: **Fix 1 of 3 (site 2 of 2): same expired claim, and here it makes the file contradict itself.**

`divide_modulus` no longer fails the sweep — #909 closed #903 yesterday, and I measure 7315/7315 sweep lanes passing on this branch merged onto current `main`.

The sharper problem is local: line 274 of this same file still says

> All constraints registered in the sweep verify under all wraps, including Abs and AllDifferent

which post-#909 is simply **true**. With this parenthetical added, the file asserts both "one constraint does fail" (line 13) and "all constraints verify" (line 274). That is the same self-contradiction between two documents that this PR is fixing here — reintroduced inside one file.

Suggested: drop the parenthetical entirely and leave line 274 standing, since it is now accurate.

gcs/constraints/innards/justify_not_in_range.hh
```diff
@@ -20,10 +20,18 @@ namespace gcs::innards
     // [other_lo, other_hi] are the bounds the range forces on `other` through the
     // equality; for a plain `pruned = other` they are [lo, hi]. The pairing
     // assumes a same-sign link: a sign-flipped link (such as Abs) needs the
-    // mirrored pairing. `other` must be a plain integer variable, equality-linked
-    // to `pruned` in the proof model. Pass the same reason the caller hands to
-    // infer_not_in_range.
-    auto justify_not_in_range_across_equality(ProofLogger & logger, const ReasonLiterals & reason, const SimpleIntegerVariableID & pruned, Integer lo,
+    // mirrored pairing. `other` must be equality-linked to `pruned` in the proof
```
ciaranm: **Fix 2 of 3: this sentence was retracted on `main` while the branch sat, and the conflict resolution is not mechanical.**

This is the only file where the branch conflicts with current `main`, and the conflict is semantic rather than textual, so it wants your hands rather than a merge-tool default.

`main` now carries #905, which **retracted** exactly this claim. Its replacement text says the opposite:

> **The pairing is the hypothesis, and a sign-flipped link does not satisfy it.** Against `pruned = -other` the row is a sum rather than a difference, the negation supplies two bounds on the same side, and the configuration is outside the theorem [...] do not reach for this helper with a mirrored pairing and expect it to hold.

So resolving in favour of this side would put a retracted claim back into the header, and #913 is currently finishing the cleanup of its last remaining site.

The good news is that the substance is compatible. The new paragraph below (lines 26-33) makes a point that is genuinely **orthogonal** to the Theorem 2.9 correction — that the sign which matters is the one on the link between the two *operands*, not either view's wrap — and that survives #905 untouched. What has to go is only the retained `a sign-flipped link (such as Abs) needs the mirrored pairing` sentence.

Two other bits of the same hunk to keep in view when you resolve:
- The widening `SimpleIntegerVariableID` -> `IntegerVariableID` is this PR's point and should win; but #905's prose asserts ``other`` "must be a plain integer variable", which this branch deliberately drops. Both halves of that need reconciling, not just the signature.
- The signature widening is source-compatible for the existing callers (`in.cc`, `global_cardinality.cc`) that this branch does not touch, so nothing else needs changing.

I resolved it locally to run the test suites and deliberately did not push that.

dev_docs/view-range-literals.md
```diff
@@ -0,0 +1,333 @@
...
+different wraps of opposite sign.
+
+| width | plain | view before | view after | negview after | mixed after |
+|---|---|---|---|---|---|
+| 10^3 | 40 | 36,012 | **106** | 106 | 109 |
+| 10^4 | 40 | 360,012 | **106** | 106 | 109 |
+| 10^5 | 40 | *timed out at 300 s* | **106** | 106 | 109 |
+| 10^6 | 40 | *timed out at 300 s* | **106** | 106 | 106 |
+
+The `before` column is `2 + 36·w` lines and stops being writable at all between
```
ciaranm: **Fix 3 of 3 (a): the formula and the table disagree — the table is right.**

`2 + 36*w` does not reproduce the cells above it:

| w | `2 + 36w` | table |
|---|---|---|
| 10^3 | 36,002 | **36,012** |
| 10^4 | 360,002 | **360,012** |

Solving for the constant gives 12 at both widths, so the table is self-consistent and the formula's constant is the typo: it should be **`12 + 36*w`**.

Worth fixing rather than dropping, since this is the figure someone will use to predict the cost before deciding to measure. The same `2 + 36·w` appears in the PR description.

dev_docs/view-range-literals.md
```diff
@@ -0,0 +1,333 @@
...
+root propagation only, `.pbp` lines, every row VeriPB-verified. `view` wraps both
+operands as `x + 10^6`, `negview` as `-x + 10^6`, `mixed` gives the two operands
+different wraps of opposite sign.
+
+| width | plain | view before | view after | negview after | mixed after |
+|---|---|---|---|---|---|
+| 10^3 | 40 | 36,012 | **106** | 106 | 109 |
+| 10^4 | 40 | 360,012 | **106** | 106 | 109 |
+| 10^5 | 40 | *timed out at 300 s* | **106** | 106 | 109 |
+| 10^6 | 40 | *timed out at 300 s* | **106** | 106 | 106 |
```
ciaranm: **Fix 3 of 3 (b): the `mixed after` column is not flat, which the sentence just below the table says it is.**

The column reads 109, 109, 109, **106** — dropping by 3 at 10^6, where `view after` and `negview after` are 106 throughout:

| width | mixed after |
|---|---|
| 10^3 | 109 |
| 10^4 | 109 |
| 10^5 | 109 |
| 10^6 | **106** |

Line 311 then says "The `after` columns are flat". Either the 10^6 cell is a transcription slip for 109, or the drop is real and wants a clause explaining it — the offset in the `mixed` wraps coinciding with the width at exactly 10^6 would be the obvious suspect, and if that is what happened it is worth a sentence, because it is the kind of coincidence that makes a reader distrust the rest of the table.

I have not tried to reproduce the table, so I cannot tell you which it is.

CLAUDE.md
ciaranm: Right, and it turned out to need no edit at all: #919 moved the sweep's build documentation out of `CLAUDE.md` into `dev_docs/building.md` while this sat, and that file already carries the ~7200 figure and makes no failure claim. So the rebase onto 7b582656 drops this hunk entirely and the branch no longer touches `CLAUDE.md`.

I get the same answer you did on the sweep: **8170/8170 on GCC** and **8174/8174 on clang**, both with the sweep on and `GCS_TEST_CAP_DEFAULTS=OFF`, no failures either side.

The other site is fixed in ddde8993 — see the reply there.

dev_docs/view-proof-logging.md
ciaranm: Dropped the parenthetical in ddde8993, leaving line 274 to stand — which is the right way round, since it is now simply true.

Agreed on the sharper point: a parenthetical that contradicts the same file three hundred lines later is worse than the two-document contradiction this branch set out to fix, because there is no version of reading the file that gets you the right answer.

gcs/constraints/innards/justify_not_in_range.hh
ciaranm: Resolved by hand, in favour of #905. The header now reads:

- #905's Theorem 2.9 justification and its **bold retraction paragraph**, kept verbatim.
- The retracted `a sign-flipped link (such as Abs) needs the mirrored pairing` sentence, gone.
- `other must be a plain integer variable`, gone, along with the `SimpleIntegerVariableID` → `IntegerVariableID` widening it contradicted — you were right that those two have to move together.
- The view paragraph kept, but reworded so it reads as *support* for #905's hypothesis rather than as an independent claim next to it: "Either operand may be a view, and a wrap does not disturb that hypothesis. [...] The sign that has to match is the one on the link between the two *operands*, not the one on either view's wrap: a negated view has the matching negation in the model's own equality line, so `pruned = other` stays a difference row however either side is wrapped."

Confirmed the widening is still source-compatible for `in.cc` and `global_cardinality.cc`, which this branch does not touch; both compile unchanged on GCC and clang.

Note the stacked #906 makes one further edit to this same paragraph — "*their conclusions* mention no range literal", since the emitted lines are written under the caller's reason, which does name it. That is on the #906 side of the stack, rebased to sit on top of this resolution rather than to fight it.

dev_docs/view-range-literals.md
ciaranm: Your arithmetic was right — the table was self-consistent and the constant was the typo. I did not patch it, though, because the rebase moved what is underneath both columns: #914 puts every one of the tracker's definition lines into VeriPB's core set, which writes an extra `core id` line for each, so every cell in the table is now wrong rather than just the formula.

Remeasured against `main` at 7b582656, in c97ad190:

| width | plain | view before | view after | negview after | mixed after |
|---|---|---|---|---|---|
| 10^3 | 68 | 71,017 | **200** | 200 | 200 |
| 10^4 | 68 | 710,017 | **200** | 200 | 200 |
| 10^5 | 68 | *timed out at 300 s* | **200** | 200 | 200 |
| 10^6 | 68 | *timed out at 300 s* | **200** | 200 | 200 |

So `17 + 71·w`, and it does reproduce: 71·10³ + 17 = 71,017 and 71·10⁴ + 17 = 710,017.

`plain` earns its column here rather than being decoration — it is the same 68 on both builds, which is what says the two are comparable at all, since the baseline build had to be a fresh one.

Two things worth having that the old table did not say. Checking is the tighter limit than writing: each `after` row checks in about 0.012 s, the 10³ `before` row takes **42 s**, and the 10⁴ one was still going after **an hour** when I stopped it — so by the width at which the fallback stops being writable it has long stopped being checkable. And the `mixed` column is now flat too; see the other thread for why it was not.

dev_docs/view-range-literals.md
ciaranm: Not a transcription slip: the drop is real, it reproduces exactly on current `main` (203, 203, 203, **200** with the same probe), and your suspect — the offsets coinciding with the width at exactly 10^6 — is the right one. But the diagnosis is worse than a coincidence in an otherwise sound measurement.

The probe's `mixed` mode paired `x + 10^6` with `-y + 2·10^6`. With `x ∈ [0, w]` and `y ∈ {0, w}` that gives `xo ∈ [10^6, 10^6 + w]` against `yo ∈ {2·10^6 − w, 2·10^6}`, and those two only meet when `w = 10^6`. At the three smaller widths the equality has no support at all, so the instance is **unsatisfiable at the root** — `conclusion UNSAT` in the `.pbp`, against `conclusion SAT` at 10^6. Three of the four `mixed` rows were measuring a refutation and the fourth the interior-hole prune the table is about; the three-line difference is the interval literal at 10^6 degenerating to an equality atom, which is cheaper by exactly its partition apparatus.

So the column was not evidence for the sentence above it, and the only symptom was the step you spotted.

Fixed by pairing `x + 10^6` with `-y + 10^6 + w`, which is the negated wrap whose two values are exactly `xo`'s endpoints, so both operands sit over the same interval at every width. The column is then a flat **200**, SAT at every row, matching `view` and `negview`. c97ad190 has the numbers; the trap itself is written into §8, since a probe whose configuration degenerates silently at three widths out of four is worth one paragraph.

ciaranm:
Agreed, and taken — e768caf1 deletes the step-1 `all_simple` gate too, along with the `each_value_mutable(result)` fallback behind it. It was a miss: I removed the gate on the single-support path because I had an argument for it, and never went back to ask whether the same argument covered the site three hundred lines earlier. It does.

Your table reproduces exactly on the rebased branch, same flags, seed 1 unless stated:

| configuration | verified | rejections |
|---|---|---|
| bare | 88 | 0 |
| `--view-wrap=3` / `9` / `11`, `--view-position=all` | 80 each | 0 |
| `--view-position=mixed`, seeds 1, 2, 3 | 80 each | 0 |

And I reproduced the discriminating half rather than only the green half, since a green run over a view is exactly the thing this branch's own §6 says not to trust: with the two bound lemmas dropped and nothing else changed, VeriPB rejects after **45** instances under `--view-wrap=9 --view-position=all` — a configuration in which every instance is wrapped, so the rejection is on the view path. That also settles "is the branch reached", without needing the counter: an unreached branch cannot be ablated into a rejection.

Same for the link clauses themselves, which the rebase moved to `ProofLevel::TopAndCore` to follow #914: re-ran the §7 ablation against the rebased tree, and with both clauses removed W6–W9 fail while W1/W2/W3/W5 stay green and `invar_view_test` still passes, exactly as the table says.

`GCS_LARGE_DOMAIN_GUARD=ON` is 827/827 with `large_domain_audit_test` green, which matters more here than it did before — `min_max.cc` now loses a second per-value walk over `dom(result)`, and that guard is what would catch it coming back.

Good to hear #912 lands in the same place for `In`'s step 1, half-reified link and all.

PR #911, opened 2026-09-09
Give multiply's and power's view sweeps the domains they claim to test (#910)

### LINKED ISSUE #910: multiply and power build their view sweep the same wrong way divide_modulus did: most of it is empty instances
Follow-up to #903 / #909, which fixed this for `divide_modulus_test`.

`multiply_test` and `power_test` are the other two tests that build their view
sweep with a local `wrap` lambda -- keeping the underlying domain and folding the
wrap into the *predicate* -- instead of
`create_integer_variable_or_constant_with_view()`, which inverts the underlying
domain so the visible value set (and so the expected solution set) is unchanged.
Every other test in the sweep uses the helper.

The consequence is that their wraps change the problem rather than the encoding,
and most of what they change it into has no solutions:

| test | uniform-sweep runs with zero solutions | always-on `_view_mixed` |
| --- | --- | --- |
| `multiply_test` | 57% | **18 of 18 runs empty** |
| `power_test` | 44% | **12 of 12 runs empty** |
| (`divide_modulus_test`, before #909) | 25% | 24 of 24 empty |

Control: `plus_minus_test`, which uses the helper, reports the same zero-solution
count (34) at every wrap with the seed pinned -- vacuity there is a property of
the test data, not of the wrap.

Both are green today, so this is not a bug report: it is that their `_view_mixed`
lane, the one that runs in the ordinary test suite and that
`gcs/CMakeLists.txt` describes as "the most likely to surface a cross-view bug",
cannot fail at all, and about half of their opt-in sweep tests the empty instance.

## What to do

The same two things #909 did, minus the GAC expectation, which is already right
in both (`multiply` and `power` claim their result slot as a `DeterminedVariable`,
so `want_tabulation()` skips that level and their Auto trees are inside the
budget -- multiply's widest sweep case is 6*6*14 = 504, which becomes 36):

- move both onto `create_integer_variable_or_constant_with_view()`, passing the
  variable names now that the helper takes them (#909);
- check the resulting solution counts are wrap-invariant, and that the
  `_view_mixed` lane is non-empty.

Worth keeping an eye on `power_test`'s negated wraps in particular: a negated
base with an even exponent is where a predicate-composed sweep and a
domain-inverted one differ most.

## Then

Once this and #909 are in, one full `GCS_ENABLE_VIEW_WRAP_SWEEP=ON` run
(`ctest -j 96 -R _view_`, `GCS_TEST_CAP_DEFAULTS=OFF`) to establish what the
sweep actually says now, since the 68 failures in #903 were its whole red and
neither of the two statements about it (`dev_docs/view-proof-logging.md`'s Status
line, the CLAUDE.md note) has been true for a while. Those two statements are
#882's to fix.

ciaranm:
Closes #910. **Stacked on #909** — base is `issue903-divide-modulus-view-sweep`,
since this uses the name parameter that PR adds to the helpers. Merge that first.

`multiply_test` and `power_test` were the other two tests building their view
sweep with a local `wrap` lambda: they kept the underlying domain and folded the
wrap into the *predicate*, so a wrap changed the problem rather than the
encoding, and most of what it changed it into had no solutions at all. Now they
use `create_integer_variable_or_constant_with_view()` like the other 41 tests in
the sweep, with the names it takes as of #909.

Three commits: the shared log label (divide_modulus grew a local copy in #909,
and a third caller is the point at which it should live in the utils), then one
test each.

## What changes, and what deliberately does not

The **GAC expectations are untouched** in both, unlike #903's. Multiply's product
slot and Power's result slot are `DeterminedVariable`s, so `want_tabulation()`
skips that level and their Auto budgets here are 3\*3, 5\*2, 6\*6 (multiply) and
the base's 7 and 6 (power), all well inside the threshold of 100 — and, unlike
Modulus, neither has a magnitude-derived aux level for a wrap's offset to
inflate, so those budgets are wrap-invariant. Their sweeps could always have
GAC-checked `Auto` safely; they just weren't checking it against anything.

The main body's hand-written view cases move to wraps too, so the file has one
convention rather than two, and multiply's random BC lane draws a random wrap
triple instead of a random view spec.

## Verification

- `ctest` **810/810** with `GCS_WERROR=ON`, both `GCS_TEST_CAP_DEFAULTS`
  settings.
- **72/72** multiply and **54/54** power sweep configurations pass (18 wraps x
  pall/p0/p1/p2 and pall/p0/p1), run through the argv flags with the sweep off.
- **No empty run left**: zero `expecting 0 solutions` across all 126
  configurations, against 57% (multiply) and 44% (power) before. Solution counts
  are now identical at every wrap.
- The always-on `_view_mixed` lanes go from 18 and 12 runs of nothing to 9/10/35
  and 6/7 solutions.
- Negative control: sabotaging `invert_view()` to drop the offset is caught by
  every offset-bearing wrap in both tests, and is a genuine no-op at wrap 1
  (offset 0), as it should be.
- clang 21 `-fsyntax-only` clean on every changed TU, and both binaries pass
  under clang + `-D_GLIBCXX_ASSERTIONS`.

Worth saying explicitly: this newly-real GAC coverage found **nothing**. Both
constraints achieve GAC through every wrap in the sweep, which is what should
have been on record for them all along.

## Next

With this and #909 in, one full `GCS_ENABLE_VIEW_WRAP_SWEEP=ON` run to put a
number on what the sweep actually says. The sweep stays off by default either
way.

### DIFF
--- a/gcs/constraints/divide_modulus/divide_modulus_test.cc
+++ b/gcs/constraints/divide_modulus/divide_modulus_test.cc
@@ -64,17 +64,6 @@ namespace
             .visit(level);
     }
 
-    auto wrap_label(const vector<ViewWrap> & wraps) -> string
-    {
-        string s = "(";
-        for (const auto & w : wraps) {
-            if (s.size() > 1)
-                s += ", ";
-            s += std::to_string(w.bare ? 1 : (w.negate ? -1 : 1)) + ", " + std::to_string(w.bare ? 0 : w.offset);
-        }
-        return s + ")";
-    }
-
     auto post_divmod(Problem & p, bool is_div, IntegerVariableID x, IntegerVariableID y, IntegerVariableID out, const DivideConsistency & level)
         -> void
     {
@@ -95,7 +84,7 @@ auto run_divmod_test(bool proofs, bool is_div, const DivideConsistency & level, 
     pair<int, int> out_range, vector<ViewWrap> wraps = {view_none(), view_none(), view_none()}) -> void
 {
     print(cerr, "{} {} {} {} {} {} views {} {}", is_div ? "divide" : "modulus", level_name(level), check_gac ? "gac-checked" : "plain", x_range,
-        y_range, out_range, wrap_label(wraps), proofs ? " with proofs:" : ":");
+        y_range, out_range, view_wraps_label(wraps), proofs ? " with proofs:" : ":");
     cerr << flush;
     set<tuple<int, int, int>> expected, actual;
 
--- a/gcs/constraints/innards/constraints_test_utils.hh
+++ b/gcs/constraints/innards/constraints_test_utils.hh
@@ -1481,6 +1481,26 @@ namespace gcs::test_innards
     }
 
     /**
+     * \brief The per-position (multiplier, offset) rendering of a wrap list, for
+     * a test's log line.
+     *
+     * Reads as the view spec the arithmetic tests used to pass around by hand:
+     * `(1, 0, -1, 5)` is a first position that is bare or identity-viewed, and a
+     * second that is `-v + 5`. A bare position and an identity view read the
+     * same, since they describe the same values.
+     */
+    inline auto view_wraps_label(const std::vector<ViewWrap> & wraps) -> std::string
+    {
+        std::string s = "(";
+        for (const auto & w : wraps) {
+            if (s.size() > 1)
+                s += ", ";
+            s += std::to_string(w.bare ? 1 : (w.negate ? -1 : 1)) + ", " + std::to_string(w.bare ? 0 : w.offset);
+        }
+        return s + ")";
+    }
+
+    /**
      * \brief Whether this config would produce no actual wrapping.
      *
      * True when wrap_index = 0 (view_none) or when single_position is set but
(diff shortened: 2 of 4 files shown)


ciaranm:
Full sweep run, on this branch (so #909 + #911 together):

```
cmake -S . -B build -DGCS_TEST_CAP_DEFAULTS=OFF -DGCS_ENABLE_VIEW_WRAP_SWEEP=ON
ctest -j 24 -R _view_

100% tests passed, 0 tests failed out of 7314
Total Test time (real) = 1376.88 sec
```

**7314 of 7314**, 23 minutes wall on 32 cores (533 minutes of CPU; the
`linear_constraint_*_view_*` entries are 70-130 s each and dominate). Confirmed
against final source -- a no-op rebuild afterwards touched nothing.

So the sweep is now entirely green and could be made a required gate, at a cost
of ~23 minutes. Worth noting for whoever picks up #882: this measurement makes
both of the statements that issue owns false rather than half-false --
`dev_docs/view-proof-logging.md`'s Status line was right about the outcome for
the wrong reason (the three arithmetic sweeps were largely verifying empty
instances), and the CLAUDE.md note's "most fail today" was never right at all;
the true figure before these two PRs was 68 of 7314.

Issue #924, opened 2026-09-11
Element's result-union rule still sends views down a per-value path, and a view on the result trips the large-domain guard


ciaranm:
## Summary

`Element`'s result-union rule keeps an install-time type test that sends the
whole rule down a per-value path whenever any operand is a view:

```cpp
// gcs/constraints/element/element.cc:653
// Views keep the per-value path. The range justification below has
// to carry an order atom from result_var across the model's
// half-reified equality to the array entry, and a view's atoms are
// spelled through the view; that crossing has not been shown to
// bridge one. Same restriction, and the same reason, as
// min_max.cc's range path. (Among's range path needs no such
// restriction, because its lines stay within one variable.)
auto all_simple = holds_alternative<SimpleIntegerVariableID>(IntegerVariableID{result_var}) &&
    all_of(considered_vars, [](const IntegerVariableID & v) { return holds_alternative<SimpleIntegerVariableID>(v); });
```

**The reason it gives no longer exists.** `min_max.cc`'s range path has no such
restriction any more: #904 gave a registered view its own range literals over its
own encoded variable, linked to the underlying one, and retired the equivalent
guards in `equals`, `all_equal` and `min_max`. (#912 retires `In`'s.) So the
comment now cites a restriction that was deleted, and `element.cc` is the last
place carrying the shape.

This is not only stale prose. The fallback the guard selects is width-proportional,
so **a view on the result turns a Clean row into a tripping one**.

## What it costs

The large-domain audit's `Element` row, with the result wrapped and nothing else
changed:

```cpp
add("Element/view-result", Expect::Clean, [](Problem & p) {
    auto result = wide_var(p);
    p.post(Element{result + 1_i, p.create_integer_variable(0_i, 2_i), narrow(p, 3, 1_i, 3_i)});
});
```

```
Element                                  Clean            survives
Element/view-result                      Clean            trips            <-- MISMATCH
```

```
Large domain guard tripped: the number of unsupported result values one Element
propagation has walked is 100001, over the limit of 100000.
```

The bare row is Clean at `0..10^9` because the remainder goes in as the two
ranges it is. Wrap the result and the same instance walks it a value at a time.
The audit has no view-wrapped probe anywhere, which is why nothing has caught
this.

## The crossing works, measured

Forcing the guard true (and letting the inner per-entry lemma test admit a view
instead of only a bare `SimpleIntegerVariableID`, since a *constant* entry
genuinely has no order-encoding flags to resolve against):

[...]

PR #925, opened 2026-09-11
Element: pick the result-union rule's form by width, not by variable kind (#924)

ciaranm:
The rule removed the result values no array entry supports, and asked whether it
could state a removal as a range by testing whether every operand was a bare
SimpleIntegerVariableID. A view anywhere sent the whole rule down a per-value
walk of the remainder, which is width-proportional: the audit's `Element` row is
Clean at 0..10^9 because the remainder goes in as the two ranges it is, and the
same instance with the result wrapped walked 100001 values and tripped the guard.

The reason the test gave stopped being true with #904, which gave a registered
view its own range literals over its own encoded variable, linked to the
underlying one -- and that is the representation the model states this equality
in, so the two bound lemmas cross on whichever encoding each operand resolves to.
(#904 retired the same test in equals, all_equal and min_max; #912 retires In's.)

What replaces it is the question that was actually being asked: is this run wide
enough to be worth stating as a range? The range form costs two bound lemmas per
feasible index tuple and the per-value form costs one line per value per tuple,
so a one-value run is pure loss -- not_in_range canonicalises to the same !=
literal there anyway. `In` draws the line in the same place.

A constant entry keeps its exemption, and now says why: `result = c` under the
tuple's guard pins every bit of result, which decides both halves of the range
disjunction outright, so it needs no lemmas. That is the one case a view is not
like a bare variable, and the inner test now names it directly rather than
catching it by accident through `holds_alternative<SimpleIntegerVariableID>`.

The per-value loop over `still_to_find_support_for.each()` is gone with the type
test. What is left is reached only when `range_lo == range_hi`, so its
`LargeDomainIterationCounter` would have been counting intervals rather than
values, and a guard that describes work it is not doing reads as coverage.

## Measured

`large_domain_audit_test` gains `Element/view-result`: the GAC `Element` row with
the result wrapped and nothing else changed. It **trips on `main`** (100001
values walked, over the limit of 100000) and survives here. It is the first probe
in that file to wrap anything, which is why nothing caught this; the lane is 192
assertions green.

Proof-scaling survey: the new row is flat (36 OPB rows, 171 steps at both 10^3
and 10^4). `Element` moves 190 -> 184 steps and `Element/holey` 84 -> 78, which
is the two lemmas per tuple that the width test stops paying on one-value runs.

`element_test var --seed=1`, whole-run proof size:

| lane | main | here |
|---|---|---|
| bare | 52147 | **50833** |
| mixed views | 57412 | 58079 |

The bare lane is 2.5% smaller for the same reason the survey rows moved. The
view lane is 1.2% larger, and that is the honest cost: at these widths a
two-value run is three lines per tuple as a range against two per value, so the
crossover is a run of three or four. The point is not the 1.2% -- it is that the
per-value form has no upper bound and this one does.

## Checks

* `ctest` 829/829, caps on and with `-DGCS_TEST_CAP_DEFAULTS=OFF`.
* `element_test` in all five modes (`var`, `const`, `constgac`, `const2d`,
  `var2d`) x {bare, mixed, `--view-wrap=8 --view-position=all`,
  `--view-wrap=10 --view-position=all`, `--view-wrap=17 --view-position=1`}: no
  VeriPB rejection.
* Sabotage: dropping the two bound lemmas is rejected bare
  (`element_test_var_w0_pall.pbp:12999`) and under views
  (`element_test_var_mixed.pbp:530`, `element_test_var_w8_pall.pbp:530`).
* Forcing the range form onto one-value runs as well still verifies, which is
  what says the width test is a size choice and not a soundness one.

## Note for whoever merges this

`dev_docs/large-domains.md`'s probe counts move here and in #912, so whichever
lands second gets a one-line conflict in the table: this branch adds one row
(76 probes, Clean 42), #912 adds two (77 probes, Clean 43). Together they are 78
and 44. The `ArrayMinMax` paragraph that says "Views keep the per-value path" is
left alone here deliberately — it is about `min_max.cc`, whose guards #904 had
already retired, and #912 rewrites it. This branch only stops `element.cc` from
citing it.

### DIFF
--- a/dev_docs/large-domains.md
+++ b/dev_docs/large-domains.md
@@ -952,8 +952,8 @@ separately, because they mean different things:
 Re-measured over all 75 probes after the interval rewrites landed for
 `ArrayMinMax`, `Table`, `Among`, `Element`, `In`, `GlobalCardinality` and
 `AllEqual/holes`, and again after #878, #875, #877 and #874 — which moved nothing
-but their own rows: `Element/holey` is flat at 41 rows and 51 steps, `Abs/hole` at
-30 and 46, `Abs/hole-preimage` at 26 and 64, `GlobalCardinality/closed` at 24 and
+but their own rows: `Element/holey` is flat at 41 rows and 78 steps, `Abs/hole` at
+30 and 78, `Abs/hole-preimage` at 26 and 113, `GlobalCardinality/closed` at 24 and
 83, and none of the rewrites changes which values get removed, so no other row
 could have moved either. (Checked, for #877, by diffing a whole survey run against
 one from `main`: identical bar the new row.) The figures move, so re-run the
@@ -962,7 +962,7 @@ loop — that is how the previous version of it went stale, and how the
 `GlobalCardinality/hall` figures below came to be corrected.
 
 **#874's two rows are the case for running this survey and not just the audit
-lane.** With the propagation fixed, `In/vars` was flat at 50 steps but
+lane.** With the propagation fixed, `In/vars` was flat at 88 steps but
 `In/vars-single-support` read 7048 → **70048**: the rule's proof was still
 per-value even though its inferences were not, because the scaffolding that rules
 out the non-supporting sources' selectors emitted one line per value of
@@ -970,7 +970,7 @@ out the non-supporting sources' selectors emitted one line per value of
 The guard cannot see either — a reason is only materialised with proofs on, and
 the lane runs without them — so the survey was the only thing that showed it.
 The walk that fixes it is the same one the conclusions use, one interval at a
-time, and the row is now flat at 53.
+time, and the row is now flat at 93.
 
 | | growth (opb / steps) | constraints |
 |---|---|---|
--- a/dev_docs/large-domains.md
+++ b/dev_docs/large-domains.md
@@ -671,14 +671,14 @@ is the part worth reading carefully:
 
 ### Where we stand
 
-77 constraint probes, plus 20 heuristic ones in the second table. The lane
+78 constraint probes, plus 20 heuristic ones in the second table. The lane
 itself is the authority — run it rather than trusting this table, which is a
 snapshot for orientation.
 
 | | constraints |
 |---|---|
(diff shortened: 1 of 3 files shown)


ciaranm:
Rebased onto `main` (now `6ee02451`, after #912) and force-pushed: `0eb9f1e3` → `15c4bd63`.

The only conflict was the one your note predicted, in `dev_docs/large-domains.md`. Resolved to **78 probes / Clean 44**, exactly as you said, and checked against the source rather than the note — `all_probes()` has 78 `add(` rows, 19 KnownTrip + 44 Clean + 15 NoWidePosition = 78. Both prose edits are kept (#912's "`In` in three rows", this branch's "with a *view* on the result"). `element.cc` is byte-identical to `0eb9f1e3`; the rebased tree matches the trial merge I tested.

Verified on that tree:

* `-DGCS_WERROR=ON -DGCS_TEST_CAP_DEFAULTS=OFF`: **0 warnings, 0 errors**, 148 binaries.
* **ctest 829/829.**
* Guard lane: **196 assertions**, 98 rows, 2 × 98 = 196. `Element/view-result` Clean, survives.
* Proof-scaling survey: the new row is flat at **36 OPB / 171 steps**, `Element` 184, `Element/holey` 78 — every figure in the body reproduces. The `neither` bucket is **69 of 78**, measured rather than assumed; `Element/view-result` is 1.0x/1.0x, so it does belong there.

The earlier build failure recorded against this branch was collateral from a machine problem here, not a compile error — it builds clean.

### One extra commit, `15c4bd63`

Five **stale proof-step figures** in the same document, corrected at ciaranm's request while the branch was open. All five are step counts; every OPB row count beside them was already right.

| row | said | measured |
|---|---|---|
| `Element/holey` | 51 | **78** |
| `Abs/hole` | 46 | **78** |
| `Abs/hole-preimage` | 64 | **113** |
| `In/vars` | 50 | **88** |
| `In/vars-single-support` | 53 | **93** |

The last two are #912's: its body records that the figure moved 53 → 93 on rebase ("what moved that figure is `main`") and the body was updated where the document was not. The other three had gone stale earlier. Measured twice to confirm the survey is deterministic. The historical `7048 → 70048` beside them is deliberately left alone — that is the before-state the paragraph is contrasting, not a current reading.

Issue #877, opened 2026-09-07
GlobalCardinality in closed mode finds its range conclusions by walking the domain


ciaranm:
## Summary

`GlobalCardinality::with_closed(true)` installs an extra propagator that walks
every variable's domain grouping runs of non-cover values:

```cpp
// gcs/constraints/global_cardinality/global_cardinality.cc:137
for (const auto & var : vars) {
    vector<pair<Integer, Integer>> runs;
    for (auto v : state.each_value_immutable(var)) {
        if (binary_search(cover, v))
            continue;
        if (! runs.empty() && runs.back().second + 1_i == v)
            runs.back().second = v;
        else
            ...
```

The *conclusions* are ranges — that is what `runs` is for — but **finding them
walks the domain**, which is exactly the shape #859 removed from `In`: the runs a
merge against the cover yields are the same runs, without the walk.

No audit row calls `with_closed`, in either consistency arm, so this is unprobed.

## Measurements

`GlobalCardinality{2 wide vars, cover {1}, count [0,2]}.with_closed(true)`, root
propagation only.

Guard build, `w = 10^9`: **`TRIPS(guard)`**, `each_value_*()` at 100001.

Guard off, linear in the width:

| `w` | wall time |
|---|---|
| 10^6 | 34 ms |
| 10^7 | 300 ms |
| 10^8 | 3006 ms |

so roughly 30 s at 10^9.

## What to do

1. An audit row for closed mode. On today's code it is a `KnownTrip`.
2. Replace the walk with `state.copy_of_values(var).each_interval_minus(cover_set)`,
   where `cover_set` is the sorted cover as an `IntervalSet` built once in
   `prepare()` — the same construction `in.cc` already does with `val_vals_set`.
   The conclusions are unchanged, so the proof should be unchanged; worth
   confirming with `GCS_PRESERVE_PROOF_FILES=all`.

Related: #833, #859 (the identical rewrite in `In`), #864.

---
Found by the per-value iteration sweep of #833 — all 216 sites in 53 files read
and classified, notes and probe harness in `tmp/sweep864/`. Measured by an AI
agent (Claude) on an AMD EPYC 7643 (192 cores, **2 TB**), one core via
`taskset -c 4`, otherwise idle, Release build of `main` at `2fb5701d` (plus
PR #873, which touches only `equals`). The memory size is load-bearing for any
"SURVIVES" above: the lane's fallback detector for an uninstrumented site is
[...]

PR #926, opened 2026-09-11
Count the audit lane's NoWidePosition rows as 15 too (#833)

ciaranm:
One character. `be55b02e` corrected the audit lane's headline counts for the
`Dag` row that #918 added after the #877 branch was cut, but not the breakdown
table immediately under them, which still added up to the old 74:

> `KnownTrip (19)` + `Clean (41)` + `NoWidePosition (14)` = **74**, under a line
> saying 75.

`Dag` is an `Expect::NoWidePosition`, so that is the count that moves.

Both numbers re-derived from a run rather than read off the page, which is what
the table's own preamble asks for ("the lane itself is the authority — run it
rather than trusting this table"). Off this commit:

* the lane is green (190 assertions) and prints **75** constraint rows, **41
  Clean / 19 KnownTrip / 15 NoWidePosition**, with `Dag` as
  `NoWidePosition survives`;
* the proof scaling survey is **75** rows and **63** strictly flat, which with
  `Multiply`, `Divide` and `Modulus` counted as the logarithmic cases the row's
  own footnote describes gives the **66 of 75** already recorded. So that half of
  `be55b02e` was right and only this one line was left behind. `Dag` is flat
  there too: 47 OPB rows and 4 proof steps at both widths.

https://claude.ai/code/session_012iUaXT5G9VBE3cJeVXvtzx


ciaranm:
Superseded by #912, which landed at 13:07 and made this exact change itself: `main` already reads `**NoWidePosition** (15)` at dev_docs/large-domains.md:682, so this branch's entire diff is already on main. (It showed as CONFLICTING only because #912 also rewrote the adjacent **Clean** row, from 41 to 43.)

Head branch `doc-audit-probe-count` left in place.

Issue #895, opened 2026-09-07
Refined watches: hoist the empty-index test, and index equality watches by value


ciaranm:
## Summary

Two separate costs in the refined-watch wake path, both measured while doing
#889, both left alone there because that PR is a correctness fix and neither
belongs in it.

## 1. The watch-index test on the inference-frequency path

#889 made `requeue_unless_already_seen` — the round-boundary replay taken
whenever some propagator in the round claimed `EnableButIdempotent` — fire
refined watches, which it had to, because a watch is a one-shot subscription and
each inference is replayed exactly once. That put

```cpp
if (v.index < _imp->refined_watches_by_var.size())
```

on a path that runs once per inference and did not have it. Instruction counts
(`perf stat`, deterministic to six figures) for that commit alone, over the
`benchmarking.md` set:

| | Δ instructions |
|---|---|
| `tsp` | +1.71% |
| `magic_square --size=5` | +1.36% |
| `ortho_latin --size=6 --all` | +0.69% |
| `qap --size=12`, `langford --size=11`, `n_queens --size=14 --all` | within ±0.15% |

Every one of those models has an **empty** watch index, so the test is pure
overhead for them, and `requeue` was already paying it before #889.

The fix is to hoist "are there any watches at all" out of the replay loop and
make it a second template parameter alongside `HonourClaims_`, so the whole
firing block compiles away for a model that arms none. It is safe *as written
today* — only a propagator arms a watch, and no propagator runs during the
replay, so `refined_watches_by_var.size()` cannot grow inside the loop — but that
is a temporal invariant about the hottest loop in the solver, and the thing it
would be protecting against getting wrong is exactly the silent wake loss #889
was. Worth doing deliberately, with the invariant written down next to it and a
test that fails if a watch is armed mid-replay.

## 2. The scan is still O(watches on the variable)

The firing loop walks every watch armed on `v` and calls `test_literal` on each
whose `trigger_mask` matches. So for a model that arms one watch per (variable,
value) — which is what #889 gives `ReifiedEquals` against a constant, and what
FlatZinc's `int_eq_reif(x, c, b)` produces per value — the scan per interior
removal is as long as the coarse trigger list it replaced. Only the per-item body
got cheaper: one `test_literal` instead of a full propagator call through the
reified dispatcher.

That is where nmseq's remaining time is. #889 took `nmseq/100` from 4.455 s to
2.097 s against Gecode's 0.808 s on the same tree; the propagator calls it
[...]

Issue #889, opened 2026-09-07
The reified equals arm wakes 23 times per domain change it makes: 6.3x Gecode on nmseq


ciaranm:
## Summary

On an **identical search tree**, gcs takes **6.3x Gecode's time** on the upstream
`nmseq` model (magic sequence, naive), and `GCS_PROPAGATOR_STATS` puts 82% of
propagation time in `Equals`. The cause is wake count, not per-call cost.

Measured with the #868 harness (`gcs-benchmarks`, `bin/tree_compare.py`), gcs
`238c7836` against Gecode 6.3.0 via the MiniZinc 2.9.7 bundle, Ryzen 9 9950X3D:

| instance | nodes (both) | gecode | gcs | ratio |
|---|---|---|---|---|
| `nmseq/100` | 767 | 0.812 s | 5.03 s | **6.20x** |
| `nmseq/200` | 1,567 | 12.53 s | 80.18 s | **6.40x** |

The model is `x == sum(i)(bool2int(xs[i] == v))` per element, so `n` reified
equalities plus one bool-sum per constraint, and it pins its own
`int_search(s, input_order, indomain_min, complete)` — so this is the same tree,
not a search difference.

## Where the time goes

`GCS_PROPAGATOR_STATS` on `nmseq/100`:

| type | constraints | calls | effectful | share of measured propagation |
|---|---|---|---|---|
| `Equals` | 20,000 | 100,290,249 | 4,378,495 (4.4%) | 5.19 s (**82%**) |
| `LinEquals` | 100 | 1,448,064 | 1,403,690 (97%) | 1.14 s (18%) |

So it is the equals family and not the linear sum. And the shape is specific:

- **130,756 `Equals` calls per node** — with 20,000 propagators installed, that
  is ~6.5 full sweeps of every propagator at every node.
- **4.4% effectful**: 22.9 wakes per domain change actually made.
- **52 ns per call**, which is *fine* — the family's unreified propagator costs
  48 ns/call on `queens`, where it is woken 20.8 times per node at 27%
  effectful.

Per-call cost is not the problem. The propagators are woken about 23 times for
every change they make, and each wake is a call that finds nothing to do.

## Why this is the same lever as #819 and #807

Both of those were trigger-granularity findings on this family: #819 moved an
unconditional `NotEquals` to `on_instantiated` because `on_change` woke every
other disequality on a neighbour for an interior removal that could not matter,
[...]

PR #934, opened 2026-09-11
Ask whether a literal is entailed, not what it is (#895)

ciaranm:
`State::test_literal` answers three ways, so a caller that only wants to know
whether a literal *holds* pays for the part of the answer it discards. For
`x != v` entailment is settled by the `in_domain` call, and `test_literal` then
calls `has_single_value` purely to separate `DefinitelyFalse` from `Undecided`;
each bounds operator reads both bounds where entailment needs one.

`State::literal_is_entailed` is the same question with that tail not computed.
**Eleven call sites were already asking it** and throwing the rest away: the
refined-watch firing loop, the two-watched-literal nogood store (5), `NegativeTable`
(3), the difference graph (2), and the `infer_all` short-circuit.

## Where it pays

The refined-watch firing loop. On `nmseq/100` — FlatZinc's `int_eq_reif` against a
constant, the model #889 and #895 are about — the loop asks the question **99.0 M
times over 5.81 M inferences, 17 per inference, and 97.8 % of those are watches
that have not fired**.

Instructions, `perf stat`, medians of two, pinned, boost off. `recursions` and
`propagations` identical in every row:

| | |
|---|---|
| `nmseq/100` | **−5.87 %** |
| `magic_series --size=300` | **−2.27 %** |
| `tsp` | +0.015 % |
| `magic_square --size=5` | +0.009 % |
| `ortho_latin --size=6 --all` | +0.023 % |
| `qap --size=12` | 0.000 % |
| `langford --size=11` | −0.008 % |
| `n_queens --size=14 --all` | +0.051 % |

The rest of the set does not move — nothing else asks often enough — and those six
rows sit well inside the ±0.15 % band `propagate()` moves by for a change that
executes nothing at all (the null-change control documented in #932).

**About as much again is still on the table**, at the firing site rather than in
this function: a prototype that inlined the check into the loop, instead of calling
out of line and visiting the `Literal` variant on the way, measured **−10.6 %** on
the same instance. Getting that means the watch carrying an
`IntegerVariableCondition` rather than a `Literal`, which belongs with a triggers
refactor, not here.

For context on what is left after both: one full scan of the watch list costs 27 %
of `nmseq/100`'s instructions (measured by running the scan twice and taking the
delta, which leaves the search untouched), and 97.8 % of it is wasted. Indexing
equality watches by value — part 2 of #895 — is what removes the rest.

## Testing

The property is the *identity* with `test_literal`, not hand-written expectations
per arm, which would let both functions be wrong in the same way. `state_test`
gains a differential over every operator, on plain variables, views, negated views
and constants, against domains that are full, holed, bound-narrowed, instantiated,
and holed-then-narrowed — **5,179 assertions**.

**Mutation tested one arm at a time** (`Equal` → `in_domain`, `NotEqual` → wrong
value, `Less`/`GreaterEqual` → the other bound, `InRange` → half the test,
`NotInRange` → `in_domain`): each is caught by between 81 and 251 assertions.

- 829/829 on `release` and `sanitize`, with `minizinc` and `cake_pb_cp` on `PATH`.
- `langford --size=11 --restarts=100` explores the same 258,670 recursions and
  learns the same 2,746 nogoods on the refined and the scan path, before and after
  — the check that matters for the nogood store's five sites.

Based on `main`, not on #932, so the two are independent; they touch neighbouring
lines in `propagator-performance.md` and `refined-triggers.md` and may want a
trivial merge in whichever lands second.

https://claude.ai/code/session_01LwP4aznqMKhCH8PLBmCdNi

### DIFF
--- a/dev_docs/propagator-performance.md
+++ b/dev_docs/propagator-performance.md
@@ -282,6 +282,36 @@ Finally: guard every reason in a file, not just the hot ones. `comparison` was
 found by a family audit rather than by a profile, and "some of them are guarded"
 is exactly what made the file look finished.
 
+### Ask the cheapest question that settles it
+
+`State::test_literal` answers three ways — `DefinitelyTrue`, `DefinitelyFalse`,
+`Undecided` — so a caller that only wants to know whether a literal *holds* pays
+for the part of the answer it then throws away. For `x != v` entailment is settled
+by one `in_domain`, and `test_literal` goes on to call `has_single_value` purely to
+separate the two negative answers; each bounds operator reads both bounds where
+entailment needs one. `State::literal_is_entailed` is the same question with that
+tail not computed, and it is what the refined-watch firing loop, the
+two-watched-literal nogood store, `NegativeTable` and the difference graph were all
+actually asking.
+
+It is worth **5.9% of `nmseq/100`'s instructions and 2.3% of `magic_series
+--size=300`'s**, because the refined-watch scan asks it about 17 times per
+inference there and 98% of those are watches that have not fired (issue #895).
+Nothing else in benchmarking.md moves, because nothing else asks often enough.
+
+About as much again sits at the call site rather than in this function: the firing
+loop reaches `literal_is_entailed` through an out-of-line call and a `Literal`
+variant visit, and a prototype that inlined the whole test into the loop measured
+10.6% on the same instance. Collecting that means the watch carrying an
+`IntegerVariableCondition` rather than a `Literal`, which belongs with a triggers
+refactor. The lesson for the figure above: a narrower entry point recovers the
+work the wide one did, not the cost of reaching it.
+
+The general shape, which is worth looking for elsewhere: when a general-purpose
+query returns more than a hot caller uses, the surplus is not free. The fix is a
+narrower entry point *beside* the wide one, not a cheaper implementation of it —
+the callers who want all three answers still need all three.
+
 ### Don't throw from a propagator that fails in bulk
 
 `inference.contradiction(...)` signals failure by throwing
--- a/dev_docs/refined-triggers.md
+++ b/dev_docs/refined-triggers.md
@@ -103,9 +103,12 @@ with the `Inference` granularity `inf` (`BoundsChanged` / `InteriorValuesChanged
 
 1. If `inf` is **not** in the watch's `trigger_mask`, skip it (see *Trigger
    masks*).
-2. Otherwise test `state.test_literal(watch.literal)`. If `DefinitelyTrue`,
-   **fire**: append the payload to the owner's inbox, wake the owner, trail the
-   removal, and swap-remove the watch (consume).
+2. Otherwise ask `state.literal_is_entailed(watch.literal)`. If it is, **fire**:
+   append the payload to the owner's inbox, wake the owner, trail the removal, and
+   swap-remove the watch (consume). (`literal_is_entailed` rather than
+   `test_literal`, because the loop does not care whether an unfired watch is false
+   or merely undecided, and finding that out is most of the cost — see
+   propagator-performance.md.)
 
 The owner processes its fired set the next time it runs in the propagation queue
 — firing and processing are *decoupled* (the owner is the schedulable unit). This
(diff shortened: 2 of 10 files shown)


ciaranm:
Pre-merge review. Holding this one on a documentation figure; the code itself checks out and is verified below.

## The blocker: the doc overstates the change by ~1.8x

`dev_docs/propagator-performance.md:297-299` says:

> It is worth **10.6% of `nmseq/100`'s instructions and 4.0% of `magic_series --size=300`'s**, because the refined-watch scan asks it about 17 times per inference there and 98% of those are watches that have not fired (issue #895).

The PR body's own measurement table gives this change as **−5.87%** on `nmseq/100` and **−2.27%** on `magic_series --size=300`.

10.6% is the body's *prototype* figure — the version that also inlined the check into the firing loop, which this PR deliberately does not do:

> **About as much again is still on the table**, at the firing site rather than in this function: a prototype that inlined the check into the loop […] measured **−10.6%** on the same instance. Getting that means the watch carrying an `IntegerVariableCondition` rather than a `Literal`, which belongs with a triggers refactor, not here.

So the sentence attributes to the landed change a number the body explicitly says belongs to work that is not in it. `4.0%` appears nowhere in the body at all; the measured `magic_series` row is 2.27%.

The body is right and the document is wrong, so this is a two-number edit — but it is a reference document other people will act on, which is why it is worth fixing before it lands rather than after.

## Smaller, entirely optional

Four places still describe the firing loop as calling `test_literal`, which the same commit changes. The statements stay true as descriptions of the *semantics* (`literal_is_entailed` is defined as `test_literal(…) == DefinitelyTrue`), so this is a readability nit rather than an error — but `dev_docs/refined-triggers.md:155` in particular is the same sentence as the `:109` one you did update:

- `dev_docs/refined-triggers.md:155` — "The firing loop tests `test_literal` only when `inf` is in the mask"
- `gcs/innards/propagators.cc:215` — "the firing loop skips the test_literal for it"
- `gcs/innards/propagators.cc:266` — "the firing loop can skip the (expensive) test_literal" — the "expensive" is the part this PR makes less true
- `gcs/innards/propagators.cc:825` — "This gates the expensive test_literal"

## What was verified, so it does not get re-raised

Checked on a local trial merge of this branch together with #932 and #933 onto `main` (`e6546277`), since #932 touches the same three files:

- **The arms are equivalent to `test_literal`'s `DefinitelyTrue` arms**, read one by one. `Less` and `GreaterEqual` drop the redundant half of the bounds pair by monotonicity (`upper < v` implies `lower < v`); `Equal` via `optional_single_value(var) == value` implies the `in_domain` test; `NotEqual`, `InRange`, `NotInRange` are unchanged predicates. `state_test` reports **5,179 assertions**, as the body says.
- **Merge with #932 is clean and semantically right**, which was the thing worth checking: #932 adds a second wake body (`requeue_coarse_only`) with *no* firing block, and this change lands inside `requeue_honouring`'s firing block, so there is no second copy of the loop left calling `test_literal`.
- Release build with `-DGCS_WERROR=ON`: clean, **0 GCC warnings**, 460 TUs. clang++-21: **36 occurrences / 3 sites**, all the known `array_has_nonconstants` baseline in `element.cc`.
- `ctest` **829/829** release and **829/829** sanitize, with `minizinc` and `veripb` on `PATH` (103/103 minizinc lanes ran, 0 skipped). Sanitize emits the known 673 GCC `<generator>` false positives and nothing else.

Happy to merge as soon as those two figures are corrected.

ciaranm:
Both fixed, amended onto `71406f56` (was `9a83c063`).

**The blocker.** You are right, and about the mechanism too: I re-measured on a
`main` base after writing that paragraph, updated the commit message and the PR
body with the −5.87% / −2.27% figures, and left the doc on the prototype's. So the
document was asserting for the landed change exactly the number the body says
belongs to work that is not in it.

`propagator-performance.md` now reads 5.9% and 2.3%, and the prototype's 10.6% has
moved into its own paragraph that says what it is — the call site, not this
function — so the two numbers can no longer be read as the same claim:

> About as much again sits at the call site rather than in this function: the
> firing loop reaches `literal_is_entailed` through an out-of-line call and a
> `Literal` variant visit, and a prototype that inlined the whole test into the
> loop measured 10.6% on the same instance. […] The lesson for the figure above: a
> narrower entry point recovers the work the wide one did, not the cost of
> reaching it.

**The nits, all four taken** — `refined-triggers.md:155` and the three
`propagators.cc` comments now say "the entailment test" rather than
"the (expensive) `test_literal`". You were right that the "expensive" was the part
this change makes less true, and `:155` being the twin of the `:109` sentence I
did update is exactly the kind of half-done edit worth catching.

**What moved since your verification**, so you can re-check cheaply rather than
re-running it: 3 files, +14/−6, and no executable change at all — the
`propagators.cc` delta is four comment lines. `literal_is_entailed`, its tests and
every converted call site are byte-identical to what you reviewed. I rebuilt and
re-ran `ctest` anyway: 829/829 release.

Thank you for the trial merge against #932 — the point that #932's
`requeue_coarse_only` has no firing block, so there is no second copy of the loop
left calling `test_literal`, is the thing I would have wanted checked and had not
checked myself.

Issue #649, opened 2026-08-03
Unification: a constraint that collapses to `x = ±y + c` cannot say so, and keeps propagating for the rest of the subtree


ciaranm:
Several constraints degenerate into an equality between two variables once
enough of their scope is fixed. `Element` with its index fixed says
`result = vars[i]`. `Plus(a,b,r)` with `a` fixed says `r = b + k`. A linear
equality down to two live terms with coefficients `±1` says `x = ∓y + c`.
`Equals` says it from the moment it is posted.

None of them can say it. The solver has no way to represent "these two are now
the same variable", so the constraint stays in the propagation queue and keeps
pushing bounds back and forth between two variables that are, from that point
on, one variable wearing two hats.

## How much this costs

Measured on 245 MiniZinc Challenge instances, flattened through our own `.msc`
and solved by `fzn-glasgow` with a 20s limit, with a read-only probe at both
linear-equality propagation entry points. Full write-up and scripts in
`~/claude/tmp/xor-scoping/census/`.

```
47,506 linear-equality constraints, 1,413,967,698 propagator invocations

2-term, all ±1 (unconditional  x = ∓y + c)
  constraints                26,623      in 148/245 instances
  invocations           171,363,127      12.1% of all linear-eq propagation

3+-term, all ±1 (Plus/Minus-shaped, conditional)
  constraints                14,208      in 210/245 instances
  ever collapse to 2 live     9,551      67.2%
  invocations         1,082,413,914      76.6% of all
    after the collapse   447,817,366     41.4% of bucket, 31.7% of all
  first-collapse depth:  7.2% at root, median 15, mean 68, max 970

removable in total      619,180,493      43.8%   (median 43.2% per instance)
```

Two thirds of the Plus/Minus-shaped constraints really do collapse. Once one
does, 41% of that constraint's remaining propagation happens after the point
where it could have been subsumed.

**Only 7.2% collapse at the root; the median is depth 15.** This is the number
that decides the design, so it is worth being explicit: a presolve-only
implementation — do it all before search, which is what Huub does — captures
about one collapse in fourteen. The value is genuinely search-time.

Read the percentages as approximate. The run list was sorted heaviest-first and
the headline moved a long way as it progressed (77% at n=7, 64% at n=43, 49% at
n=132, 44% at n=245). The *shape* conclusions are stable from n=43 on; the
absolute figures are not.

### What this measurement does not say

It is a share of **linear-equality propagation**, not of total solver work. The
[...]

PR #938, opened 2026-09-11
Certified GF(2) reasoning over a conjunction of XORs (#647)

### LINKED ISSUE #647: Parity: `ParityOdd` reasons one XOR at a time; the system over GF(2) is tractable, GAC, and certifiable
`ParityOdd` reasons about one XOR at a time. The propagator in
`gcs/constraints/parity/parity.cc:108` walks the literals, and bails out the
moment it sees a second undecided one:

```cpp
case Undecided:
    // two or more undecided literals? can't do anything
    if (++how_many_unknown > 1)
        return PropagatorState::Enable;
```

That is the right answer for a single parity constraint — GAC on one XOR *is*
unit propagation, because any value has a support as long as two variables are
still free. All of the available inference lives in the conjunction, and we
currently throw it away.

## Why the system is worth having

Over GF(2) the conjunction of parity constraints is one of the rare tractable
ones. Row-reduce the system:

- `x_i` is implied iff the reduced row echelon form contains the unit row `e_i`;
- `x_i = x_j` (or `x_i = ¬x_j`) is implied iff `e_i + e_j` is in the row space.

So system-level GAC reduces to detecting implied literals, and Gauss-Jordan
gives all of them. The SAT literature calls the levels here plain unit
propagation, equivalence reasoning, and complete parity reasoning (Laitinen,
Junttila and Niemelä's DPLL(XOR) work); we are at the bottom of that ladder.

The tractability is specific to GF(2), and it is specific to the *conjunction*.
It does not survive going to `mod p` — see #648.

## Where everyone else is

| Solver | What exists | Strength |
|---|---|---|
| **gcs** | `ParityOdd`, `gcs/constraints/parity/parity.cc:108` | one constraint, UP |
| ACE | `constraints/global/Xor.java`, two watched sentinels, `TagAC` | one constraint, UP |
| Gecode | `rel(home, BOT_XOR, BoolVarArgs, y)` | one constraint, UP |
| Chuffed | decomposes `array_bool_xor` into a binary accumulator chain, `chuffed/redefinitions.mzn:69` | one constraint, and the row is gone |
| CPMpy | `Xor` decomposes to a chain of sums; maps to native xor where the backend has one | one constraint |
| Huub | XOR is a model-level `BoolFormula`; its `propagate` is `unreachable!()` | encoded to CNF |
| OR-Tools CP-SAT | `AddBoolXOr` → `LiteralXorIs`, plus `PresolveBoolXor` | one constraint |

Nobody in mainline CP does the system. CP-SAT knows it, and says so in
`ortools/sat/cp_model_presolve.cc:258`:

> `// TODO(user): More generally, if a variable appear in only bool xor`
> `// constraints, we can simply eliminate it using linear algebra on Z/2Z.`
> `// This should solve in polynomial time the parity-learning*.fzn problems`
> `// for instance. This seems low priority, but it is also easy to do. Even`
> `// better would be to have a dedicated propagator with all bool_xor`
> `// constraints that do the necessary linear algebra.`

The two research artifacts that do exist are `abstractXOR` (Rouquette and
Solnon, CP 2020, a Choco 4 prototype that never landed in mainline Choco) and
MiniCPBP's linear modular constraint. `abstractXOR` is not plain GF(2) — its
variables are "is this byte difference nonzero" indicators over an underlying
system, satisfaction is NP-complete for a finite difference bound and only
becomes polynomial when the bound is dropped — but its section 4 is a directly
usable algorithm: maintain the matrix in RRE form, check that every row has at
least two non-zero cells for feasibility, and use its Theorem 4 for the implied
equalities. Complexities are `O(n·n₁·m₁)` to reach RRE, `O(n₁m₁)` to propagate
an assignment to 0, and `O(m·n₁m₁)` for an assignment to 1, with `n₁`/`m₁` the
maximum non-zeros in a row/column.

The SAT side is far ahead: CryptoMiniSat runs Gauss-Jordan during search, WDSat
does XOR reasoning at the ANF level, and Potassco's `xorro` does it for ASP.

## The modelling route already exists

The XOR structure survives MiniZinc flattening intact. Flattening
`constraint xorall(b);` gives back exactly `constraint array_bool_xor(b);`, and
we already map that to `ParityOdd` in `minizinc/fzn_glasgow.cc:614` (with
`bool_xor` at `:651`), alongside XCSP3 at
`xcsp/xcsp_glasgow_constraint_solver.cc:1518` and `:1640`, CPMpy via
`python/gcspy.cc:761`, and `.scp` at `gcs/scp_reader.cc:918`.

So the plumbing is done end to end and only the strength is missing. Gathering
the individual constraints into a system is a presolver job, and
`gcs/presolvers/auto_table.{hh,cc}` is the template — the same weak-propagator
plus opt-in-presolver shape we use for tables.

## Proof logging: already solved, but not in our current encoding

Gocht and Nordström, *Certifying Parity Reasoning Efficiently Using
Pseudo-Boolean Proofs* (AAAI 2022, arXiv:2209.12185), encode
`x₁ ⊕ … ⊕ xₖ = b` as

```
sum_{i in [k]} x_i = b + 2 * sum_{i in [floor(k/2)]} y_i
```

with fresh, otherwise unconstrained `y` variables. Under that encoding **a
Gaussian elimination step is adding the two pseudo-Boolean equalities: one
`pol` line.** Reason and conflict clauses come out by literal axioms. Same
shape as the difference-logic proofs.

Two caveats:

1. **This conflicts with the encoding we have.** `ParityOdd::define_proof_model`
   (`gcs/constraints/parity/parity.cc:80`) uses cake_pb_cp's accumulator chain —
   `x[id][k] = x[id][k-1] XOR l_k`, four labelled clauses per literal, then an
   `acc` row pinning the final accumulator. That is the right definitional
   encoding for cake faithfulness, and it does *not* make elimination one `pol`
   line. Section 4.4 of the paper derives the slack form from a clausal encoding
   using extension variables and redundance-based strengthening, so we can keep
   the definitional rows and derive the slack form once per constraint — but
   that is real work and should be costed separately, not assumed free.
2. **Cutting planes alone is not enough.** Tseitin (mod 2) formulas are the
   classic hard case; Dadush and Tiwari's quasi-polynomial cutting-planes proofs
   are still superpolynomial. The extension variables are load-bearing, not a
   convenience.

## What would exercise it

- MiniZinc Challenge 2012 `parity-learning` — minimal disagreement parity, i.e.
  learning parity with noise. This is the family the CP-SAT TODO names. It is
  the *optimisation* variant, so Gauss-Jordan alone will not crack it; the noise
  is the hard part. Still the obvious first benchmark.
- MiniZinc Challenge 2016 `cryptanalysis` (Gérault, Minier and Solnon's AES
  related-key Step 1 model). Note that the 2017/2018/2021 `opt-cryptanalysis`
  model is DDT-`table` plus a bit permutation with no XOR system in it, so the
  honest challenge count is two families, both old.
- Hashing-based model counting and sampling, which is where CryptoMiniSat's
  Gauss-Jordan bought ApproxMC two to three orders of magnitude.
- LDPC decoding as XOR-SAT; low-weight codeword and syndrome decoding.
- Index calculus on binary elliptic curves (Trimoska, Ionica and Dequen,
  CP 2020).

## Open questions before committing

- **Where do implied equivalences go?** The real payoff over unit propagation is
  the implied `x = y` and `x = ¬y`. Implied *literals* prune domains for free,
  but we have no equivalence store, so those inferences currently have nowhere
  to live. `abstractXOR` gets value from them because Δ-equality is semantically
  meaningful in its model. This is the main design risk and it is an
  architecture question, not an algorithms one. To be split into its own issue.
- **Backtracking.** `abstractXOR` reports that sparse sets do not trail cleanly
  here, because XORing two rows *creates* new non-zero cells; they undo
  operations explicitly instead. Our trailing model needs a plan for that.
- **What do we actually beat?** On pure XOR systems the competitor is
  CryptoMiniSat, which is very good. The differentiator here is the certificate,
  not the speed — worth being upfront about when framing it. "First CP solver
  with certified system-level parity reasoning" is the claim; "fastest" is not.

## References

- Gocht and Nordström, *Certifying Parity Reasoning Efficiently Using
  Pseudo-Boolean Proofs*, AAAI 2022. https://arxiv.org/abs/2209.12185
- Rouquette and Solnon, *abstractXOR: A global constraint dedicated to
  differential cryptanalysis*, CP 2020. https://hal.science/hal-02899338v1/document
- Laitinen, Junttila and Niemelä, *Extending Clause Learning SAT Solvers with
  Complete Parity Reasoning*. https://arxiv.org/abs/1207.0988
- Trimoska, Ionica and Dequen, *Parity (XOR) Reasoning for the Index Calculus
  Attack*, CP 2020. https://arxiv.org/abs/2001.11229
- Fleming, Göös et al., *On the Power and Limitations of Branch and Cut*,
  CCC 2021 (for the Dadush–Tiwari Tseitin result).
  https://drops.dagstuhl.de/entities/document/10.4230/LIPIcs.CCC.2021.6

ciaranm:
Closes #647.

`ParityOdd` reasons about one XOR at a time, and that is the right answer for one
XOR — unit propagation on a single parity constraint *is* GAC, because any value
has a support while two literals are still free. So every bit of inference a
system of them offers lives in the conjunction, and we were throwing all of it
away. Over GF(2) that conjunction is one of the rare tractable cases.

Two things here. **`ParitySystem`** eliminates over a set of rows and reads off
every literal the system implies, which is GAC. **`ParitySystemGathering`** is a
presolver that collects posted `ParityOdd` and Boolean `Equals`/`NotEquals`
constraints into per-component systems, so a model arriving through a frontend
gets this without being re-modelled — XOR structure survives MiniZinc flattening
as individual `array_bool_xor` constraints.

Nobody in mainline CP does the system; CP-SAT has a TODO saying it should. The
claim here is **the certificate, not the speed** — CryptoMiniSat is very good at
this and is not being competed with.

## The proof

Gocht and Nordström (arXiv:2209.12185). Their slack form `Σx = b + 2B` makes a
Gaussian elimination step one `pol` line, but it cannot go in the `.opb`: cake
re-derives its own OPB from the `.scp`, and a presolver has no `ProofModel` at
all. So it is derived in-proof at `Top` from rows each donor already emitted —
which is the chain-portable option anyway.

That derivation is cheaper than the paper's. Their §4.4 recovers the form from a
CNF encoding they did not write, by brute force over all assignments; our
accumulator chain already *is* the partial-parity split of their (4.15), so each
step is a 3-XOR needing two `red`s — one discharged by RUP against the step's own
`1_1` clause, one carrying a five-line `pol` subproof whose three divide-by-2s are
where the parity argument lives — and the steps then telescope. Linear, not
exponential.

§4.3's clause extraction folds into the same RPN expression as the elimination, so
**an inference is one `pol` plus one RUP** whatever the size of the row
combination.

A two-literal donor is cheaper still: `B` is zero, so the slack form is just
`l1 + l2 = 1` and both halves are plain RUP against rows the donor emitted —
negating either fixes both operands and walks into one of them.

## What is verified

- **Zero assertions.** 61 veripb 3.0.2 runs across the two test binaries, all
  `s VERIFIED`, and `grep -c '^a '` zero on a preserved `.pbp`. With and without
  the view wrapper.
- **GAC is checked, not asserted** — `solve_for_tests_checking_gac`, and
  selecting `ParitySystemPropagation::CheckOnly` makes it fail, which is what
  says the check has teeth.
- **Mutation testing: 5 of 7 rejected.** The two survivors only *weaken* the
  derived row, by a term the `.opb` pins anyway, so the wrapping RUP replaces it
  — `veripb-facts.md`'s "a `pol` only has to get close", not a defect. The five
  that corrupt the *claim* are all caught. Written up as a table so nobody reads
  5/7 as a gap.
- **The presolver's tripwire holds node-for-node.** Retiring the gathered donors'
  propagators leaves solutions *and* recursions identical on every fixture while
  propagations drop by a third to a half, which is what says the system subsumes
  a single XOR's unit propagation.
- **It buys something.** On a system unsatisfiable while no single row conflicts,
  gathering takes the refutation from 3 recursions to 1. Two such fixtures, one
  unsatisfiable only once an equality has identified two XOR rows' variables — so
  it separates "the Boolean family is counted" from "the Boolean family is doing
  work".

## Two API extensions

The presolver API is very much "what has been needed so far", and this needed two
things it did not have.

- `ConstraintProofModelData<ParityOdd>` publishes a naming **object**,
  `ParityChainNaming`, rather than loose role functions. `Cumulative` publishes
  loose ones and that is right for it; here every name in the family is indexed
  the same way, by the same optional row index, and bundling them stops a citer
  picking up one without the rest or mixing an unprefixed role with a prefixed
  flag. `define_parity_chain` and `find_parity_chain` both build every name
  through it.
- **`ProofFlagKey` gained a `family`, and `find_proof_flag_values` became
  `find_proof_flag`.** This is a latent defect fixed rather than a feature added:
  a key carried numbers and an annotation but not which of cake's flag families
  it meant, so `v[id][1]` and `x[id][1]` — genuinely different flags — had the
  same key. The lookup hard-coded `v`, which worked only while nothing needed the
  other one. `ParityOdd`'s accumulators are cake's `x[id][k]` and cannot be
  anything else. The creation side already indexed both families, so only the
  lookup was narrow; `Values` is the default, so every existing published key is
  unchanged.

## What is NOT done, and should be read before merging

- **No sanitize build, and not the full test suite.** 372 related `release` tests
  pass (parity, scp chain, constraint enumeration, equals, cumulative,
  disjunctive, difference, presolvers, propagators, abs — everything the rename
  touched), and the whole tree builds. But `CONTRIBUTING.md` asks for `release`
  *and* `sanitize`, both full, and for `-DGCS_TEST_CAP_DEFAULTS=OFF` when a
  propagator has changed. None of those three has been run.
- **No benchmark numbers at all.** Everything measured is three- and
  four-variable fixtures. The `Top` footprint, the per-inference line count and
  the wall clock are all claims this design makes and nothing has checked at
  scale.
- **The `Top` footprint is still `7n` lines per row, not 2.** The per-step `red`s
  and their subproof lines are only needed until the telescoping `pol`s have run
  and are meant to be deleted afterwards; that is not implemented. Worth doing
  before pointing this at anything large, since every line left at `Top` taxes
  every later unhinted RUP (#666).
- **No frontend exposure**, so `parity-learning` and `cryptanalysis` cannot be
  run yet. That is the next thing, and it is also what would turn the numbers
  above into real ones.
- Implied equivalences (`x = y`, `x = ¬y`) are the real payoff over unit
  propagation and have nowhere to live — #649 scopes an equivalence store.
  Deliberately out of scope; detecting implied literals alone is already GAC.

`dev_docs/parity-system.md` has all of the above with the reasoning, including
the two things that bit during bring-up (line references inside a `red` subproof
must be absolute, and a row over no literals has no chain to telescope).

## AI use

Written with **Claude Code**, model **Claude Opus 5 (1M context)**
(`claude-opus-5[1m]`); every commit carries a matching `Co-Authored-By:` trailer.
The slack-row derivation was additionally put in front of a critical-review
subagent on the Fable model before being implemented, per `CLAUDE.md`'s "When to
ask for a second opinion"; it found the `proofgoal`-must-end-in-contradiction
defect and the empty-row exclusion, both of which are recorded in the design
note. **This has not yet been checked by a human**, which given
`CONTRIBUTING.md`'s bar — being able to explain why each change is correct — is
the main thing outstanding alongside the sanitize run.

https://claude.ai/code/session_01Ayd9oqNB3Qwif1MpYy1ySh

### DIFF
--- a/dev_docs/README.md
+++ b/dev_docs/README.md
@@ -342,6 +342,19 @@ library. For an introduction to *using* the solver, start with the top-level
   punched, because it crossed a bounds-consistent `int_lin_eq`, and fixing that
   took mario from 4/15 optimal to 9/15 — the same bug is in four more
   redefinitions (#803).
+- [`ParitySystem`: GF(2) reasoning over a conjunction of XORs](parity-system.md) —
+  the design and as-built note for issue #647: why a single `ParityOdd` is
+  already GAC and all the remaining inference lives in the conjunction, the
+  Gocht-Nordstrom slack form under which a Gaussian elimination step is one
+  `pol`, and why it has to be *derived in-proof* from cake's accumulator chain
+  rather than written into the `.opb` (a presolver has no `ProofModel`, and a
+  row cake did not emit is not chain-portable). Records the per-step `red` pair
+  and the five-line subproof that bridges the two encodings — linear, where the
+  paper's CNF recovery is exponential, because our chain already *is* its
+  partial-parity split — the §4.3 fold that makes every inference one `pol` plus
+  one RUP whatever the size of the combination, and the two mutations that
+  survive because they only remove slack the wrapping RUP replaces. Read before
+  touching `ParityOdd`'s encoding, whose rows the derivation cites.
 - [`MinDistance`: encoding and proofs](min-distance-proofs.md) — the definitional
   OPB encoding for `min_distance(D, x, z)` (site-selection flags, per-site counts,
   pair clauses, and the min-attained ladder), the justification for each of the
--- a/dev_docs/README.md
+++ b/dev_docs/README.md
@@ -353,8 +353,15 @@ library. For an introduction to *using* the solver, start with the top-level
   paper's CNF recovery is exponential, because our chain already *is* its
   partial-parity split — the §4.3 fold that makes every inference one `pol` plus
   one RUP whatever the size of the combination, and the two mutations that
-  survive because they only remove slack the wrapping RUP replaces. Read before
-  touching `ParityOdd`'s encoding, whose rows the derivation cites.
+  survive because they only remove slack the wrapping RUP replaces. Also covers
+  `ParitySystemGathering`, the presolver that collects posted `ParityOdd` and
+  Boolean `Equals` / `NotEquals` constraints into per-component systems, why a
+  two-literal donor needs two RUP lines where a chain needs a `red` per step, the
+  measured node-for-node tripwire and the strength differentials that say the
+  gathering buys anything at all, and the two API extensions it forced — a
+  published naming *object* for a family of rows, and a `family` on
+  `ProofFlagKey`, without which `v[id][1]` and `x[id][1]` had the same key. Read
+  before touching `ParityOdd`'s encoding, whose rows the derivation cites.
 - [`MinDistance`: encoding and proofs](min-distance-proofs.md) — the definitional
   OPB encoding for `min_distance(D, x, z)` (site-selection flags, per-site counts,
   pair clauses, and the min-attained ladder), the justification for each of the
--- a/dev_docs/parity-system.md
+++ b/dev_docs/parity-system.md
@@ -1,7 +1,7 @@
 # `ParitySystem`: GF(2) reasoning over a conjunction of XORs
 
-Working note for issue #647. **Status: the constraint is built and its proofs
-verify with zero assertions. `ParitySystemGathering` is not written yet.**
+Working note for issue #647. **Status: built. `ParitySystem` and
+`ParitySystemGathering` both verify with zero assertions.**
 
 `ParityOdd` reasons about one XOR at a time (`gcs/constraints/parity/parity.cc`,
 the `if (++how_many_unknown > 1) return` bail-out). That is GAC for a single
@@ -284,13 +284,10 @@ component with at least two rows. A single-row component is left to its own
 
 `ParitySystemGathering`, in `gcs/presolvers/parity_system_gathering/`, built to
 the `DifferenceLogic` pattern: a shared `ComponentStats` block registered
-unconditionally at the top of `run()`, donor discovery through
-`Problem::each_constraint_of_type_with_proof_data` so the rows it cites are
-named by the donor's published role rather than by a label it built itself, and
-one bucket per reason a candidate was passed over — because a presolver that
-silently gathered nothing passes every solution-equivalence, OPB byte-diff and
-VeriPB check there is, and the counts are the only thing that tells "working"
-from "no-op".
+unconditionally at the top of `run()`, and one bucket per reason a candidate was
+passed over — because a presolver that silently gathered nothing passes every
+solution-equivalence check, OPB byte-diff and VeriPB run there is, and the counts
+are the only thing that tells "working" from "no-op".
 
 Donors come from two families.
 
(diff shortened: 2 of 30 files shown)


ciaranm:
Cross-reference for whoever picks these up: #940 vendors the `<generator>` shim
and is what unblocks macOS builds on current Xcode (libc++ gained
`std::ranges::elements_of` in LLVM 22, which collides with the polyfill this tree
fetched). It is independent of this PR — nothing here needs it and it needs
nothing here — but it is worth landing first, so that this branch can be rebased
onto it and have its own macOS lanes test the arrangement `main` will actually
have.

CI is green on both today, since the GitHub runners predate LLVM 22; the
collision is currently only visible on a local Xcode that is ahead of them.

ciaranm:
Pre-merge review. The code is in good shape — everything the body claims that I could check, checked out exactly (see the bottom). Five things to fix, none of them merge-blocking: four are `dev_docs/parity-system.md` disagreeing with itself, one is a coverage gap.

## 1 & 2. `## The constraint` (L403–L414) describes the design that "Decisions taken" says was abandoned

Two wrong claims in one paragraph, and I think they are one stale edit rather than two:

- **L403:** *"`prepare()` installs one child `ParityOdd` per row (the `path.cc` / `tree.cc` pattern)"*. There is no `prepare()`. `ParitySystem::define_proof_model` loops `define_parity_chain` over the rows. **L524** records exactly this: *"installing child `ParityOdd` constraints does not work, because a child takes the parent's `ConstraintID` and k identical children would collide on every row label"*.
- **L411:** *"`s_expr()` throws: the `.scp` grammar has `parity` for one row and nothing for a system, and `write_scp` renders the whole file before opening it, so a throw leaves no `.scp` behind."* It does not throw — it writes the row list, and this PR adds the reader branch for it. **L434**, four lines later, says so: *"`gcs/scp_reader.cc` reads the `parity_system` form back, so a `.scp` this writes round-trips."*

Observed, not inferred — a `.scp` preserved from `parity_system_test`:

```
(constraints
    (_1 parity_system (((_1 != 0) (_2 != 0) (_3 != 0))))
)
```

and `glasgow_scp_solver --all` on it returns the four odd-parity solutions.

The sentence that survives is *"The SCP chain lane does not cover `ParitySystem`"*, which is true — there is no `scp_cases` fixture for it.

## 3. L170–L177 states the `Top` scaffolding deletion as done; L543 says it is not implemented

L170: *"so they are deleted immediately afterwards"*; L172: **"Steady-state `Top` footprint: two lines per donor."**; L175 then explains at length why the deletion is safe and unchecked. "Residue" at L543 says *"that is not implemented yet, so the current `Top` footprint is the full `7n` lines per row rather than two"* — which matches the body of this PR, and matches the code: there is no proof-line deletion anywhere in `parity_chain.cc`, `gf2_system.cc` or `parity_system_gathering.cc`.

L543 is candid that "everything above says to delete them", so the two halves are not equally wrong — but §2 reads as a description of what the code does, and a reader who stops before "Residue" takes away the opposite of the truth. Worth a "not yet implemented" on L172 specifically, since that is the sentence stated as a measured fact.

## 4. L450: *"All five are done for the constraint; the presolver has not been started."*

Contradicted by L3 (*"Status: built. `ParitySystem` and `ParitySystemGathering` both verify with zero assertions"*), by the whole of "The presolver" with its measured tables, and by L481's *"The presolver came after all five, with its own three gates"*.

## 5. The new `parity_system` reader branch has no test

`gcs/scp_reader.cc:1314` adds the `parity_system` op. Nothing exercises it: no case in `scp_reader_test.cc` (which has 87, including one for the sibling `parity` op at L326), no `verified_encodings/scp_cases/*.scp`, and nothing else in the tree reads a `.scp` containing one. A typo in the `terms.size() != 3` arity check or in the row walk would go unnoticed.

Flagging rather than proposing a fix, but the sibling's test is the obvious shape. I did check the branch works by hand — one-row and two-row systems both enumerate correctly, and a row of constants over no variables gives the single empty solution.

---

## What I verified, so it does not get re-raised

Measured on the trial merge onto `main` `620a62e0` (the merge needed conflict resolution; see the following comment):

- **Builds and tests.** GCC release 0 warnings / 680 targets; clang-21 36 occurrences from 3 sites, which is exactly the pre-existing `element.cc` `-Wunused-lambda-capture` baseline and nothing new. **Release 878/878, sanitize 878/878, and release with `-DGCS_TEST_CAP_DEFAULTS=OFF` 878/878** — all with 0 skips, MiniZinc on PATH, and `CAKE_PB_CP` set, so all 170 `scp_chain` lanes ran the full workflow-2 chain with **zero** fallbacks (CakePB-dev `a402078`). That covers the three things the body lists as not run.
- **Zero assertions, confirmed the positive way.** 25 runs in `parity_system_test` + 36 in `parity_system_gathering_presolver_test` = the body's 61, every one `s VERIFIED`, no `s UNDER ASSERTIONS`, and `grep -c '^a '` zero across all 61 preserved `.pbp`.
- **`ParityOdd`'s refactor into `define_parity_chain` is byte-identical.** All 128 artefacts (`.opb` / `.pbp` / `.scp` / `.varmap`) hash-identical to `main`'s at `--seed=424242`, compared by content rather than filename.
- **Every number in the document's tables is exact**: all four tripwire rows (`chain` 2/3/14 → 2/3/5, `camouflage` 0/1/4 → 0/1/1, `two_components` 4/7/24 → 4/7/14, `bool_bridge` 2/3/14 → 2/3/5), both strength differentials (3 → 1), stage 3's 23 + 2 = stage 5's 25.
- **No search-tree movement.** `equals.hh` gains only const accessors, but since it is in the MiniCP-benchmark trigger set I ran them: `magic_series` 1193, `qap` 123333, `tsp` 4285745, `magic_square` 6042079 all match their references, and `n_queens --size=14 --all` is an identical A/B against `main` at 6391931 recursions / 365596 solutions.
- Build wiring complete (new sources, `parity_system_test`, the presolver test, and the view tests all registered); `gcs/gcs.hh` needs nothing, as L436 says; trailers on 4/4 commits in the house form with AI declared in the description; no stray files; no `AssertProofRule` / `AssertRatherThanJustifying` use; all four new `overloaded{//}` blocks pinned correctly.

One note that belongs to neither PR: `main`'s `NamesAndIDsTracker::ensure_flag_defined` builds its memo key with a hard-coded `'v'` — the same latent ambiguity `ProofFlagKey::family` exists to remove, one function away. Not live today (only `Cumulative` publishes a flag definer and its keys are all `Values`), and it arrived after this branch forked, so neither review could have seen it. Worth a follow-up rather than anything here.

The `dev_docs/frontend-support-matrix.md` row L437 says is still needed: the matrix has no `ParityOdd` row either, so that gap predates this PR.

ciaranm:
Follow-up opened as #981, covering all five items above plus the closing note.

- **1–4** (`parity-system.md` disagreeing with itself): all four corrected. Item
  3 is the one with teeth — "Steady-state `Top` footprint: two lines per donor"
  was the sentence stated as measured fact, so the correction lands there, and
  §2 now reads as the argument for why the deletion *would be* sound rather than
  a description of what runs.
- **5** (no test for the `parity_system` reader arm): two cases in
  `scp_reader_test.cc`, one counting solutions and one round-tripping a posted
  `ParitySystem`. No `scp_cases` fixture — cake has no rule for a system, and
  the CMakeLists' own `lin_greater_equal` note says the chain cannot check a
  reader anyway. Mutating the reader to flatten rows and to accept a trailing
  term fails 13 of the 30 assertions.
- **The closing note** (`ensure_flag_defined`'s hard-coded `'v'`): fixed there
  too, since it is the same latent ambiguity `ProofFlagKey::family` exists to
  remove. Both sites now build the name through one helper.
- **The matrix row**: added, since the gap predates this PR. `ParityOdd` is
  reachable from both frontends; the system and the presolver from neither.

#981 is based on `main` rather than on `issue647-parity-system`, so nothing here
has to be sequenced around it and deleting this branch on merge cannot close it
(`CONTRIBUTING.md` L178). Its diff shows this PR's four commits until this one
lands, then collapses to three files.

PR #940, opened 2026-09-11
Vendor the <generator> shim instead of fetching a polyfill

ciaranm:
**This unblocks macOS builds on current Xcode.** It is a build-system change with
no behaviour change, and it is independent of #938 — but it should land first, so
that #938 can be rebased onto it and have its macOS lanes test the arrangement
we'll actually have.

## The breakage

`#include <__generator.hpp>` alone no longer compiles:

```
./__generator.hpp:224:11: error: type constraint differs in template redeclaration
.../c++/v1/__ranges/elements_of.h:32:11: note: previous template declaration is here
```

libc++ has added `std::ranges::elements_of` — `libcxx/include/__ranges/elements_of.h`
is absent in `llvmorg-19/20/21.1.0` and present in `main` — and exposes it from
`<ranges>`, while still not providing `<generator>`. The polyfill includes
`<ranges>` itself at line 41 and then defines its own `elements_of` at line 224,
so the clash is self-contained.

**CI is currently green**, because the `macos-15` and `macos-26` runners predate
LLVM 22. This was a local Xcode getting there first; it will reach them.

## Why vendor rather than patch

The dependency was worth removing regardless. It was an `ExternalProject_Add`
fetching `github.com/todoubaba/generator` at configure time — a fork with no
releases and no other users, last pushed December 2023, existing for three small
commits (a missing `<atomic>`, some warning fixes). Its own upstream has not been
touched since January 2025 and has five open issues, so there was nowhere to send
a fix either. Reported anyway, as
[lewissbaker/generator#17](https://github.com/lewissbaker/generator/issues/17).

So the header is vendored as `util/p2168_generator.hpp`, under the Boost Software
License 1.0 it already carried, with the licence file alongside. The `try_compile`
probe, the `ExternalProject`, the `generator_include` target and its install rules
are gone; `util/generator.hh` makes the choice in the preprocessor instead, and
the seven call sites that each repeated the conditional now include that.

## The one change to the vendored copy

`std::ranges::elements_of` and the three `yield_value` overloads that consume it
are **removed**, so `co_yield std::ranges::elements_of(...)` does not compile
against the shim. Nothing in the tree does that — there are no references to
`elements_of` anywhere.

Guarding the definition and deferring to libc++'s would not work: libc++'s is an
aggregate with public `range` / `allocator` members where this one has `get()` /
`get_allocator()`, and libc++'s `std::allocator<std::byte>` default collapses
exactly the distinction this one's `use_allocator_arg` default drew between
`yield_value(elements_of<_Rng>&&)` and `yield_value(elements_of<_Rng, _Allocator>&&)`.
Adapting means reworking that allocator dispatch, for a facility this tree never
uses and could not test here — so the shim is honestly missing it, loudly, rather
than quietly carrying template code nobody can check.

The diff against the pinned upstream source is **66 lines removed and 31 added,
and every one of the 31 is a comment**. The file keeps its third-party `.hpp`
extension, which also puts it outside what the clang-format lane globs.

## Verification

`build/` deleted entirely, then: configures with no network fetch, builds the
whole tree on the SDK that was broken, and `ctest --preset release` passes
**833/833**.

Not run locally: the `sanitize` build, and `-DGCS_TEST_CAP_DEFAULTS=OFF`. CI
covers the first.

`dev_docs/building.md` has all of the above under "The vendored `<generator>`
shim", including the version boundaries, so the next person meeting this
diagnostic finds the answer.

## AI use

Written with **Claude Code**, model **Claude Opus 5 (1M context)**
(`claude-opus-5[1m]`); the commit carries a matching `Co-Authored-By:` trailer.
The version boundaries, the repro and the licence provenance above were executed
and checked rather than inferred. Not yet reviewed by a human.

https://claude.ai/code/session_01Ayd9oqNB3Qwif1MpYy1ySh


ciaranm:
Pre-merge review. No blocking findings; one reproducible off-by-one worth correcting whenever the file is next touched.

## The provenance claim checks out, and it is the one worth checking

I fetched the pinned fork source (`todoubaba/generator` at `a849a206`, `include/__generator.hpp`, 805 lines) and diffed it against `util/p2168_generator.hpp`. Four hunks, and nothing else in 770 lines:

```
0a1,30        the vendored header comment
222,259d251   ranges::elements_of + its two deduction guides   (38)
462,476d453   two yield_value(elements_of<...>) overloads      (15)
503,514d479   yield_value(elements_of<_Rng>&&)                 (12)
```

- **Every added line is a comment or blank**, checked mechanically rather than by eye (`grep '^>' | grep -vE '^\s*(//|$)'` is empty).
- The removals are exactly `elements_of` and the three overloads consuming it. Nothing else is modified anywhere.
- `elements_of` appears nowhere in the tree outside that header's own comments, so "nothing in this tree does that" holds.

## The one nit: the line counts are 65 and 30, not 66 and 31

Both `dev_docs/building.md` ("the diff against the pinned upstream source is 66 lines removed and 31 added, and every one of the 31 is a comment") and the PR body give 66/31. Two independent tools agree on 65/30:

```
diff upstream __generator.hpp vendored p2168_generator.hpp   ->  65 '<' lines, 30 '>' lines
git diff --numstat (same two files as one commit)            ->  30     65
```

38 + 15 + 12 = 65. The *substance* is right — every added line really is a comment, and the removal really is only `elements_of` — so this is a stale count rather than a wrong claim, and it blocks nothing. Worth a one-character fix next time `building.md` is open, since it is a figure a reader can reproduce in one command.

## What I verified, so it does not get re-raised

Trial-merged onto `main` at `71642ee0` (after #937), which is not the base CI built against.

- **The install tree, which no CI lane touches** and which this PR's install-rule rewrite could have broken. `cmake --install` into a clean prefix puts `generator.hh`, `p2168_generator.hpp` and `p2168_generator.LICENSE` in `include/util/`; no stale `__generator.hpp` reference and no `generator_include` survives in the exported targets. Then a consumer `#include <gcs/gcs.hh>`, compiled and linked against that prefix, builds and runs correctly. The `PATTERN "*.hh" PATTERN "*.hpp" PATTERN "*.LICENSE"` glob is exactly right for `util/`, which holds nothing else matching.
- **The shim itself compiles and works**, which is worth doing because it has *zero* coverage on a libstdc++ box — `__cpp_lib_generator` is defined here, so `util/p2168_generator.hpp` is never reached by any of the 829 lanes. Forced directly: a `std::generator<int>` coroutine over it builds and yields correctly under **GCC 15.2 and clang++-21**, at `-W -Wall`, 0 warnings each. (The macOS lanes are the real coverage, and they are green.)
- No fetch remains: `configure.log` clones catch2, cxxopts, json, small_vector and xcsp3parser, and no `generator`.
- `-DGCS_WERROR=ON` release: **0** warnings, 460 TUs. clang++-21: **36 occurrences / 3 sites**, all the known `array_has_nonconstants` baseline in `element.cc`, none from the vendored header. Sanitize: the known 673 GCC `<generator>` false positives and nothing else.
- `ctest` **829/829** release; sanitize running at the time of writing and green so far.

## What I could not check, and neither can CI

The actual breakage. I have no LLVM 22 libc++ here, and the `macos-15` / `macos-26` runners predate it — as the PR says. So neither of us can demonstrate the *fix*; what is demonstrable is the mechanism, and it is demonstrable by construction: the shim no longer declares `std::ranges::elements_of` at all, so it cannot redeclare libc++'s. Worth stating plainly rather than leaving the green macOS lanes to imply more than they show.

Removing the overloads rather than adapting them reads as the right call for the stated reason — a shim that is loudly missing a feature nothing uses beats one carrying template code nobody here can exercise.

PR #942, opened 2026-09-12
State an at-least-one over an interval cover, not per value (#939)

### LINKED ISSUE #939: A variable's at-least-one constraint is built over its whole definition range, one term per value
## Summary

`NamesAndIDsTracker::need_constraint_saying_variable_takes_at_least_one_value()`
builds its at-least-one over the variable's whole **definition range**, one term
and one `eq` atom per value:

```cpp
// names_and_ids_tracker.cc — the simple branch and the view branch, identically
auto [lower, upper] = _imp->integer_variable_definition_bounds.at(var);
for (Integer v = lower; v <= upper; ++v)
    al1s += 1_i * (var == v);
auto line = _imp->logger->emit_rup_proof_line(al1s >= 1_i, ProofLevel::TopAndCore);
```

So asking for a variable's AL1 costs a proof line with `ub - lb + 1` terms, and
creates that many `eq` atoms, **regardless of how many values the variable's
domain actually holds**. It is cached per variable, so it is paid once — but once
is enough, and the atoms it leaves behind are the expensive part.

## It is not just the one line — the atoms poison what comes after

Those `eq` atoms seed the variable's **interval partition**
(`define_invar_with_covering` → `init_interval_partition`, "the variable's first
range literal creates its containment tree, seeded with the eq atoms that already
exist"). So after an AL1 over a wide range, the *first range literal* on that
variable has a covering over span-many cells, and each cell is defined.

Measured on a `GlobalCardinality` over four variables, three of which have
**exactly two values in their domain** (`{1, W}`), the fourth `1..W`:

| `W` | `.opb` | `.pbp` | `need_invar` calls | invars actually defined |
|---:|---:|---:|---:|---:|
| 100 | 7 162 B | 278 738 B | | |
| 1 000 | 9 133 B | **3 226 308 B** | **3** | **2 991** (997 per variable) |
| 10 000 | 12 076 B | 37 618 315 B | | |
| 100 000 | — | >170 MB, did not finish | | |

**The encoding is flat (1.2x per 10x) and the proof grows 11.6x per 10x.** Three
range literals are requested; 2 991 get defined, because the covering has to span
a partition with a cut at every one of the ~1 000 `eq` atoms the AL1 created.
`red` lines are 17 956 of the 65 891 proof lines at `W = 1000`.

## Attribution

Traced with a stack trace on the 500th `need_direct_encoding_for` call:

```
need_direct_encoding_for            names_and_ids_tracker.cc:788
xliteral_for_ensuring               names_and_ids_tracker.cc:2034
pb_file_string_for_ensuring         names_and_ids_tracker.cc:2027
emit_inequality_to                  emit_inequality_to.cc:41
ProofLogger::emit_rup_proof_line    proof_logger.cc:691
need_constraint_saying_variable_takes_at_least_one_value  names_and_ids_tracker.cc:506
```

The caller in this instance is `emit_gcc_capacity_pol`, which adds one AL1 per
confined variable, but the primitive is generic: `gac_all_different`'s Hall
reasoning and `recover_am1` are the other significant users, and any future rule
that wants "this variable takes some value" pays the same.

## Why nothing caught it

* The **audit lane** runs without proofs, so it never reaches a tracker that
  emits anything. `GlobalCardinality` and `GlobalCardinality/closed` are `Clean`.
* The **proof-scaling survey** does run with proofs, but its `GlobalCardinality`
  rows use a narrow value list (`{1_i}`, `{1_i, 2_i}`), so the definition range
  the AL1 spans is small. Nothing in either instrument varies the *spread* of a
  cover, and the spread is what this is proportional to.

## What would fix it, roughly

The AL1's purpose is "this variable takes at least one value", and the terms for
values the variable can never take are dead weight the checker has to carry. Two
directions, neither obviously right:

* Build the AL1 over the variable's **initial domain** rather than its definition
  bounds, where those differ. The values in between are excluded by the model
  already, so the line would be no weaker; the question is whether every consumer
  can still cancel against it (`#882`/#904 made the view branch emit in V-form
  for exactly this kind of cancellation reason, so the constraint is real).
* Or state it over **runs** the way range literals do, which needs the
  at-least-one to be expressible over intervals at all.

**A trap either way, and it is the reason #936 and this issue are coupled.** The
span-many `eq` atoms this creates are what lets a *range* literal propagate at
all: `link_immediate_containment` emits `~[j=s] \/ [j in g]` per contained eq
atom, so unit propagation descends the containment tree and zeroes the gap terms
one at a time. #936 replaces a run of disequalities with one range literal in the
capacity reason, and that is only as strong as the disequalities **because this
AL1 has already shattered the gap into singletons**. Narrow the AL1 without
changing the reason and the closure RUP stops propagating. The two have to move
together.

The shape that moves them together, and the one I would try first: replace the
definition-range AL1 with a per-firing at-least-one over the hall set, stated
under exactly the confinement literals #936 already builds —

```
rup  Sum_{v in W} [j=v] + ~[j >= lb] + [j >= ub+1] + Sum_g [j in g] >= 1
```

which is RUP (negate it: the eq-atom reverse implication at `lb` gives
`ge_{lb+1}`, each gap literal's forward reification carries `ge_{g_lo}` to
`ge_{g_hi+1}`, each hall value steps on, and `ge_{ub+1}` conflicts), touches only
thresholds the count lines' atoms already created, names no gap `eq` atom, and is
O(|confined| x |W|) per firing rather than O(span) once. Its literal list is
exactly `append_confined_to_hall_reason`'s output, so the reason and the
at-least-one become one object.

Related: #833, #936 (the neighbouring per-value loop in the same rule, which is
what I was measuring when this turned up), #857, #876.

---

Measured by an AI agent (Claude) on this machine, Release build, g++ 15.2.0, at
`main` `8f261250`. All probe edits were reverted; nothing here is committed.

---

## CORRECTION (fixed by #942)

Two things in "Why nothing caught it" above were wrong, and both matter for
anyone reproducing this.

**1. The survey explanation was wrong.** I wrote that the survey's
`GlobalCardinality` rows miss this because their value lists (`{1_i}`,
`{1_i, 2_i}`) are narrow, "so the definition range the AL1 spans is small". The
definition range the AL1 spans is the *variable's*, not the cover's, and those
rows' variables are declared over the full `0..width`. The real reason is
simpler: those rows never make a variable **confined** (domain ⊆ hall set), so
`emit_capacity_pol` is never reached and no at-least-one is emitted at all. The
longest proof line `GlobalCardinality/hall` writes is 190 characters.

**2. `GlobalCardinality/hall` does not grow because of this.** I expected its
10.0x-per-10x proof growth in the survey to be this bug. It is not: applying
#942 leaves that row at exactly 43 988 → 439 988 lines, unchanged. Its growth is
the demand-side per-value pruning already documented as its `KnownTrip` — the
at-most-one over the hall set plus the single removed value, one value at a time.

So this issue's own measurement (four variables, three holding `{1, W}`) was
sound, and the survey simply has no row of that shape. The fix therefore adds
one: `TEST_CASE("Large domain proof sizes")`, a wide **definition range** with a
narrow **domain**, which is the shape no existing probe produces — every other
probe in the file gets its wide definition range *by* wanting a wide domain.

**What the fix actually buys**, on a `GlobalCardinality` with three variables
declared over `0..width` and confined to `{1, 2}`:

| width | before | after |
|---|---|---|
| 10^3 | 66 057 lines, 0.14 s | 255 lines |
| 10^4 | 660 057 lines, 10.9 s | 255 lines |
| 10^5 | 2 528 821+ lines, >600 s | 255 lines |
| 10^9 | — | 255 lines |

**And one thing the issue did not anticipate**: the rewrite has to be gated on a
width threshold, not applied unconditionally. The per-value line is emitted once
per variable and then serves every cover anyone asks for; a cover line is
specialised, so a caller whose cover changes per firing pays one per distinct
cover. Over a narrow definition range that is a *regression* — 16 % more proof
lines on `sudoku`, 3 % on `ortho_latin` — and caching recovers only a fifth of
it. With the threshold those examples are byte-identical.

`recover_am1` is named above as a significant user; it is not a caller of this
primitive. The actual callers were `GlobalCardinality` (both arms),
`gac_all_different`, `Among` (two sites), `min_distance` and `subcircuit`. All
but `subcircuit` are converted — its pigeonhole needs every value named, and a
successor's definition range *is* the node set.

ciaranm:
Fixes #939. **Stacked on #941** (`issue936-gcc-capacity-reason-runs`) — review that first; this PR's diff against it is the 11 files below.

## The problem

`need_constraint_saying_variable_takes_at_least_one_value` spells its at-least-one out one term per value over the variable's whole **definition** range. Asking for those eq atoms also splits the variable's interval partition into singletons, so every later covering is span-proportional too. Neither cost has anything to do with how many values the variable can still take.

A `GlobalCardinality` whose three variables are declared over `0..width` and confined by `In` to `{1, 2}`:

| width | before | after |
|---|---|---|
| 10^3 | 66 057 lines, 0.14 s | 255 lines, 0.001 s |
| 10^4 | 660 057 lines, 10.9 s | 255 lines, 0.001 s |
| 10^5 | 2 528 821+ lines, >600 s | 255 lines, 0.001 s |
| 10^9 | — | 255 lines, 0.001 s |

`AllDifferent` and `Among` in the same shape go 659 937 → 133 and 659 940 → 135 at 10^4. All verified by VeriPB, with real conclusions (`COMPLETE ENUMERATION OF 6 SOLUTIONS` and `UNSATISFIABLE`), not `NO CONCLUSION`.

## The fix

`..._over_cover(var, singled_out)` states the same fact over an interval cover: the caller's values of interest as their own eq atoms, the maximal runs between and around them as one range literal each.

```
[x in 0..2] + [x = 3] + [x in 4..6] + [x = 7] + [x in 8..10^9] >= 1
```

A pol that adds it still gets `+[x = v]` for each named value, so cancellation against per-value at-most-ones and count lines is unchanged. What differs is the residue: one `+[x in run]` per run instead of one `+[x = w]` per unnamed value. The caller's reason rules those out exactly when it pins the variable's bounds and excludes its holes — a run outside the bounds dies on the order chain, a run inside one sits inside an excluded hole and containment falsifies it. `generic_reason` and the Hall-set reasons all qualify.

**Converted**: both `GlobalCardinality` arms, `gac_all_different`'s Hall set, both `Among` sites, `min_distance`. **`subcircuit` deliberately keeps the per-value form** — its pigeonhole needs every value named, and a successor's definition range *is* the node set, so there is no width there to save. (That call site was not in my first survey of consumers: I had grepped the shared main checkout, whose working tree is weeks stale. Grep the worktree.)

## Why it is a threshold, not an unconditional rewrite

This is the part I would not have predicted. The per-value line is emitted **once per variable** and then serves every cover anyone asks for, because it names every value. A cover is specialised, so a caller whose cover changes from firing to firing — a Hall set — pays a line per distinct one. Over a *narrow* definition range each of those lines is about as big as the single line it replaces, and there are many of them:

| example | base | cover always | cover + cache | with width test |
|---|---|---|---|---|
| `sudoku` | 21 050 | 25 211 (+16.3%) | 24 474 | **21 050** |
| `ortho_latin 5 --all-different gac` | 123 288 | 127 423 (+3.4%) | 127 423 | **123 288** |
| `crystal_maze` | 12 395 | 13 414 | 13 305 | **12 395** |
| `colour` | 1 187 | 1 187 | 1 187 | **1 187** |
| `n_fractions` | 3 677 826 | 3 678 077 | 3 678 018 | **3 677 826** |

Caching the cover lines recovered only about a fifth of the regression, because the Hall sets genuinely differ. So the cover is stated only above a definition-range width threshold, and every example above is byte-identical to before. A rewrite that is asymptotically better can still be worse everywhere anyone actually is.

## A new lane, because nothing could see this class

The audit lane solves **without** `ProofOptions`, so no justification is ever emitted and no reason materialised — it structurally cannot see per-value work done while writing a proof. The `[.proofscaling]` survey can, but it is run by hand and pins nothing. `TEST_CASE("Large domain proof sizes")` closes that gap: each row is a wide definition range with a narrow domain, solved with proofs, and its proof both counted and checked with VeriPB. Registered unconditionally, since it needs no guard.

It discriminates: **3 failures and 39 s without this change, 0 failures and 0.18 s with it.**

That shape is also why the class stayed hidden. Every other probe in the file gets its wide definition range *by* wanting a wide domain, so a variable whose domain is narrow while its definition range is wide never arose.

## A correction to #939's own premise

I filed #939 saying this was what drives `GlobalCardinality/hall`'s 10× proof growth in the survey. **It is not.** That row never reaches an at-least-one at all — its longest proof line is 190 characters — and its growth is the demand-side per-value pruning already documented as a `KnownTrip`. Applying this fix leaves that row at exactly 43 988 → 439 988. The at-least-one cost is real and much worse, but it needed a probe nobody had written. I have corrected the issue.

## Gates

- release ctest **829/829**, *nothing skipped* (MiniZinc 2.9.7 and the real `cake_pb_cp` verified chain both on `PATH`)
- `-DGCS_TEST_CAP_DEFAULTS=OFF` **829/829**
- Sanitize **829/829**
- clang **833/833**
- `-DGCS_WERROR=ON` clean, clang-format 21.1.8 clean
- audit lane with the guard **ON**: 211 assertions, 3 test cases, no verdict moved
- merges cleanly with current `main` (which has #937 and #940)

---

### Naming only the still-possible values

Each caller now names only the values **its variable can still take**, not its whole value set. They differ whenever that set is a Hall set, a cover, or a union of domains and one variable's domain is a small part of it — which is every caller here. On a Hall violator over eight values the at-least-ones drop from ten terms to four:

```
-  rup 1 i[_1][eq0] 1 i[_1][eq1] ... 1 i[_1][eq8] 1 i[_1][in9_1000] >= 1;
+  rup 1 i[_1][eq0] 1 i[_1][eq1] 1 i[_1][eq2] 1 i[_1][in3_1000] >= 1;
```

Naming a value the variable cannot take was only wasteful, not wrong — its pol term is zero under the reason either way. What the tightening actually buys is the correspondence: the residue is now exactly the complement of the domain, which is exactly what the reason states, so the leftovers discharge by construction rather than by an argument about which runs happen to sit inside which holes.

**Measured honestly, it is 0.5–0.8% of proof size** on a violator whose hall set is 4–64× any one domain. Small, because the at-least-ones are not where the volume is (the pairwise at-most-ones are, and they are cached at Top). **It also does not remove the need for the width threshold**: it makes each line smaller, and the narrow-domain regression is about how *many* lines there are, so `sudoku` only improves from +16.3% to +10.4% with the threshold off. The threshold stays; the examples stay byte-identical.

### The thing that was actually costing AllDifferent

`AllDifferent::prepare` built its compressed value set — the union of the initial domains, which is the right-hand side of GAC's bipartite graph — by walking every value of every domain and **linearly scanning what it had collected so far** for each one. O(values × distinct values): asymptotically worse than the algorithm it feeds.

| initial domain | before | after |
|---|---|---|
| `0..10^4` | 67 ms | **2 ms** |
| `0..10^5` | 6 637 ms | **38 ms** |
| `0..10^6` | did not finish in 200 s | **528 ms** |

Membership now goes through a set, leaving the order — first-seen, and so the value indices the propagator's graph uses — exactly as it was. **340 proof artefacts across the two AllDifferent test binaries are byte-identical at a pinned seed**, which is what "the order is unchanged" has to mean. `AllDifferentExcept` had the same loop and gets the same treatment.

This **does not change the audit lane's verdict, and should not**: GAC still wants a graph vertex per value, so a genuinely wide domain remains the `KnownTrip` the lane records, and what it needs is stage 5's weaker arm rather than a faster setup. `consistency::VC` on the same probe was 0 ms at every width throughout, which is what made the attribution unambiguous.

### Gates re-run for this commit

release **830/830** nothing skipped · caps-off **830/830** · Sanitize **830/830** · clang **834/834** · `-Werror` clean · clang-format clean · guard-on audit lane 211 assertions, no verdict moved · `ortho_latin`, `sudoku`, `crystal_maze`, `colour`, `n_fractions` byte-identical · merges cleanly with current `main`

https://claude.ai/code/session_01AsUZjNThjBFs7VsFkCQfux

### DIFF
--- a/dev_docs/large-domains.md
+++ b/dev_docs/large-domains.md
@@ -966,6 +966,81 @@ ns/node before and 3 912 after, i.e. no difference outside noise, because a
 branching decision is a per-cent of what a node costs. The trade is a constant
 against an asymptote, which is the right way round.
 
+## The proof-size lane
+
+The audit lane above solves **without** `ProofOptions`, so nothing that happens
+only while a proof is being written is reachable from it: no justification is
+emitted and no reason is materialised. The `[.proofscaling]` survey below can see
+that work, but a survey is run by hand and pins nothing. Between them sat a class
+with no instrument at all, and #935, #936 and #939 all came out of it.
+
+`TEST_CASE("Large domain proof sizes")` in the same file is the gate for it. Each
+row is one shape — a variable **declared** over a wide range, a domain confined by
+`In` to a couple of values inside it, and an inference whose justification names
+those values — solved with proofs, and its proof both counted and checked with
+VeriPB. The bound is loose on purpose: it separates "a few hundred lines" from
+"one line per value", which is orders of magnitude, not a byte count to retune
+whenever the encoding shifts. Checking as well as counting matters, because a
+small proof that does not verify is a worse outcome than a big one that does.
+
+**This shape is why the class was invisible.** Every other probe in the file gets
+its wide definition range *by* wanting a wide domain, so a variable whose domain
+is narrow while its definition range is wide never arose — and that is exactly the
+gap between a domain (what a propagator sees) and a definition range (what the
+proof encoding names). Before #939 each of these rows wrote about 66 × width proof
+lines; the lane fails in about 39 seconds if that comes back, against 0.2 seconds
+when it has not.
+
+### At-least-one constraints span the definition range (#939)
+
+`need_constraint_saying_variable_takes_at_least_one_value` spells its at-least-one
+out one term per value over the variable's whole **definition** range, and asking
+for those eq atoms also splits the variable's interval partition into singletons,
+so every later covering is span-proportional too. Neither cost has anything to do
+with how many values the variable can still take.
+
+`..._over_cover(var, singled_out)` states the same fact over an interval cover
+instead: the caller's values of interest as their own eq atoms, the maximal runs
+between and around them as one range literal each. A pol that adds it gets the
+same `+[x = v]` for each named value, so cancellation against per-value
+at-most-ones and count lines is unchanged; what differs is the residue, one term
+per run rather than one per unnamed value. The caller's reason has to rule those
+runs out, which it does exactly when it pins the variable's bounds and excludes
+its holes — a run outside the bounds dies on the order chain, and a run inside one
+sits inside an excluded hole, so containment falsifies it. `generic_reason` and
+the Hall-set reasons all qualify.
+
+Measured on a `GlobalCardinality` whose variables are confined to two values:
+
+| width | before | after |
+|---|---|---|
+| 10^3 | 66 057 lines, 0.14 s | 255 lines, 0.001 s |
+| 10^4 | 660 057 lines, 10.9 s | 255 lines, 0.001 s |
+| 10^5 | 2 528 821+ lines, >600 s | 255 lines, 0.001 s |
+| 10^9 | — | 255 lines, 0.001 s |
+
+`AllDifferent` and `Among` in the same shape go 659 937 → 133 and 659 940 → 135 at
+10^4.
+
+**It is a threshold, not an unconditional rewrite, and that is the interesting
+part.** The per-value line is emitted once per variable and then serves every
+cover anyone asks for, because it names every value; a cover is specialised, so a
+caller whose cover changes from firing to firing — a Hall set — pays a line per
+distinct one. Over a *narrow* definition range each of those lines is about as big
+as the single per-value line it replaces, and there are many of them: going to
+covers unconditionally cost `sudoku` 16 % more proof lines and `ortho_latin` 3 %,
+for variables declared over nine values, where naming every value was never the
+problem. Caching the cover lines recovered only a fifth of that, because the Hall
+sets genuinely differ. With the width test the examples are byte-identical again
+(`ortho_latin`, `sudoku`, `crystal_maze`, `colour`, `n_fractions` all unchanged to
+the line) and the wide case is flat. **A rewrite that is asymptotically better can
+still be worse everywhere anyone actually is**, and the only way to find that out
+is to measure the narrow case as well as the wide one.
+
+`subcircuit` deliberately keeps the per-value form: its pigeonhole needs every
+value named, and a successor's definition range *is* the node set, so there is no
+width there to spend on values the counting does not use.
+
 ## Proofs
 
 **Out of scope for fixing.** Several of these have no viable fix today, and a
--- a/dev_docs/large-domains.md
+++ b/dev_docs/large-domains.md
@@ -1041,6 +1041,47 @@ is to measure the narrow case as well as the wide one.
 value named, and a successor's definition range *is* the node set, so there is no
 width there to spend on values the counting does not use.
 
+Each caller names only the values **its variable can still take**, not its whole
+value set. The two are different whenever the value set is a Hall set, a cover or
+a union of domains and one variable's domain is a small part of it, and the
+correspondence it buys is worth more than the terms: the residue is then exactly
+the complement of the domain, which is exactly what the reason states, so the
+leftovers discharge by construction instead of by an argument about which runs
+happen to sit inside which holes. Measured, it is a **0.5–0.8 % proof-size**
+saving on a Hall violator whose hall set is four to sixty-four times any one
+domain — real, but small, because the at-least-ones are not where the volume is
+(the pairwise at-most-ones are, and they are cached at Top). It does **not**
+remove the need for the width threshold: it makes each line smaller, and the
+narrow-domain regression is about how *many* lines there are, so `sudoku` only
+improves from +16.3 % to +10.4 % with the threshold off.
+
+### AllDifferent's compressed value set was quadratic
+
+Separately from anything proof-shaped, `AllDifferent::prepare` built its
+compressed value set --- the union of the initial domains, which is the right-hand
+side of GAC's bipartite graph --- by walking every value of every domain and
+doing a **linear scan of what it had collected so far** for each one. That is
+O(values x distinct values), so it was asymptotically worse than the algorithm it
+feeds:
+
+| initial domain | before | after |
+|---|---|---|
+| `0..10^4` | 67 ms | 2 ms |
+| `0..10^5` | 6 637 ms | 38 ms |
+| `0..10^6` | did not finish in 200 s | 528 ms |
+
+Membership now goes through a set, leaving the order --- first-seen, and so the
+value indices the propagator's graph uses --- exactly as it was: 340 proof
+artefacts across the two `AllDifferent` test binaries are byte-identical at a
+pinned seed. `AllDifferentExcept` had the same loop and gets the same treatment.
+
+**This does not change the audit lane's verdict, and should not.** GAC still wants
+a graph vertex per value, so a genuinely wide domain is still the `KnownTrip` the
+lane records, and what it needs is stage 5's weaker arm rather than a faster
+setup. What the fix removes is only the part that was gratuitous: `consistency::VC`
+on the same probe was 0 ms at every width throughout, which is what made the
+attribution unambiguous.
+
 ## Proofs
 
 **Out of scope for fixing.** Several of these have no viable fix today, and a
--- a/gcs/CMakeLists.txt
+++ b/gcs/CMakeLists.txt
@@ -201,6 +201,13 @@ if(GCS_BUILD_TESTS)
         add_test(NAME large_domain_audit_test COMMAND $<TARGET_FILE:large_domain_audit_test>)
     endif()
 
+    # The proof-size case in the same binary is registered unconditionally,
+    # because it needs no guard: it asserts how big a proof is, which is as
+    # meaningful with the guard off as on, and it is the only lane that can see
+    # per-value work done while writing a proof (the audit above solves without
+    # ProofOptions, so it never emits a justification at all).
+    add_test(NAME large_domain_proof_size_test COMMAND $<TARGET_FILE:large_domain_audit_test> "Large domain proof sizes")
+
     # Checks that the GCS_VERBOSE_LOGGING proof annotation resolves gcs source
     # frames, which depends on the debug info the Release build keeps (-g1 on
     # GCC/Clang, /Z7 + /DEBUG on MSVC; see the top-level CMakeLists.txt). It
--- a/gcs/constraints/all_different/all_different.cc
+++ b/gcs/constraints/all_different/all_different.cc
@@ -15,6 +15,7 @@
 #include <algorithm>
 #include <map>
 #include <memory>
+#include <set>
 #include <utility>
 #include <variant>
 #include <vector>
@@ -27,6 +28,7 @@ using std::make_shared;
 using std::make_unique;
 using std::map;
 using std::move;
+using std::set;
 using std::unique_ptr;
 using std::vector;
 using std::ranges::adjacent_find;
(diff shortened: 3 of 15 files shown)


ciaranm:
Merged (`099cbf71`), after a trial merge onto current `main` — worth doing here because `main` had gained #937 and #940 since the branch point, and four files overlapped, `names_and_ids_tracker.hh` included. Release with `-DGCS_WERROR=ON`: 0 warnings, 830/830, 0 skips with all 103 MiniZinc lanes run. Sanitize: 830/830, and 673 warnings from 37 sites, which is the known `<generator>` false-positive baseline exactly. clang-21: 36 occurrences from 3 sites, all `element.cc`, also unchanged. `git diff` between the tested tree and `main` is empty.

One thing to fix afterwards, in `dev_docs/large-domains.md`: **three figures there disagree with the lane this PR adds.** Measured on the merged tree, at the lane's own width of 10^4:

| doc | line | says | lane prints |
|---|---|---|---|
| GlobalCardinality, before | L1039 | 660 057 | **659 934** |
| GlobalCardinality, after | L1038-L1041 | 255, at every width | **129** |
| AllDifferent, after | L1043 | 133 | **132** |

`AllDifferent`'s before (659 937) and `Among`'s 659 940 → 135 are both exact, which is what makes the GlobalCardinality row look like it was measured before #941 was rebased underneath: #941 changes the GCC capacity reason, so both of that row's columns move, and no other row's do.

**The claim the table supports is intact** — the after column really is flat in width. I measured GlobalCardinality at 129 lines at 10^3, 10^4, 10^5 and 10^9, verifying each time. So this is a two-number edit, not a doubt about the work.

How I got the current figures, if it helps:

```
GCS_PRESERVE_PROOF_FILES=all ./build/large_domain_audit_test "Large domain proof sizes"
GlobalCardinality/confined                      129 proof lines
AllDifferent/confined                           132 proof lines
Among/confined                                  135 proof lines
```

and the "before" column, by forcing `worth_a_cover` to return false so every call falls back to the per-value form:

```
GlobalCardinality/confined                   659934 proof lines
AllDifferent/confined                        659937 proof lines
Among/confined                               659940 proof lines
```

That last run is also the lane's negative control, and it is a good one: it fails in **38.3 s**, so the "about 39 seconds if that comes back" in the same section is exact.

Two other claims I ran rather than took, both of which hold, so they need no rechecking:

- **#941's helper is load-bearing.** Widening its emitted range by one on each side reddens both `bounds_global_cardinality_constraint` and `gac_global_cardinality_constraint` — VeriPB rejects the false reason. The new `{1, 5}` row earns its place too: with it, every run emits `[2,4]` deterministically, where without it 2 of 5 random seeds produced no interval at all. (So the arm *could* reach one before, just not reliably — slightly stronger than "nothing in the suite built an interval here at all".)
- **The byte-identity claim for the `AllDifferent` setup fix.** 340 artefacts across `all_different_test` and `all_different_except_test` at `--seed=424242`, 0 missing and 0 extra by content hash — the count matches the document as well as the verdict.

Last, an aside rather than a finding: the proof-size probes at 10^9 hang in `AllDifferent::prepare` for >600 s even with the set fix, since GAC still wants a graph vertex per value. That is the `KnownTrip` the audit lane already records, and the shipped lane is at 10^4, so nothing is affected — but it is why I could only confirm the width-flatness claim for GlobalCardinality at 10^9, not for the other two.

ciaranm:
Fixed in #945. Thanks — and the diagnosis needed one correction, which is the interesting bit.

**It was not measured before #941 was rebased underneath.** The 255 / 660 057 pair came from a *standalone probe of a different instance*: satisfiable, counts `0..2`, six solutions, where the lane's row is UNSAT with counts `0..1`. On the merged tree that same probe gives **252**, not 255 — so it was also stale, by the follow-up commit that made each caller name only the values its variable can still take. `AllDifferent`'s 133 has that second cause alone, which is why its *before* matched and its *after* did not.

So it was two different instances sitting one paragraph apart under one heading, with one of them stale on top. Every individual number had been measured; that is what let it through. The table now quotes the lane, which is one command away and cannot drift from it.

Re-measured rather than copied: 659 934 / 659 937 / 659 940 before, 129 / 132 / 135 after, negative control 39.7 s (text now says "about 40 seconds").

**One addition to your 10^9 aside**: `Among` does hold at 10^9 — 135 lines, VeriPB-verified — so it is only `AllDifferent` that cannot be checked there, for the GAC-setup reason you give. The old table asserted a single 10^9 figure for all three rows; it is now per-row:

| row | 10^3 | 10^4 | 10^5 | 10^9 |
|---|---|---|---|---|
| `GlobalCardinality` | 129 | 129 | 129 | 129 |
| `Among` | 135 | 135 | 135 | 135 |
| `AllDifferent` | 132 | 132 | 132 | does not finish |

**Your `{1, 5}` finding is in too.** "Two of five seeds produced no interval at all" is a better claim than my "no row above this one exercises the interval form at all", and it makes the row's actual value clearer: what it buys is determinism, not first coverage. Comment corrected to say that.

The two claims you re-ran rather than took — #941's helper being load-bearing, and the 340-artefact byte-identity — I have left alone, since your checks were stronger than mine (a content-hash comparison and a widened-range mutation, against my count-and-verdict). I also re-checked the five example proof sizes and the `AllDifferent` setup timings on the merged tree; both unchanged.

PR #945, opened 2026-09-15
Quote the lane, not a probe, for the proof-size figures

ciaranm:
Fixes the three figures in `dev_docs/large-domains.md` that Ciaran found disagreeing with the lane, on the #942 merge.

## What was wrong, and why

| row | doc said | lane prints |
|---|---|---|
| `GlobalCardinality`, before | 660 057 | **659 934** |
| `GlobalCardinality`, after | 255, at every width | **129** |
| `AllDifferent`, after | 133 | **132** |

**The cause is not a stale rebase.** The 255 / 660 057 pair came from a standalone probe of a *different instance* — satisfiable, counts `0..2`, six solutions — where the lane's row is UNSAT with counts `0..1`. On the merged tree that same probe gives **252**, not 255, because it also predated the follow-up commit that made each caller name only the values its variable can still take. So two different instances sat one paragraph apart under one heading, and the number was stale on top of that. `AllDifferent`'s 133 has only the second cause.

That is worth naming precisely, because it is the kind of error that survives review: every individual number was measured, and the conclusion the table supports was correct.

## The fix

The table now quotes the lane, which is one command away and cannot drift from it:

```shell
./build/large_domain_audit_test "Large domain proof sizes"
```

| row | before | after |
|---|---|---|
| `GlobalCardinality/confined` | 659 934 | **129** |
| `AllDifferent/confined` | 659 937 | **132** |
| `Among/confined` | 659 940 | **135** |

Re-measured here rather than copied from the review comment: the before column by forcing `worth_a_cover` to return false, which is also the lane's negative control, and it takes **39.7 s** (the text said "about 39 seconds"; now "about 40").

## The width-flatness claim is narrowed to what is checkable

The old table asserted one 10^9 figure for all three rows. Verified individually instead:

| row | 10^3 | 10^4 | 10^5 | 10^9 |
|---|---|---|---|---|
| `GlobalCardinality` | 129 | 129 | 129 | **129** |
| `Among` | 135 | 135 | 135 | **135** |
| `AllDifferent` | 132 | 132 | 132 | **does not finish** |

All VeriPB-verified where they run. `AllDifferent` cannot be checked at 10^9 because its GAC setup still wants a graph vertex per value — the `KnownTrip` the audit lane already records, and nothing to do with this work. (You noted you could only confirm 10^9 for `GlobalCardinality`; `Among` does hold there too.)

## One more overstated claim, same class

The spread-cover row's comment in `bounds_global_cardinality_test.cc` said no row above it exercised the interval form *at all*. Your measurement — two of five seeds built no interval, the rest did so incidentally — is the accurate version, and it makes the row's real value clearer: what it buys is **determinism**, every run emitting `[2,4]`, which is what a regression test for the spelling has to be. Comment corrected to say that.

## Re-checked on the merged tree and unchanged

- the five example proof sizes are still byte-identical (`ortho_latin` 123 288, `sudoku` 21 050, `crystal_maze` 12 395, `colour` 1 187, `n_fractions` 3 677 826)
- the `AllDifferent` setup timings are still 2 / 38 / 532 ms at 10^4 / 10^5 / 10^6

## Gates

Release ctest **830/830**, nothing skipped, with MiniZinc 2.9.7 and the verified `cake_pb_cp` chain on `PATH`. clang-format clean. Docs plus one test comment, so no further lanes run.

https://claude.ai/code/session_01DMSzBpTxfSYdetu8wWvkCX


ciaranm:
The three replacement figures are right — re-measured on this branch rather than compared against the review comment, and the table matches exactly: 659 934 → 129, 659 937 → 132, 659 940 → 135. Quoting the lane instead of a standalone probe is the right fix, and the diagnosis is better than the one in my comment: I guessed a stale rebase, and two different instances a paragraph apart under one heading — SAT with counts `0..2` against the lane's UNSAT with `0..1` — explains it properly.

The narrowed flatness claim holds too, **including the row I said I could not check**. `Among` does hold at 10^9: 135 lines, `s VERIFIED`. Dropping the `AllDifferent` probe from the list is enough to reach it, since that is what does not finish. So the table's 10^9 column is confirmed on both rows that claim it.

## One correction, and it is to my number, not yours

The comment this adds to `bounds_global_cardinality_test.cc` says:

> measured over five seeds, two of them built no interval here at all, and the rest did so only incidentally

That "two" comes from my #942 comment, and it was wrong twice over. The five runs I actually had were 3 with no interval and 2 with one — so **three of five**, not two — and five runs of a randomised suite is too small to quote as a figure at all. I should have said so rather than putting a count in a review comment.

Re-measured properly on this branch, with the spread-cover row removed and the helper instrumented to report every range it emits, over **30 seeds**:

```
runs with NO wide interval: 16 of 30 ; runs with at least one: 14 of 30
```

so it is about half, and the per-run counts are spread widely (0 to 8 wide intervals). The substantive point is unaffected and the new comment is a real improvement on "no row above this one exercises the interval form at all" — it is only the count that should not go into the tree, because a reader will take it as a property of the suite rather than as one sample of a random process.

Something like "on about half of random seeds the rows above this one build no interval here at all, and the rest do so only incidentally" would be true and would not drift. Or drop the frequency entirely and keep the second half, which is the part that justifies the row: what it buys is determinism, every run emitting `[2,4]`.

Nothing else to raise. `AllDifferent`'s 2 / 38 / 528 ms setup timings and the five example proof sizes are untouched by this diff, and the "about 39" → "about 40" second edit is within what I measured for the negative control (38.3 s here, 39.7 s in your run) — a timing, so it will differ between sittings either way.

ciaranm:
Agreed on all counts, and fixed in `35f52483`.

You are right that the number should not be in the tree, and right that it was wrong — but the mistake is mine rather than yours: I took the count out of a review comment and committed it without measuring it. That is the same lesson as the rest of this branch, one level along. Quote the instrument; and if a figure is worth stating at all, derive it rather than inherit it.

I have dropped the frequency rather than restating it as "about half", taking your second option. Half is still a property of a random process, and the comment does not need it — what justifies the row is the determinism, not how often the rows above happen to miss:

> so the rows above this one reach the interval form only incidentally, if at all --- which of them do depends on the seed. This row makes it deterministic: every run emits `[2,4]`, which is what a regression test for the interval spelling has to be.

No number to drift, and it still stops a future reader concluding either that nothing above ever builds an interval or that this row is redundant because something above might.

I did not re-run your thirty seeds. The edit removes the claim rather than restating it, so there is nothing left for a measurement to support — and spending thirty randomised runs to justify a sentence that no longer contains a figure would be the wrong trade. The determinism half is the one that needed evidence and it has it twice over: instrumented at `[2,4]` three times when the row was added in #941, and confirmed deterministic in your own #942 run.

`bounds_global_cardinality_test` still 47 VeriPB runs, all `s VERIFIED`, on a comment-only change.

Issue #972, opened 2026-09-20
Disjunctive2D: the cumulative relaxation, derived in the proof, to bring 2D the whole certified ladder


ciaranm:
`Disjunctive2D` has pairwise time-table strength and nothing else. It has no
`with_rules` method at all, so unlike 1D the ladder is not merely off by default
there — it does not exist. This issue is the route to giving 2D the whole
certified cumulative ladder in one step, by deriving the cumulative relaxation
inside the proof rather than by building four more 2D rules.

## The relaxation

Project the `diffn` onto one axis. On x, task `i` has start `x_i`, duration
`w_i`, **height** `h_i` (its y-size), and the capacity is `H`, the y-extent of
the rectangles' own domains.

The validity argument is three steps:

1. Two rectangles both x-active at time `t` can satisfy neither x-disjunct of
   the 4-way separation clause, so the clause forces
   `before_{i,j,y} ∨ before_{j,i,y}` — they are y-separated.
2. So the set active at `t` is pairwise y-disjoint, and every member lies
   inside the y-window.
3. A set of pairwise-disjoint intervals of lengths `h_i` inside a window of
   height `H` has `Σ h_i ≤ H`.

Step 3 is the whole difficulty, and we already have it.

## Step 3 is `ComparatorNetwork`, unchanged

`gcs/innards/proofs/comparator_network.hh` proves exactly this statement, and it
is already general enough in every dimension that matters:

- its telescoping `chain`, plus `top_upper_bound` and `bottom_lower_bound`, sum
  to `window_hi − window_lo − Σ durations ≥ 0`;
- it handles **unequal** durations — they are muxed through the network with one
  sum-preservation row per comparator, and the `SwapDurations` mutation exists
  precisely to demonstrate that the construction really does unequal ones rather
  than only the equal-duration case #730 first verified;
- it is **duration-magnitude invariant**: cost depends on the number of tasks and
  only logarithmically on their range. That property is why it was built — it is
  the certificate of choice "where a time-indexed re-encoding would be too wide",
  which is exactly the y axis here;
- a caller supplies its own separation rows through `add_separation` and "never
  has to know how they are moved onto the sorted wires".

Substitute *duration* := y-size and *window* := y-extent and it is the row we
want, with no change to the class.

Steps 1 and 2 are also mostly built. Minting an activity flag per (task, time)
by `red` over order literals the encoding already has is what #737's time-indexed
[...]

Issue #984, opened 2026-09-20
Disjunctive2D: the flagged capacity row, for the relaxation's energetic rungs


ciaranm:
The 2-D cumulative relaxation now exists as far as *time-tabling* goes (#972,
PR #985): project a `Disjunctive2D` onto an axis, and the set whose mandatory
parts cover a time is pairwise separated on the other axis, so a comparator
network sorts it and telescopes to "the window is at least as tall as what is
in it". The overflow contradiction and both bound pushes are derived from that,
per firing, with nothing in the OPB and no activity flag minted anywhere.

That is the first rung. This issue is the rest of the ladder, and the reason it
is a separate piece of work rather than more of the same.

## Why the energetic rungs need a different row

Route B's endgame lands on a statement about the **mandatory** set, with no
activity flags in it:

    Σ_{i ∈ S} h_i ≤ H     (under the firing's guard)

Time-tabling wants exactly that, because its set *is* the mandatory one. Every
energetic rule — overload, edge-finding, TTEF, KAOC — instead sums a capacity
row across a whole window, over tasks that are only *possibly* active at each
time in it. What it needs is the **flagged** row

    Σ_i h_i · active_{i,t} ≤ H

and no amount of guarding gets there from the mandatory-set form: a task
outside `S` has an activity flag the reason says nothing about, so it cannot be
weakened in and it cannot be pinned out.

## The construction that does reach it

Route A, recorded in a comment on #972 and repeated here so this issue stands
alone. Per rectangle, two wires and no new model content:

- `P_i = a_i ? y_i : y_hi` — the position muxed on the activity flag, so an
  inactive rectangle becomes a dummy parked at the top of the window;
- `D_i = h_i · a_i` — the duration as a *wire* rather than a pinned constant.

Then every pair is separated under every assignment, not merely the active
ones: both active is the model's own separation, one inactive puts the active
one entirely below the dummy, both inactive is a tie of two zero-duration wires
at `y_hi`. The endgame telescopes to `y_hi − y_lo − Σ D_i ≥ 0`, which **is**
`Σ h_i·a_i ≤ H`, unconditional, with the wires deleted afterwards.

The semantics were brute-forced over 200,000 random instances with no
counterexample. It has **never been run through VeriPB**.

## What it would take

1. **Lift `ComparatorNetwork`'s pinned-positive duration.** `add_task` pins a
   constant `≥ 1` and the gap lemma consumes that positivity on its tie side.
[...]

Issue #973, opened 2026-09-20
Derived Cumulatives from a donor that has no per-time capacity rows


ciaranm:
Prerequisite for bringing `Disjunctive2D` into the certified ladder via its
cumulative relaxation — see the tracker.

## The assumption, and where it is written down

`gcs/constraints/cumulative/derived_cumulative.hh` lets a presolver post an
implied `Cumulative` that adds no rows to the OPB and establishes its per-time
capacity rows inside the proof instead. Every donor it can read, though, is a
*posted `Cumulative`*:

- `DerivedCumulativeTask` carries a donor `ConstraintID` and a `position`,
  because "a derived constraint creates no flags of its own, so every task has
  to point at a posted Cumulative that already encoded it ... the flag keys
  `ConstraintProofModelData` publishes are by task position";
- `DerivedCumulativeRows` is `std::map<ConstraintID, ProofLine>`, documented as
  "the donors' capacity rows for one time point", i.e. rows that already exist
  and are cited;
- `CumulativeDonorView` is "a posted Cumulative as a would-be deriver can speak
  about it".

All three assume the donor already has a time index in the model.

## What a non-Cumulative donor needs

`Disjunctive2D`'s OPB is purely pairwise — `before_{i,j,d}` per pair and axis
plus one 4-way separation clause per pair, and no time index anywhere. So its
per-time capacity rows have to be **derived**, not cited, and its activity flags
**minted**, not looked up.

Both halves already exist elsewhere: #737's time-indexed bridge mints an activity
flag per (task, time) by `red` over order literals for the 1D energetic rules,
and `ComparatorNetwork` supplies the row's content.

## The change

Give the deriver a donor abstraction with two implementations — one that cites a
posted `Cumulative`'s row and flags (today's behaviour, unchanged), one that
derives them. This is the only real design decision in the 2D arc; everything
above and below it is existing machinery.

Worth getting right rather than quickly, because it is shared innards and
`InferredDisjunctive`, `InferredCumulative` and `CumulativeStrengthening` all sit
on it. A change that makes the Cumulative path go through an indirection it did
not have should be checked for proof-byte drift on the existing derived-constraint
lanes, which compare output closely.

Part of #976.

---

*Drafted by Claude (Opus 5) during a scoping pass for the scheduling paper, from
`origin/main` at `620a62e0`.*

Issue #922, opened 2026-09-10
GlobalCardinality loses solutions and emits an unRUPable inference when a cover value is repeated


ciaranm:
## Summary

`GlobalCardinality` assumes its cover values are distinct, and nothing enforces
it. Give it a repeated value and **both** propagators lose solutions and emit an
inference VeriPB rejects. The OPB encoding is fine; it is the propagation that
is wrong.

`bounds_global_cardinality.cc:160` states the assumption in a comment — part 2's
Hall reasoning is over "contiguous runs of the **(sorted, distinct)** cover
values" — but `sort_cover_values()` only sorts, `GlobalCardinality` has no
`prepare()`, and no constructor or `post` path rejects the input.

## Repro

```cpp
Problem p;
auto a = p.create_integer_variable(0_i, 1_i, "a");
auto c1 = p.create_integer_variable(0_i, 1_i, "c1");
auto c2 = p.create_integer_variable(0_i, 1_i, "c2");
p.post(GlobalCardinality{vector<IntegerVariableID>{a}, {1_i, 1_i}, {c1, c2}});
```

`values = {1, 1}` means `c1 = #{v in vars : v = 1}` and `c2` the same, so
`(a=1, c1=1, c2=1)` satisfies the constraint. All four configurations drop it:

| | solutions found | expected |
|---|---|---|
| BC, open | 1 — `(a=0,c1=0,c2=0)` | 2 |
| BC, closed | **0** | 1 — `(a=1,c1=1,c2=1)` |
| GAC, open | 1 — `(a=0,c1=0,c2=0)` | 2 |
| GAC, closed | **0** | 1 |

## It is an invalid proof, not just a lost solution

Run either with `--prove` and VeriPB refuses:

```
Error: Checking error at dupcover_open.pbp:21
	The constraint is not implied by reverse unit propagation (RUP) from core
	and derived database.
```

The rejected step is the propagator concluding `a != 1`:

```
rup 1 ~i[a][b0] 1 ~i[a][b0] >= 1;
```

which the closed row alone refutes. **The encoding is correct** — the closed
at-least-one comes out as

```
@c[_1][0_al1] 1 i[a][b0] 1 i[a][b0] >= 1;
```

i.e. the same literal twice, which VeriPB normalises to `2 a >= 1` and which
still says exactly `a = 1`; and the two count rows are duplicates of each other,
which is consistent rather than wrong. So this is a propagator bug on both arms,
[...]

PR #985, opened 2026-09-20
Derive a cumulative relaxation per firing, so 2-D can time-table each axis

ciaranm:
**Stacked on #982** (guarded separation clauses), which is stacked on #980 —
merge top-down. This adds one commit on top.

Project a `Disjunctive2D` onto one axis and the rectangles become a
`Cumulative`: rectangle `i` is a task with start `pos_i` and duration `size_i`
on that axis, of height its size on the other, on a resource whose capacity is
the extent the set is confined to there. This adds that relaxation's
time-tabling — the overflow contradiction and both bound pushes, on each axis
in turn — behind a new `Disjunctive2DRules::cumulative_relaxation`, off by
default.

It sees conflicts the pairwise rule cannot: three rectangles whose mandatory
parts share a time, no two of whose mandatory *boxes* overlap, can still be too
tall between them to fit. That is what a square packing is made of.

## The certificate

Nothing reaches the OPB, and nothing can — one encoding per constraint is the
#922/#780 rule, and a capacity row is not something `cake_pb_cp` could
re-derive from the `.scp`. It is derived per firing instead, which is route B
of the two #972 records:

1. Two rectangles that both occupy `t` can satisfy neither time-axis disjunct
   of their 4-way separation clause. Each is refuted by `emit_before_pol`'s
   shape, and what is left of the clause is the pair's separation **on the
   resource axis**.
2. The set is then pairwise separated, and `ComparatorNetwork` sorts it and
   telescopes to "the window is at least as tall as what is in it", which the
   firing says it is not.

The members are the ones whose *mandatory* parts cover `t`, so their occupancy
follows from their bounds and **no activity flag is minted anywhere** — no time
index, no `Top`-level vocabulary, nothing that outlives the firing. The
separation clauses are derived rather than model rows, so they carry the guard,
which is what #982 exists for.

Three things that were not obvious:

- **State the time-axis bounds tightly**, at `t − len + 1` and `t` rather than
  at the rectangle's own bounds. Not merely a better nogood: it is what makes
  each refutation cancel to degree **exactly one**, so that adding it to the
  4-way clause leaves a *clause* rather than something of higher degree the
  next `pol` cannot use. The reason is therefore an `ExplicitReason` over those
  literals rather than `generic_reason`.
- **A push's one synthetic fact is the negated conclusion**, and otherwise a
  push and a contradiction build the same certificate.
- **The non-strict zero-size escape needs *both* the guard and a pin**, and
  this is the one thing that cost a rejected proof. The escape arrives in each
  pair's clause from the model's own 4-way row, so the guard has to carry it or
  the network's arithmetic does not cancel — but the closing RUP then has to
  get from `size ≥ 1` to `~escape`, and its only route is bit arithmetic. The
  network's final row can force **one** unassigned escape flag and no more, so
  a firing with two of them was rejected. Pinning each escape false under the
  reason as well, exactly as the pairwise rule's `pin_escapes` does, makes
  every one of them follow by unit propagation.

Who may take part is settled in `prepare()`, from the model alone — a positive
constant size on the resource axis, plain variables for both positions, a
non-negative and not-too-wide resource domain, and no position handle shared
with another rectangle. Deciding that without the logger is what keeps the rule
drawing the same inferences whether or not proofs are on.

## What this does not reach, and why it is not the whole ladder

#972's text supposed route B would bring 2-D the entire certified ladder. It
does not, and the reason is worth recording. The endgame lands on a statement
about the **mandatory** set with no activity flags in it, which is exactly
enough for time-tabling. Every energetic rung — overload, edge-finding, TTEF,
KAOC — sums `capacity_row(t)` across a window over tasks that are only
*possibly* active at each time, so it needs the flagged row
`Σ h_i·active_{i,t} ≤ H`. That is route A, it needs the muxed construction
(an inactive rectangle becomes a zero-duration dummy parked at `window_hi`),
and that needs `ComparatorNetwork`'s pinned-positive duration lifted.

Filed as #984, with the construction, what it would take, and three invalid
variants of it. #973 is not needed for this rung — there is nothing for a donor
abstraction to carry when no per-time row exists — and under route A it is #973
as filed.

## Measurements

`examples/squares` is new, because shikaku cannot answer the question: its
rectangles have *variable* sizes and the relaxation sorts on a constant one.
`--relaxation` switches the rule, so both directions run on one model.

| instance | pairwise | with the rule |
|---|---|---|
| 3 squares of 3 in 5×8 (`tight`) | 202 recursions | **1** |
| 4 squares of 4 in 7×15 | 29,013 | **1** |
| 5 squares of 3 in 5×14 | 156,856 | **1** |
| proof, 4 squares of 4 | 25.7 MB / 596,725 lines | **112 KB / 1,695 lines** |
| proof, 5 squares of 3 | 145.7 MB / 3,254,334 lines | **218 KB / 3,098 lines** |

Both arms verify in every row. **And the other half, which is why it is off by
default**: on a solve where it barely fires — 4 squares of 2 in 7×7, 404,016
packings, 0.4% fewer recursions — it costs **1.74× wall clock**, 2.03 s against
3.54 s. `perfect21`, Duijvestijn's order-21 squared square, is out of reach for
both arms at 25 s.

## The lanes

`disjunctive_2d_relaxation_test`, whose every fixture comes with a control —
the same instance with the rule off — because a green suite says nothing about
a rule nothing else fires. The fixtures are built so the pairwise rule is
*silent* on them by construction: no rectangle has a mandatory part on the
resource axis, so there is no mandatory box to overlap and no blocker to push
away from.

| fixture | what it says |
|---|---|
| `sharp` | nine units of height do not fit in eight; the root closes, and does not with the rule off |
| `loose` | one unit more and it fits, so the refutation is not a technicality |
| `push` | the third rectangle's origin moves to 3, and does not with the rule off |
| `push_at_zero` | the same with the blocked time at the domain's start, so the pushed rectangle's tight lower bound lands *below* its domain |
| `push_transposed` | the mirror, with y playing the part of time |
| `two_escapes` | two zero-size escapes in one firing, the shape that made the pins necessary |
| `shared_width` | the same with one width variable shared between two rectangles, so the guard names its literal twice |
| `varwidth_strict` / `varwidth_nonstrict` | variable time-axis sizes, so each refutation has a size term to cancel, and in non-strict mode two zero-size escapes to discharge |
| `dup` | rectangles sharing a position handle take no part, so the rule must leave such a model exactly as it was |
| `enumerate` | 2,292 solutions against brute force |
| `--search 24` lane | 24 random instances, each enumerated against brute force and verified; the sweep has to fire the rule and to have pruned at the root somewhere |

Node counts are *not* compared: the harness branches randomly and reseeds every
run, so what the rules did at the **root** is the reproducible thing. Both
lanes clear the default caps, the way the `disjunctive_optional` lanes do, and
fail loudly if a cap ever fires — a truncated enumeration has checked no
completeness.

Four mutation lanes, all rejected by VeriPB: `emit_nothing` (the control — if
this verified, the whole derivation would be decoration), `skip_refutation`
(one time-axis disjunct left standing, so the "separation" is not one),
`skip_guard_weakening` (a clause carrying only its own share of the guard) and
`skip_escape_pins`, which runs on its own two-escape instance because a lane
with one escape passes with the pins gone.

## What a review caught

A Fable consultant was asked to attack the derivation, as `CLAUDE.md`
prescribes, and found the escape defect above — reproduced here before acting
on it. Worth recording is *how* it hid: the first `varwidth_nonstrict` fixture
gave one rectangle a width lower bound of two, which forces an encoding bit,
which pins one escape, after which the final row can force the other. Drop that
bound to one and the same fixture fails. It now uses one, and `two_escapes`
plus the `skip_escape_pins` lane pin the behaviour from both sides. The same
review also confirmed the degree-one cancellation for every reachable shape
(including literals outside a declared domain), that `set_bounds` closes over
bits rather than order-atom chains, that the pushes have no off-by-one, and
that the guard coefficient can never exceed the network's `big()`.

## Gates

- **Build**: GCC 15.2 release, exit 0. Also `-DGCS_WERROR=ON`, clean, zero
  warnings. Also clang 21.1.8.
- **`ctest --preset release`: 885/885, 0 skipped**, with MiniZinc 2.9.7 and
  `cake_pb_cp` on `PATH`. The same under clang (889/889 there --- its four extra are lifetime-annotation lanes).
- **Nothing else moves**: `disjunctive_2d_test`'s 118 proof files per mode are
  **byte-identical** either side of this change under a pinned `--seed=4242`,
  in both strict and non-strict mode. Asking the reifier for its big-M, which
  is the only change to `define_proof_model`, changes nothing it writes.
- Sanitize build clean.

Closes #972.

https://claude.ai/code/session_018bFVzk6osgMSxoovgWiCcv


ciaranm:
Holding this one for a test, at ciaranm's request. Everything else below it in the stack (#979, #980, #982) and #978 has landed; this PR is now standalone on `main`.

## The optional-rectangle guard is reached by no test

The rebase onto #979 added this, which is exactly what the old *"WHOEVER ADDS OPTIONAL RECTANGLES HERE (#974)"* comment asked for:

```cpp
// gcs/constraints/disjunctive_2d/disjunctive_2d.cc:253
if (_presence[i] && ! (is_constant_variable(*_presence[i]) && constant_value_of(*_presence[i]) == 1_i))
    continue;
```

It reads as correct, and this is not a report of a bug. It is a report that **nothing would notice if it were deleted**. I deleted those two lines, rebuilt `disjunctive_2d_relaxation_test` and `disjunctive_2d_optional_test`, and ran every lane both register: the relaxation test plain and with `--search 24`, and the optional test's `strict` / `nonstrict` / `falsify`. **Every one passed.**

The reason is structural, and it is the shape a stacked rebase tends to create:

- the relaxation is opt-in — `Disjunctive2DRules::cumulative_relaxation = false` (`disjunctive_2d.hh:86`);
- the only test that posts optional rectangles, `disjunctive_2d_optional_test`, never turns it on;
- the only test that turns it on, `disjunctive_2d_relaxation_test`, never posts a presence.

The guard only runs where those two meet, and no test puts them together. Before this rebase that intersection could not exist, which is why nothing covers it yet.

## Why it wants an enumeration test, not another proof lane

The comment at L218–L222 already says why a proof cannot catch this: counting an undecided rectangle's height makes the overflow conclusion unsound, and *"a solve would lose the placements that need the rectangle absent"* — while the certificate would carry the same wrong member set and verify against it. So a verify lane passes either way.

What would pin it is a case with **optional rectangles and `cumulative_relaxation` on together, where some solutions need a rectangle absent**, checked by enumeration against the expected solution set. With the guard deleted it should lose exactly those solutions and fail.

Two things would make it conclusive rather than merely present:

1. **Mutation-test it the same way** — delete L253–L254 and confirm the new case goes red, not just that it passes with the guard in.
2. **Cover the constant-present case too.** The guard deliberately *keeps* a rectangle whose presence is the constant 1, and the comment says that gives a byte-identical OPB to the non-optional form. A case where every presence is constant 1 should enumerate identically with the relaxation on and off — which would also catch the guard being tightened too far in the other direction.

## Also on this branch

I pushed **`b107a69f`** on top, at ciaranm's instruction: the rebase left the header of the old five-mode element loop with an empty body at `gcs/CMakeLists.txt:819` (`foreach(mode const constgac const2d var var2d)` / `endforeach()`). It was inert — `ctest -N` reports the same 902 tests before and after, identical lists — and the real eight-mode loop below it is untouched. It is a plain commit rather than a rewrite, so your history is as you left it.

https://claude.ai/code/session_01CHeiDuLnCj1ZBpEGi1uxD3

ciaranm:
Thank you — you were right that nothing reached the guard, and it was worse than a missing test: **the `optional` fixture this PR's description advertised was never on the branch.** During the rebase, `git apply --3way` failed on the `.cc` half of my patch and, being atomic, rolled back the test half as well while still printing "Applied patch … cleanly". My follow-up grep came back empty and I read it as a pass. The description is corrected.

Everything below is pushed as ``710dc452``.

## The guard now has a test that goes red

`optional_undecided`: two unconditionally-present rectangles that fit, plus one with a presence *variable* that would overflow, so every solution needs it absent. Enumerated against brute force with the rule on. **Delete the decline and it loses all six solutions** — checked against the code as shipped, not just against the original two lines.

## "A verify lane passes either way" — that premise was mine, and wrong

It came from my dev doc, which said the certificate "would carry the same wrong set and verify against it". I ran it instead of reasoning about it: with the decline deleted, **VeriPB rejects** the proof, at a proofgoal subproof that cannot close — the presence disjunct in that pair's separation clause is in no goal the comparator network offers. So the proof catches it with proofs on; with proofs off the solve loses the solutions silently. Your conclusion stands for exactly that reason, and the doc now says so.

## The constant-present suggestion — its premise was mine too

A constant presence never reaches the guard. `task_presence` resolves a constant 1 to *no literal* (a plain rectangle) and a constant 0 to `never_present`, dropped from `_active_rects` first. So the guard's constant-1 branch was **dead code**, and I have simplified it to `if (_presence[i])`, which is what it always meant. The comment claiming the guard "keeps" constant-present rectangles described a case it never sees.

And an on-and-off enumeration can't catch the guard being made too coarse, since leaving a rectangle out only weakens propagation. Checked rather than asserted: under a realistic over-tightening — skip the rule for any `Disjunctive2D` posted with presences — the constant-present enumeration **passes, 2292 of 2292**, and only an assertion that the rule *fired* notices. So the two constant-present fixtures assert firing:

| fixture | goes red under |
|---|---|
| `optional_undecided` | deleting the decline (all 6 solutions lost) |
| `optional_constant_sharp` — `sharp` with every presence the constant 1 must close the root, and must not with the rule off | skipping the rule for optional-form models |
| `optional_constant_enumerate` — `small` the same way, brute force plus "the rule fired" | the same, via the firing check alone |

## What is still coarse

A presence *variable* whose domain is `{1}`, or one fixed to 1 during search, is present and still left out for the whole solve, because membership is settled once in `prepare()`. Sound, recorded in the comment and the dev doc, not addressed here — taking such a rectangle part properly means carrying its presence literal through every fact list.

## `b107a69f`

Thank you — that stray loop was mine, from the rebase's conflict resolution. My `foreach`/`endforeach` balance check couldn't see it, because an empty loop is itself balanced.

## Gates

All on `710dc452`: GCC `ctest --preset release` **902/902**, clang **906/906** (its four extra are lifetime-annotation lanes), 0 skipped either way; `-DGCS_WERROR=ON` clean with zero warnings; sanitize **13/13** on the 2-D and `squares` lanes, which include the four proof-mutation lanes, all still rejected.

https://claude.ai/code/session_018bFVzk6osgMSxoovgWiCcv

